PR #41, opened 2023-10-15
new (interim) mbin icons

asdfzdfj:
tried to redraw/recreate mbin logo, looks decent enough

some notes:
- I kind of expecting this icon not to stay, but this should do until a better designed and accepted icons exist
- still need to find a good place to place the svg source file, as it's an inkscape svg and it have some hidden/invisible stuff I used as a reference, not really suitable for serving as is

speaking of which, should the filenames be renamed to be mbin related? currently it's a bit of a rush job overwriting existing kbin files and I think that renaming should be done.


melroy89:
Cool! Don't forget the image in docs/images as well (which is indeed already renamed to `mbin.png`).

asdfzdfj:
Updated some more icon images, haven't touch og image yet, not sure how to go with it, but maybe that can be done later

at least I think this should be ready for review

melroy89:
This is really awesome work! Thanks. I added you as owner as well to the GitHub Org.

Let me do some searching on disk if you maybe missed something.

melroy89:
It's indeed good to just leave the images in `assets/images` as is..

melroy89 (approved):

PR #43, opened 2023-10-15
Revert "Update .env.example_v2"

melroy89:
- This reverts commit 567de4b2435147771ad222a73c0b14036b60c87c.
- And.. reverts commit 3139a41964fa9605b28c6bc9a77092101865ad87.
- Change some safe values to Mbin


nobodyatroot (changes requested):
Shouldn't this commit actually revert to the KBIN naming of the variables? I think the MBIN_ naming will still break everything referenced outside of the .env file.

nobodyatroot (dismissed):
Much better

nobodyatroot (approved):
Looks good to me

PR #44, opened 2023-10-15
Improve Nginx config

melroy89:


asdfzdfj (approved):
since it looks like it was sourced from your existing nginx config this should be good

at least it won't match `/m/firefoxcss` or `.../Tommy-Tallruleico` anymore

asdfzdfj:
this one is a bit of less importance but at https://github.com/MbinOrg/mbin/blob/67d642151af9ad60bc33813ce2d039738c1a01fc/docs/admin_guide.md?plain=1#L484-L490 the `?$query_string` may be changed with `$is_args$args` (if there's no args the entire thing disappears).

and this one doesn't really improve its working in any way but there're a few `kbin`s to rename if you wish so.

melroy89:
> try_files $uri $uri/ /index.php?$query_string;

Is `$query_string` even needed?

melroy89:
> the `?$query_string` may be changed with `$is_args$args` (if there's no args the entire thing disappears).

@asdfzdfj  I like the idea.. but I don't use S3 storage for caching. So I can not fully verify this. Please, open a new PR for this and make sure it's tested on a real instance incl S3 storage.

PR #51, opened 2023-10-16
new/alternate mbin opengraph image

asdfzdfj:
proposing a new image for use in opengraph (+ source svg), a leftover task from #41 

the reason that I didn't update it at the time was because I kinda want the og image to have a bit of 'flair' somewhat like the kbin og image, and not just a ~~seal on bedsheet~~ logo on a plain background kind of image.


melroy89 (changes requested):
I do like the retro top & bottom, but see me remarks above. Maybe it's not just the font type, but also letter-spacing. Adding more space between each letter?...

I just find the B en N looking weird.

asdfzdfj:
things I could do:
- scaling up both the logo *and* the text, to about 1/2 or 2/3 of the image height should be good? (personal opinion but don't want them to be too big)
- add more space between letters

as for changing font face, that would be a bit harder as there is no font, those were actually drawn, but I'll see what I can do.  
perhaps increasing the text size while trying to retain text weight and adjust spacing should do? (note that I'm very much not a designer)

melroy89:
> things I could do:
> 
> * scaling up both the logo _and_ the text, to about 1/2 or 2/3 of the image height should be good? (personal opinion but don't want them to be too big)
> * add more space between letters
> 
> as for changing font face, that would be a bit harder as there is no font, those were actually drawn, but I'll see what I can do. perhaps increasing the text size while trying to retain text weight and adjust spacing should do? (note that I'm very much not a designer. designing is hard, and trying to appease human vision is not any easier)

Sounds great to me.

vintprox:
Text doesn't need to be hand-drawn, even though argument can be made that it's just a custom text that appears once in a logo. There will be times again where people promoting this software would return to this dreaded search of a type to make something even slightly customized. Given this, I would implore to find one that:
- fits the overall motiff (not too shabby, not too formal)
- has wider Unicode support

At least, these are pain points that I've encountered while customizing existing products.

melroy89:
I force myself to remove from "Changes requested" in GitHub.

Not that I don't want my proposal implemented (I want). But.. simply because I do not want to block PRs just by me alone. I think a discussion already started in Matrix, so I will not block this PR anymore.

asdfzdfj:
> Text doesn't need to be hand-drawn, even though argument can be made that it's just a custom text that appears once in a logo. There will be times again where people promoting this software would return to this dreaded search of a type to make something even slightly customized. Given this, I would implore to find one that:
> 
>   * fits the overall motiff (not too shabby, not too formal)
>   * has wider Unicode support
> 
> 
> At least, these are pain points that I've encountered while customizing existing products.

fair point, I think what compelled me to hand draw them at the time was:
- I wanted to avoid another license violation situation in here
  - see how well it went for us with Postmill, and that was a more permissive license.
  - I was thinking to use this font: [Novecento Sans Wide](https://www.fontspring.com/fonts/synthview/novecento-sans) (wide variant is available at no cost except for ultralight/ultrabold weights) and I wasn't so sure about its usage in vector graphics (the [license](https://www.fontspring.com/lic/l0g3bolf1h) says it's fine for fixed-size images, but then there's a part that say things like no transform or embed allowed, though giving it another read it may be fine for us if the letters were vectorized into outlines and not including the font files in here).
- initially I wanted to have the `M` play a bit with the logo shape/parts, drawn the `M`, but couldn't make it work in the end, so I just shift the letter out of the logo.

anyway I made some changes according to the remarks:
- increased the logo size to about 300px, as this is about the biggest size I could increase without it looks like it's about to spill over,  
- replacing the hand drawn letter with the letters from an actual font, Novecento Sans Wide, Bold, with `case` (all caps), `ss02` (alternate style `N`, yes I got that funky `N` from here) features enabled, and increase the letter spacing a bit (the default spacing is even tighter that my hand drawn spacing)

vintprox:
Thank you for getting back to this, @asdfzdfj. You've clearly put some thought into this!

1. That being said, reproductions with **Novecento Sans Wide** require "checking out" the font with the acceptance of End-User License Agreement, forfeiting some of the sweet data from designers having to use it. The license doesn't seem to align with SIL Open Font License - or it's the impression I get from this business, as far as my Legalese goes.
   
   Plus, by some weird coincidence, this font supports many language except Russian. I would close my eyes on this, as there can always be alternative to propose. The fact is it's just not the Unicode support I personally expected.

3. Getting playful by fusing logo with text doesn't sound bad - when done with caution. You're creating another logo with one derived, which sets up an expectation that it will have a solid shape as well.
   
   But if you remove all color from the latest sketch, all that remains is the shadow where triangle from former logo and `M` are barely distinguishable, because their parts were fused **through** without air to breathe. Silhouettes don't have layers. But again, maybe I'm just too focused on the shapes and it looks fine in the eye of the beholder:
   
   Shadow of the proposed Mbin logo with text

To address the 1st conundrum, let's find a font that looks as cool as the one you picked and has most languages.

2nd problem will require cleverly integrating both triangle and text with the room to breathe - or avoid fusion by just placing text to the right.

melroy89:
Why is https://github.com/MbinOrg/mbin/pull/51/files#diff-f3f89d5c092398cbb91f82acec60c43f6124a03cf924d43c13da1ac6262db177 not using the same rounded letters? Also feel free to update all logos with the new font.

asdfzdfj:
whoops I forgot to vectorize the text object to outline. fixed.  
I guess that's what you see when you didn't have the font it used installed.

asdfzdfj:
updated text logo to use the new font as requested

also I took liberty to change the image used in README to the OG image, let me know if you want to retain the old one.

vintprox:
Slightly rounded corners of the Rubik font give out this so-needed friendly vibe in the details. Plus the problem with intersecting graphics has been solved, which adds points to modularity. If need be, the logo can be easily swapped.

Thank you for your work!

melroy89:
Just some question/remark, maybe it's just me...

But when looking at these two images:

image

The logo in the top image seems fine to me. However, the logo in the image below seems a bit stretched vertically. Or is that optical illusion? The bottom logo seems a bit "pointy" or stretched vertically. Is that correct?

melroy89:
Anyhow, that being said. You did a wonderful job @asdfzdfj ! Thank you so much. And feel free to update ALL icons/images with the new letter type.

vintprox:
> The logo in the top image seems fine to me. However, the logo in the image below seems a bit stretched vertically. Or is that optical illusion? The bottom logo seems a bit "pointy" or stretched vertically. Is that correct?

I think eyes play tricks because of the old top logo that was squished?

melroy89:
> I think eyes play tricks because of the old top logo that was squished?

Should I maybe measure it haha.. And see that ratio of the two logos.

asdfzdfj:
there's a good chance it's just our eyes, human vision is hard

and just to check in case Inkscape was drunk, I recheck and the dimensions is the same when scaled to a common height
(currently Inkscape scale propotionally checkbox in transform window seems to only work once)

image

melroy89:
OK no worries.

melroy89 (approved):

PR #53, opened 2023-10-16
Fix consistenty with VisibilityInterface code

melroy89:
Let's avoid using `self::` where possible, so it's much more clear what kind of interface these enums belong to. Just like the rest of the PHP.

This was most likely introduced with this API PRs, but the code is now no longer consistent with the existing code (see `restore()` method below for example). Hence my PR.


nobodyatroot (commented):
Agreed, self-commenting code is always a good thing.

nobodyatroot (approved):

PR #54, opened 2023-10-16
Restore danger buttons to user purge actions

nobodyatroot:
The danger button class assignments were inadvertently removed in one of the kbin commits before the fork, this PR reverts this so admins don't accidently click on the wrong button.


melroy89 (approved):
Agreed. These are actions that should the user and/or admin should pay extra attention to.

Issue #35, opened 2023-10-14
Admin settings page submission and "other" user actions cause reroute to /all


nobodyatroot:
This bug is also present in the latest kbin codebase, but submitting/saving from the admin page as well as sometimes subscribing to magazines that don't already exist cause the user to be rerouted to the /all page. While not a 500 error, it's still just as irritating. It seems like the code is defaulting to /all instead of just returning the user to the page they were already on.

nobodyatroot:
Interestingly, I cannot replicate this behavior on a fresh build using the docker environment. I'm almost starting to wonder if this is related to something in the bare metal nginx configuration. However, I know this use to work without issue so something must have changed at some point in time causing this "bug" to happen. It may be be a simple nginx config change to remedy, but further investigation is necessary.

nobodyatroot:
After further investigation, I've tracked down the Nginx setting that was causing this issue...

The controller code (and in multiple places, **not just the admin settings controller**) explicitly calls a function to redirect to the referrer "or home":

`src\Controller\Admin\AdminSettingsController.php`:

```
if ($form->isSubmitted() && $form->isValid()) {
	$this->settings->save($dto);

	return $this->redirectToRefererOrHome($request);
}
```

`src\Controller\AbstractController.php`:
```
protected function redirectToRefererOrHome(Request $request, string $element = null): Response
{
	if (!$request->headers->has('Referer')) {
		return $this->redirectToRoute('front'.($element ? '#'.$element : ''));
	}

	return $this->redirect($request->headers->get('Referer').($element ? '#'.$element : ''));
}
```

The guilty Nginx config line as suggested in the latest revision of the admin guide:

`add_header Referrer-Policy "no-referrer" always;`

Commenting out that line from mbin's Nginx config resolves the issue. @melroy89, I suggest a follow on to https://github.com/MbinOrg/mbin/pull/44... otherwise this `redirectToRefererOrHome` function will need to be updated in order to comply with the latest recommended Nginx security headers in https://github.com/MbinOrg/mbin/blob/main/docs/admin_guide.md.

PR #55, opened 2023-10-16
Remove Nginx referrer policy setting from admin guide

nobodyatroot:
The controller code requires the referrer header to be set for submission redirects to process correctly. This will close https://github.com/MbinOrg/mbin/issues/35.

### DIFF
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -437,7 +437,6 @@ server {
     add_header X-Frame-Options "DENY" always;
     add_header X-XSS-Protection "1; mode=block" always;
     add_header X-Content-Type-Options "nosniff" always;
-    add_header Referrer-Policy "no-referrer" always;
     add_header X-Download-Options "noopen" always;
     add_header X-Permitted-Cross-Domain-Policies "none" always;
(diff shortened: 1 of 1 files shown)


melroy89:
Interesting find. I got those security headers by default included on all my domains and I never experienced any problems with it.. Until now 😆 . Let me check if I can reproduce the problem and solution..

melroy89:
@nobodyatroot  Yup nailed it! This will fix the problem with the redirect.. I will approve to improve the user experience right away..

However, .. this is a security header for a reason. Could we try to implement it differently.. so we can use this header, while also having a working product..?

melroy89 (approved):

nobodyatroot:
> @nobodyatroot Yup nailed it! This will fix the problem with the redirect.. I will approve the PR.
> 
> However, .. this is a security header for a reason. Could we try to implement it differently.. so we can use this header, while also having a working product..?

agreed, it will require some additional work to figure out what the proper configuration should be.

melroy89:
I just found out `same-origin`.. Let's do that in a new PR:

image

PR #56, opened 2023-10-16
Add Referrer-Policy "same-origin"

melroy89:


nobodyatroot (approved):
Yep, better than nothing.

PR #57, opened 2023-10-17
docker guide - remove duplicate system requirements section

e-five256:
It looked like the system requirements section was duplicated in this guide. There was also the same level header above each section. Looking at the history, it might have happened during merging changes from multiple sources.


nobodyatroot (commented):
We should also change the hard coded UID and GID in the [Mbin installation section](https://github.com/MbinOrg/mbin/blob/main/docs/docker_deployment_guide.md#mbin-installation)... I know the UID is supposed to be the mbin user (I think...), but what is group 82 supposed to be? Didn't really seem to matter in my environment, I set both to the mbin user/group.

Change:
`sudo chown 1000:82 storage/media storage/caddy_config storage/caddy_data`

To:
`sudo chown mbin:mbin storage/media storage/caddy_config storage/caddy_data`

Maybe it doesn't matter if the mbin user is part of the Docker group? Not sure.

nobodyatroot (approved):

melroy89 (commented):
> Maybe it doesn't matter if the `mbin` user is part of the `docker` group? Not sure.

No user id 1000 is linked to the first user created under Linux. If that user is not mbin, you might got the wrong permissions to read/write. So the chances that there is a mbin user when using Docker is 0% at this moment in time. And what group id 82 means here, I also don't understand.

melroy89:
@e-five256 I think this PR is now causing regression. Please, change it to: `sudo chown $USER:$USER` instead.

PR #62, opened 2023-10-17
Do not allowed spaces in tags

melroy89:
Do not allow spaces in tags.

It's currently possible to add tags with spaces, these tags are eventually ignored and actually not allowed by AP protocol. Most of the cases users can use `camelCase` or `PascalCase` to add tags with multiple words in the same tag. This PR will trigger on a space and insert the tag.

Thank you @simonrcodrington for the contribution. I know you are currently busy, so here is my PR using your code ;).

### DIFF
--- a/src/Form/DataTransformer/TagTransformer.php
+++ b/src/Form/DataTransformer/TagTransformer.php
@@ -10,7 +10,7 @@ class TagTransformer implements DataTransformerInterface
 {
     public function transform($value): ?string
     {
-        return $value ? implode(',', $value) : null;
+        return $value ? implode(' ', $value) : null;
     }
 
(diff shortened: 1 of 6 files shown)


simonrcodrington (approved):

nobodyatroot (approved):

melroy89:
I will very briefly check this patch on my instance. Just to be sure. Then I will merge it.

simonrcodrington:
Thanks for that one. That should fix up the space issues on tags, so at least visually it breaks when you press space on the keyboard now (where beforehand it looks like it's going to work but when it gets submitted only the first part of the word was preserved)

Cheers

PR #63, opened 2023-10-17
Add footer to sidebar on mobile

melroy89:
Add footer to sidebar on mobile. This fixed the issue with infinite scrolling on mobile! 

This is a good solution for now, we are agile, so if we find a better way of showing the footer on mobile with infinite scrolling we can always improve further I think.

Once again thank you @simonrcodrington for the contribution. This PR is for you!

### DIFF
--- a/templates/layout/_sidebar.html.twig
+++ b/templates/layout/_sidebar.html.twig
@@ -131,3 +131,6 @@
             })|raw }}</p>
     </div>
 </div>
+
+{% include 'layout/_topbar.html.twig' %}
+{% include 'layout/_footer.html.twig' %}
--- a/templates/layout/_sidebar.html.twig
+++ b/templates/layout/_sidebar.html.twig
@@ -132,5 +132,4 @@
     </div>
 </div>
 
-{% include 'layout/_topbar.html.twig' %}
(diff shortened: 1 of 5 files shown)


templates/layout/_sidebar.html.twig
```diff
@@ -130,4 +130,7 @@
                 '%link_end%': '</a>'
             })|raw }}</p>
     </div>
-</div>
\ No newline at end of file
+</div>
+
+{% include 'layout/_topbar.html.twig' %}
```
melroy89, line 135: Top bar as well?

simonrcodrington: This should be fine as it keeps the layout consistent. I'll mark this this one as approved and we can make more changes down the track if needed

simonrcodrington (approved):

templates/layout/_sidebar.html.twig
melroy89: It doesn't seem to do anything. Even if you enable topbar, it will show at the top (hence the name), but it doesn't got inserted in the footer. I will remove this line I think.

nobodyatroot (approved):
Tested on my instance, looks good!

PR #65, opened 2023-10-18
sidebar - remove instances placeholder

e-five256:
In the federation section of the sidebar is an "Instances:" entry which, as far as I can tell, does not get used. My [thought had been](https://codeberg.org/Kbin/kbin-core/issues/968#issuecomment-1033431) this was a placeholder to list either all instances or at least ones you are looking at maybe? But now that the federation list page was added, I'm not sure what purpose this serves. 

This removes it so it's less confusing to users. Maybe the translation should be kept in case it becomes useful later? As far as I could tell there weren't tests for this part of the sidebar to remove either. But I could definitely use a spot-check on whether this makes sense, isn't actually in use via something like mercure that I might not realize, and doesn't break things.

(My current IDE appears to be a little opinionated on trailing whitespace and line endings, I'm not sure if that's a good thing or a bad thing.)


nobodyatroot:
It looks to be a placeholder that has since been implemented by the federation page. Instead of removing it, should we use it to link to the federation page for those too lazy to scroll down to the footer? I'm indifferent about it either way and I honestly wonder how many users know about that menu in the first place. :-)

nobodyatroot (approved):

PR #77, opened 2023-10-19
fix inconsistent first/last rounding edges for settings row

asdfzdfj:
saw cooperaj mentioned back in kbin matrix room and decided to investigate

turns out the rounded bottom styling rules of setting row is hardcoded by its position, and it's possible to have this index be out of sync: there's an extra checkbox row for enabling dynamic list if mercure is enabled.

I wrapped the settings rows in a container by its section and changed styling rules to be based on `:first-child`/`:last-child` instead, which should be more robust against adding/removing setting rows.  
also I moved these rules to be in one place/file (previously the rule that define rounded top and bottom were in a separate files).

before:
image

after:
image

### DIFF
--- a/assets/styles/components/_settings_row.scss
+++ b/assets/styles/components/_settings_row.scss
@@ -6,13 +6,17 @@
   width: 100%;
   background: var(--kbin-sidebar-settings-row-bg);
 
-  &:nth-child(7),
-  &:nth-child(13),
-  &:nth-child(15) {
-    .rounded-edges & {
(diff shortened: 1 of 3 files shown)


melroy89:
Maybe add a before & after screenshot? That help others to visualize the change better. thx!

asdfzdfj:
added before/after screenshot as requested.

melroy89 (approved):
Clear! Thanks.

PR #82, opened 2023-10-20
List our unique features

melroy89:
List our features we have after only 6 days of development. This will be a living document and will be updated later down the line.


nobodyatroot (approved):
Good first cut, obviously we'll be adding more as time goes on.

melroy89:
Exactly, and any non feature might be removed again in the future.

PR #87, opened 2023-10-20
Adjust User class constructor to support type field during object instantiation

nobodyatroot:
This PR adjusts/updates the `User` class constructor for cleaner code when instantiating a `User` object. This was motivated by updating the user fixture code to support the new user `type` field requirement that cannot be set to null. This will fix the regression currently present in the user fixture code that is used to generate random data during development and functional testing.

### DIFF
--- a/tests/FactoryTrait.php
+++ b/tests/FactoryTrait.php
@@ -84,20 +84,22 @@ trait FactoryTrait
             'username' => 'adminUser',
             'password' => 'adminUser123',
             'email' => 'adminUser@example.com',
+            'type' => 'Person',
         ];
 
         yield [
             'username' => 'JohnDoe',
             'password' => 'JohnDoe123',
             'email' => 'JohnDoe@example.com',
+            'type' => 'Person',
         ];
     }
 
-    private function createUser(string $username, string $email = null, string $password = null, $active = true, $hideAdult = true, $about = null): User
+    private function createUser(string $username, string $email = null, string $password = null, string $type, $active = true, $hideAdult = true, $about = null): User
     {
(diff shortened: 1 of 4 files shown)


tests/FactoryTrait.php
```diff
@@ -84,20 +84,22 @@ private function provideUsers(): iterable
...
         yield [
             'username' => 'JohnDoe',
             'password' => 'JohnDoe123',
             'email' => 'JohnDoe@example.com',
+            'type' => 'Person',
         ];
     }
 
-    private function createUser(string $username, string $email = null, string $password = null, $active = true, $hideAdult = true, $about = null): User
+    private function createUser(string $username, string $email = null, string $password = null, string $type, $active = true, $hideAdult = true, $about = null): User
```
melroy89, line 98:
```suggestion
    private function createUser(string $username, string $email = null, string $password = null, string $type = 'Person', $active = true, $hideAdult = true, $about = null): User
```

This will set a default value as input parameter.

tests/FactoryTrait.php
```diff
@@ -84,20 +84,22 @@ private function provideUsers(): iterable
...
     {
         $manager = $this->getService(EntityManagerInterface::class);
 
-        $user = new User($email ?: $username.'@example.com', $username, $password ?: 'secret');
+        $user = new User($email ?: $username.'@example.com', $username, $password ?: 'secret', $type ?: 'Person');
```
melroy89, line 102: Assuming you set a default above, you can just use this I think:

```suggestion
        $user = new User($email ?: $username.'@example.com', $username, $password ?: 'secret', $type);
```

melroy89 (approved):

PR #88, opened 2023-10-20
Small docker improvements and guide

melroy89:
- Docker v2 rename Docker image to `mbin`, updating guide and composer file (but leave the rest alone for now)
- Setting `POSTGRES_VERSION` env variable as the value in Docker Compose (keep the default/fallback v13), that environment variable is coming from the `.env` file.
- Remove unused forgejo directory (we're using GitHub Actions now)


nobodyatroot:
> * Upgrading PostgreSQL image to v16

we may want to put a "gotcha" on this, though. upgrading to 16 on any existing installs is going to break the db environment unless we introduce a migration step... somehow, although, i'm not sure (and haven't tried) how easy that's going to be. fresh installs should be ok, however.

also, the env files will need to be updated to call out the right DB version...and actually, it's set to 14 in the example env but the docker-compose file (as is currently checked in) uses a 13 image.....wth.

```
POSTGRES_VERSION=**16**
DATABASE_URL="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}/${POSTGRES_DB}?serverVersion=${POSTGRES_VERSION}&charset=utf8"
```

melroy89:
> > * Upgrading PostgreSQL image to v16
> 
> we may want to put a "gotcha" on this, though. upgrading to 16 on any existing installs is going to break the db environment unless we introduce a migration step... somehow, although, i'm not sure (and haven't tried) how easy that's going to be. fresh installs should be ok, however.
> 
> also, the env files will need to be updated to call out the right DB version...and actually, it's set to 14 in `.env.example_v2` but the docker-compose file (as is currently commited) uses a 13 image.....wth.
> 
> ```
> POSTGRES_VERSION=16
> DATABASE_URL="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@${POSTGRES_HOST}/${POSTGRES_DB}?serverVersion=${POSTGRES_VERSION}&charset=utf8"
> ```

That is indeed now fixed as well..

nobodyatroot (approved):
Cool! And TIL about being able to put the env vars in the docker-compose file, very convenient.

PR #90, opened 2023-10-21
Improve error warning log pollution (Part 1)

melroy89:
- Adding actor null checks
- Link array null check
- Refactoring some code (moving methods around, adding types, etc.)


nobodyatroot (approved):
So far so good, I've been running it for about an hour and a lot less spammy already. :-)

melroy89:
Come on.. Unit-test please don't do this to me now:

image

PR #93, opened 2023-10-21
Cleaning up v1. Make v2 setup the default Docker setup.

melroy89:
- Remove Docker v1 setup
- Make Docker v2 the default, and move out of the v2 folder
- Clean-up the documentation of Docker
- Fix bare metal documentation: it should use `.env.example` as a template, and _not_ the Docker dot env file.


nobodyatroot (dismissed):
Followed new steps from scratch, docker environment is working as it was prior to v2 folder removal.

nobodyatroot (approved):

PR #95, opened 2023-10-21
Figure out thumbnail issues possibly related to max_image_bytes 

TheVillageGuy:
I don't know why this exists, as you can set a max upload size in PHP and it's been driving crazy from day one. This is the value I have on my instance


melroy89:
_Note:_ This is 40MB for the record

nobodyatroot:
Is it possible to get the max upload size from the PHP environment so as to not hard code it... and only have to change it in one place - the `php.ini` file?

nobodyatroot:
Since this is now labeled as a WIP, I converted the PR to draft status.

melroy89:
Drupal code (GPL):

```
// Returns a file size limit in bytes based on the PHP upload_max_filesize
// and post_max_size
function file_upload_max_size() {
  static $max_size = -1;

  if ($max_size < 0) {
    // Start with post_max_size.
    $post_max_size = parse_size(ini_get('post_max_size'));
    if ($post_max_size > 0) {
      $max_size = $post_max_size;
    }

    // If upload_max_size is less, then reduce. Except if upload_max_size is
    // zero, which indicates no limit.
    $upload_max = parse_size(ini_get('upload_max_filesize'));
    if ($upload_max > 0 && $upload_max < $max_size) {
      $max_size = $upload_max;
    }
  }
  return $max_size;
}

function parse_size($size) {
  $unit = preg_replace('/[^bkmgtpezy]/i', '', $size); // Remove the non-unit characters from the size.
  $size = preg_replace('/[^0-9\.]/', '', $size); // Remove the non-numeric characters from the size.
  if ($unit) {
    // Find the position of the unit in the ordered string which is the power of magnitude to multiply a kilobyte by.
    return round($size * pow(1024, stripos('bkmgtpezy', $unit[0])));
  }
  else {
    return round($size);
  }
}
```

TheVillageGuy:
Notes for self:

php.ini
    upload_max_filesize = 10M
    post_max_size = 10M

imageManager.php
	// add more image types (why aren't they allowed in the first place?)
    public const IMAGE_MIMETYPES = ['image/jpeg', 'image/jpg', 'image/gif', 'image/png'];
    public const IMAGE_MIMETYPE_STR = 'image/jpeg, image/jpg, image/gif, image/png';
	
	//default = 6000000
    public const MAX_IMAGE_BYTES = 40000000;
	
	
	 public function store(string $source, string $filePath): bool
    {
        $fh = fopen($source, 'rb');

        try {
            if (filesize($source) > self::MAX_IMAGE_BYTES) {
                throw new ImageDownloadTooLargeException();
            }
			
			$response = $this->httpClient->request(
                'GET',
                $url,
                [
                    'timeout' => 5,
                    'max_duration' => 5,
                    'headers' => [
                        'Accept' => implode(', ', self::IMAGE_MIMETYPES),
                    ],
                    'on_progress' => function (int $downloaded, int $downloadSize) {
                        if (
                            $downloaded > self::MAX_IMAGE_BYTES
                            || $downloadSize > self::MAX_IMAGE_BYTES
                        ) {
                            throw new ImageDownloadTooLargeException();
                        }
                    },
                ]


max_duration 5? If this is for downloading 35MB pictures, 5 seconds isn't always going to cut it, especially on busy servers

current situation seems to be:
public const MAX_IMAGE_BYTES = 40000000; seems to be for both uploads and downloads.

A picture post is created. Upload fails if php limits are exceeded, then if MAX_IMAGE_BYTES is exceeded.
If the upload succeeds and the creation is done, it is announced to other instances. They try to download it and fail, likely due to size and possibly time limits. Then they store nothing and don't go for the thumbnail instead.

Thumbnails seem to be created when accessing the creation. This can lead to crashes on very high res (not necessarily large) files, (jpg bombs) because multiple thumbnails can be created at the same time due to long creation time, causing the server to run out of memory. 

Possible solution:
remote instance checks filesize before downloading. Can this information be announced? Can thumbnail location be announced?
If the file is too big, only download the thumbnail*, avoiding possible thumbnail creation failure (due to PHP memory limit). Or, just always download the thumbnail? (Thumnail filename=picture filename anyway..)
if the file is not too big download full file and depending on always thumbnail downloading or not, create a thumbnail right away**, instead of when accessing.

Option: adjust max_duration relative to filesize, assuming filesize can be announced?

Where does $downloadsize come from? Is it always available? E.g. can $downloadsize be aquired somehow before commencing get? This would be a way of filesize announcement. Can max_duration be adjusted during request?

* Can announcement wait until thumbnail is fully created? This would be good for the instance creating the thumbnail regardless, as thumnail creation is intense. Or can the response to an announcement be delayed? (not the ideal solution)

** preferably locking thumbnail creation, so it can only be done once to avoid server crashes.

Is the lightbox addition involved in any of this? Not likely.

More information can be aquired by adjusting limits for tests. Are ImageDownloadTooLargeException() thrown into prod.log, always?

melroy89:
You sure it's in seconds?

melroy89:
Any process or help needed here? I would personally like to see a draft PR not being open for more than a week or so.

TheVillageGuy:
I'm not going to start on this until I have returned from my holiday, so
I'll close it for now

PR #99, opened 2023-10-22
Check if ENABLE_ACL is set before variable access

nobodyatroot:
$ENABLE_ACL is read even when it's not set, leading to shell error (sh: 1: unknown operand) during docker entry. PR checks to see if $ENABLE_ACL exists before evaluation.

### DIFF
--- a/docker/docker-entrypoint
+++ b/docker/docker-entrypoint
@@ -51,8 +51,8 @@ if [ "$1" == "php-fpm" ] || [ "$1" == "php" ] || [ "$1" == "bin/console" ]; then
         bin/console doctrine:migrations:migrate --no-interaction
     fi
 
-    if [ -n "${ENABLE_ACL+1}" ]; then
-        if [ $ENABLE_ACL = 1 ]; then
+    if [ -n "$ENABLE_ACL" ]; then
+        if [ "$ENABLE_ACL" -eq 1 ]; then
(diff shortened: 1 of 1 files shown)


docker/docker-entrypoint
```diff
@@ -51,9 +51,15 @@ if [ "$1" == "php-fpm" ] || [ "$1" == "php" ] || [ "$1" == "bin/console" ]; then
         bin/console doctrine:migrations:migrate --no-interaction
     fi
 
-    if [ $ENABLE_ACL = 1 ]; then
-        setfacl -R -m u:www-data:rwX -m u:"$(whoami)":rwX var
-        setfacl -dR -m u:www-data:rwX -m u:"$(whoami)":rwX var
+    if [ -n "${ENABLE_ACL+1}" ]; then
```
melroy89, line 54:
```suggestion
    if [ -n "$ENABLE_ACL" ]; then
```

This will do it as well, right?

docker/docker-entrypoint
```diff
@@ -51,9 +51,15 @@ if [ "$1" == "php-fpm" ] || [ "$1" == "php" ] || [ "$1" == "bin/console" ]; then
...
+        if [ $ENABLE_ACL = 1 ]; then
```
melroy89, line 55: I would change this to:

```suggestion
        if [ "$ENABLE_ACL" -eq 1 ]; then
```

melroy89 (approved):

PR #101, opened 2023-10-22
Update FAQ.md

TheVillageGuy:
added a line break to avoid confusion and small grammar fix


TheVillageGuy:
another change needs to made, what's the best way to force this?

( 
sudo /usr/sbin/rabbitmq-plugins enable rabbitmq_management
)
is the line it should be to work

FAQ.md
```diff
@@ -80,7 +80,8 @@ Execute: `sudo rabbitmqctl status`, that should provide details about your Rabbi
 
 Ensure you do not see any connection errors in your `var/log/prod.log` file.
 
-Talking about plugins, we advise to also enable the `rabbitmq_management` plugin by execute: `rabbitmq-plugins enable rabbitmq_management`.
+Talking about plugins, we advise to also enable the `rabbitmq_management` plugin by executing: 
+`rabbitmq-plugins enable rabbitmq_management`.
```
TheVillageGuy, line 84: sudo /usr/sbin/rabbitmq-plugins enable rabbitmq_management

melroy89: Fixed.

melroy89:
`/usr/sbin/` is not needed, that is part of the default path of the root user.

TheVillageGuy:
> `/usr/sbin/` is not needed, that is part of the default path of the root user.

You assume I run it as root, but I don't

melroy89:
> > `/usr/sbin/` is not needed, that is part of the default path of the root user.
> 
> You assume I run it as root, but I don't. It's not in my path.

I mean you run via `sudo` right? It should be part of that.

melroy89:
Check your `/etc/sudoers` file:

```
Defaults        secure_path="/usr/local/sbin:/usr/local/bin:/usr/sbin:/usr/bin:/sbin:/bin:/snap/bin"
```

TheVillageGuy:
You are right, I was blind

melroy89 (approved):

PR #100, opened 2023-10-22
Improve error warning log pollution + composer recipes (Part 3)

melroy89:
- Recipe updates (doctrine-migrations-bundle, flystem-bundle, monolog-bundle, stimulus-bundle, translation)
- **Enable log rotation** on production (max 7 files, one per dag)
- Enable log rotation in _development_ mode as well (max 10 files)
- **Full stacktraces** in development mode, strict logging on action level, buffer size, and deprecation to stderr
- Change default **development log level** from `debug` to `info`, with a remark the developer can always enable `debug`. Warning: `debug` is very very verbose, hence `info` as default.

PR #112, opened 2023-10-23
Change from info to debug level

melroy89:
Reduce log pollution when running "info" by changing from `info` to `debug` logging.

The reason is that running `info` can help debugging issues in other parts of the code like federation issues of messenger. The current `info` lines in Mbin cause log pollution. If the development wants, you can always enable debug logging.


melroy89:
Note: in PR: https://github.com/MbinOrg/mbin/pull/100 I plan to also increase the default log level from debug to info. So you don't see these messages in development mode, unless the developer actually set the logging to `debug`. If the user wants to debug..

nobodyatroot (approved):

PR #113, opened 2023-10-24
adding display and overflow properties to images with alt text

e-five256:
mainly focused on user avatars, covers, and magazine icons

this is a followup to #81 

wanted to make a more generic approach to images with alt text rather than handle it on a case by case basis. It semi-reverts the changes from that PR in that, rather than reuse the `user-avatar` class everywhere (left it alone on `user_image_component`), makes a generic image class for handling the display of them.

At first I put it in `components/_main.scss`, which is why that file has changes, but that felt a little off as it isn't related to a single component, so I opted to make a new layout file to include

I wanted to use `display: block` in places where it made sense but upon doing that to everything that wasn't an inline `twig` file, I realized that made things worse as they were no longer centered (such as in the sidebar). I feel like I might be overusing `inline-block` but it appears to work in all use cases.

I was going to make a before and after of the images when loaded for comparison to show nothing changed but that felt a bit overkill as there wouldn't be much to compare, but I could still do it if people are concerned. I did check all images against computed css of regular mbin instances and didn't notice any changes besides the extra properties added by this change.

alt_overflow_beforeafter


nobodyatroot (approved):
Nice! I like that it's now a generic implementation.

PR #121, opened 2023-10-24
adding sidebar setting for controlling container width

e-five256:
the proposal here is to add a new sidebar setting to allow users to control the width of the content

1. wanted to get opinions on whether people thought this was a worthwhile change to have
2. ~~the icons are placeholders, don't think they're very descriptive of what the setting does but didn't want to spend time on it if this wasn't a change people wanted (similar to this, alt text is placeholder and untranslated if changes are needed)~~
3. ~~the sizes were meant to reflect monitor resolutions (max, 8k, 4k, 2k, 1080) but I can see going a different path, maybe just specific widths or tied to something else. The 2k appears better on my 1080 monitor than the default setting was so maybe some tweaking is necessary.~~ changed sizing to max (100%), auto (85%), fixed (1360px)
4. the settings don't appear on mobile as the width is always 100% at mobile resolutions

**Alternative:**

Maybe instead of all these settings it could be brought down to something like: Max, Auto, Fixed. Max being 100%, auto being... something, either less like 80% or use `@media` to set a bunch of automatic max widths up, and then fixed being what there is now.

image

<details>
<summary>These are the old implementation images</summary>

examples:

2k width setting
image
max width setting
image
mobile
image

</details>


e-five256:
here's an alternate implementation with it as a max (100%), auto (85%), and fixed (1360px as it is now) as an enum setting (2nd option under the general header). It also goes away on mobile (and thanks to asdfzdfj's change, the rounded edges can handle that as long as it's not the top or bottom option)

image

simonrcodrington:
I think having that in the settings section like your updated comment is probably the best approach. 

We could have 3 options for that setting along with a description eventually, but if we get more options we might have to switch to a drop-down or something (that sidebar wasn't really made for multiple inline options at once it seems)

Having this seems like a decent feature, I'm primarily only using the mobile profile so I'll defer feedback to others for now

TheVillageGuy:
I think the current layout situation is fine as is, of course improvement is always welcome. As said in the chat if I'd like the subs on by default so that users can get used to it right from the start. I'd also like them separate by default for the same reason. However, providing it'll just be a little button to click to pop them out and then an arrow button to move it to the other side, by default in the regular sidebar would be good compromise. This feature should not be kept hidden in the settings, even from existing users.

e-five256:
I think most of your comment was feedback for the subscription widget that @BentiGorlich added, correct? I think he was looking for feedback so I'm sure there can be iterations made, but I'm not sure that this PR is related, unless your first part is saying you don't believe this change should be made (adding a sidebar setting to allow for users to select options to have less deadspace on either side of the web ui)

TheVillageGuy:
Correct, I thought this was related because you invited me from chat, I
should have paid better attention.

To this pr I have nothing to add, it looks like a great idea to me,
providing there's enough room for both sidebars

melroy89:
This PR is heavily correlated with this other PR IMO: https://github.com/MbinOrg/mbin/pull/161 . In fact I propose that you two work together in a single PR maybe? If we want to enable this additional sidebar by default, I think the default width needs to be adjusted maybe as well. 

What are your thoughts?

e-five256:
I prefer to keep PRs as small as possible to make them easier to review. I agree that this PR helps alleviate some of the issues with moving to a three column layout, but I also believe this change helps everyone regardless of that, and I'm not sure what else I have to do to get this PR approved, as the only feedback I have gotten is that it looks good without approvals

BentiGorlich:
I'd say I consider my PR blocked until this is on main and I've tested them together. 
As to the missing approvals: you actually didn't request anyone's review other than [simonrcodrington](https://github.com/simonrcodrington) :D

e-five256:
I didn't add him, debounced did 😆 but there are five of you here now, yet still no reviews 😝 really not sure what is needed to move this forward. is it because it's on my fork? it predates me becoming a member, so I could close it and move it to a branch on this repo so it's easier to test if that's what's holding it up.

BentiGorlich:
I just didn't have time and wanted to work on my own stuff :D 
You can request the review of people on your pr. That way they get notified. I always add everbody to my PRs 😅

Am 29. Oktober 2023 18:34:59 MEZ schrieb e-five ***@***.***>:

nobodyatroot (approved):
Sorry, I wanted to give someone else a chance to panic merge. :-)

TheVillageGuy:
What's going on?  If you feel you're waiting for an approval too long just remind the community on matrix. Personally I don't really understand what this pr does so I didn't look into it and currently I don't have the time. If I'm completely misreading the room here, just ignore this message

PR #124, opened 2023-10-25
Fix App_Env = dev loading in docker

CocoPoops:
Fix App_Env = dev loading in docker


nobodyatroot (approved):
Confirmed this fixes dev mode launch

melroy89:
Ow so auto loader is needed? But not for prod?? How does composer work with this flag.

PR #130, opened 2023-10-25
Explain increasing verbosity level of monolog 

melroy89:
When in development mode; add an optional step if you want to increase the verbosity level to `debug`.

### DIFF
--- a/config/packages/monolog.yaml
+++ b/config/packages/monolog.yaml
@@ -48,8 +48,8 @@ when@prod:
                 type: fingers_crossed
                 action_level: error
                 handler: nested
-                buffer_size: 50 # How many messages should be saved? Prevent memory leaks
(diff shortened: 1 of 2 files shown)


config/packages/monolog.yaml
```diff
@@ -48,8 +48,8 @@ when@prod:
                 type: fingers_crossed
                 action_level: error
                 handler: nested
-                buffer_size: 50 # How many messages should be saved? Prevent memory leaks
                 excluded_http_codes: [404, 405]
+                buffer_size: 50 # How many messages should be saved? Prevent memory leaks
```
melroy89, line 52: I just moved these lines around, so it's the same as a Symfony from scratch. Nothing special here.

nobodyatroot (approved):

PR #133, opened 2023-10-26
Ensure doctrine migration (and ACL set if enabled) only happens once when php-fpm is started

nobodyatroot:
Docker-entrypoint fix to check and make sure doctrine migration only happens when the `php-fpm` command is executed (will only happen once during stack startup). Also moved ACL set/check under same condition since it should only happen once during startup.

### DIFF
--- a/docker/docker-entrypoint
+++ b/docker/docker-entrypoint
@@ -28,28 +28,28 @@ if [ "$1" == "php-fpm" ] || [ "$1" == "php" ] || [ "$1" == "bin/console" ]; then
         composer dump-env dev
     fi
 
-    if [ "$1" == "php-fpm" ]; then
-        echo "Waiting for db to be ready..."
-        ATTEMPTS_LEFT_TO_REACH_DATABASE=60
-        until [ $ATTEMPTS_LEFT_TO_REACH_DATABASE -eq 0 ] || DATABASE_ERROR=$(bin/console dbal:run-sql "SELECT 1" 2>&1); do
-            if [ $? -eq 255 ]; then
-                # If the Doctrine command exits with 255, an unrecoverable error occurred
-                ATTEMPTS_LEFT_TO_REACH_DATABASE=0
-                break
-            fi
-            sleep 1
-            ATTEMPTS_LEFT_TO_REACH_DATABASE=$((ATTEMPTS_LEFT_TO_REACH_DATABASE - 1))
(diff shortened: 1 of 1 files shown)


docker/docker-entrypoint
```diff
@@ -28,40 +28,42 @@ if [ "$1" == "php-fpm" ] || [ "$1" == "php" ] || [ "$1" == "bin/console" ]; then
...
-        if [ $? -eq 255 ]; then
-            # If the Doctrine command exits with 255, an unrecoverable error occurred
-            ATTEMPTS_LEFT_TO_REACH_DATABASE=0
-            break
-        fi
-        sleep 1
-        ATTEMPTS_LEFT_TO_REACH_DATABASE=$((ATTEMPTS_LEFT_TO_REACH_DATABASE - 1))
-        echo "Still waiting for db to be ready... Or maybe the db is not reachable. $ATTEMPTS_LEFT_TO_REACH_DATABASE attempts left"
-    done
+    if [ "$1" == "php-fpm" ]; then
```
melroy89, line 31: The if needs to be lower I think. We still want to have a blocking waiting before those services continue to boot further.  That was fine to have.

I think the `if` should only be added around the migration part. So before line 53, and also directly the `fi` after around line 56.

melroy89 (changes requested):
We should keep the waiting for the DB to be ready for all services, just don't perform a db migration on other services.

melroy89:
Acl I don't now to be honest. Should that only happen on php?

nobodyatroot:
> Acl I don't now to be honest. Should that only happen on php?

I was thinking keeping with php-fpm since it should only happen once like the doctrine migration... dunno if it matters.

melroy89:
> > Acl I don't now to be honest. Should that only happen on php?
> 
> I was thinking keeping with php-fpm since it should only happen once like the doctrine migration... dunno if it matters.

Sure then keep that together

melroy89 (approved):
I assume you tested it locally? I'm going to sleep now 😴😉

nobodyatroot:
> I assume you tested it locally? I'm going to sleep now 😴😉

yep, i'll test it again in both `prod` and `dev` modes before merging.

melroy89:
I will test it today. Let's see if migration only happens once with a new PostgeSQL database

melroy89:
Yep works as expected. Thanks @nobodyatroot

PR #134, opened 2023-10-26
Include bundles directory to copy over necessary files for API in docker environment

nobodyatroot:
Bundles directory wasn't being copied over to `www` container in docker environment, this PR updates the dockerfile to include the proper copy command from the build environment.


e-five256:
Can confirm this fixed swagger api for my local docker env

lilfade (approved):

PR #135, opened 2023-10-26
Deprecation: Remove Awards & Badges

lilfade:
This commit removes more unimplemented code from the code base and database.


melroy89:
Wow ok.. Could you maybe add why this was maybe considered used for? Maybe it's still something we want to work on as a community?

Cardano was clear to me, we just removed that. But what are badges? What was the idea or rationale behind badges?

lilfade:
No idea for that you'd have to ask Ernest as he never answered when I asked. From what I can tell awards and badges are a copy of Facebook badges for fans and awards like reddits post awards. Neither was being used and with the amount of in general unused code this is one of them so I vote to remove it or have someone take it over and finish it out if we agree it's needed.

An example for awards, you make a post on your home server and the community home is on server x and I want to give you an award. Which server profits from the donation? Does the award get federated to every subscribing server?

Badges I can see if they worked but they seem to be half implemented as a duplicate of tags to a degree.

My $0.02.

nobodyatroot:
There's little value added with these "features" and is not really useful across the federation, I'm 100% on board with removing them.

asdfzdfj:
(quite a bit of my own opinion mixed down below)

dunno where I heard this from but apparently badges was meant (or has been understood) to be something similiar to (post level) reddit flairs where it's most common application seems to be a kind of per sub/forum submission categorization. and additionally for reddit it can be used for submission filtering within sub/forum, like to either only show entries containing that flair or to exclude them from the listing.

take for (non reddit) example: [Raddle's /f/egg_irl](https://raddle.me/f/egg_irl), it's those little boxes with short text besides the title. although for them it appears to just simply exist and cannot be used for filtering at the moment.

image

If we are to keep the badges then I think getting them to work just to this point would be a good start (although this would be a site specific features as I'm not too sure how to translate it to something that could be AP federated).

-----

awards on the other hand though can get axed for all I care, from what I can guess from global award page it looks like a top tier clout chasing feature and isn't much useful.

lilfade:
Seems a general consensus so far on awards. I like your angle of the badges. Do we have any ideas on how to go about implementation I assume this would require a bit of the UI devs input here for possible options. 

As for propagating to other sites that would indeed be a bit of a kerfuffle as we have to send more data to and from and then updates for these, likely getting the "flairs" to update to all places. Though that might touch on relaying and I'm sure that would need another topic/discussion itself.

melroy89:
I like the idea of _badges/flares_ still in that case, let's see how we can shape this (we could use sub-tags to make it AP compatible). For instance we can leverage the AP tags, but instead of showing all the tag content within the magazine, we could leverage badges/flares to make it easier to filter if a magazine has multiple tags. My feeling is that we can definitely make this happen.

And I agree on the removal of _awards_, that can be fully removed from the code.

asdfzdfj:
> Seems a general consensus so far on awards. I like your angle of the badges. Do we have any ideas on how to go about implementation I assume this would require a bit of the UI devs input here for possible options.

for full implementation details I think we might need a place to fully discuss and come up with a feature specification of sorts, maybe hash out some more details before starting serious implementation

as for some basic ideas I have floating around right now:

### 0. get it working first

- overall goal is to get a small text badge to display besides the entry title, similar to how one might use `[this construct]` in the title to categorize their entries.
- each magazine can have a set of predefined badges that can be set by magazine mod/owner
- the current badge settings UI in magazine panels is rather barebones, and I think changing it to works in a similiar way to adding tags when making entries would be an improvement
- when user is making a new entry, they could pick which badge (or up to 2) to apply to the entry, using autocomplete textbox that offers preset badges for selection should be sufficient. note that adding badges is entirely optional.
- (my opinion) maybe impose some restriction on badge text length, generally it should be succinct, about 40-50 chars should be plenty.

### 1. make it meaningful

- find a way to federate these badges over to other mbin instances, otherwise these badges is only useful on the origin instance and not so useful in the grand scheme
  - maybe these could be put into `tag` section of AP representation of an entry, perhaps something like this:
    ```jsonc
    {
      "@context": { /* ... */ },
      "tag": [
        {
          "type": "Badge",
          "name": "badge text",
          "magazine": "https://mbin.instance/m/magazine"
        }
      ]
    }
    ```
  - using native activitystream or existing well known json-ld contexts and constructs is preferred...
  - ...but do prepare to make our own extension
    - honestly, this might make our lives easier and more preferable that trying to gymnastic existing vocabs and properties to fit our semantic if a perfect fit term couldn't be found
    - come up with a common base indentifier for mbin specific json-ld context and extensions,
      - perhaps we might be able to leverage our github pages domain as namespace base if this additional github dependence is deemed acceptable
      - other projects tends to use their landing/main project domain for this, and we don't have any for the moment
- allow basic filter to either list entries containing a certain badge, or to remove them from listing.

### 2. add some possible useful enhancements

- allow badges with fully custom text, with setting to enable per magazine
  - sometimes some sub/forums doesn't really use badges/flairs for categorization like previously described,  
    and on the occasion that it comes up the texts within is most likely for fun, for actual "flair".
- allow multiple badges to be applied to one entry
  - configurable upper limit either per instance or magazine
  - tread carefully: (my opinion) overall badges should be succinct. this is not a substitute or another space for hashtags
- allow more complex entry filtering criteria with badges
  - selecting or excluding multiple badges at once
  - selecting with one badges not if another badges is also applied
- colorable badges for easier reading/grouping

PR #140, opened 2023-10-26
Improve error warning log pollution + dev setup + composer update (part 5)

melroy89:
We are at part 5 already of fixing errors, warning,  deprecation messages/clean-ups inside our code base:

- Ran: `composer update`
- Remove deprecation channel to stderr from prod completely in `monolog.yaml`
- Introduce deprecation channel to stderr in dev 
- ~~Ran composer recipe `symfony/webpack-encore-bundle`~~ Despite it still "worked".. This recipe upgrade caused some strange page-reload behavior. So I rolled-back.
- Fix various deprecation problems in our code-base
- Null check on mercure-url in `event-source.js` (return `null` otherwise)
- Increase mercure reconnect setTimeout to 10s (10000ms) plus `null` check
- README.md improvements (dev setup)
- Contributing.md improved with unit-testing section

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -36,6 +36,30 @@ There are three levels of tests that we distinguish between:
 
 For more info read: [Symfony Testing guide](https://symfony.com/doc/current/testing.html).
 
+#### Unit Tests
+
+- First increase execution time in your PHP config file: `/etc/php/8.2/fpm/php.ini`:
+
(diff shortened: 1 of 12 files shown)


nobodyatroot (approved):
I'll let this run overnight, but I'll approve preemptively since I won't be able to check on it until later tomorrow night... traveling all day tomorrow.

PR #141, opened 2023-10-26
Revert "Limitation of random sidebar content"

melroy89:
I never agreed with [this change of Ernest](https://codeberg.org/Kbin/kbin-core/pulls/881), because this impact small/medium size local instances. This PR in Mbin will revert that PR from Kbin.

This reverts commit 4d81b090d1f62cdb77bbb838c6a05cf6e2e60149.


BentiGorlich:
I really hate that it only takes local content. I agree that this should be reverted :+1:

BentiGorlich (approved):

melroy89:
> I really hate that it only takes local content. I agree that this should be reverted 👍

Exactly, used to get all the content from your instance. Ernest did this "quick fix" (which is still in the code till this day), to get rid of spam I think.

nobodyatroot (approved):
Agreed

PR #143, opened 2023-10-27
Futher increase setTimeout Mercure reconnect

melroy89:
Every second is I think a bit too much. I increased it now to 10 seconds (10000ms).


melroy89:
I will apply this in PR: https://github.com/MbinOrg/mbin/pull/140 instead

PR #146, opened 2023-10-27
fix inline link preview basically imploded when trying open preview

asdfzdfj:
so basically this is a revert/reimplementation of #109, plus some of my personal touches.

the merged changes probably worked nicely when previewing from an entry listing/card, but it seems to cause regression in inline link preview as it basically made the link implodes when trying to open the preview

link-implode-demo

- revert/reimplement changes that #109 set out to do (or at least how I understand what it should do)
- fix inline link preview self destruct when trying to open preview
- change the preview container selection/manipulation to be stimulus target based instead of the old sibling node dancing, which should be more robust and simpler work with, though this does come with some tradeoff: the entry preview is now inside the entry itself instead of after or between the entry card.
- when fetching embed, only wrap embed link when it's image embed as it doesn't make much sense for other type for embed, and can cause some usability issues like suddenly being able to drag a `<video>` element when trying to drag its seek bar.

image  
(the preview is now inside the entry card)

image  
(both entry preview (top) and inline preview (middle, bottom) expanded)

### DIFF
--- a/assets/controllers/preview_controller.js
+++ b/assets/controllers/preview_controller.js
@@ -18,28 +18,21 @@ export default class extends Controller {
     async show(event) {
         event.preventDefault();
 
-        let element = this.element;
+        let container = this.element.nextElementSibling && this.element.nextElementSibling.classList.contains('js-container')
+            ? this.element.nextElementSibling : null;
 
-        if (element.classList.contains('preview')) {
-            element = element.parentElement.previousElementSibling;
-            this.element.remove();
+        if (null === container) {
+            container = document.createElement('div');
+            container.classList.add('js-container');
+            container.style.display = 'none';
+            this.element.insertAdjacentHTML('afterend', container.outerHTML);
         } else {
-            let container = this.element.nextElementSibling && this.element.nextElementSibling.classList.contains('js-container')
-                ? this.element.nextElementSibling : null;
-
-            if (null === container) {
-                container = document.createElement('div');
-                container.classList.add('js-container');
-                container.style.display = 'none';
-                this.element.insertAdjacentHTML('afterend', container.outerHTML);
-            } else {
-                if (container.querySelector('.preview')) {
-                    container.querySelector('.preview').remove();
-                    if (0 === container.children.length) {
-                        container.remove();
-                    }
-                    return;
+            if (container.querySelector('.preview')) {
+                container.querySelector('.preview').remove();
+                if (0 === container.children.length) {
+                    container.remove();
                 }
+                return;
             }
         }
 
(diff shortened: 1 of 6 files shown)


melroy89:
I tried this on my instance and open: https://kbin.melroy.org/m/cat@lemmy.world/t/57388/27th-October-is-National-Black-Cat-Day-UK

Preview button now doesn't work anymore... 
preview_broken

Here another one when the patch was installed (https://kbin.melroy.org/m/piracy@lemmy.dbzer0.com/t/57155/UNPACKED-team-of-UNPACKERS-who-love-to-share-releases-unpacked):

preview_broken_2

melroy89:
Another link on a local magazine (of vinxprox): https://kbin.melroy.org/m/drbboard/t/50852/Icon-Leomard
js_error

melroy89 (changes requested):
I see too much regression and issues happening. I removed the patch again from my instance.

asdfzdfj:
ok new regressions confirmed, made some changes:

### first 2 cases

most likely cargo cult programming moment on my part: basically I kept the `.js-container` class for the preview container target that was present on the original, but that class have 'display: none' style defined and probably meant to be changed by editing `style` property from js, but I was controlling visibility by modulating `.hidden` class instead (and I was making these change on top of another changeset that have the `display: none` here changed/removed)

so technically it *might* be working, but the result was hidden by the class, and I didn't noticed it when tearing off commits and turn it into this branch/PR

right now I opted to remove the `.js-container` class from the container target (personally not too keen on editing style peroperty directly)

### third/vintprox case

this seems to be something different, saw the preview controller complained that it couldn't find the preview container target.  
I *did* add the preview container target element for both entry preview and the inline embed preview, so perhaps try clearing the cache and see if it's still happened, and if that was the case then I think I'm going to need some help identifying which part/component that I've missed.

melroy89 (changes requested):
I still have regression on the preview button inside the thread description of vintprox: `Uncaught (in promise) Error: Missing target element "container" for "preview" controller` (yes I cleared the cache).

asdfzdfj:
I don't think mbin currently allows custom js to be used? sure the field's there, but last time I checked it basically connects to nowhere and isn't being used in templating either (honestly, good).

the more confusing thing is that I somehow managed to [pull vintprox entry over to my instance](https://kbin.asdfzdfj.xyz/m/drbboard@kbin.melroy.org/t/7035/Icon-Leomard) and I couldn't reproduce the missing target problem.

vintprox-no-target-reproc

I could try to include a workaround code that will create the target itself if it couldn't find the target to use, but I honestly think that such code shouldn't be needed in the first place. and at this point I'm contemplating if I should just give up reworking the target-based preview for now and just doing [the bare minimum](https://github.com/asdfzdfj/mbin/tree/fix/preview-link-imploded-immediate) for fixing the problem at hand instead...

melroy89:
You forked the project instead of creating a branch in the MbinOrg/mbin repo. Making it harder to switch branches.

melroy89:
I left your patch on my instance now.. You can see it fails still on my side: https://kbin.melroy.org/m/drbboard/t/50852/Icon-Leomard

asdfzdfj:
some observation that I see from your instance, as the time of writing:

- thankfully this seems to only affect inline preview link, the entry preview seems to be working properly, and for this I think I could afford to add a workaround just for this setup
- inline embed preview link in newer posts/entries seems to be working properly
  - found an interesting case: [this comment](https://kbin.melroy.org/m/opensource@lemmy.ml/t/58425/OSS-Notetaking-App-Notesnook#entry-comment-448182) is having the missing preview target problem, while [this comment](https://kbin.melroy.org/m/opensource@lemmy.ml/t/58425/OSS-Notetaking-App-Notesnook#entry-comment-448723) from the same entry but made later (presumably after you applied the patch) have a working embed preview with the container target present
  - for your instance, it looks like the old posts before the applying patch _never_ gets to be rendered again, like I managed to dug out [this old comment](https://kbin.melroy.org/m/technology@beehaw.org/t/26/Redditors-how-do-you-like-Lemmy#entry-comment-27185) which still have this problem, post this old should get to be rendered again and the embed link would get the missing preview target back.
  - for comparison, I found [this comment](https://kbin.asdfzdfj.xyz/m/196@lemmy.blahaj.zone/t/28/Pirulecy#entry-comment-65) on my instance which is probably one of the the oldest embed link that exist for me, and the new embed link works properly.
- although I don't wish to reassert my belief and/or contradicts you in the process, the closest guess I could think of on why this is the case is likely have something to do with caching rendered markdown body results. you said you cleared the cache, but then the embed links should get rendered again and get their missing preview container back, but this is currently not the case.

anyway I rebased the changes and added the workaround for inline preview to give itself the preview container target if it couldn't find one (still not happy, but better than having to handle every case, which feels like a step back)

melroy89:
Yes, now it is working!

melroy89 (approved):

asdfzdfj:
finally, thank you. hope nothing more breaks.

Issue #148, opened 2023-10-27
Settings menu doesn’t display properly on IOS mobile 


CocoPoops:
**Describe the bug**
Settings menu doesn’t display properly on iOS mobile 

![image](https://github.com/MbinOrg/mbin/assets/7891055/515317a7-c36a-4ada-8520-afdc5cb89ecb)

Smartphone
 - Device: IPhone 13 Pro Max
 - OS: iOS 17.0.3
 - Browser: Safari

melroy89:
Works fine here...Not sure with instance you are on (I will add this to the ticket template). But did you perform a rebuild of all the assets and cleared the cached (as well as clearing the browser cache on your mobile in this case)?

![works_fine_here](https://github.com/MbinOrg/mbin/assets/628926/6d638fe2-136f-4fb4-b1f9-1008559bc763)

CocoPoops:
just tested on 2 iPhones and a Pixel 7 Pro phone 
seems to just be a iOS/Safari Browser thing as it worked just fine on the pixel

CocoPoops:
also tried it on a few instances
 mbin.cocopoops.com | kbin.dk | kbin.melroy.org and they all do it on iOS/Safari

melroy89:
I don't own any Apple product 🥳 . But since you mentioned Safari, which is based on WebKit. I installed a WebKit browser under Linux like: https://wiki.gnome.org/Apps/Web (also known as Epiphany). 

Which allows me to reproduce the issue under Linux:

![image](https://github.com/MbinOrg/mbin/assets/628926/9d0d39a0-b2ee-4ede-b200-294b2ac746a1)

melroy89:
Simple debugging caused me to find that `display: flex` on `#sidebar .settings-list .settings-section` is causing the issue:

![image](https://github.com/MbinOrg/mbin/assets/628926/b020806a-8b6b-4052-8719-5d1c46074b0d)

melroy89:
I also found another solution to fix the problem, by keeping this display: flex. Which is disabling `height: 100%` on `.settings-row .label`:
![image](https://github.com/MbinOrg/mbin/assets/628926/f6766016-8a03-475d-92e5-e0e7f69ce150)

melroy89:
See my PR: https://github.com/MbinOrg/mbin/issues/148

PR #150, opened 2023-10-28
Remove height 100% from label class in settings

melroy89:
Fixes: https://github.com/MbinOrg/mbin/issues/148

Before:

image

After:

image

Ps. Yes this is WebKit under Linux.

### DIFF
--- a/assets/styles/components/_settings_row.scss
+++ b/assets/styles/components/_settings_row.scss
@@ -27,7 +27,6 @@
   .label {
     grid-area: label;
     line-height: normal;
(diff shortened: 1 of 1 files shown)


simonrcodrington:
I'll approve this one. I haven't ran across it on both chrome + Firefox on Ubuntu but it seems like a simple fix. Might need to keep an eye open to see if anyone reports an issue once this one gets merged (no idea why the 100% was even in there)

simonrcodrington (approved):

PR #151, opened 2023-10-28
add some missing localization messages

e-five256:
looking through pages to find missing localization tokens

These entries are for:
- the close label of the sidebar on mobile
- the bottom label in the sidebar for comment reply box position
- the admin reports page

from my understanding this will add them to weblate to allow translations into other languages, but let me know if that is incorrect and it's supposed to go the other way


melroy89:
Correct adding them to the English language, which is configured as base lanaguage will sync towards weblate in mbin. This is the power of configuring the tools correctly. 💪

nobodyatroot (approved):

PR #153, opened 2023-10-28
Explain Nginx setup for Docker (and bare metal for www)

melroy89:
- Explain how to use Nginx if you are running the Docker setup
- Add missing section about the `www` subdomain, remove `www` from the server section, this caused a lot of issues for @Jerry-infosecexchange I think.. Sorry.

### DIFF
--- a/docs/docker_deployment_guide.md
+++ b/docs/docker_deployment_guide.md
@@ -181,11 +181,6 @@ files.
 NGINX reverse proxy example for the Mbin Docker instance:
 
 ```conf
-map $http_upgrade $connection_upgrade {
-    default upgrade;
-    ''      close;
-}
-
 # Redirect HTTP to HTTPS
 server {
     server_name domain.tld;
@@ -225,9 +220,6 @@ server {
     access_log /var/log/nginx/mbin_access.log;
 
     location / {
-        proxy_http_version 1.1;
-        proxy_set_header Upgrade $http_upgrade;
-        proxy_set_header Connection "Upgrade";
         proxy_set_header HOST $host;
         proxy_set_header X-Forwarded-Host $host;
         proxy_set_header X-Real-IP $remote_addr;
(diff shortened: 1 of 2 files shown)


docs/docker_deployment_guide.md
```diff
@@ -170,12 +170,87 @@ If you created the file `compose.override.yml` with your configs (`cp compose.pr
...
+        proxy_set_header HOST $host;
+        proxy_set_header X-Forwarded-Host $host;
+        proxy_set_header X-Real-IP $remote_addr;
+        proxy_set_header X-Forwarded-For $proxy_add_x_forwarded_for;
+        proxy_set_header X-Forwarded-Proto $scheme;
+        proxy_pass http://127.0.0.1:8008;
+    }
+
+    location /.well-known/mercure {
+        proxy_pass http://127.0.0.1:8080$request_uri;
```
nobodyatroot, line 240: Only thing here is the port needs changed to match above, 8008 not 8080.

nobodyatroot (changes requested):

docs/docker_deployment_guide.md
melroy89: Oopsy. you are right! Fixed now. But nothing else? Can we remove the HTTPS upgrade section maybe?

docs/docker_deployment_guide.md
```diff
@@ -170,12 +170,87 @@ If you created the file `compose.override.yml` with your configs (`cp compose.pr
...
+    add_header Strict-Transport-Security "max-age=31536000; includeSubDomains; preload" always;
+
+    client_max_body_size 20M; # Max size of a file that a user can upload
+
+    # Logs
+    error_log /var/log/nginx/mbin_error.log;
+    access_log /var/log/nginx/mbin_access.log;
+
+    location / {
+        proxy_http_version 1.1;
```
melroy89, line 228: so maybe this is not required, same for the two headers below this `proxy_http_version` line. Out of my head, this is only needed when working with websockets.

nobodyatroot: yep, you're correct. it's for websocket support, so it's safe to remove.

melroy89: Done :)

nobodyatroot (approved):

PR #156, opened 2023-10-28
Handle errrors and successes on forms

melroy89:
Handle potential errors or success messages on forms that can could a failure in the event handlers (eg. during a `onPostSubmit` event an invalid image type can thrown an exception).

- Show the user an error when something went wrong after a form submission (bad-weather behavior), instead of showing 500 error.
   - Add missing flash twig template where needed in that case as well..
- Also show success message to user in case of a successful change after a form submission, I only show a flash message when there is no ajax call involved (when ajax/JSON responses are created to the frontend, we currently are unable to easily inform the user; due to the fact that Twig templating system is not executed in those cases. after all it's only an ajax call)
- Further improve `.editorconfig` (if you know how to _stop_ doing a new line with yaml multi-line, let me know, since that is what Weblate is using..)
-  Renamed `UserEditController::general` to `UserEditController::profile`
-  Additional null checks (in case the image is null)
- Add code documentation

Example:

example

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -5,3 +5,4 @@ indent_size = 4
 [*.yaml]
 indent_style = space
 indent_size = 4
+quote_type = single
--- a/.editorconfig
+++ b/.editorconfig
@@ -6,3 +6,8 @@ indent_size = 4
 indent_style = space
 indent_size = 4
 quote_type = single
+
+[translations/*.yaml]
+indent_style = space
+indent_size = 2
+quote_type = single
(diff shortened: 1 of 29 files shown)


TheVillageGuy (approved):

nobodyatroot (approved):
Looks good to me, running without any issues in `prod` after testing a few cases to exercise the code.

melroy89:
Fun fact, there was an error in the title.

PR #157, opened 2023-10-29
comment reply position - translate top label text

e-five256:
focused a bit too much on missing loc entries and missed that top wasn't being translated at all. this translates it, which already existed for "Top" regarding posts. I'm not quite sure if the two usages have the same meanings in all languages so if they don't, we might want to break this out to be a different loc token

### DIFF
--- a/templates/layout/_options_appearance.html.twig
+++ b/templates/layout/_options_appearance.html.twig
@@ -44,7 +44,7 @@
     </div>
     <strong>{{ 'single_settings'|trans }}</strong>
     <div class="settings-section">
-        {{ component('settings_row_enum', {label: 'comment_reply_position'|trans, help: 'comment_reply_position_help'|trans, settingsKey: 'KBIN_COMMENTS_REPLY_POSITION', values: [ {name: 'top'|trans , value: 'TOP'}, {name: 'bottom'|trans , value: 'BOTTOM' } ], defaultValue: 'TOP' } ) }}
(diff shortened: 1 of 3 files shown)


nobodyatroot (dismissed):

e-five256:
oops, sorry, looking at translations, it might be better to change the token after all, the translations for `top:` seem to be for "Outstanding", which makes sense. Maybe I should change this to `position_top` and change bottom as well before people translate it

melroy89:
Or we just change the translation of top.

e-five256:
It's used in post sorting (Top, Hot, New) etc. I think it makes sense to use two different loc tokens in this case, as non-English languages probably have distinct words for each

melroy89:
I mean the other place where top is used, where apparently it is used in the context for "outstanding".

e-five256:
Spanish (Outstanding), Italian (Better), Polish (Important), Chinese (Optimal), according to google translate. top definitely has multiple meanings in english so it makes sense it doesn't translate exactly for both scenarios

melroy89:
Don't change other translation files apart from English. This will give issues in Weblate two-away sync. Only English is allow to be changed in the repository. [Weblate should be used for all translations](https://hosted.weblate.org/projects/mbin/)...

You're introducing regression now:

image

e-five256:
Sorry about that, thanks for fixing it. I thought the concept of two-way meant changes on both ends worked for any files. I'm surprised it tries to take changes from branches, as I would've thought it'd only sync the default branch as tons of changes can happen while work is in progress on branches. And I guess there's no way to lock those files from changes on this end.

nobodyatroot (approved):

e-five256:
@melroy89 does this change seem ok to you now? Don't want to make any mistakes with the localization files again. not sure what happens to the dutch translation if the english token is removed

melroy89:
@e-five256 I was also surprised. I still do not recommend changing the translation files except for the English version to limit the sync issues. But it might have been just something strange sync bug on the github main branch and the weblate remote main branch. Yes it's fixed. I think the change is fine ... Let's see how it goes.

melroy89 (approved):

PR #161, opened 2023-10-29
SubscriptionSidebar: improvements and default changes

### LINKED ISSUE #154: Improvements for subscriber box
A few ideas for the new subscriber box as discussed on matrix

> <@thevillageguy:matrix.org> Or, even better, now that's still new and users haven't gotten used to it yet in it's current form;
> turn it on by default, including for existing users,
> by default seperate,
> Give it a move button (An arrow point to the other size of the screen)
> Give it a close button

>I thought about turning it on by default, but not in the separate layout. >Worth an issue as well

So my final thoughts,

On by default
Give it a pop out button
Once popped out, give it a button to move to other side

I think it will be much appreciated, so a hide button should not necessary

BentiGorlich:
- show subs by default
- add icons in the header of the panel to pop it in and out and, if popped out, to change the side the panel is on
- add a setting to make the inline sub panel larger (the former size)
- set the default `max-height` of the inline panel to `20em` which is half the size
- make the popped out panel smaller (half of the sidebar and a 6th of the main content -> double the ratio on everything except the sub sidebar
- hide the placeholder div in popped out state so no space is wasted
- add the sort values to the `VALUES` array in `ThemeSettingsController`

Fix #154

A lot of this is up for debate. I just implemented it in a way that should satisfy @TheVillageGuy and the way I'd want it myself.

# Images

1. Freshly logged in: 
grafik

2. After clicking the left arrow:
grafik

3.After clicking the right arrow:
grafik

4. After clicking the right arrow again:
grafik

4. After clicking the left arrow:
grafik

6. After changing the "Subscriptions panel large" setting to true:
grafik

7. After changing the "Page width" to "Auto":
grafik

8. After changing the "Page width" to "Max":
grafik

9. After clicking the left arrow again:
grafik

10. After clicking the right arrow:
grafik

### DIFF
--- a/assets/styles/components/_sidebar-subscriptions.scss
+++ b/assets/styles/components/_sidebar-subscriptions.scss
@@ -63,6 +63,9 @@
 
     .subscription {
       padding: 0.5rem;
+      overflow: hidden;
+      text-overflow: ellipsis;
+      white-space: nowrap;
 
       @include media-breakpoint-down(lg) {
         min-width: max-content;
@@ -79,8 +82,11 @@
       }
 
       .magazine-subscription-avatar {
+        margin: 0;
         height: 1.5rem;
+        max-height: 1.5rem;
         width: 1.5rem;
+        max-width: 1.5rem;
         object-fit: scale-down;
         border-radius: 50%;
         display: inline;
--- a/assets/styles/components/_sidebar-subscriptions.scss
+++ b/assets/styles/components/_sidebar-subscriptions.scss
@@ -5,9 +5,28 @@
 .sidebar-subscriptions {
   height: fit-content;
 
+  .sidebar-subscriptions-icons {
+    float: right;
+
+    @include media-breakpoint-down(lg) {
+      display: none;
+    }
+  }
+
   &.inline .subscription-list {
-    max-height: 40em;
+    max-height: 20em;
     overflow: auto;
+    &.lg {
+      max-height: 40em;
+    }
+  }
+
+  &.inline .magazine-subscription-avatar-placeholder {
+    display: inline-block;
+  }
+
+  &:not(.inline) .magazine-subscription-avatar-placeholder {
+    display: none;
   }
 
   &.section {
@@ -70,7 +89,6 @@
 
       .magazine-subscription-avatar-placeholder{
         width: 1.5rem;
-        display: inline-block;
         @include media-breakpoint-down(lg) {
           display: none;
         }
@@ -83,7 +101,8 @@
 
       .magazine-name {
         font-weight: bold;
-        padding: 0.25rem;
+        // padding: 0.25rem;
+        display: inline;
 
         &.has-image {
           @include media-breakpoint-down(lg) {
--- a/assets/styles/components/_sidebar-subscriptions.scss
+++ b/assets/styles/components/_sidebar-subscriptions.scss
@@ -95,7 +95,6 @@
         max-width: 1.5rem;
         object-fit: scale-down;
         border-radius: 50%;
-        display: inline;
         vertical-align: middle;
       }
 
--- a/assets/styles/components/_sidebar-subscriptions.scss
+++ b/assets/styles/components/_sidebar-subscriptions.scss
@@ -2,6 +2,36 @@
   border-radius: 0.5rem;
 }
 
+.kbin-container.width--fixed {
+  .sidebar-subscriptions:not(.inline) .subscription-list .subscription a {
+    max-width: calc(max(305px, 1360px / 6) - 3rem);
+  }
+
+  .sidebar-subscriptions.inline .subscription-list .subscription a {
+    max-width: calc(max(305px, 1360px / 4) - 3rem);
+  }
+}
+
+.kbin-container.width--auto {
+  .sidebar-subscriptions:not(.inline) .subscription-list .subscription a {
+    max-width: calc(max(305px, 85vw / 6) - 3rem);
+  }
+
+  .sidebar-subscriptions.inline .subscription-list .subscription a {
+    max-width: calc(max(305px, 85vw / 4) - 3rem);
+  }
+}
+
+.kbin-container.width--max {
+  .sidebar-subscriptions:not(.inline) .subscription-list .subscription a {
+    max-width: calc(max(305px, 100vw / 6) - 3rem);
+  }
+
+  .sidebar-subscriptions.inline .subscription-list .subscription a {
+    max-width: calc(max(305px, 100vw / 4) - 3rem);
+  }
+}
+
 #sidebar .sidebar-subscriptions,
 .sidebar-subscriptions {
   height: fit-content;
--- a/assets/styles/components/_sidebar-subscriptions.scss
+++ b/assets/styles/components/_sidebar-subscriptions.scss
@@ -96,6 +96,11 @@
         object-fit: scale-down;
         border-radius: 50%;
         vertical-align: middle;
+        &.onlyMobile {
+          @include media-breakpoint-up(lg) {
+            display: none;
+          }
+        }
       }
 
       .magazine-subscription-avatar-placeholder{
@@ -103,12 +108,17 @@
         @include media-breakpoint-down(lg) {
           display: none;
         }
+        &.onlyMobile {
+          @include media-breakpoint-up(lg) {
+            display: none;
+          }
+        }
       }
 
       &.active {
-        border: var(--kbin-header-border);
-
+        background-color: var(--kbin-bg);
       }
+
       a {
         display: inline-block;
         overflow: hidden;
@@ -121,6 +131,11 @@
 
           &.has-image {
             margin-left: .25rem;
+            &.onlyMobile {
+              @include media-breakpoint-up(lg) {
+                margin-left: 0;
+              }
+            }
             @include media-breakpoint-down(lg) {
               display: none;
             }
--- a/assets/styles/components/_sidebar-subscriptions.scss
+++ b/assets/styles/components/_sidebar-subscriptions.scss
@@ -6,6 +6,14 @@
 .sidebar-subscriptions {
   height: fit-content;
 
+  &:not(.inline) {
+    padding: 0 0.5rem;
+  }
+
+  &.inline {
+    padding-bottom: unset;
+  }
+
   .sidebar-subscriptions-icons {
     float: right;
 
@@ -14,7 +22,7 @@
     }
   }
 
-  &.inline .subscription-list {
+  .inline .subscription-list {
     max-height: 20em;
     overflow: auto;
     &.lg {
@@ -22,15 +30,15 @@
     }
   }
 
-  &.inline .magazine-subscription-avatar-placeholder {
+  .inline .magazine-subscription-avatar-placeholder {
     display: inline-block;
   }
 
-  &:not(.inline) .magazine-subscription-avatar-placeholder {
+  :not(.inline) .magazine-subscription-avatar-placeholder {
     display: none;
   }
 
-  &.section {
+  .section {
     padding: 0.5rem 0.5rem 0;
 
     @include media-breakpoint-down(lg) {
@@ -47,7 +55,7 @@
   }
   .subscription-list {
 
-    :first-child {
+    &.meta :first-child {
       border-top: unset;
     }
 
--- a/assets/styles/components/_sidebar-subscriptions.scss
+++ b/assets/styles/components/_sidebar-subscriptions.scss
@@ -2,6 +2,7 @@
   border-radius: 0.5rem;
 }
 
+#sidebar .sidebar-subscriptions,
 .sidebar-subscriptions {
   height: fit-content;
 
@@ -30,29 +31,25 @@
   }
 
   &.section {
-    padding: 0.5rem 0.5rem 1rem;
+    padding: 0.5rem 0.5rem 0;
 
-      &:not(.inline) {
-        @include media-breakpoint-down(lg) {
-          margin: 0.5rem;
-          padding: 0.5rem;
-        }
+    @include media-breakpoint-down(lg) {
+      padding: 0.5rem;
     }
-
   }
 
   h3 {
-    border-bottom: var(--kbin-sidebar-header-border);
-    color: var(--kbin-sidebar-header-text-color);
-    font-size: .8rem;
-    margin: 0;
-    text-transform: uppercase;
+    margin: 0 0 .5rem;
 
     @include media-breakpoint-down(lg) {
       margin: 0;
     }
   }
   .subscription-list {
+
+    :first-child {
+      border-top: unset;
+    }
 
     @include media-breakpoint-down(lg) {
       display: flex;
@@ -63,9 +60,6 @@
 
     .subscription {
       padding: 0.5rem;
-      overflow: hidden;
-      text-overflow: ellipsis;
-      white-space: nowrap;
 
       @include media-breakpoint-down(lg) {
         min-width: max-content;
@@ -79,6 +73,10 @@
         @include media-breakpoint-down(lg) {
           border-right: var(--kbin-meta-border);
         }
+      }
+
+      &:last-child {
+        border-bottom: unset;
       }
 
       .magazine-subscription-avatar {
@@ -104,15 +102,21 @@
         border: var(--kbin-header-border);
 
       }
+      a {
+        display: inline-block;
+        overflow: hidden;
+        text-overflow: ellipsis;
+        white-space: nowrap;
 
-      .magazine-name {
-        font-weight: bold;
-        // padding: 0.25rem;
-        display: inline;
+        .magazine-name {
+          font-weight: bold;
+          display: inline;
 
-        &.has-image {
-          @include media-breakpoint-down(lg) {
-            display: none;
+          &.has-image {
+            margin-left: .25rem;
+            @include media-breakpoint-down(lg) {
+              display: none;
+            }
           }
         }
       }
--- a/assets/styles/layout/_layout.scss
+++ b/assets/styles/layout/_layout.scss
@@ -36,7 +36,7 @@ body {
     @include media-breakpoint-up(lg) {
       .subs-show & {
         grid-template-areas: 'subs main sidebar';
-        grid-template-columns: 1fr 6fr 2fr;
+        grid-template-columns: minmax(305px, 1fr) 4fr minmax(305px, 1fr);
       }
 
       .sidebar-left & {
@@ -46,18 +46,18 @@ body {
 
       .sidebar-left.subs-show & {
         grid-template-areas: 'sidebar main subs';
-        grid-template-columns: 2fr 6fr 1fr;
+        grid-template-columns: minmax(305px, 1fr) 4fr minmax(305px, 1fr);
       }
 
 
       .sidebars-same-side.subs-show & {
         grid-template-areas: 'main sidebar subs ';
-        grid-template-columns: 6fr 2fr 1fr;
+        grid-template-columns: 4fr minmax(305px, 1fr) minmax(305px, 1fr);
       }
 
       .sidebars-same-side.sidebar-left.subs-show & {
         grid-template-areas: 'subs sidebar main';
-        grid-template-columns: 1fr 2fr 6fr;
+        grid-template-columns: minmax(305px, 1fr) minmax(305px, 1fr) 4fr;
       }
     }
 
--- a/assets/styles/layout/_layout.scss
+++ b/assets/styles/layout/_layout.scss
@@ -36,7 +36,7 @@ body {
     @include media-breakpoint-up(lg) {
       .subs-show & {
         grid-template-areas: 'subs main sidebar';
-        grid-template-columns: 1fr 3fr 1fr;
+        grid-template-columns: 1fr 6fr 2fr;
       }
 
       .sidebar-left & {
@@ -46,18 +46,18 @@ body {
 
       .sidebar-left.subs-show & {
         grid-template-areas: 'sidebar main subs';
-        grid-template-columns: 1fr 3fr 1fr;
+        grid-template-columns: 2fr 6fr 1fr;
       }
 
 
       .sidebars-same-side.subs-show & {
         grid-template-areas: 'main sidebar subs ';
-        grid-template-columns: 3fr 1fr 1fr;
+        grid-template-columns: 6fr 2fr 1fr;
       }
 
       .sidebars-same-side.sidebar-left.subs-show & {
         grid-template-areas: 'subs sidebar main';
-        grid-template-columns: 1fr 1fr 3fr;
+        grid-template-columns: 1fr 2fr 6fr;
       }
     }
 
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -24,31 +24,6 @@
 
     {% block stylesheets %}
         {{ encore_entry_link_tags('app') }}
-            <style>
-                {#
-                This is needed so the magazine names in the subscription panel are correctly overflowing.
-                The max width depends on the max-width of the whole container, which is controlled by a setting (KBIN_PAGE_WIDTH)
-                 #}
-                .subscription-list .subscription a {
-                    {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_SUBSCRIPTIONS_IN_SEPARATE_SIDEBAR')) is same as 'true' %}
-                        {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::MAX') %}
-                            max-width: calc(100vw / 9 - 3rem);
-                        {% elseif app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::AUTO') %}
-                            max-width: calc(85vw / 9 - 3rem);
-                        {% else %}
-                            max-width: calc(1360px / 9 - 3rem);
-                        {% endif%}
-                    {% else %}
-                        {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::MAX') %}
-                            max-width: calc(100vw / 4 - 3.5rem);
-                        {% elseif app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::AUTO') %}
-                            max-width: calc(85vw / 4 - 3.5rem);
-                        {% else %}
-                            max-width: calc(1360px / 4 - 3.5rem);
-                        {% endif%}
-                    {% endif %}
-                }
-            </style>
     {% endblock %}
 
     {% block javascripts %}
--- a/templates/base.html.twig
+++ b/templates/base.html.twig
@@ -30,13 +30,13 @@
                 The max width depends on the max-width of the whole container, which is controlled by a setting (KBIN_PAGE_WIDTH)
                  #}
                 .subscription-list .subscription a {
-                    {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS_IN_SEPARATE')) is same as 'true' %}
+                    {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_SUBSCRIPTIONS_IN_SEPARATE_SIDEBAR')) is same as 'true' %}
                         {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::MAX') %}
-                            max-width: calc(100vw / 9);
+                            max-width: calc(100vw / 9 - 3rem);
                         {% elseif app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::AUTO') %}
-                            max-width: calc(85vw / 9);
+                            max-width: calc(85vw / 9 - 3rem);
                         {% else %}
-                            max-width: calc(1360px / 9);
+                            max-width: calc(1360px / 9 - 3rem);
                         {% endif%}
                     {% else %}
                         {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::MAX') %}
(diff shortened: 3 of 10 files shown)


melroy89:
You could also just trim it at the end, like a max. number of characters are allowed and if that exceeds you just limit the chars then just add `...` at the end. Something like that. 

If we enable this by default.. I really really believe that this needs to work together with a bigger width in general therefore see related PR: https://github.com/MbinOrg/mbin/pull/121

BentiGorlich:
> You could also just trim it at the end, like a max. number of characters are allowed and if that exceeds you just limit the chars then just add `...` at the end. Something like that.

I considered that, but
1. I didn't manage to implement it. What CSS would that be or would you do it in the backend?
2. I think it looks just fine even with the longes name, but maybe a lot of people disagree with that :D

TheVillageGuy:
The button system is spot on.

TheVillageGuy:
The bar seems narrower than the regular sidebar? I liked it symmetrical, but this isn't bad either. However, many magazine names will be too wide and the way it works now, for instance with "Langer Magazine Name", does not look good. Instead perhaps we could use a lighter font, end the line with ... and with a hover show the full name,

melroy89:
> The bar seems narrower than the regular sidebar? I liked it symmetrical, but this isn't bad either. However, many magazine names will be too wide and the way it works now, for instance with "Langer Magazine Name", does not look good. Instead perhaps we could use a lighter font, end the line with ... and with a hover show the full name,

True.. I actually requested to have the width smaller, in order to have the main body the same size, but I was aware this will create a asymmetrical look & feel as a down-side. Unless there is a better solution.

melroy89:
> > You could also just trim it at the end, like a max. number of characters are allowed and if that exceeds you just limit the chars then just add `...` at the end. Something like that.
> 
> I considered that, but
> 
>     1. I didn't manage to implement it. What CSS would that be or would you do it in the backend?
> 
>     2. I think it looks just fine even with the longes name, but maybe a lot of people disagree with that :D

https://stackoverflow.com/questions/11426275/how-can-i-show-dots-in-a-span-with-hidden-overflow

templates/layout/sidebar_subscriptions.html.twig
```diff
@@ -1,6 +1,43 @@
...
+                </a>
+            {% else %}
+                <a href="#" title="{{ 'subscription_sidebar_pop_in'|trans }}" aria-label="{{ 'subscription_sidebar_pop_in'|trans }}"
+                   onclick="fetch('{{ path('theme_settings', { key: constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS_IN_SEPARATE'), value: constant('App\\Controller\\User\\ThemeSettingsController::FALSE')}) }}').then(() => document.location.reload())">
+                   <i class="fa-solid fa-right-to-bracket"></i>
+                </a>
+            {% endif %}
+        </span>
+    </h3>
+    <div class="subscription-list {{ app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS_LARGE')) is same as 'true' ? 'lg' : '' }}">
```
e-five256, line 40: I think if you add the class `meta` here, you'll get a lot of the styling to match the other widgets. For example the `.75` opacity and the text color

before:
image

after:
image

e-five256:
on https://github.com/MbinOrg/mbin/pull/161/files#diff-238b2d267292fab3559e0698428914b62472452bb0a82a78128257dafe967505R46 do you have thoughts on adding

```
class="image-inline"
  width="30" height="30"
  style="max-width: 30px; max-height: 30px;"
```

to match https://github.com/MbinOrg/mbin/blob/main/templates/components/magazine_inline.html.twig#L5-L7

---

I would like to add at least `image-inline` as that protects against overflow when images fail to load, but not sure about opinions on the size of the icon

BentiGorlich:
The image overflow problem should be fixed now by setting max-width and -height

BentiGorlich:
I also rebased the branch and it looks fine with the code from #121

melroy89:
Are the screenshots still accurate in your description?

e-five256:
I, personally, am not sure it should be included on the magazine, entry, and non-top level microblog pages. (basically everywhere random doesn't appear in the sidebar). if you're already at a magazine view or in a thread I think you don't really need to see all your subscriptions, it sort of makes the page too busy / confusing UX flow

BentiGorlich:
Imo it should appear. I want to be able to switch between my subbed magazines easily and to remove the panel from everything but the main page is not ideal...

BentiGorlich:
> Are the screenshots still accurate in your description?

I'll update them.

assets/styles/components/_sidebar-subscriptions.scss
```diff
@@ -59,9 +75,16 @@
...
+
       .magazine-subscription-avatar {
+        margin: 0;
         height: 1.5rem;
+        max-height: 1.5rem;
         width: 1.5rem;
+        max-width: 1.5rem;
         object-fit: scale-down;
         border-radius: 50%;
         display: inline;
```
e-five256, line 90: I think this is overriding the `image-inline` class, so the alt text is overflowing (`image-inline` sets a `inline-block` display)

so it's doing this when the images break
image

rather than
image

BentiGorlich:
> > Are the screenshots still accurate in your description?
> 
> I'll update them.

done

assets/styles/components/_sidebar-subscriptions.scss
BentiGorlich: you're right, I'll remove that :+1:

e-five256:
The styling when separate looks a bit different to me, which I'm not sure if it's because of the meta thing I mentioned earlier or not, still looking into it, but this is what I see

image

templates/layout/_options_appearance.html.twig
```diff
@@ -17,16 +17,16 @@
...
-            {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS')) is same as 'true' %}
+            {{ component('settings_row_switch', {label: 'show_subscriptions'|trans, settingsKey: 'KBIN_GENERAL_SHOW_SUBSCRIPTIONS', defaultValue: 'true'}) }}
+            {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS')) is not same as 'false' %}
                 {{ component('settings_row_enum', {label: 'subscription_sort'|trans, settingsKey: 'KBIN_GENERAL_SHOW_SUBSCRIPTIONS_SORT', values: [ {name: 'alphabetically'|trans , value: 'ALPHABETICALLY'}, {name: 'last_active'|trans , value: 'LAST_ACTIVE' } ], defaultValue: 'LAST_ACTIVE'}) }}
-            {% endif %}
-            {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS')) is same as 'true' %}
                 {{ component('settings_row_switch', {label: 'subscriptions_in_own_sidebar'|trans, settingsKey: 'KBIN_GENERAL_SHOW_SUBSCRIPTIONS_IN_SEPARATE'}) }}
```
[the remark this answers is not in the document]
BentiGorlich: I thought it would be to make the settings appear to more users. But I think you're right and they could be removed.

BentiGorlich:
But I really don't know why the icons appear so weird...

BentiGorlich:
> > The styling when separate looks a bit different to me, which I'm not sure if it's because of the meta thing I mentioned earlier or not, still looking into it, but this is what I see
> > image
> > not sure why the dotted line is appearing above all the content rather than at the bottom, and where the icons are
> 
> That's still an error. The CSS that adds that was more specific for the popped out panel than mine which reverts it.

fixed

templates/layout/_options_appearance.html.twig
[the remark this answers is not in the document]
e-five256: I think this is my only outstanding comment: I would say get rid of those two options in the sidebar options as they've moved to the widget itself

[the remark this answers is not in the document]
BentiGorlich: I'll do that then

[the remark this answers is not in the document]
e-five256: wait which one? :stuck_out_tongue_closed_eyes: if the latter and hiding it on mobile, I did that with the page width toggle in this same file, you could potentially rename that from `width-setting` to something generic like `hide-on-mobile` and use it for all 3 toggles. if the former can ignore this comment

e-five256 (dismissed):
looks good to me

templates/layout/_options_appearance.html.twig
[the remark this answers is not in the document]
BentiGorlich: I just hid the two options, but now I realize, that you have no way of making the sidebar appear at the top on mobile, because you can't make the sidebar separate from the rest :D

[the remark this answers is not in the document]
BentiGorlich: If I could easily just always put the sidebar on top if the user is on mobile that would be very nice... But nobody told me an easy way to recognize mobile user agents :sweat_smile:

translations/messages.en.yaml
```diff
@@ -794,6 +794,11 @@ subscription_sort: Subscription sort
 alphabetically: Alphabetically
 subscriptions_in_own_sidebar: Subscriptions separate
 sidebars_same_side: Sidebars on the same side
```
e-five256, line 796: I think this and the above translation can be cleaned up but doesn't need to be part of this PR, maybe we can wait and see if people disagreed with me and want it back :laughing:

BentiGorlich: you can always change translations in weblate :D

BentiGorlich:
I want to wait for @TheVillageGuy and @melroy89 to review the changes before I merge them. It is not that small of a change anymore :D

asdfzdfj:
my 2c after test driving for a day for consideration (desktop, fixed width):

- given the amount of associated configuration options I think it should be grouped into a new section, if anything this should give more space for description and not having to include the word 'subscription' in almost all of them.
- highlighting current active magazine is a nice touch, however I find using hard border to mark it is quite jarring (more than once that I mistook this as browser keyboard navigation border that's stuck), I tried modifying it to have it change background color of active magazine instead and I find it less jarring (and ideally using the swap/popout icon color for text highlight too, but css's a mess and I don't know which color is it)
  image
- the split panel width feels a bit scuffed, from my experiment I think the column width could be safely raised to about `1.25fr` and still looks fine without eating too much into main content's width (see above image)
- in single entry/post view, the inline subscription panel should be at the top, or at least not between thread and magazine info
  image
  (a dirty inspector edit to show what it could looks like)
- this one's a bit of edge case but for the left split panel case, both side toggle and inline/split toggle have a right pointing arrow in its icon, and that could be a bit hard to read at a glance (occasionally I have to stop and think which one will do what I want at the moment), I wonder if there's other icon pair that could be used for the inline/split panel toggle instead of the current one

translations/messages.en.yaml
e-five256: can `en` translations be removed from the weblate side? I wasn't sure if it was the "sentinel" version and is only editable on this end. I don't want to break weblate again :grimacing:

melroy89:
> I want to wait for @TheVillageGuy and @melroy89 to review the changes before I merge them. It is not that small of a change anymore :D

Thanks. I will come back soon with a list of feedback/ideas/proposals I have for this PR if you don't mind 👍🏽 . Thanks for waiting on me and TheVillageGuy.

TheVillageGuy:
> I want to wait for @TheVillageGuy and @melroy89 to review the changes before I merge them. It is not that small of a change anymore :D

Is there somewhere where I can test it? I don't have much time and my Dev server isn't up yet

TheVillageGuy:
I just tried it on Melroy's instance and it looks great! The pop/in pop out button could be improved by simply removing it by using just arrows, moving it from left to right. 

So when it''s on the left clicking -> will dock it. When it's docked <- | -> to move it either left or right. When it's on the right <- will dock it

BentiGorlich:
> I just tried it on Melroy's instance and it looks great! The pop/in pop out button could be improved by simply removing it by using just arrows, moving it from left to right.
> 
> So when it''s on the left clicking -> will dock it. When it's docked <- | -> to undock it either left or right. When it's on the right <- to dock it

I don't know 🤔 I think its not super intuitive if you don't have to option to just change the side its on when it is already undocked

BentiGorlich:
I noticed I never repleid to you @asdfzdfj 
> my 2c after test driving for a day for consideration (desktop, fixed width):
> 
> * given the amount of associated configuration options I think it should be grouped into a new section, if anything this should give more space for description and not having to include the word 'subscription' in almost all of them.

Agreed, I removed some options, but I need to put them back for the mobile view, as the icons are hidden when on mobile

> * highlighting current active magazine is a nice touch, however I find using hard border to mark it is quite jarring (more than once that I mistook this as browser keyboard navigation border that's stuck), I tried modifying it to have it change background color of active magazine instead and I find it less jarring (and ideally using the swap/popout icon color for text highlight too, but css's a mess and I don't know which color is it)

I can definetely understand that. Now I just need a CSS class which achieves what you want, so the themes still look nice

> * the split panel width feels a bit scuffed, from my experiment I think the column width could be safely raised to about `1.25fr` and still looks fine without eating too much into main content's width (see above image)

I don't have an opinion on that. I think it looks fine the way it is, but we can vote on that or smth :D

> * in single entry/post view, the inline subscription panel should be at the top, or at least not between thread and magazine info

Absolutely.
 
> * this one's a bit of edge case but for the left split panel case, both side toggle and inline/split toggle have a right pointing arrow in its icon, and that could be a bit hard to read at a glance (occasionally I have to stop and think which one will do what I want at the moment), I wonder if there's other icon pair that could be used for the inline/split panel toggle instead of the current one

Ok yeah. What do you think about the suggestion from @TheVillageGuy ?

TheVillageGuy (dismissed):

melroy89:
@BentiGorlich one question: Despite the subs being showed by default now. It won't show as a separate panel, right?

melroy89:
@BentiGorlich Another question/request is what if people disabled the icons "Show magazines’ icons" to "False".. Should this sidebar with subs also stop showing the icons on desktop?

melroy89:
Another feedback would be that by default it the sidebar seems like 150px width, but if you have a magazine with a big name it can get extended to 167 px. So my proposal would be: increase the sidebar to this width in all cases ORR.. limit the chars earlier to keep the width the same.

melroy89:
If the sidebar is enabled and put in a separate sidebar, the right sidebar becomes SMALLER. Causing this to happen:

image

melroy89 (dismissed):
See my remarks above.

TheVillageGuy:
> Last but not least can we remove the LEFT padding on the right panel, and the right padding on the left panel. Making the spacing between the main content and the sidebars the same as the action bar or alerts on the top. Basically now it's double the padding.. image
> 
> Currently the padding is applied to both left and right on the sidebar classes (eg. `#middle #sidebar`). So only add right padding to the right sidebar (same idea on the left sidebar (`#sidebar .sidebar-subscriptions:not(.inline), .sidebar-subscriptions:not(.inline)` .. why so complex; but I digress?); only add left padding on the left sidebar), and keep `#middle #main` alone (do **not** change the padding on `#middle #main`)
> 
> Ps. so maybe also make it CSS cleaner, why not just add padding to `#sidebar, .sidebar-subscriptions` and remove the padding when inline. Sounds easier for me.

Padding on my instance is asymmetrical, but I assume that's been fixed.

I would like the subscription bar to be the same width as the sidebar. Or at least when it's on the opposite side, for symmetry but also so that longer magazine names will fit

melroy89:
> > Last but not least can we remove the LEFT padding on the right panel, and the right padding on the left panel. Making the spacing between the main content and the sidebars the same as the action bar or alerts on the top. Basically now it's double the padding.. image
> > Currently the padding is applied to both left and right on the sidebar classes (eg. `#middle #sidebar`). So only add right padding to the right sidebar (same idea on the left sidebar (`#sidebar .sidebar-subscriptions:not(.inline), .sidebar-subscriptions:not(.inline)` .. why so complex; but I digress?); only add left padding on the left sidebar), and keep `#middle #main` alone (do **not** change the padding on `#middle #main`)
> > Ps. so maybe also make it CSS cleaner, why not just add padding to `#sidebar, .sidebar-subscriptions` and remove the padding when inline. Sounds easier for me.
> 
> Padding on my instance is asymmetrical, but I assume that's been fixed.
> 
> I would like the subscription bar to be the same width as the sidebar. Or at least when it's on the opposite side, for symmetry but also so that longer magazine names will fit

the padding is now the same on the left and right side.  But.. my proposal here was specifically to change the sidebar paddings in general. Which would result into this (preview):

image

TheVillageGuy:
> > > Currently the padding is applied to both left and right on the sidebar classes (eg. `#middle #sidebar`). So only add right padding to the right sidebar (same idea on the left sidebar (`#sidebar .sidebar-subscriptions:not(.inline), .sidebar-subscriptions:not(.inline)` .. why so complex; but I digress?); only add left padding on the left sidebar), and keep `#middle #main` alone (do **not** change the padding on `#middle #main`)
> > > Ps. so maybe also make it CSS cleaner, why not just add padding to `#sidebar, .sidebar-subscriptions` and remove the padding when inline. Sounds easier for me.
> > 
> > 
> > Padding on my instance is asymmetrical, but I assume that's been fixed.
> > I would like the subscription bar to be the same width as the sidebar. Or at least when it's on the opposite side, for symmetry but also so that longer magazine names will fit
> 
> the padding is now the same on the left and right side. But.. my proposal here was specifically to change the sidebar paddings in general. Which would result into this (preview), do you like this or not?:
> 
> image
> 
> Ps. maybe it's a bad idea.. I was just sharing my ideas.

Looks good to me

BentiGorlich:
> @BentiGorlich one question: Despite the subs being showed by default now. It won't show as a separate panel, right?

correct

BentiGorlich:
> @BentiGorlich Another question/request is what if people disabled the icons "Show magazines’ icons" to "False".. Should this sidebar with subs also stop showing the icons on desktop?

No not right now. Should propably be the case.

BentiGorlich:
> Another feedback would be that by default it the sidebar seems like 150px width, but if you have a magazine with long names it can get extended to 167 px. So my proposal would be: increase the sidebar to this width in all cases ORR.. limit the chars earlier to keep the width the same.

I thought I fixed that... Need to look at that again

BentiGorlich:
> If the sidebar is enabled and put in a separate sidebar, the right sidebar becomes SMALLER. Causing this to happen:
> 
> image

Ok, but what solution do you propose? I personally don't mind the little overflow there :D

BentiGorlich:
I'd say general sidebar padding belongs in its own issue :D

melroy89:
I'm using you latest changes now, look this is what I mean:

fixed_right_sidebar_width_decreases

templates/layout/sidebar_subscriptions.html.twig
```diff
@@ -1,30 +1,71 @@
-<div class="sidebar-subscriptions section {{ app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS_IN_SEPARATE')) is not same as 'true' ? 'inline' : 'section--top' }}">
-    <h3>{{ 'subscriptions'|trans }}</h3>
-    <div class="subscription-list">
-        {% for magazine in magazines %}
-            <div class="subscription {{ openMagazine and openMagazine.name is same as magazine.name ? 'active' : '' }}">
-                <a href="/m/{{ magazine.name }}">
-
-                    {% if magazine.icon %}
-                        <img src="{{ asset(magazine.icon.filePath) | imagine_filter('avatar_thumb') }}" class="magazine-subscription-avatar" alt="{{ magazine.name }}'s icon" />
+<div class="sidebar-subscriptions {{ app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS_IN_SEPARATE')) is not same as 'true' ? 'inline' : '' }}">
```
melroy89, line 1: Make this also a `aside` element instead of `div` just like the other sidebar. So it's easier for browsers /read mode to know it's related to the main content, but aside of the main.

melroy89: ow I see.. that is not possible or is it? Because you reuse the same layout for inside the aside :\

melroy89: I was even considering make the left sidebar more generic, so we can use it not only for subscriptions but other things...

melroy89:
I don't agree. but I don't want to be blocking. I remove my blocked for changes.

templates/layout/sidebar_subscriptions.html.twig
BentiGorlich: well firefox doesn't complain about an `aside` in an `aside` :D

BentiGorlich:
> I don't agree. but I don't want to be blocking. I remove my blocked for changes.

What is bugging you? I am not particularly fond of the small sidebar, but the option to make it larger always exist. There already is a setting that could do that when popped out

melroy89:
> I am not particularly fond of the small sidebar,

OK let's make in symmetrical again, but I agree in that case that both sidebars are allowed to become smaller in width, to still give enough space in the main section.

assets/styles/layout/_layout.scss
```diff
@@ -36,7 +36,7 @@ body {
     @include media-breakpoint-up(lg) {
       .subs-show & {
         grid-template-areas: 'subs main sidebar';
-        grid-template-columns: 1fr 3fr 1fr;
+        grid-template-columns: 1fr 6fr 2fr;
```
melroy89, line 39: So I think that will become this again:

```suggestion
       grid-template-columns: 1fr 3fr 1fr;
```

Or we could even for go: `300px auto 300px;`

asdfzdfj:
> > my 2c after test driving for a day for consideration (desktop, fixed width):
> > 
> > * given the amount of associated configuration options I think it should be grouped into a new section, if anything this should give more space for description and not having to include the word 'subscription' in almost all of them.
> 
> Agreed, I removed some options, but I need to put them back for the mobile view, as the icons are hidden when on mobile

that'd be great, also I think it would be nice to have all the options available when the subscription panel is enabled instead of showing/hiding options depending on which settings is enabled like it is now, unless you have a need to enforce possible setting state and its transition, which is understandable

(this one is _mainly my personal opinion_ due to a bit of my browsing quirk: I have them set to clear all data on closing, and so when I come back again I have to re-set all options and while the panel button is nice and interactive, it immediately reloads which means I have to set one option at a time, and I'd like to be able to set all needed options and only reload once to apply)
 
---

> Seems super intuitive to me, the arrows indicate in which direction the panel will move. Especially if you make the texts clear, so from left to right
> -> dock in sidepanel
> 
> <- undock to the left
> -> undock to the right
> 
> <- dock in sidepanel
> 
> The next step is to just be able to slide the panels around :)

> > * this one's a bit of edge case but for the left split panel case, both side toggle and inline/split toggle have a right pointing arrow in its icon, and that could be a bit hard to read at a glance (occasionally I have to stop and think which one will do what I want at the moment), I wonder if there's other icon pair that could be used for the inline/split panel toggle instead of the current one
> 
> Ok yeah. What do you think about the suggestion from @TheVillageGuy ?

It's a bit hard for me to imagine how will it work or looks like but it might work out, I think I'll need either working prototype or at least some illustrative figures to give a better assessment

at the very least I was thinking of just changing the panel pop out (inline -> split) button to something like [arrow-up-right-from-square](https://fontawesome.com/icons/arrow-up-right-from-square?f=classic&s=solid) and the reattach (split -> inline) to what should be the inverse of the pop out button, but I couldn't find one and the next best thing I could find is [down-left-and-up-right-to-center](https://fontawesome.com/icons/down-left-and-up-right-to-center?f=classic&s=solid) and for me this is sufficient.

image
image
(dirty inspector edit again to show what it might look like)

---

> > If the sidebar is enabled and put in a separate sidebar, the right sidebar becomes SMALLER. Causing this to happen:
> > image
> 
> Ok, but what solution do you propose? I personally don't mind the little overflow there :D

you could try to play around with [`minmax()`](https://developer.mozilla.org/en-US/docs/Web/CSS/minmax) in grid column size definition, maybe something like `grid-template-columns: 1fr 6fr minmax(320px, 2fr);` and see how it works out

TheVillageGuy:
> > I am not particularly fond of the small sidebar,
> 
> OK let's make in symmetrical again, but I agree in that case that both sidebars are allowed to become a smaller in width, to still give enough space in the main section. What do you think? You agree as well?
> 
> EDIT: Sorry that I changed my mind, you convinced me ;)

Why, it works and looks fine as is, why should a website be narrow?

templates/base.html.twig
```diff
@@ -24,6 +24,31 @@
...
+                        {% if app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::MAX') %}
+                            max-width: calc(100vw / 4 - 3.5rem);
+                        {% elseif app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_PAGE_WIDTH')) is same as constant('App\\Controller\\User\\ThemeSettingsController::AUTO') %}
+                            max-width: calc(85vw / 4 - 3.5rem);
+                        {% else %}
+                            max-width: calc(1360px / 4 - 3.5rem);
+                        {% endif%}
+                    {% endif %}
+                }
+            </style>
```
asdfzdfj, line 51: highly recommended to find a way to put this rule into the stylesheet

from what I could see, setting page width to either auto or max would set `.width--{auto,max}` on the parent container of the subscription sidebar, and when subscription sidebar is inline it'll have `.inline` class set, so I think it should be possible to define the default values for fixed width container and then add rules after to make setting specific override:

```scss
.subscription-list {
  /* defaults for fixed width container */
}

// setting specific override rules
.subscription-list {
  .width--auto & {
    // auto container panel override
    &.inline {
      // auto inline panel override
    }
  }
  .width--max & {
    // max container panel override
    &.inline {
      // max inline panel override
    }
  }
}
```

melroy89:
> Why make the bars smaller, it works and looks fine as is, why should a website be narrow?

No, the **main** content should be bigger, _not_ narrower, that was my point when two sidebars are active.

templates/base.html.twig
BentiGorlich: I'll look into that. That would be a far better solution, I agree

BentiGorlich:
> at the very least I was thinking of just changing the panel pop out (inline -> split) button to something like [arrow-up-right-from-square](https://fontawesome.com/icons/arrow-up-right-from-square?f=classic&s=solid) and the reattach (split -> inline) to what should be the inverse of the pop out button, but I couldn't find one and the next best thing I could find is [down-left-and-up-right-to-center](https://fontawesome.com/icons/down-left-and-up-right-to-center?f=classic&s=solid) and for me this is sufficient.
> 
> image image (dirty inspector edit again to show what it might look like)

@TheVillageGuy @melroy89 @e-five256 do you agree that these icons would be better suited? I am not really sure. I like both the way it is right now and @asdfzdfj suggestion

BentiGorlich:
I am playing around with the width right now and I gotta say I am tempted to just hardcode the width of both sidebars to 305px. The "active user" panel has a min width of 302.xx px anyway. I don't know if it is all that useful to have the sidebars grow with the middle section when one chooses a wider page width...

e-five256:
> I am playing around with the width right now and I gotta say I am tempted to just hardcode the width of both sidebars to 305px. The "active user" panel has a min width of 302.xx px anyway. I don't know if it is all that useful to have the sidebars grow with the middle section when one chooses a wider page width...

That seems fine to me. When I added the page width thing, I did notice the sidebar grew in width as well, but I didn't have a strong opinion about that as I thought maybe it could add some breathing room; but the main goal of page width to me was to allow a larger main content box so text wasn't so squished (especially when you get into super nested comment chains). The sidebar width staying consistent in each page width setting seems fine

e-five256:
> do you agree that these icons would be better suited?

to me, the pop out icon is usually very specific to opening something in a separate window; so if I clicked that I would expect the subscription panel to open in a small new browser window that was separate but clicking on it could target the main browser or something. I feel like what's happening here is more of a "split" than a "pop out", but that's just my opinion

BentiGorlich:
> > do you agree that these icons would be better suited?
> 
> to me, the pop out icon is usually very specific to opening something in a separate window; so if I clicked that I would expect the subscription panel to open in a small new browser window that was separate but clicking on it could target the main browser or something. I feel like what's happening here is more of a "split" than a "pop out", but that's just my opinion

Makes sense to me. I'll implement the button system @TheVillageGuy suggested and maybe the confusion will gone afterwards :D

translations/messages.en.yaml
```diff
@@ -790,10 +790,15 @@ password_and_2fa: Password & 2fa
...
+subscription_sort: Sort
 alphabetically: Alphabetically
-subscriptions_in_own_sidebar: Subscriptions separate
+subscriptions_in_own_sidebar: In separate sidebar
 sidebars_same_side: Sidebars on the same side
+subscription_sidebar_pop_out_right: Move to separate sidebar on the right
+subscription_sidebar_pop_out_left: Move to separate sidebar on the left
+subscription_sidebar_pop_in: Move subscriptions to the inline panel
+subscription_panel_large: Large panel
+subscription_header: Subs
```
nobodyatroot, line 801: Since we have the space (and to be consistent with the "random magazines" header), should we rename this to "subscribed magazines"?

templates/layout/sidebar_subscriptions.html.twig
```diff
@@ -1,30 +1,78 @@
-<div class="sidebar-subscriptions section {{ app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_GENERAL_SHOW_SUBSCRIPTIONS_IN_SEPARATE')) is not same as 'true' ? 'inline' : 'section--top' }}">
-    <h3>{{ 'subscriptions'|trans }}</h3>
-    <div class="subscription-list">
-        {% for magazine in magazines %}
-            <div class="subscription {{ openMagazine and openMagazine.name is same as magazine.name ? 'active' : '' }}">
-                <a href="/m/{{ magazine.name }}">
+<aside class="sidebar-subscriptions {{ app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::KBIN_SUBSCRIPTIONS_IN_SEPARATE_SIDEBAR')) is not same as 'true' ? 'inline' : '' }}">
```
e-five256, line 1: small nitpick, could we add a `{% if magazines|length %}` and `{% endif %}` at the end? I _think_ that should exclude this template if the user has no subscriptions. not a big deal but seems like it can be hidden until they sub to something

image

e-five256: actually that might be confusing with the options in the sidebar still... maybe we can hold off on that for another PR pass, this one has already gotten quite a lot in it

BentiGorlich: I'd put it in a new PR as well. But I'd prefer it if there was just a text inside: "you did not yet subscribe to any magazine".

translations/messages.en.yaml
BentiGorlich: Yeah since I have changed the sizes again this doesn't make any sense any more

melroy89 (approved):

PR #165, opened 2023-10-29
Introduces docker build and publish job

melroy89:
Introduces a Docker build (and publish when on main branch) in GitHub Actions. However, it does take quite some time to build. I'm doubting whenever or not we should include this big docker build (took 5 min) inside each PR...

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -11,3 +11,9 @@ quote_type = single
 indent_style = space
 indent_size = 2
 quote_type = single
+
+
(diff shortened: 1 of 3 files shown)


nobodyatroot (commented):
Is it possible to "cron" this so that it only builds an image after x amount of time since the last merge... or it might be good to tie it to the Mbin version so that it only builds after significant enough changes.

nobodyatroot (approved):

PR #168, opened 2023-10-29
Enable avatars by default

melroy89:
- Enable profile avatars by default
- Enable magazine avatars/icons by default


simonrcodrington (approved):
I agree having these on by default makes the experience feel a bit more "modern" and friendly. People can still revert and turn them off if they don't like it so it's a good change.

melroy89:
> I agree having these on by default makes the experience feel a bit more "modern" and friendly. People can still revert and turn them off if they don't like it so it's a good change.

That was my thoughts as well..

PR #172, opened 2023-10-30
Introducing NodeInfo v2.1 support (+refactoring)

melroy89:
- Introducing NodeInfo v2.1 support
- Add missing `services` and `metadata`
- Small refactoring

### DIFF
--- a/config/kbin_routes/activity_pub.yaml
+++ b/config/kbin_routes/activity_pub.yaml
@@ -13,9 +13,15 @@ ap_node_info:
   path: '/.well-known/nodeinfo'
   methods: [ GET ]
 
-ap_node_info_v2:
-  controller: App\Controller\ActivityPub\NodeInfoController::nodeInfoV2
+ap_node_info_v20:
+  controller: App\Controller\ActivityPub\NodeInfoController::nodeInfoV20
   path: '/nodeinfo/2.0.{_format}'
+  methods: [ GET ]
+  format: json
+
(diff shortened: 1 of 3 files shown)


BentiGorlich (changes requested):
As far as I can tell the schema definition requires `services` and `metadata` properties. 
It is fine to have empty in- and outbound services and empty metadata tags, but they are required

melroy89:
> As far as I can tell the schema definition requires `services` and `metadata` properties. It is fine to have empty in- and outbound services and empty metadata tags, but they are required

Ow let me check with: https://nodeinfo.diaspora.software/schema.html

melroy89:
> Yeah mastodon keeps them empty as well. They are required in v2.0 as well. Funny that they were just ignored...

Let me see what this JSON formatter does in PHP, since the meta-data should be an object (in JSON) while the inbound and outbound should be of type array (in JSON). But PHP well just only has `[]` lol..

BentiGorlich (approved):

BentiGorlich (approved):

melroy89:
> > Yeah mastodon keeps them empty as well. They are required in v2.0 as well. Funny that they were just ignored...
> 
> Let me see what this JSON Response formatter does in PHP, since the meta-data should be an object (in JSON) while the inbound and outbound should be of type array (in JSON). But PHP you just only have `[]` lol..

I fixed this by using `(object)[]` in PHP, that did the trick:

image

PR #173, opened 2023-10-30
remove active class addition when too many subscriptions

e-five256:
the section for too many subscriptions was checking against the `magazine` iterator which was used above it, but it exists outside the for loop. The active class doesn't need to be added to the show more section of the sidebar


BentiGorlich (approved):

melroy89:
Thanks! Works again now.

PR #176, opened 2023-10-30
fix page width not affecting header and footer

### LINKED ISSUE #175: Top Navbar doesn't scale with the container width
**Describe the bug**
When I switch the container width to "auto" or "max" the top navbar with the site name doesn't scale accordingly:

gehirneimer.de

**To Reproduce**
1. Got in the theme settings
2. switch "page width" to anything other than "fixed"

**Expected behavior**
It should scale accordingly

**Screenshots**
Screenshot from 2023-10-30 22-20-19

 - OS: Fedora 38
 - Browser Firefox

**Additional context**
#121

e-five256:
missed adding the user setting to these templates
also do not allow the 85% setting on mobile. should not happen as the option does not show on mobile but when going from large view to small it was being kept, which was not intended

fixes #175

### DIFF
--- a/assets/styles/layout/_layout.scss
+++ b/assets/styles/layout/_layout.scss
@@ -15,8 +15,10 @@ body {
     max-width: 100%;
   }
 
-  &.width--auto {
-    max-width: 85%;
+  @include media-breakpoint-up(lg) {
+    &.width--auto {
+      max-width: 85%;
+    }
   }
 }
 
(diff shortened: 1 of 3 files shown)


BentiGorlich:
I'll have a look at it tomorrow. Feel free to request the review of more people :)

melroy89:
Let me check it..

melroy89 (approved):
Indeed this PR fixes the mobile view again where you always want to have 100% width.

PR #179, opened 2023-10-31
Update admin_guide.md

TheVillageGuy:
Added npm to main apt-get line


melroy89 (changes requested):
Follow the nodejs guide to install nodejs with npm. The node package and npm from apt is too old.

TheVillageGuy:
npm is installed by
sudo apt-get install nodejs -y
which is a different version than the default debian 12 version so my edit is not justified, closing it

PR #180, opened 2023-10-31
Update FAQ.md

TheVillageGuy:
Proposed additional instructions for opening the rabbitmq web interface

### DIFF
--- a/FAQ.md
+++ b/FAQ.md
@@ -86,10 +86,18 @@ Talking about plugins, we advise to also enable the `rabbitmq_management` plugin
 sudo rabbitmq-plugins enable rabbitmq_management
 ```
 
-Give the kbin user the management tag:
+Let's create a new admin user in RabbitMQ (replace `<user>` and `password` with a username & password you like to use):
 
 ```sh
-sudo rabbitmqctl set_user_tags "kbin" management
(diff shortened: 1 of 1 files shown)


FAQ.md
```diff
@@ -86,7 +86,13 @@ Talking about plugins, we advise to also enable the `rabbitmq_management` plugin
 sudo rabbitmq-plugins enable rabbitmq_management
 ```
 
-This allows you to go to the RabbitMQ management page: `http://<server-ip>:15672`. [More info can be found here](https://www.rabbitmq.com/management.html#getting-started). See screenshot below of a typical small instance of Mbin running RabbitMQ management interface:
+Give the kbin user the management tag:
```
melroy89, line 89: I could also create a separate user with management permissions, that least that is what I did via: `sudo rabbitmqctl add_user <user>`

And then give this user management permissions: `sudo rabbitmqctl set_user_tags <user> administrator`

TheVillageGuy: Hmm, a bit more work but worth it for qol and safety to prevent having to log in with the 'secret'. Good idea

melroy89: dedicated admin account is indeed a best pratice.

melroy89 (approved):

PR #183, opened 2023-10-31
Prepare for tombstone support

melroy89:
- Do not fail on 410 status code (used for [Tombstone](https://www.w3.org/TR/activitypub/#delete-activity-outbox))
- Read also 410 error status HTTP content (which contains the tombstone object), passing `false` to `getContent()`
- Do **not** throw error / 500 error message

### DIFF
--- a/src/Service/ActivityPub/ApHttpClient.php
+++ b/src/Service/ActivityPub/ApHttpClient.php
@@ -58,13 +58,16 @@ class ApHttpClient
                 'headers' => $this->getInstanceHeaders($url),
             ]);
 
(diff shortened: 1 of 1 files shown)


melroy89:
You wanna test? sure try this tombstone url: `https://lemm.ee/post/2871450`

nobodyatroot (approved):
Less 500s, more better.

melroy89:
We still need to improve the fetching (making it blocking fetch) and after that, we can improve the frontend, informing the user that this post/thread/.. is no longer there in case of a tombstone.

PR #190, opened 2023-11-01
Introducing Open URL link to thread/post/comment

melroy89:
The "Copy" link is nice, but users including myself also want to open the original link sometimes directly into a new window (yes you can use middle mouse button click, if you have that). 

- Introducing a Open URL menu item
- Reorder the menu items
- Introducing a separator

image

And when there is no edit, no actor and no moderator, it looks like this:

image

In another PR I might also want to automatically close the drop-down menu when a user clicked on "Copy ... URL".

### DIFF
--- a/templates/components/entry.html.twig
+++ b/templates/components/entry.html.twig
@@ -177,8 +177,10 @@
                                 {% endif %}
 
                                 <li>
-                                    <a data-action="clipboard#copy"
-                                       href="{{ entry_url(entry) }}">{{ 'copy_url'|trans }}</a>
+                                    <a target="_blank"
+                                       href="{{ entry.apId ?? path('ap_entry', {magazine_name: entry.magazine.name, entry_id: entry.id}) }}">
+                                        {{ 'open_url_to_fediverse'|trans }}
+                                    </a>
                                 </li>
                                 <li>
                                     <a data-action="clipboard#copy"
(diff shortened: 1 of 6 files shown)


TheVillageGuy (approved):

e-five256:
I'm wondering if the more drop down is getting a bit bloated with too many options. An alternative could be something like `Original URL (Copy) (Open)`. But that might be a little difficult to hit on mobile screens 🤔

melroy89:
> I'm wondering if the more drop down is getting a bit bloated with too many options. An alternative could be something like `Original URL (Copy) (Open)`. But that might be a little difficult to hit on mobile screens 🤔
> 
> Edit: Looking at mastodon, it is two different options in the drop down, so maybe being consistent with them is better, or maybe other options can be moved out some other time

Fair points. Yes, you are right Mastodon also does this in a similar way. I could add maybe some horizontal lines for better readability? Something that Mastodon also does (Better a good steal than a bad idea, right?)

melroy89:
Reordered the menu items to make it more logical for me and it's also matching Mastodon more.

templates/components/entry.html.twig
```diff
@@ -176,16 +176,24 @@
...
                                     <a data-action="clipboard#copy"
                                        href="{{ entry.apId ?? path('ap_entry', {magazine_name: entry.magazine.name, entry_id: entry.id}) }}">
                                         {{ 'copy_url_to_fediverse'|trans }}
                                     </a>
                                 </li>
+                                <li>
+                                    <a data-action="clipboard#copy"
+                                       href="{{ entry_url(entry) }}">{{ 'copy_url'|trans }}</a>
+                                </li>
+                                <li class="dropdown__separator"></li>
```
e-five256, line 196: I would move this below the `is_granted` for each file, I don't think it's needed if there aren't more entries, the end of the list is a decent enough separator

melroy89: You are right!

melroy89: but uh.. I need to check now on if it's `is_granted` for edit OR `app.user and entry.isAuthor(app.user)` OR `is_granted` for moderator... Since you just want to add 1 separator here.

melroy89: `if is_granted('edit', entry) or if app.user and entry.isAuthor(app.user) or is_granted('moderate', entry)`  wonderful

melroy89: Done

e-five256 (approved):

PR #192, opened 2023-11-02
Change federation setting to match other enum styling

e-five256:
Personally, I found the federation status setting to be difficult to read. Was it on or off? What does on or off even mean?

This PR changes it to have a consistent style matching the settings panel. It also changes "On" and "Off" to terms used elsewhere (in magazine search): "Local" and "Local and federated". It does switch the option sides (enabled now appears on right rather than left), that felt more natural to the words being used and matches the current defaults of the other setting enums, don't think it has a big effect on usability.

Before:

image

After:

image

when changing settings (changed themes to give a different example) reload appears on both settings tab, but that seemed accurate to me as it's still true:

image

Conversely, these terms might not be a perfect fit as well. "Local" will still show posts from local users but to remote magazines, as your local users posts to the local copies of the magazines your instance has.

(As an aside, the enums don't seem quite readable to me in dark theme, it's probably the worst, it looks much better in the other themes, but at least now it matches either way)

<details>
<summary>screenshot in tokyo night</summary>

image

</details>

### DIFF
--- a/assets/controllers/settings_row_enum_controller.js
+++ b/assets/controllers/settings_row_enum_controller.js
@@ -9,7 +9,10 @@ export default class extends Controller {
     change({params: {actionPath, reloadRequired}}) {
         return fetch(actionPath).then(() => {
             if (reloadRequired) {
-                document.querySelector('.settings-list').classList.add('reload-required');
+                document.querySelectorAll('.settings-list')
+                  .forEach((el) => {
+                    el.classList.add('reload-required');
+                });
             }
         });
     }
--- a/templates/layout/_sidebar.html.twig
+++ b/templates/layout/_sidebar.html.twig
@@ -52,19 +52,8 @@
     <div id="federation" class="section" data-options-target="federation">
         <h3>{{ 'federation'|trans }}</h3>
         <div class="settings-list">
-            <div class="row">
-                <span>{{ 'status'|trans }}:</span>
-                <div>
-                    <a href="{{ path('theme_settings', {key: constant('App\\Controller\\User\\ThemeSettingsController::KBIN_FEDERATION_ENABLED'), value: 'true'}) }}"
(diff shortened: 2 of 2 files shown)


melroy89:
If this has something to do with sync status. Making make it indeed clear it's a "sync option" that can be set to only local or also local and remote? Just thinking out loud.

e-five256:
It's unrelated to the admin setting enabling or disabling federation as a whole. It's a near closer relation to Lemmy's `Local` and `All` which are just filters users can set for viewing content

image

The main difference here is that local means both local magazines and local users, which I don't believe is true in Lemmy. I like that distinction, as I feel it can help build a sense of community when switching to local you can see `localuser1`'s posts to a local magazine _and_ their posts to a remote `funny@remote.url`. But it might be confusing for people used to Lemmy's terms

melroy89:
ah I see.. well maybe in that case (it was confusing indeed), should we change the "status" label to something else? 

Like, uh... "Show user content" instead of "Status"? Then you have the option "local", which basically translates to "the local user content" (despite the user could indeed be commenting or sharing a remote magazine).

Just thinking out loud again. If you think 'status' is fine that is also OK, but I try to come up with a better label name.

melroy89:
No?

e-five256:
I think you're right that the setting is very confusing as it is now. Maybe debounced is right and I'm the only one who even knows it exists and uses it. It's possible a better approach is moving it to the other user filters like subscribed in the main content. But that would just leave the configuration settings in the sidebar, and a lot of the code is hardcoded to have 2 panels in the sidebar. Maybe this is where @lilfade 's idea of moving the footer would come in, but it's clear there's a lot of work/thought here for potentially not much gain, so starting from a different angle might be better.

melroy89:
I also now start to believe this federation setting can just be a filter option. And then remove this tab fully.

PR #193, opened 2023-11-02
Set a limit on the Bolt mercure db

melroy89:
By default there is no size limit (`0`), this change proposes a limit on the mercure size/DB. See also: https://mercure.rocks/docs/hub/config#bolt-adapter

This will set a size limit of "1000 events".

Plus some other minor improvements in the Caddyfile.

### DIFF
--- a/docker/caddy/Caddyfile
+++ b/docker/caddy/Caddyfile
@@ -33,7 +33,7 @@
 
         mercure {
             # Transport to use (default to Bolt)
-            transport_url {$MERCURE_TRANSPORT_URL:bolt:///data/mercure.db}
+            transport_url {$MERCURE_TRANSPORT_URL:bolt:///data/mercure.db?size=1000}
             # Publisher JWT key
             publisher_jwt {env.MERCURE_PUBLISHER_JWT_KEY} {env.MERCURE_PUBLISHER_JWT_ALG}
             # Subscriber JWT key
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -705,16 +705,23 @@ nano metal/caddy/Caddyfile
 
 The content of the `Caddyfile`:
 
-```
+```conf
 {
         {$GLOBAL_OPTIONS}
+        # No SSL needed
         auto_https off
         http_port {$HTTP_PORT}
         persist_config off
+
         log {
-                output file /var/www/kbin/var/log/mercure.log
                 # DEBUG, INFO, WARN, ERROR, PANIC, and FATAL
                 level WARN
+                output discard
+                output file /var/www/kbin/var/log/mercure.log {
+                        roll_size 50MiB
+                        roll_keep 3
+                }
+
                 format filter {
                         wrap console
                         fields {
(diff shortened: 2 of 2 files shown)


e-five256 (approved):
I'm not an instance admin running mercure so not sure I can give a complete stamp of approval, but this looks to be common practice in all the answers I see online (people also set a higher cleanup frequency (e.g. 0.3 default to 0.5 or 1) it seems but again, no idea if that fits mbin's design)

nobodyatroot (approved):
I don't use Mercure either but the db size issue should definitely be addressed.

melroy89:
I know the PR is merged now. But I did put out a question to Mercure repo over here: https://github.com/dunglas/mercure/issues/74#issuecomment-1791037000

If I get an answer back that might results into a different size (either or or lower) I can always create a new PR with a different size limit.

emomaliev:
My mercure.db file currently takes up 25 GB of memory. After which I set the restrictions as indicated in the documentation:

bolt:///var/run/mercure.db?size=3000&cleanup_frequency=0.4

```
  mercure:
    container_name: ${DOCKER_NAME}_mercure
    image: dunglas/mercure:v0.17
    restart: unless-stopped
    tty: true
    environment:
      SERVER_NAME: '****'
      MERCURE_TRANSPORT_URL: bolt:///var/run/mercure.db?size=3000&cleanup_frequency=0.4
      MERCURE_PUBLISHER_JWT_KEY: ****************
      MERCURE_SUBSCRIBER_JWT_KEY: ****************

```

In theory, this should help and the mercure.db file should no longer endlessly consume more and more memory. 

Now the problem is cleaning up mercure.db, are there any mechanisms for freeing already used memory?

BentiGorlich:
The easiest (and only) way I know is to just delete the mercure db... You can really drastically limit it, since the contents just are irrelevant. The only relevant thing is that the update is pushed to the clients...

FYI we are not continuing developement of the mercure realtime features

emomaliev:
> The easiest (and only) way I know is to just delete the mercure db..

If I delete the mercure.db file, will this affect currently active subscriptions?



> FYI we are not continuing developement of the mercure realtime features

that is, the Mercure project will no longer be updated and is abandoned?

BentiGorlich:
> If I delete the mercure.db file, will this affect currently active subscriptions?

Since you have to restart the mercure server, probably

> that is, the Mercure project will no longer be updated and is abandoned?

The mbin features using it, yes. On all the main servers of devs and also on fedia mercure is already disabled. It had too much overhead and caused too much browser cache problems.

melroy89:
ps. The documentation already documented the `size` parameter to limit the size: https://github.com/MbinOrg/mbin/blob/ec8bc2d5539ab821132fafb3ae9b491b4c8d8217/docs/02-admin/03-optional-features/01-mercure.md?plain=1#L72
It's just not documenting the `cleanup_frequency` option. Maybe still a good idea to add?

Since without `&cleanup_frequency=0.4` it will never trigger a clean-up ;)

melroy89:
Ow nevermind, the default of `cleanup_frequency` is apparently 0.3. Which is fine.

So there should already be a limit of 1000 now.

PR #195, opened 2023-11-02
Prepare for magzine removal Mbin (partial kbin sync)

melroy89:
Cherry-picked changes from [PR-1207](https://codeberg.org/Kbin/kbin-core/pulls/1207) and [PR-1208](https://codeberg.org/Kbin/kbin-core/pulls/1208) I do want to have in Mbin, but not the changes I don't like (yet). As discussed in Matrix, I would like to have a correct cascading DB schema first. If the database design is correct, the code can be much simpler, easier to understand and thus cleaner. 

- Adding `remove` cascading in Doctrine annotation  
- Prepare for magazine removal by admin role
- This will give the admin already access to local magazines the admin isn't owner of
- Clean-up dead code
- Minor fix; CSS class on delete account button


nobodyatroot (commented):
Agreed, lets take the time to implement this feature correctly and most efficiently.

nobodyatroot (approved):

PR #196, opened 2023-11-03
minor contributing doc change

e-five256:
remove duplicated word, change note to match style of other note


nobodyatroot (approved):
I dunno, this seems like too many changes. Better call a group meeting! jk, approved!

PR #197, opened 2023-11-03
Valuation: combine up votes and favourites

### LINKED ISSUE #127: Adjust valuation of threads to fit with Lemmy
In the "hot" category the current system works for me, although "boosts" are a lot more important than just upvotes, but I gues the amount of comments is relevant as well. However in the "top" category it just doesn't work. As far as I can tell it only values boosts and ignores comment count and upvotes, which just hides basically all lemmy posts, as they only upvote.

Replace boosts with upvote in the valuation formular

Add upvotes to the evaluation formular

BentiGorlich:
- change the way the score is calculated for `Entry` and `Post`. Added a simple addition of the favourites. Add a migration for it.
- Add `updateScore` method to the `FavouriteInterface`, so the score can be updated when one adds or remove a favourite
- The favourite count is no longer used in the `RankingTrait`, because it is used in the score which is the baseline for the ranking
- The "hot" and "top" (same in this case) sorts of `PostComment`s and `EntryComment`s now use the favourites as well

Close #127

### DIFF
--- a/migrations/Version20231103004800.php
+++ b/migrations/Version20231103004800.php
@@ -1,5 +1,7 @@
 <?php
 
+declare(strict_types=1);
+
 namespace DoctrineMigrations;
 
 use Doctrine\DBAL\Schema\Schema;
--- a/migrations/Version20231103004800.php
+++ b/migrations/Version20231103004800.php
@@ -1,0 +1,26 @@
+<?php
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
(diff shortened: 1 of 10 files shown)


melroy89:
Just one question.. how long with the top be at the top? this is a 3 month old thread for instance (I hope we have some SQL query that limit the record in terms of creation/update date):

image

BentiGorlich:
I honestly don't know, but my guess is that the filter (you can filter for time) will filter by the creation date of a thread...

melroy89:
ah yes the time filter.. Also still buried too far in the UI. Well, I think my question was a bit off-topic anyway. Thanks!

melroy89 (approved):

PR #198, opened 2023-11-03
remove extra subscribe in resubscribe function

asdfzdfj:
@melroy89 copy paste error when adding a null check? ended up with every call to resubscribe making extra connection again


melroy89 (approved):
Thanks!

PR #199, opened 2023-11-03
switch default theme from light to dark

e-five256:
this is an attempt at solving what originalucifier brought up in matrix:

- `theme--default` (and `theme--light`) aren't defined classes in the repo, they just use the configured root variables
- other themes have named files which override 93 of these variables (94 in both solarzied themes)
- there is [javascript](https://github.com/MbinOrg/mbin/blob/main/assets/controllers/kbin_controller.js#L50) that sees `theme--default` and checks your system settings, switching to `theme--light` or `theme--dark` dependingly

the result was that if you were visiting for the first time (no cookies for saved theme) and javascript was slow to load, you would get blinded by `theme--default` for a bit before it swaps to `theme--dark`. you can see this also by disabling javascript in your browser.

this change makes `theme--default` and `theme--dark` equivalent (and also easier to change in the future as it's defined now), and moved the light theme into being an actual theme like the others (thus making the opposite situation happen, you'll see `theme--default` as dark, and if your system is light themed, it will swap to `theme--light`. that seemed less shocking, but I'm not a light system user so not sure if that has the same flashbang effect for people who are.)

there were a bunch of variables that only appeared in `_variables` side by side with the light theme, I've attempted to keep those in the root file. (both solarized override the topbar setting, but it's common to all themes besides that)

### DIFF
--- a/assets/styles/_variables.scss
+++ b/assets/styles/_variables.scss
@@ -15,150 +15,29 @@ $aspect-ratios: (
 ) !default;
 
 :root {
-  --kbin-body-font-family: system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue", "Noto Sans", "Liberation Sans", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
-  --kbin-body-font-size: 1rem;
-  --kbin-body-font-weight: 400;
-  --kbin-body-line-height: 1.5;
-  --kbin-body-text-align: left;
-  --kbin-body-bg: #fff;
+  // these variables are common to all themes
 
-  --kbin-bg: #ecf0f1;
-  --kbin-bg-nth: #fafafa;
   --kbin-shadow: rgba(0, 0, 0, 0.25) 0px 54px 55px, rgba(0, 0, 0, 0.12) 0px -12px 30px, rgba(0, 0, 0, 0.12) 0px 4px 6px, rgba(0, 0, 0, 0.17) 0px 12px 13px, rgba(0, 0, 0, 0.09) 0px -3px 5px;
-
-  --kbin-text-color: #212529;
-  --kbin-link-color: #37769e;
-  --kbin-link-hover-color: #275878;
-  --kbin-outline: #ff8c00 solid 4px;
-
-  --kbin-primary-color: #61366b;
-  --kbin-text-muted-color: #95a6a6;
-
-  --kbin-success-color: #0f5132;
-  --kbin-danger-color: #842029;
-
-  --kbin-own-color: #0f5132;
-  --kbin-author-color: #842029;
 
   --kbin-rounded-edges-radius: .5rem;
 
-  // section
-  --kbin-section-bg: #fff;
-  --kbin-section-text-color: var(--kbin-text-color);
-  --kbin-section-title-link-color: var(--kbin-link-color);
-  --kbin-section-link-color: var(--kbin-link-color);
-  --kbin-section-link-hover-color: var(--kbin-link-hover-color);
-  --kbin-section-border: 1px solid #e5eaec;
-  --kbin-author-border: 1px dashed var(--kbin-author-color);
-  --kbin-own-border: 1px dashed var(--kbin-own-color);
-
-  // meta
-  --kbin-meta-bg: none;
-  --kbin-meta-text-color: #606060;
-  --kbin-meta-link-color: #606060;
-  --kbin-meta-link-hover-color: var(--kbin-link-hover-color);
-  --kbin-meta-border: 1px dashed #e5eaec;
-  --kbin-avatar-border: 3px solid #ecf0f1;
-  --kbin-avatar-border-active: 3px solid #d3d5d6;
-  --kbin-blockquote-color: #0f5132;
-
-  // options
-  --kbin-options-bg: #fff;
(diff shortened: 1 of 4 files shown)


nobodyatroot (approved):

e-five256:
@melroy89 @asdfzdfj @simonrcodrington wanted to check if there were any concerns with this change. I think it's a good change overall, keeping all themes equal with each other (which I think is less confusing from a code organization perspective), and opens up further discussion to be had about the default theme and potentially changing it, this was just a quick start of handling the flashbang issue on first site load. I've tried to go around to all pages and actions I could find in light theme and I don't _think_ I missed anything but will try to fix quickly if someone does notice a style issue

assets/styles/_variables.scss
```diff
@@ -15,150 +15,29 @@ $aspect-ratios: (
 ) !default;
 
 :root {
-  --kbin-body-font-family: system-ui, -apple-system, "Segoe UI", Roboto, "Helvetica Neue", "Noto Sans", "Liberation Sans", Arial, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
```
melroy89, line 18: I think you should keep the common settings/CSS variables in this `assets/styles/_variables.scss`.

melroy89 (changes requested):
Restore the common variables in variables.scss. Change default theme selection in the sidebar as well. This requires changes in the template. EDIT: and see my comment below as well to change the default to Tokyo Night.

melroy89:
Furthermore, my proposal is not to use the dark theme by default, but instead a much better theme and fit, which is the Tokyo Night theme (later we have NeonTokyo), but for now I prefer to have Tokyo Night theme (which is also dark) over the dark theme.

assets/styles/_variables.scss
melroy89: yea I see, you are saying this is causing this flash issues?

e-five256:
I think changing the default theme might be a larger discussion for the entire group:

When I spoke to a friend who uses light themes if they ever experienced the same "flashbang" sense the opposite produces, their response to me was "??? sites don't randomly change themes"

That made me think, do we really want to have this javascript which changes the theme based on system settings? Should we just get rid of it and swap `theme--default` to something else, tokyo night you suggest or any of the other themes.

Right now it'd be incredibly jarring on a first visit, in my opinion, if the default theme was tokyo night and then javascript suddenly changed it to default dark because of your system setting. I don't believe we should have both the system theme detection javascript **and** a default theme that isn't either dark or light.  One of the other would have to go

e-five256:
> Restore the common variables in variables.scss

Do you mean all settings, or just the font-family one? I'm not sure I see a point to either. Every single theme defines the font family one and all others that I moved out:

https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_dark.scss#L2
https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_kbin.scss#L2
https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_solarized-dark.scss#L19
https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_solarized-light.scss#L19
https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_tokyo-night.scss#L2

melroy89:
Agreed we might want to remove the detection and just make Tokyo night the default. Unless of course the user changed it (which will create a cookie).

melroy89:
> > Restore the common variables in variables.scss
> 
> Do you mean all settings, or just the font-family one? I'm not sure I see a point to either. Every single theme defines the font family one and all others that I moved out:
> 
> https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_dark.scss#L2 https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_kbin.scss#L2 https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_solarized-dark.scss#L19 https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_solarized-light.scss#L19 https://github.com/MbinOrg/mbin/blob/main/assets/styles/themes/_tokyo-night.scss#L2
> 
> Unless I'm missing something about how Sass is compiled, there being a "special" theme in `_variables` you have to be aware of rather than that theme having its own file just adds to confusion (because right now in main, there is no "light" theme
> 
> Edit: To clarify, I kept the ones that are shared between all themes. I only moved out the 92 variables that are overridden in every single theme, as they were used to craft the "light" theme
> 
> Edit 2: I guess in the worst case that for some reason `theme--default` class is removed from the body tag, having them in `_variables` would allow the site to still function. But is your suggestion to then duplicate those settings into a new `_light` file, or continue with the path of light being a special case theme that has no file and falls back to the root level vars?

I would imagine variables file had a good use case. But apparently not.

e-five256:
I think there's a larger discussion people want to have about a default theme, and potentially making a new theme altogether. I was hoping to do a small little stop gap in between then and now to handle the flashbang issue, but it might not be a big problem. I feel like I tend to run into these weird edge cases a lot (e.g. I open everyone's mbin instances in private windows to check on them so I see this a lot), whereas it's probably not even seen by people using their single instance driver daily as usual and having a theme set. Since the change was basically to revert the entire commit, I think it might be easier to restart this approach once there's a theme that we want to make default.

melroy89:
Feel free to make Tokyo night the default for now

PR #201, opened 2023-11-04
Fix duplicate key value for url_idx

melroy89:
- Finally fix those errors: `"message":"Error thrown while handling message App\\Message\\LinkEmbedMessage. Removing from transport after 0 retries. Error: \"Handling \"App\\Message\\LinkEmbedMessage\" failed: An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint \"url_idx\"\nDETAIL:  Key (url)=(https://www.....) already exists.\"","context":{"class":"App\\Message\\LinkEmbedMessage","retryCount":0,"error":"Handling \"App\\Message\\LinkEmbedMessage\" failed: An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint \"url_idx\"\nDETAIL:  Key (url)=(https://www.....) already exists.`

- Clean-up dead code? Unnecessary amount of DB connection for each link message event. It doesn't seem to use this DB connection?!?

### DIFF
--- a/src/MessageHandler/LinkEmbedHandler.php
+++ b/src/MessageHandler/LinkEmbedHandler.php
@@ -7,8 +7,6 @@ namespace App\MessageHandler;
 use App\Message\LinkEmbedMessage;
 use App\Repository\EmbedRepository;
 use App\Utils\Embed;
-use Doctrine\DBAL\DriverManager;
-use Doctrine\ORM\EntityManagerInterface;
 use Psr\Cache\CacheItemPoolInterface;
 use Symfony\Component\Messenger\Attribute\AsMessageHandler;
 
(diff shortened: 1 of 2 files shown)


BentiGorlich (changes requested):
The code looks fine to me however I think before a `LinkEmbedMessage` is dispatched the code should check whether the url already has a preview image. Can you add that as well?

BentiGorlich (approved):
After looking at the code I a bit more (especially how when the message gets dispatched) I changed my mind. It is always send when an entry, post or comment to eigher is created. So I think its fine this way

melroy89:
> After looking at the code I a bit more (especially how when the message gets dispatched) I changed my mind. It is always send when an entry, post or comment to eigher is created. So I think its fine this way

Yes indeed. It just create a entry object, which is fine. But in the [EmbedRepository.php](https://github.com/MbinOrg/mbin/pull/201/files#diff-7999d918a1ca6e4795e33b6eaaba5f94644b2a8ecdb8ed12a83333a7e9246b6b) file I retrieve the `url` from this entity and check it against the DB. If it's returning null the object doesn't yet exists in PostgreSQL. And thus insert a new DB record.

PR #205, opened 2023-11-04
swap readme github links from closed to merged

e-five256:
I think the intent with these links was to show the work and changes that had made it into the repo, so I think showing the merged PRs rather than merged or closed might be more accurate, as closed may return results that did not happen


melroy89 (approved):
Make sense yes!

Issue #69, opened 2023-10-18
Make it easier to add profile picture


TheVillageGuy:
To change the profile picture you now have to go through settings -> profile, though intuitively people go directly to profile, where it can't be changed.

Either add an edit profile button there, or add an edit profile option to the top right user menu that brings you straight to settings->profile (my preference)

Issue opened 3 weeks ago on codeberg

simonrcodrington:
Yeah this one is a bit confusing. The /profile link makes sense because when you go to someone else's profile, you expect to see all of their posts etc.

But when you're logged in, you kinda assume that if you press profile it would take you to your admin profile edit page under settings (why would you want to view your regular front end facing /profile link if you're in that menu?)

I think maybe changing the wording or adding better links here might clarify what's going on? Maybe even rename *profile* to *view profile* and add a new *edit profile* link also? Thoughts on what we can do here?

e-five256:
Definitely super common in my experience for clicking on your own avatar on your own profile page to bring you to the place to change it (can even do so here on github), so I feel like that seems to make sense UI/UX wise

PR #206, opened 2023-11-05
add quick links to change avatar and cover when viewing your own profile

e-five256:
if viewing your own profile, clicking on your avatar or cover will bring you to the page to change them

viewing someone else's profile does not show this link

image
image

(if I do show pointer, it won't show labels, if I do selection, it won't show pointer... just trust me the pointer is over avatar and cover where the label appears. there is a slight no-click zone to the right of the avatar due to the HTML element spacing, but I feel like the cover is a large enough target that users can click anywhere else)

related #69

not sure if it completely solves what @TheVillageGuy wanted to see, but this was one of my suggestions to help, as I feel it's fairly common to have this flow (you can see it here on github if you go to your profile, hovering over your avatar if you have one says "Change your avatar" (I wonder why they went with your rather than my :thinking: ) and clicking that just directs you to the settings page

### DIFF
--- a/templates/components/user_box.html.twig
+++ b/templates/components/user_box.html.twig
@@ -1,7 +1,7 @@
 <div{{ attributes.defaults({class: 'user-box'}) }}>
     <div class="{{ html_classes({'with-cover': user.cover, 'with-avatar': user.avatar }) }}">
         {% if user.cover %}
-          {% if app.user is same as user %}
+          {% if app.user is same as user and is_route_name('user_overview') %}
             <a href="{{ path('user_settings_profile') }}" title="{{ 'change_my_cover'|trans }}" aria-label="{{ 'change_my_cover'|trans }}">
           {% endif %}
             <img class="cover image-inline"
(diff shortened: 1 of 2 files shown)


nobodyatroot (approved):
Makes sense to me, verified it works as expected.

melroy89:
Ow btw... This now only works when you are on profile overview page. Let's say I'm on a sub-page like profile comments or profile followers, this quick links will NOT appear..

e-five256:
I think that was 100% my intention, I don't think they should appear everywhere. Only for users that click the profile link expecting it to be a way to change their avatar. I might be misunderstanding, which routes are those?

melroy89:
ok so besides `user_view`, the routes are: `user_entries`, `user_comments`, `user_posts`, `user_replies`, `user_boosts`, `user_following`, `user_followers`, `user_subscriptions` and `user_reputation` (so basically those sub-profile-pages)..

e-five256:
following up from matrix convo, these routes are likely gaining the avatar / cover soon, but they don't on main yet. once it is I'll do a followup to add it to those routes as well

PR #209, opened 2023-11-05
Improve .gitignore & FAQ

melroy89:
- Add the file `public/media` to `.gitignore` as well, since I use a symbolic link under Linux now for the media directory. Which is seen as "a file".
- FAQ changed due to PR merged: https://github.com/MbinOrg/mbin/pull/208

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -17,6 +17,7 @@ tools/vendor/
 # Mbin specific
 .env
 /public/media/*
+/public/media
 docker/compose.override.yml
 yarn.lock
 /metal/
--- a/FAQ.md
+++ b/FAQ.md
@@ -124,10 +124,12 @@ That said, if you are _experiencing serious issues_ with your instance which you
 
 ## I changed my .env configuration but the error still appears/new config doesn't seem to be applied?
 
-After you edited your `.env` configuration file on a bare metal/VM setup, you always need to execute the dump-env command manually (this is **not** needed on Docker):
+After you edited your `.env` configuration file on a bare metal/VM setup, you always need to execute the `composer dump-env` command (in Docker you just restart the containers).
+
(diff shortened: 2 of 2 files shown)


.gitignore
```diff
@@ -17,6 +17,7 @@ tools/vendor/
 # Mbin specific
 .env
 /public/media/*
+/public/media
```
e-five256, line 20: I think you could also do a single rule of `/public/media*` to cover both situations but that is a minor nitpick

FAQ.md
```diff
@@ -124,10 +124,12 @@ That said, if you are _experiencing serious issues_ with your instance which you
...
 ## I changed my .env configuration but the error still appears/new config doesn't seem to be applied?
 
-After you edited your `.env` configuration file on a bare metal/VM setup, you always need to execute the dump-env command manually (this is **not** needed on Docker):
+After you edited your `.env` configuration file on a bare metal/VM setup, you always need to execute the `composer dump-env` command (in Docker you just restart the containers).
+
+Running the `post-upgrade` script will also execute `composer dump-env` for you:
 
 ```bash
-composer dump-env prod
+./bin/post-upgrade
```
e-five256, line 132: I'm not super knowledgeable on the command as a non-docker user but I assume there isn't a downside to running it when unneeded

e-five256 (approved):

.gitignore
melroy89: but that could potentially also ignore media-blabla files in the public folder right?

FAQ.md
melroy89: no there is no real down-side, maybe some minor cache clean-up, but in general there is no down side in running this command whenever you please.

.gitignore
melroy89: or media.css .. just an example.

PR #211, opened 2023-11-06
Add postmill to credits

melroy89:
Missing credit to Postmill.


nobodyatroot (approved):
Agreed, should be credited since several classes are lifted verbatim. And I think the code blocks in question have already been properly credited with the zlib license attribution.

PR #213, opened 2023-11-06
Rename uk to ua according to ISO 3166

melroy89:


melroy89:
Second try: https://github.com/MbinOrg/mbin/pull/218

PR #207, opened 2023-11-05
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [Mbin/Mbin](https://hosted.weblate.org/projects/mbin/kbin/).

Current translation status:

Weblate translation status

PR #218, opened 2023-11-07
Move Ukrainian translation to other filename (ua)

melroy89:
See title. This is because `uk` is the ISO 3166 country code for United Kingdom and not Ukraine. So that would be `ua` instead.


melroy89:
I will do a weblate update directly after this PR merged.

zen-tools:
@melroy89 its ISO 639-1 codes are used for localizations, not ISO 3166. Please see my comment [here](https://github.com/MbinOrg/mbin/pull/207#issuecomment-1798305689).

melroy89:
> Weblate prefers two letter codes as defined by ISO 639-1

You're right. I will revert again I think.

PR #223, opened 2023-11-08
update sidebar labels for new loc token

e-five256:
minor change, it looks like the translations were changed to `subscription_sidebar_pop_out_left` and `subscription_sidebar_pop_out_right` but the `aria-label` fields were left with the loc token `subscription_sidebar_pop_out` which does not exist


melroy89 (approved):
Yup. Oopsy.

PR #224, opened 2023-11-08
update changes from sync 7

e-five256:
un-lower the menu items
update cross post template "more" menu to match mbin (now that it's copy pasted in two locations it might be worth moving to a component if possible, but that strays further from kbin, wasn't sure on this one)
update tables to use `th` in `thead` sections, match magazine data, use component

will swap base to main once ready, just wanted to make a WIP


melroy89:
you might need to change the merge into branch, since I will merge PR 212 I think now. So you need to merge to `main` as well with this PR :)

melroy89:
THIS IS NOT GOOD.. You should have waited maybe.. Now you have all the changes of kbin sync 7 in this PR as well :(

e-five256:
Well I can't log in to matrix to discuss, but no worries, this should be finnneeee. I've done it a million times. Once you merge main back in all the commits that it shows disappear. Probably. Actually I never work with squash so that might be why this is acting a bit off now that I think about it. But I'm fine with conflicts, not the end of the world, always able to resolve them. I'm more concerned about RTR7 (6 appears to have been skipped)

e-five256:
Ok so in hindsight it doesn't work like that, which I'm guessing is due to the squash, now that the commits are gone it doesn't seem them as the same thing. Just to keep the PR looking clean I can just redo the changes in a new branch, doesn't take much time at all

melroy89:
You can cherry pick commits from this branch into your new branch. If that helps you.

e-five256:
I just redid the changes in #225, was only a handful of files so wasn't a big deal in this case, I'll keep it in mind for the future. I was expecting some turbulence anyways, I just wanted to get a draft up for people to check if I had missed anything obvious that needed tweaking from the RTR

PR #225, opened 2023-11-08
update changes from sync 7

e-five256:
un-lower the menu items
update cross post template "more" menu to match mbin (now that it's copy pasted in two locations it might be worth moving to a component if possible, but that strays further from kbin, wasn't sure on this one)
update tables to use `th` in `thead` sections, match magazine data, use component

### DIFF
--- a/templates/admin/_options.html.twig
+++ b/templates/admin/_options.html.twig
@@ -26,7 +26,7 @@
         <li>
             <a href="{{ path('admin_moderators') }}"
                class="{{ html_classes({'active': is_route_name('admin_moderators')}) }}">
-                {{ 'moderators'|trans|lower }}
+                {{ 'moderators'|trans }}
             </a>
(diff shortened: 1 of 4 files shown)


nobodyatroot (approved):

e-five256:
My only concern is merging this I will make conflicts with RTR7 (6?) https://codeberg.org/Kbin/kbin-core/pulls/1217/files#diff-6073986cdd23e99006fca1d041c3707f09d22958 so I wonder if I should only do the remove-lower and more menu and leave fixing the invalid `td`s in the `thead`s until after RTR7

e-five256:
@melroy89 since you've been taking care of the syncs wanted to get your feelings. Of course if conflicts ever happen because of changes I make, feel free to just throw the conflict at me and tell me to fix it, I don't want to put any extra burden on people

PR #229, opened 2023-11-09
add change my avatar and cover quick links to all user pages

e-five256:
there's nothing fundamentally different in this than #206 but sync 7 added the user avatar / cover to all user non-settings pages (threads, comments, posts, replies, boosts, following, followers, subscriptions, reputation, moderated). this adds the quick links to all those pages when viewing your own profile


nobodyatroot (approved):
Makes sense to me

PR #234, opened 2023-11-10
Fix logic precedence in admin panel sidebar

nobodyatroot:
Minor fix to admin panel sidebar logic so that admins and moderators don't accidently ban, delete or purge themselves. The bug was introduced after introducing OR logic to check for moderator role.


e-five256 (approved):

melroy89:
Late to this party as well. I just wanna say this might also needs a backend change if we want to be more secure. Removing or hiding the button is only one aspect.

PR #235, opened 2023-11-10
Feature remote retrieval mod followers

### LINKED ISSUE #217: Correct Information about remote magazines
The amount of subscribers of remote magazines and users is wrong. It only shows the the amount of local subscribers/followers. 
Additionally the moderator of remote magazines is always the local admin, not the actual moderators of the magazine.

The correct info should be fetched when initially looking a magazine or user up. Afterwards it should periodically refetch this information

Imo the solution above is the only one that is correct

BentiGorlich:
Fetch the remote follower count on updateActor and initial fetch. The remote follower count is used instead of the local follower count if it could be fetched. 

The moderators of a magazine are fetched on updateActor and initial fetch. The admin of a local instance will still be the owner of remote magazines and therefore listed in the `moderator` table. There are just too many places in the code that would need to be adjusted, because they rely on the fact that a magazine always has an owner. The owner of remote magazines is therefore hidden in the moderators list. 

Implement the `/m/{name}/moderators` endpoint and change the `moderators` property to `attributedTo` in the magazine fetch like other services use (e.g.: Lemmy)

Fix #217


src/Factory/ActivityPub/GroupFactory.php
```diff
@@ -48,7 +48,7 @@ public function create(Magazine $magazine): array
             ],
             'summary' => $magazine->description,
             'sensitive' => $magazine->isAdult,
-            'moderators' => $this->urlGenerator->generate(
```
e-five256, line 51: One thing I was wondering is if we should be setting both, to try to work with both lemmy and kbin in the future, as this seems like sort of an API break where we're saying we'll go with fep-1b12 rather than what kbin was planning?

BentiGorlich: I'd say we wait until kbin implements something. Then we can look how we can support it

BentiGorlich:
ToDo List:
- add event for adding moderators and add a message for that
- add event for removing moderators and add a message for that
- figure out a way to regularly update remote magazines
- add magazines to the `kbin:ap:actor:update` command

BentiGorlich:
I think we need to implement #289 before this can hit main. Otherwise the remote magazines on instances that upgrade do not have a visibile moderator anymore (which is kinda true atm) but that is not the intended behaviour. Problem is that actors are only refreshed if one searches for them (which doesn't really make sense)...

FYI @e-five256 @nobodyatroot

BentiGorlich:
I have it up and running on my instance <gehirneimer.de> now. It will take a while to show anything since I put a 1 day limit on the magazine fetch. When a lemmy magazine posts something it should show up quicker, since the time limit in that code path is at 1 hour. 

Should I set it to 1 hour, too?

nobodyatroot:
I updated my "in house" docker cluster with the branch (they're federated and subscribed to each other) and I'm seeing this error? Is this normal before an update happens the first run or something else?

image

image

BentiGorlich:
I really don't know :thinking: 
Maybe it is because I did something wrong copying this:
```php
    #[ManyToOne(targetEntity: User::class, inversedBy: 'moderatorTokens')]
    #[JoinColumn(nullable: false)]
    public User $user;
    #[ManyToOne(targetEntity: User::class, inversedBy: 'moderatorTokens')]
    #[JoinColumn(nullable: true)]
    public ?User $addedByUser;
```
My guess is that the `inversedBy: 'moderatorTokens'` shouldn't be on the new column, but I am not sure ...

BentiGorlich:
I get these kind of errors on my server:
> An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "moderator_magazine_user_idx"                                                                                                                                                      
Detail  Key (magazine_id, user_id)=(51, 16481) already exists.

I guess it is caused by parallel execution of multiple `UpdateActorMessages` with the same target. How do I prevent this?

BentiGorlich:
yep that was the problem

nobodyatroot:
> > My guess is that the `inversedBy: 'moderatorTokens'` shouldn't be on the new column, but I am not sure ...
> 
> yep that was the problem

Confirmed the fix, nice! And everything else in the PR seems to work. After remote posting a thread from one Mbin instance to another the remote mod showed up right away.

Do you plan on adding anything else to this PR or are you pretty happy with where it is?

BentiGorlich:
No I won't add anymore to this PR. It is big enough as it is XD
I will start tro work on more federation stuff. Mainly federating the mod actions and handling the inbox mod actions.

nobodyatroot (dismissed):
Confirmed  federated 3 instance Mbin cluster under docker works as expected with this PR.

**EDIT:** Now running live on [kbin.run](https://kbin.run/)

BentiGorlich:
> Confirmed federated 3 instance Mbin cluster under docker works as expected with this PR.

Can you make a code review as well?
I will definitely let my instance run 1-2 days and look at the failed queue to see if there are more problems

src/Controller/SearchController.php
```diff
@@ -68,8 +73,13 @@ public function __invoke(Request $request): Response
                         }
                     }
                 } catch (\Exception $e) {
+                    $this->logger->warning("an error occurred during lookup of '$query': ".\get_class($e).": {$e->getMessage()}", [$e]);
                 }
+            } else {
+                $this->logger->debug("query doesn't match the pattern...", [$matches]);
             }
+        } else {
+            $this->logger->debug('something bla bla bla', [str_contains($query, '@'), !$this->settingsManager->get('KBIN_FEDERATED_SEARCH_ONLY_LOGGEDIN'), $this->getUser()]);
```
e-five256, line 82: should this string be cleaned up?

src/Service/ActivityPubManager.php
```diff
@@ -370,6 +413,82 @@ public function updateMagazine(string $actorUrl): ?Magazine
             $magazine->apFetchedAt = new \DateTime();
             $magazine->isAdult = (bool) $actor['sensitive'];
 
+            if (null !== $magazine->apFollowersUrl) {
+                try {
+                    $this->logger->info("updating remote followers of magazine $actorUrl");
```
e-five256, line 418: suggestion of switching this and ln 430 to debug level

src/Controller/ActivityPub/Magazine/MagazineModeratorsController.php
```diff
@@ -4,16 +4,60 @@
...
+            $actor = $mod->user;
+            $items[] = $this->manager->getActorProfileId($actor);
+        }
+
+        $id = Uuid::v4()->toRfc4122();
+
+        return [
+            '@context' => ActivityPubActivityInterface::CONTEXT_URL,
+            'type' => 'OrderedCollection',
+            'id' => $this->urlGenerator->generate('ap_object', ['id' => $id], UrlGeneratorInterface::ABSOLUTE_URL),
```
e-five256, line 58: I notice this evaluates a little differently for us than lemmy:

```
"id": "http://127.0.0.1/f/object/a6cb5314-87e5-46d2-9555-513c08bfc15f",
```

vs

```
"id": "https://lemmy.world/c/memes/moderators",
```

src/MessageHandler/ActivityPub/Inbox/AddHandler.php
```diff
@@ -0,0 +1,48 @@
...
+    ) {
+    }
+
+    public function __invoke(AddMessage $message): void
+    {
+        $payload = $message->payload;
+        $actor = $this->activityPubManager->findUserActorOrCreateOrThrow($payload['actor']);
+        $targetMag = $this->magazineRepository->getMagazineFromModeratorsUrl($payload['target']);
+        if (!$targetMag) {
+            throw new \LogicException("could not find a magazine with moderators url like: '{$payload['target']}'");
```
e-five256, line 32: mentioned in matrix but wanted to track in the PR, curious if these are bubbled up to users and need translations or should be using translation tokens, or if they just go to logs

src/Controller/SearchController.php
BentiGorlich: yes it should :D

src/Controller/ActivityPub/Magazine/MagazineModeratorsController.php
BentiGorlich: you're right. I switched it to a generated id so it wouldn't be cached, but now that I clear the cache properly I should be able to change it back

src/MessageHandler/ActivityPub/Inbox/AddHandler.php
BentiGorlich: This is in a Handler, so it will just log to a file

e-five256 (approved):

BentiGorlich:
@e-five256 can you approve again? I pushed some AP fixes. Probably should have kept those out of this PR, but I was live testing :D

migrations/Version20231107204142.php
```diff
@@ -0,0 +1,32 @@
...
+class Version20231107204142 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add column remote_followers_count';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE "user" ADD ap_followers_count INT DEFAULT NULL');
```
melroy89, line 19: If you add new columns, try to add them to the entity in `src/Entity/<class>.php` instead, to avoid DB migration diff issues between Doctrine and the DB.

melroy89: You added it to ActivityPubActorTrait instead I see? hm. OK. not sure if that is good or bad 😖

BentiGorlich: It is added there through the `ActivityPubActorTrait`. Didn't know that that would lead to an update (does it?)

src/Controller/SearchController.php
```diff
@@ -44,14 +46,17 @@ public function __invoke(Request $request): Response
             );
         }
 
+        $this->logger->debug("searching for $query");
```
melroy89, line 49: I know it's "just" debugging. But maybe something like this format might be safer?

```php
   $logger->debug('User {userId} has logged in', [
        'userId' => $this->getUserId(),
    ]);
```

migrations/Version20231107204142.php
melroy89: Yes it's used here: https://github.com/MbinOrg/mbin/blob/239071ad2822acc3f05273da5df8a238343af9ac/src/Entity/User.php#L44 I think this is expected behavior. Assuming this is what we want.

After the main branch changes merged into this branch (which I just did recently). You can verify yourself by executing: `php bin/console doctrine:migrations:diff`

If this result into a new file with additional up and down queries.. well it shouldn't, since then you forgot still something.

src/Controller/SearchController.php
melroy89: Ps. same remark of course on all the other places like line 55, 59, 76, etc.

src/EventSubscriber/ActivityPub/UserWebFingerProfileSubscriber.php
```diff
@@ -40,28 +44,44 @@ public function buildResponse(WebfingerResponseEvent $event): void
...
-                        $this->imageManager->getUrl($actor->avatar),
-                    ); // @todo media url
+                $link->setRel('self')
+                    ->setType('application/activity+json')
+                    ->setHref($accountHref);
                 $jsonRd->addLink($link);
+
+                if ($actor->avatar) {
+                    $link = new JsonRdLink();
+                    $link->setRel('http://webfinger.net/rel/avatar')
```
e-five256, line 79: not quite sure what this hardcoded link is... oh it was in original code? I just noticed it's `http`.... but not sure how it's used

src/Controller/SearchController.php
BentiGorlich: good call

src/EventSubscriber/ActivityPub/UserWebFingerProfileSubscriber.php
BentiGorlich: yeah https wouldn't hurt I think

migrations/Version20231107204142.php
BentiGorlich:
```
docker compose exec php php bin/console doctrine:migrations:diff

In NoChangesDetected.php line 13:
                                                    
  No changes detected in your mapping information.  
                                                    

doctrine:migrations:diff [--configuration CONFIGURATION] [--em EM] [--conn CONN] [--namespace NAMESPACE] [--filter-expression FILTER-EXPRESSION] [--formatted] [--line-length LINE-LENGTH] [--check-database-platform [CHECK-DATABASE-PLATFORM]] [--allow-empty-diff] [--from-empty-schema]
```

e-five256 (dismissed):

e-five256 (approved):

nobodyatroot (approved):
Functional verification after a few days of live testing in production looks good!

PR #237, opened 2023-11-10
Do not automatically subscribe admin OR users to new federated magazine lookups

nobodyatroot:
Disables the default /kbin behavior of automatically subscribing the admin to federated magazine lookups. The searching user must explicitly hit the subscribe button in order for the Mbin instance to pull in federated updates.

This change would also seem to remove the need for the admin panel option to explicitly allow only logged in users to search for federated content since you have to be logged in to subscribe anyway, but I left it intact incase we want to modify this behavior in the future.

### DIFF
--- a/src/Service/MagazineManager.php
+++ b/src/Service/MagazineManager.php
@@ -72,7 +72,7 @@ class MagazineManager
         $this->entityManager->persist($magazine);
         $this->entityManager->flush();
 
-        $this->subscribe($magazine, $user);
+        //$this->subscribe($magazine, $user);
 
         return $magazine;
     }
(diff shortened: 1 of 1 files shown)


e-five256:
Super curious how this flow looks now (especially with the described async annoyance in that:) user searches for new remote mag, gets no results because the process is async, searches again, and... does it show up and load that time? but doesn't subscribe them, so they have to click into that mag and subscribe

At the point before any subscribers, the magazine exists as a new entry in magazine, but without any subscribers it won't get activity events?

BentiGorlich:
Does this work when you are creating a new magazine? I did try it in my PR, but reverted it, because newly created local magazines would always throw a 500 because the owner was missing

nobodyatroot:
I added an condition where local mag creations still get auto subscribed to the user, _now_ this will only impact federated magazines. :-)

BentiGorlich (approved):

melroy89:
Thank god.

PR #240, opened 2023-11-10
Change 500 error message

melroy89:
- Change the 500 error description
- Change 500 error title
- Add link to other instances (introducing `%link_start%` and `%link_end%` to avoid XSS attacks via Weblate)
- Fix security issues in Symfony Composer (ran `composer update symfony/*`):
  - CVE-2023-46733: Possible session fixation
  - CVE-2023-46734: Potential XSS vulnerabilities in CodeExtension filters
- Ran PHPUnit recipe

### DIFF
--- a/templates/bundles/TwigBundle/Exception/error500.html.twig
+++ b/templates/bundles/TwigBundle/Exception/error500.html.twig
@@ -15,6 +15,8 @@
 {% block body %}
     <div id="content" class="section section--top section--muted">
         <p>{{'errors.server500.title'|trans}}</p>
-        <small>{{'errors.server500.description'|trans}}</small>
+        <small>{{ 'errors.server500.description'|trans({
+                '%link_target%': 'https://fedidb.org/software/mbin'
+            })|raw }}</small>
     </div>
 {% endblock %}
--- a/templates/bundles/TwigBundle/Exception/error500.html.twig
+++ b/templates/bundles/TwigBundle/Exception/error500.html.twig
@@ -16,7 +16,8 @@
     <div id="content" class="section section--top section--muted">
         <p>{{'errors.server500.title'|trans}}</p>
         <small>{{ 'errors.server500.description'|trans({
-                '%link_target%': 'https://fedidb.org/software/mbin'
+                '%link_start%': '<a href="https://fedidb.org/software/mbin">'
+                '%link_end%': '</a>'
             })|raw }}</small>
     </div>
 {% endblock %}
--- a/templates/bundles/TwigBundle/Exception/error500.html.twig
+++ b/templates/bundles/TwigBundle/Exception/error500.html.twig
@@ -16,7 +16,7 @@
     <div id="content" class="section section--top section--muted">
         <p>{{'errors.server500.title'|trans}}</p>
         <small>{{ 'errors.server500.description'|trans({
-                '%link_start%': '<a href="https://fedidb.org/software/mbin">'
+                '%link_start%': '<a href="https://fedidb.org/software/mbin">',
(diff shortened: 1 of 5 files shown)


translations/messages.en.yaml
```diff
@@ -524,41 +538,49 @@ federation_page_allowed_description: Known instances we federate with
...
   want your account deleted.
 request_account_deletion_button: Request account deletion
 more_from_domain: More from domain
-errors.server500.title: 500 Server error
-errors.server500.description: Sorry, something went wrong on our end. We're working
-  on resolving this issue, please check back soon.
+errors.server500.title: 500 Internal Server Error
+errors.server500.description:
+  Sorry, something went wrong on our end. If you continue to see this error, try contacting the instance owner.
+  If this instance is not working at all, check out <a href="%link_target%">other Mbin instances</a> in the meanwhile until the problem is resolved.
```
melroy89, line 549: It's about this... ignore the rest of the indents I guess.

translations/messages.en.yaml
```diff
@@ -524,41 +538,49 @@ federation_page_allowed_description: Known instances we federate with
...
   want your account deleted.
 request_account_deletion_button: Request account deletion
 more_from_domain: More from domain
-errors.server500.title: 500 Server error
-errors.server500.description: Sorry, something went wrong on our end. We're working
-  on resolving this issue, please check back soon.
+errors.server500.title: 500 Internal Server Error
+errors.server500.description:
+  Sorry, something went wrong on our end. If you continue to see this error, try contacting the instance owner.
+  If this instance is not working at all, check out %link_start%other Mbin instances%link_end% in the meanwhile until the problem is resolved.
```
melroy89, line 549: It's about this... ignore the rest of the indents I guess.

templates/bundles/TwigBundle/Exception/error500.html.twig
```diff
@@ -15,6 +15,9 @@
 {% block body %}
     <div id="content" class="section section--top section--muted">
         <p>{{'errors.server500.title'|trans}}</p>
-        <small>{{'errors.server500.description'|trans}}</small>
+        <small>{{ 'errors.server500.description'|trans({
+                '%link_start%': '<a href="https://fedidb.org/software/mbin">',
```
e-five256, line 19: my only concerns are linking to a domain not under the control of the instance owner. Not that I imagine dansup would ever do something malicious, but if anything happened to the domain or it was ever released, there is a risk involvement there, not sure how instance owners feel about it

melroy89: Technically it's under the control of the instance owner if something is really wrong with this domain, also the owner can still change the url by just changing this template file now and updating the URL.

I'm personally fine for now to link to this domain of Daniel for a list of Mbin instances.

nobodyatroot (approved):
I'm indifferent to this change either way. It's probably good to remind the user there are other instances out there, but I ultimately think the onerous is on them to vet whichever instance they end up using... which also should have been the same process used to end up at the starting instance in the first place. Frequent tagged Mbin releases should at least let the users know which servers have been recently updated so as to not pick a "dead" or unmaintained instance.

melroy89:
For the record this message is not to intend users to migrate and use the another instance indefinitely. Instead it's just to inform the user to temporarily give them the option to visit content on a working instance.

templates/bundles/TwigBundle/Exception/error500.html.twig
melroy89: Ps. Feel free to give it a follow up pr introducing an admin panel option and db record.

PR #243, opened 2023-11-11
Fix returning deleted admin

### LINKED ISSUE #67: New magazines after a search are created with deleted user and related problems
I suspect by default user 1 is used. **Update: it appears now the latest admin is used, maybe as a fix to this original problem**. A search for the first active admin should be done and applied. Fixing this can also fix the ghost moderators, if there's ever going to be automatic maintenance/data integrity checks.

I've fixed it by editing the database, but they were all linked to user 1 who no logger exists due to user verification issues at the time.
Then I removed the admin role from (deleted) user 1
**Update: this led to searching for remote magazines no longer working at all.**

Issue opened 3 weeks ago by me on codeberg

BentiGorlich:
- the `findAdmin` method didn't filter for `isDeleted`, yet

Fix #67

### DIFF
--- a/src/Repository/UserRepository.php
+++ b/src/Repository/UserRepository.php
@@ -442,10 +442,16 @@ class UserRepository extends ServiceEntityRepository implements UserLoaderInterf
     public function findAdmin(): User
     {
         // @todo orderBy lastActivity
-        return $this->createQueryBuilder('u')
+        $result = $this->createQueryBuilder('u')
             ->andWhere("JSONB_CONTAINS(u.roles, '\"".'ROLE_ADMIN'."\"') = true")
+            ->andWhere('u.isDeleted = false')
             ->getQuery()
-            ->getResult()[0];
+            ->getResult();
+        if (0 === \sizeof($result)) {
+            throw new \Exception('the server must always have an active admin account');
+        }
+
+        return $result[0];
     }
 
     public function findUsersSuggestions(string $query): array


nobodyatroot (approved):

src/Repository/UserRepository.php
```diff
@@ -442,10 +442,16 @@ public function findBannedPaginated(
     public function findAdmin(): User
     {
         // @todo orderBy lastActivity
-        return $this->createQueryBuilder('u')
+        $result = $this->createQueryBuilder('u')
             ->andWhere("JSONB_CONTAINS(u.roles, '\"".'ROLE_ADMIN'."\"') = true")
+            ->andWhere('u.isDeleted = false')
             ->getQuery()
-            ->getResult()[0];
```
e-five256, line 448: I know I'm late but I wonder why this didn't have a limit applied to it :thinking:

BentiGorlich: I mean how many admins will an instance have :D 
I'd say most instances are running with exactly one admin

nobodyatroot: hmm, i wonder what happens where there's more than 1, though... it might hose whatever is expecting only 1 like assigning a magazine owner? you're probably right about there usually only being 1 admin, but these edge cases might come back to bite later.

e-five256: It's not important, I just feel like if you already know the limitations you expect, it's good to declare, so I might want to add `->setMaxResults(1)` to this or something, but also like I said not important

BentiGorlich:
well it just takes the first onw, so the edge case is kinda solved. The whole owner thing justneedstobe removed... Maybe we could say the mod that has been a mod the longest is the "owner", but I digress 😅

Am 11. November 2023 01:39:15 MEZ schrieb debounced ***@***.***>:

src/Repository/UserRepository.php
TheVillageGuy: > I mean how many admins will an instance have :D I'd say most instances are running with exactly one admin

Mine has three

PR #245, opened 2023-11-11
cleanup from sync 8

e-five256:
- hide abandoned magazine link while logged out
- hide request ownership from non visible accounts (marked for deletion, suspended)
- `th` tag usage in `thead` for admin panels
- unlower new "mod requests" panel entry
- add dumpster icon to purge to fix spacing
- disallow admin panel (ban / suspend user) when viewing the admin by either the global mods or another admin

### DIFF
--- a/templates/user/overview.html.twig
+++ b/templates/user/overview.html.twig
@@ -11,7 +11,7 @@
 {% endblock %}
 
 {% block sidebar_top %}
-    {% if (is_granted('ROLE_ADMIN') or is_granted('ROLE_MODERATOR')) and app.user != user and 'ROLE_ADMIN' not in user.roles %}
+    {% if (is_granted('ROLE_ADMIN') or is_granted('ROLE_MODERATOR')) and app.user != user and not user.admin() %}
         <div class="section magazine">
             <h3>{{ 'admin_panel'|trans }}</h3>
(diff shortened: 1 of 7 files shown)


melroy89 (dismissed):

templates/user/overview.html.twig
```diff
@@ -11,7 +11,7 @@
 {% endblock %}
 
 {% block sidebar_top %}
-    {% if (is_granted('ROLE_ADMIN') or is_granted('ROLE_MODERATOR')) and app.user != user %}
+    {% if (is_granted('ROLE_ADMIN') or is_granted('ROLE_MODERATOR')) and app.user != user and 'ROLE_ADMIN' not in user.roles %}
```
e-five256, line 14: it seems I'm the first one to use `in user.roles` which really surprised me (and also sets off red flags). I'm not sure if there is any reason it was avoided

e-five256: seems this was accurate, @BentiGorlich found an easier / perhaps more accurate if ever changes way of calling `.admin()` which calls `isAdmin()` (which basically does the same thing `return \in_array('ROLE_ADMIN', $this->getRoles());`) but seems safer way of doing it

melroy89 (approved):

PR #246, opened 2023-11-11
nsfw entries - blur by default, unblur via JS hover event

e-five256:
This blurs nsfw threads, microblogs, and comments by setting the blur on the html element

currently this is only done in https://github.com/MbinOrg/mbin/blob/main/assets/controllers/subject_controller.js#L419
this means if JS is slow to load for any reason, the user gets to see all nsfw images until JS finishes loading and then blurs the images (technically, this always happens to some extent, the js tags are `defer` so wait until the DOM is finished loading before executing, how long that takes seems to vary)

potential downsides:
if the user has JS disabled, hovering over the image won't unblur it, it will be stuck in a blurred state rather than the current always unblurred with JS disabled, but expanding the preview or, of course, going to the image url would produce the unblurred image

### DIFF
--- a/templates/components/entry.html.twig
+++ b/templates/components/entry.html.twig
@@ -88,6 +88,7 @@
                                rel="{{ get_rel(is_route_name('entry_single') ? entry.url : entry_url(entry)) }}">
                                 <img class="thumb-subject"
                                      src="{{ asset(entry.image.filePath) |imagine_filter('entry_thumb') }}"
+                                     {% if entry.isAdult %}style="filter: blur(8px)"{% endif %}
                                      alt="{{ entry.image.altText }}">
                             </a>
                         </figure>
@@ -100,6 +101,7 @@
                                class="{{ html_classes({'thumb': is_route_name('entry_single')}) }}">
                                 <img class="thumb-subject"
                                      src="{{ asset(entry.image.filePath) |imagine_filter('entry_thumb') }}"
+                                     {% if entry.isAdult %}style="filter: blur(8px)"{% endif %}
                                      alt="{{ entry.image.altText }}">
                             </a>
                         </figure>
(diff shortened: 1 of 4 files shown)


e-five256:
Earlier we had a conversation about making sure the site always works without javascript, and then I went and made this PR, so I'm trying to think if there is an alternative to still not have to deal with waiting for JS to load, but also not having the issue of showing NSFW until the DOM fully loads

nobodyatroot:
Either way, I think we need to do something different than what's currently implemented... and for now this seems like a fair compromise. I have it running on kbin.run.

nobodyatroot:
Also, we should probably consider adding the blur to the magazine icons on 18+ mags since those can and often are NSFW themselves.

e-five256:
You're running a draft?! 😆 I'll try to test this a bit more. I want to run it by asdfzdfj when there's time. I also guess it might be nice to see how any thing else handles nsfw, like mastodon, but the problem is I imagine I need accounts and such to see how they handle it, but I don't have anything besides mbin anymore >.< so if anyone knows better implementations, let me know

e-five256:
Doing magazine box icons, and I assume like magazine inline as well, might be a bit more work, it's very tied to the DOM structure currently for unblurring, it's possible it can be rewritten to be more generic on things, but this acts as a stop gap so people don't have to deal with constantly seeing nsfw for a second

nobodyatroot (approved):
I think this is a much needed change vs. the behavior right now... the blur is basically useless until the page is finished loading. While it may not be the final solution to this "problem", I think it's worth merging it as at least a temporary stop gap measure until a better solution is found.

e-five256:
Think this is a quick enough revert if peeps disagree, I'm looking into change it all to `data-action` events and adding magazine icons to that where they appear now

PR #247, opened 2023-11-11
Add badges to admins, mods and OPs in comments

BentiGorlich:
Looks like this:
1. kbin theme:  grafik
2. light theme: grafik
3. dark theme: grafik
4. sunny light theme: grafik
5. sunny dark theme: grafik
6. Tokyo Night theme: grafik

### DIFF
--- a/assets/styles/components/_comment.scss
+++ b/assets/styles/components/_comment.scss
@@ -61,6 +61,13 @@ $comment-margin-sm: .3rem;
         font-weight: normal;
       }
     }
+
+    .user-badge {
+      border: var(--kbin-section-border);
+      padding: 0.25rem 0.5rem;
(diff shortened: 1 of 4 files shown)


e-five256:
Would we want a label for global mods as well?

BentiGorlich:
> Would we want a label for global mods as well? Edit, or maybe that's what `user.moderator()` is? I wonder if we'd want it to say something else if so, to clarify they aren't someone in the mod list in the sidebar but someone with global powers

Yes that exactly what the `user.moderator()` call is doing

e-five256:
I think it'd be better to have a Global Mod label / translation in that case, at least in my opinion

e-five256 (approved):

melroy89:
Love it!

PR #252, opened 2023-11-12
Improve performance of subscription panel

BentiGorlich:
- move the array sorting to SQL for performance reasons

### DIFF
--- a/src/Twig/Components/SidebarSubscriptionComponent.php
+++ b/src/Twig/Components/SidebarSubscriptionComponent.php
@@ -6,8 +6,9 @@ namespace App\Twig\Components;
 
 use App\Controller\User\ThemeSettingsController;
 use App\Entity\Magazine;
-use App\Entity\MagazineSubscription;
 use App\Entity\User;
+use App\Repository\MagazineRepository;
+use App\Utils\SubscriptionSort;
 use Symfony\UX\TwigComponent\Attribute\AsTwigComponent;
 use Symfony\UX\TwigComponent\Attribute\PostMount;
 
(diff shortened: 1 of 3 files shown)


src/Twig/Components/SidebarSubscriptionComponent.php
```diff
@@ -25,22 +26,22 @@ class SidebarSubscriptionComponent
...
-        }
         if (ThemeSettingsController::ALPHABETICALLY === $this->sort) {
-            usort($this->magazines, fn ($a, $b) => $a->name > $b->name ? 1 : -1);
+            $this->magazines = $this->magazineRepository->findMagazineSubscriptionsOfUser($this->user, SubscriptionSort::Alphabetically, $max);
         } else {
-            usort($this->magazines, fn ($a, $b) => $a->lastActive < $b->lastActive ? 1 : -1);
+            $this->magazines = $this->magazineRepository->findMagazineSubscriptionsOfUser($this->user, SubscriptionSort::LastActive, $max);
         }
-        if (\sizeof($this->magazines) > $max) {
+        if (\sizeof($this->magazines) === $max) {
```
melroy89, line 43: sorry equal to max?

melroy89: maybe `>=` ?

melroy89: is there a type check in PHP like `>==` xD?

src/Twig/Components/SidebarSubscriptionComponent.php
```diff
@@ -25,22 +26,22 @@ class SidebarSubscriptionComponent
...
             $this->tooManyMagazines = true;
         }
-        $this->magazines = \array_slice($this->magazines, 0, $max);
```
melroy89, line 44: yea this of course was indeed not ideal...  lol

src/Twig/Components/SidebarSubscriptionComponent.php
BentiGorlich: The reason is, the max is 50, so the array can never be bigger than 50. Before the change the array was not limited in size

src/Twig/Components/SidebarSubscriptionComponent.php
BentiGorlich: yep. Fetching **all** magazines, then sorting them and then discarding all but 50 of them is really not ideal performance wise

src/Twig/Components/SidebarSubscriptionComponent.php
melroy89: Yea ok..

src/Twig/Components/SidebarSubscriptionComponent.php
melroy89: Not exactly.. In fact first fetching all magazines by looping over all elements and adding them manually to an array is also less than ideal xD

PR #251, opened 2023-11-12
Kbin sync 9

melroy89:
- Introduce HTML Twig cache for cross-posting threads
- Improve request magazine ownership buttons (for none-magazine owners)
- Only check on thread url (`e.url`) if present otherwise use the thread title (`e.title`)
- Change template regarding cross-posts
- Filter on magazine on moderator_requests page

_Note:_ For now still skipping the user visibility where clauses.  

PR:
- https://codeberg.org/Kbin/kbin-core/pulls/1227
- https://codeberg.org/Kbin/kbin-core/pulls/1229

OUT OF SCOPE:
- https://codeberg.org/Kbin/kbin-core/pulls/1224
- https://codeberg.org/Kbin/kbin-core/pulls/1225

Co-committed-by: Ernest Wiśniewski <ernestwisniewski2@gmail.com>

PR #253, opened 2023-11-12
button changes to request for mod or ownership

e-five256:
Right now on `m/{magazine}/moderators`:

Apply for moderator: Appears when logged out, or logged in and not a moderator of the magazine
Request magazine ownership: Appears when logged out (always, regardless of abandoned, it's missing a `()`, likely a bug upstream) or when logged in and magazine is abandoned

After this:

Apply for moderator: Appears only when logged in and not moderator of the magazine and account is visible (not marked for delete, not suspended)
Request magazine ownership: Appears only when magazine is abandoned and logged in and not moderator of the magazine and account is visible (not marked for delete, not suspended)

This PR is pointed at #251

### DIFF
--- a/templates/magazine/moderators.html.twig
+++ b/templates/magazine/moderators.html.twig
@@ -15,7 +15,7 @@
 {% block body %}
     <h1>{{ 'moderators'|trans }}</h1>
     <div class="flex">
-        {% if not app.user or not magazine.userIsModerator(app.user) %}
+        {% if app.user and not magazine.userIsModerator(app.user) %}
             <form action="{{ path('magazine_moderator_request', {name: magazine.name}) }}"
                   method="POST"
                   class="float-end mb-2"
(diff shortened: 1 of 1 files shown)


melroy89:
> not marked for delete, not suspended

You mean not marked for delete, but suspended? Since currently we use that visibility field for that.

e-five256:
If you request your account to be deleted, I believe your visibility changes to `VISIBILITY_SOFT_DELETED`. if you are suspended, I think it changes to `VISIBILITY_TRASHED`

so neither of those can request to be mod or take ownership

hmm there is also `VISIBILITY_PRIVATE`, I missed that, so they would also not be able to do those... what is that used for...

e-five256:
I don't mind if this isn't found useful (can revert what I did on the abandoned list page as well if desired)

I sort of have a different view to UX than kbin I think, which causes these things where I believe:

Don't give users too many systems at once (imagine getting tons of new terms you haven't seen before, not understanding exactly what each means, it's easy to be overloaded). Not trying to stop people from having options, but seeing terms like abandoned or requesting ownership or request moderator, when you're just browsing the site with not account and can't even act on it seems like a bad experience to me. I guess you can argue it's inconsistent because if the guest user does make an account, they might not know the button is now there, so I guess there is a balance.

The other part of this change was, it just seems strange to allow someone that wants their account deleted to be able to request ownership of magazines. I would also consider closing all open requests to do so but obviously there's a desire to rewrite account deletion on our end

nobodyatroot:
I think it makes sense, but I'll let @melroy89 make the ultimate decision since it's his branch this would merge into.

melroy89 (approved):

PR #254, opened 2023-11-12
nsfw - add blurring of nsfw magazine icons

e-five256:
add controller for blurring nsfw images
blur nsfw magazine icons, inline icons
replacing blurring of entries, entry comments, posts, post comments

followup to #246 where it was brought up to blur magazine icons

image
image
image

entry (thread) comment
image

post (microblog) comment
image

<details>
<summary>hover states of the new mag blurring:</summary>

image
image
image

</details>

### DIFF
--- a/templates/magazine/_list.html.twig
+++ b/templates/magazine/_list.html.twig
@@ -14,7 +14,7 @@
                             {% if magazine.icon %}
                                 <figure>
                                     <img width="32" height="32"
-                                         class="{{ magazine.isAdult ? 'image-adult' : '' }}"
+                                         class="image-inline {{ magazine.isAdult ? 'image-adult' : '' }}"
                                          src="{{ asset(magazine.icon.filePath) | imagine_filter('avatar_thumb') }}"
                                          {% if magazine.isAdult %}data-controller="thumb" data-action="mouseover->thumb#adult_image_hover mouseout->thumb#adult_image_hover_out"{% endif %}
                                          alt="{{ magazine.name ~' '~ 'avatar'|trans|lower }}">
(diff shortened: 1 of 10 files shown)


templates/magazine/_list.html.twig
```diff
@@ -14,7 +14,9 @@
                             {% if magazine.icon %}
                                 <figure>
                                     <img width="32" height="32"
+                                         class="{{ magazine.isAdult ? 'image-adult' : '' }}"
                                          src="{{ asset(magazine.icon.filePath) | imagine_filter('avatar_thumb') }}"
+                                         {% if magazine.isAdult %}data-controller="thumb" data-action="mouseover->thumb#adult_image_hover mouseout->thumb#adult_image_hover_out"{% endif %}
```
e-five256, line 19: I was trying to track down where this is used. I might be missing it, but it looks like the magazine list page has table view and card view. I do not see a "columns" view, so I'm not sure this `elseif` gets called,  unless I'm on the wrong page

e-five256: Looks like you can still route to it manually so I went ahead and made it work and fixed the alt overflow

image

e-five256:
I currently chose to not blur the magazine icon on your subscription sidebar or subscription page in settings. Let me know if people think I should.

nobodyatroot (approved):
Nice follow on!

And +1 about hiding **all** NSFW stuff for logged out users. Maybe we should even have an option in the future to let the instance admin define this behavior, but you'll hear no argument from me about making this the default.

e-five256:
Things to follow up on:

- Blur magazine subscriptions for users
- Blue magazine blocks for users
- Potentially add `18+` in some locations like the subscription columns / card view, user subscriptions, maybe next to the magazine name inline on entries? (Or is that too redundant since every nsfw magazine will have its entries already marked nsfw)
- Potentially do more filtering / something else. The blur isn't super effective
- We hide nsfw entries when logged out, which appears to be different than lemmy, should we?
  - if so, should we hide nsfw magazines completely or just hide their icons

PR #256, opened 2023-11-13
Allow admins to create a sitewide notification below the header

nobodyatroot:
Until now there hasn't been a way for admins to make site wide announcements to notify users of system maintenance or other serious news. This PR adds a notification bar below the header only if an announcement message has been set.

Screenshot 2023-11-13 150225

I didn't want to make a separate admin panel tab for it, so it's under the pages tab for now... our admin panel needs some serious TLC anyway.

I've also fixed a bug in the other the admin pages where once something is set you were unable to clear it/set it to null again without a 500 error being generated.

### DIFF
--- a/assets/styles/app.scss
+++ b/assets/styles/app.scss
@@ -14,6 +14,7 @@
 @import 'layout/alerts';
 @import 'layout/forms';
 @import 'layout/images';
+@import 'components/announcement';
 @import 'components/topbar';
 @import 'components/header';
 @import 'components/sidebar';
--- a/assets/styles/components/_announcement.scss
+++ b/assets/styles/components/_announcement.scss
@@ -1,0 +1,24 @@
+.announcement {
+  padding: 0.75rem;
+  position: relative;
+
+  p {
+    margin: 0;
+    text-align: center;
+  }
+
+  a{
+    font-weight: bold;
+  }
+
+  &__info {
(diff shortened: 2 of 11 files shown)


melroy89:
> I didn't want to make a separate admin panel tab for it, so it's under the pages tab for now... 

I understand, but pages isn't really the correct place either I think, because it's not a page (nor the template component is stored in the `page` folder). What about putting it under _Settings_ tab?

nobodyatroot:
> > I didn't want to make a separate admin panel tab for it, so it's under the pages tab for now...
> 
> I understand, but pages isn't really the correct place either I think, because it's not a page (nor the template component is stored in the `page` folder). What about putting it under _Settings_ tab?

That works too, I'll make that adjustment.

nobodyatroot:
so that settings page is a little more tricky than i thought... im not seeing an easy way to port over what i did without reworking a bunch of stuff in there. the settings page also mostly consists of things we have env vars assigned to... so we can keep this PR open if you want to take a crack at fixing what i did or we can make a follow on PR to adjust this and add the twig cache + whatever else needs fixing.

melroy89 (approved):
Ah fine for now.

melroy89:
.. Assuming it doesn't hurt performance. . ;P

PR #257, opened 2023-11-13
remove system-ui font due to display issues

e-five256:
@BentiGorlich noticed that there were some display issues he was seeing. After investigating, I noticed it was due to the `system-ui` font, which led me to google and finding this blog post about avoiding the font and how it has been removed from github / bootstrap js: https://infinnie.github.io/blog/2017/systemui.html

before:

image

after:

image

@simonrcodrington @asdfzdfj not sure if either of you have experience with this. I feel like the change makes the fonts look slightly worse on my browser / os at least, like slightly less anti-aliased, so maybe it needs an additional change like putting a different font first or some other adjustments


assets/styles/components/_header.scss
```diff
@@ -288,8 +288,7 @@
     height: 100%;
 
     span {
-      font-weight: lighter;
```
e-five256, line 291: this part felt a bit strange for what it was doing (the `/m/` in the header when viewing a magazine) `lighter` may or may not work in chromium, seems to but it all depends on the font I think, if there is no `lighter` version of the font it can't do anything, and the `font-size` was offsetting the `/m/` in other fonts (making it raised above like an exponent or something). I'll double check around but feel like this sort of preserves the way it looked for me, with other fonts

simonrcodrington (dismissed):
I think the fact that bootstrap has taken it out of us defaults is probably a good indicator as to what we should do, if using that gives inconsistency across different OS versions / languages then it's probably best we get rid of it also. We can also come back and readdress this down the road if need to 🦙

melroy89:
Looks like some Zalgo text to me: https://lingojam.com/ZalgoText

e-five256:
Yea, haha, that's what Benti noticed was causing display issues for us where it doesn't for others.

So this did fix it for me... in linux. I just booted up windows and it's looking a bit different. Maybe I should've opened this as a draft, there's obvious concerns with changing the font of the entire site everywhere. I think it objectively looks worse now, but there are _so many_ issues for this font linked on github. I did slightly lie, or rather the blog post is out of date, bootstrap added this font back in v5, but even that PR to add it back has a ton of linked issues saying them doing so broke things. I might have to toy with this a bit more, difficult to find fonts that work for all devices of all years and OSs

e-five256:
Looking around how a lot of sites do things, I ran across a mastodon PR where they mention they have a user setting to enable system ui font instead. This might be the direction I go if that's still the case, need to make sure that wasn't outdated info but that makes sense to me, allow the system ui font to be opt-in for people who have a specific font they want to use, but don't use it by default due to the display issues in certain languages / regions

e-five256:
Alright based on the recent convos about issues with emojis and reading a bunch more, just going to mark this as ready for review now. If we want to bring system ui back can always revert or add a user setting as well. I've tested this in firefox / chromium / linux / windows / android but not mac or ios so it might be worth checking with that or we can just see if there's user feedback on it as we go

---

fonts were changed based on what github / wordpress uses; removing system-ui and the noto color emoji that seemed to be causing issues as well. slightly reordered things based on how github / wordpress order them as well, but don't think that should make a super noticeable change

changed some font weights in locations where 500 was being used, for reference 400 is normal, 700 is bold, so moved these slightly closer to bold which matches github bolding for headers. Also reduced their size a bit to make up for the extra spacing that can add in certain fonts, `1.0rem` seems ~ `16px` which also matches github header sizing

nobodyatroot (approved):
I don't have an iOS device or Mac to test either, so I guess we'll source complaints if anyone has issues. Otherwise looks good to me, been running it for a couple hours so far.

PR #262, opened 2023-11-14
fix embed fetching occasionally persist results between calls

asdfzdfj:
adjust the way fetched embed data is assigned and returned to the caller

if I understand this correctly it seems to be doing clone returning in reverse: generally the object is cloned first and then value is assigned to the clone, however this class assigned results to itself (`$this`) first and then return a clone of self, which in some cases would results in prior results being persisted between calls and then in the next unfortunate fetch calls (that presumably failed when trying to retrive embed, judging from the code) would get a wrong embed data that was unintentionally persisted returned to them

this is most likely why some entries would get a wrong thumbnail, as this is the place where the thumbnail is being retrieved, and a said failure could results in a less than desirable or appropiate thumbnail image returned back

### DIFF
--- a/src/Utils/Embed.php
+++ b/src/Utils/Embed.php
@@ -8,6 +8,7 @@ use App\Entity\Entry;
 use App\Service\ImageManager;
 use App\Service\SettingsManager;
 use Embed\Embed as BaseEmbed;
+use Psr\Log\LoggerInterface;
 use Symfony\Contracts\Cache\CacheInterface;
 use Symfony\Contracts\Cache\ItemInterface;
 
@@ -19,8 +20,18 @@ class Embed
     public ?string $image = null;
     public ?string $html = null;
 
-    public function __construct(private CacheInterface $cache, private SettingsManager $settings)
+    public function __construct(
+        private CacheInterface $cache,
+        private SettingsManager $settings,
+        private LoggerInterface $logger,
+    ) {
+    }
+
+    public function __clone(): void
     {
+        unset($this->cache);
+        unset($this->settings);
+        unset($this->logger);
     }
 
     public function fetch($url): self
@@ -28,6 +39,14 @@ class Embed
         if ($this->settings->isLocalUrl($url)) {
             return $this;
         }
+
+        $this->logger->debug('Embed:fetch: leftover data', [
+            'url' => $this->url,
+            'title' => $this->title,
+            'description' => $this->description,
+            'image' => $this->image,
+            'html' => $this->html,
+        ]);
 
         return $this->cache->get(
             'embed_'.md5($url),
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):
Nice, glad this is finally getting fixed!

src/Utils/Embed.php
```diff
@@ -38,29 +57,35 @@ function (ItemInterface $item) use ($url) {
...
+                $c->html = $this->cleanIframe($oembed->html('html'));
 
                 try {
-                    if (!$this->html && $embed->code) {
-                        $this->html = $this->cleanIframe($embed->code->html);
+                    if (!$c->html && $embed->code) {
+                        $c->html = $this->cleanIframe($embed->code->html);
                     }
                 } catch (\TypeError $e) {
+                    $this->logger->info('Embed:fetch: html prepare failed: '.$e->getMessage());
```
melroy89, line 79: I tried to set this logging to `debug` across Mbin, to avoid too much log pollution.. Hence the default log value for the dev mode is currently `info`.

src/Utils/Embed.php
```diff
@@ -74,18 +99,6 @@ private function cleanIframe(?string $html): ?string
         }
 
         return $html;
-
-        //        $types = [
```
e-five256, line 78: can this block be cleaned up? not sure if it was used for debugging purposes

src/Utils/Embed.php
e-five256: it is an error though, I would actually have expected this to be error rather than debug, but not familiar with the log generation

src/Utils/Embed.php
melroy89: these lines are removed already right? It's in red.

e-five256: 🤦

src/Utils/Embed.php
asdfzdfj: should this one be downgraded to `debug` level then? 

the way I see it is that this a rather minor error, not a showstopper, but it's still an error and so it should be logged with a bit higher level compared to the standard debug/trace/diagnostic logs, and for this I think `info` is fine

e-five256: I agree with info, @nobodyatroot not sure if you've been running this and also asdfzdfj do either of your happen to know how often it comes up? If it's too much, we can always tone it down. I guess if an instance is having issues it might make it a bit spammy

asdfzdfj: I have it run on my instance since yesterday and I haven't seen it even once. maybe @nobodyatroot might have more info as my single user selfhosted instance and is a rather quiet one

actually, given the current code I don't event know what could trigger this exception in the first place

e-five256: oops, I misread this comment as ln 60, that's the info one I might expect to see it fetching fails

nobodyatroot: i think this one happens so infrequently (but super annoying when it does and we notice more when it f-s up and puts a NSFW image in there) that it's going to need extended live testing. i'd be fine with merging it and let it run wild for awhile as it doesn't seem to hurt anything right now, can't say i notice any increased log file size... only 6 MB in the last 24 hours.

e-five256 (approved):

src/Utils/Embed.php
melroy89: Sure ok. Let's leave it like this.

nobodyatroot:
Merge it, merge it, merge it ;-)

I temporarily reverted the patch just to see what would happen and sure enough it didn't take long for out of place NSFW thumbnails to start reappearing...

PR #107, opened 2023-10-23
enable auto rotate filter when making thumbnail

asdfzdfj:
enable auto rotate filter to properly make thumbnails for images that have rotation data stored in exif, such as those coming from phones.

note that this required php exif extension for it to work (added extension in docker build, basic php install for bare metal guide should already included exif extension)

PR #263, opened 2023-11-14
exiftool based image exif cleaner

asdfzdfj:
something I mentioned to put up [weeks ago](https://github.com/MbinOrg/mbin/pull/107#issuecomment-1775592752)

add an image metadata sanitizer/cleaner based on exiftool, comes with following modes, configurable via env variables:

- `none`: disable image metadata cleaning. default for external images.
- `sanitize`: strips just GPS data and what might looks like serial numbers. default for uploaded images.
- `scrub`: strip most of metadata save for those needed for proper image rendition and some attribution tags.

I tried to rig this up as an optional feature, as in it would only work when a valid `exiftool` binary is found and is executable, and it's effectively a no op when that's not the case without impeding the entire image processing flow.

---

<details><summary>concerns about where to plug the processor</summary>
<p>

now the hard part that I could use some suggestion: right now I plugged the cleaner inside `ImageRepository::findOrCreateFromPath`, right just before storing the image, and I'm not sure if there's more appropiate place to put this:

- putting it here means all image introduced would be cleaned, including federated images, which may not be desired for busy instances
- putting it before `findOrCreateFrom{Path,Upload}` could results in image lookup by hash being less effective because the input image will have to be cleaned before checking for hash although the same, cleaned image might already exists, which means what's supposedly fast path is now slower and the cleaning job is practically wasted
- right now, the function now looks like at least 3 functions stacked together and looks like it could use some refactoring, but I'm not sure how to go with it (and for some inexplicable reasons I feel like event dispatcher could be used here)

</p>
</details> 

ended up refactored/rearranged the `findOrCreateFrom{Path,Upload}` functions, reviewed separately at #287

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -73,6 +73,16 @@ OAUTH_KEYCLOAK_VERSION=
 # If true, sign ins and sign ups will only be possible through the OAuth providers configured above
 SSO_ONLY_MODE=
 
+# image exif cleaning options
+# available value: none, sanitize, scrub
+# can be set differently for user uploaded and external media
+EXIF_CLEAN_MODE_UPLOADED=sanitize
+EXIF_CLEAN_MODE_EXTERNAL=none
+# path to exiftool binary, leave blank for auto PATH search
+EXIF_EXIFTOOL_PATH=
+# max execution time for exiftool in seconds, defaults to 10 seconds
+EXIF_EXIFTOOL_TIMEOUT=10
+
 ###> symfony/framework-bundle ###
 APP_ENV=prod
 APP_SECRET=!CHANGE_SECRET!
--- a/src/Utils/ExifCleaner.php
+++ b/src/Utils/ExifCleaner.php
@@ -1,0 +1,102 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Utils;
+
+use Psr\Log\LoggerInterface;
+use Symfony\Component\DependencyInjection\ParameterBag\ContainerBagInterface;
+use Symfony\Component\Process\Exception\ProcessFailedException;
+use Symfony\Component\Process\ExecutableFinder;
+use Symfony\Component\Process\Process;
+
+class ExifCleaner
+{
+    protected const EXIFTOOL_COMMAND_NAME = 'exiftool';
+    protected const EXIFTOOL_ARGS_COMMON = [
+        '-overwrite_original', '-ignoreminorerrors',
+    ];
+    protected const EXIFTOOL_ARGS_SANITIZE = [
+        '-GPS*=', '-*Serial*=',
+    ];
+    protected const EXIFTOOL_ARGS_SCRUB = [
+        '-all=',
+        '-tagsfromfile', '@',
+        '-colorspacetags', '-commonifd0', '-orientation', '-icc_profile',
+        '-XMP-dc:all', '-XMP-iptcCore:all', '-XMP-iptcExt:all',
+        '-IPTC:all',
+    ];
+    protected const EXIFTOOL_TIMEOUT_SECONDS = 10;
+
+    private readonly ?string $exiftoolPath;
+    private readonly ?string $exiftool;
+    private readonly int $timeout;
+
+    public function __construct(
+        private readonly ContainerBagInterface $params,
+        private readonly LoggerInterface $logger,
+    ) {
+        $this->exiftoolPath = $params->get('exif_exiftool_path');
+        $this->timeout = $params->get('exif_exiftool_timeout') ?? self::EXIFTOOL_TIMEOUT_SECONDS;
+        $this->exiftool = $this->getExifToolBinary();
+    }
+
+    public function cleanImage(string $filePath, ExifCleanMode $mode)
+    {
+        if (ExifCleanMode::None === $mode) {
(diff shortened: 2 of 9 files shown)


e-five256:
> putting it here means all image introduced would be cleaned, including federated images, which may not be desired for busy instances

I definitely think it's best to only handle local images and not make it the job of the instance owner to scrub all remote media. I don't think anything legally harmful can come from this metadata, it's more of a personal data / security thing? I would expect most incoming to already be scrubbed, it looks like `pict-rs` does for lemmy and I _think_ imgur did unless they changed something

asdfzdfj:
> > putting it here means all image introduced would be cleaned, including federated images, which may not be desired for busy instances
> 
> I think it might be best to only handle local images if possible and not make it the job of the instance owner to scrub all remote media. I don't think anything legally harmful can come from this metadata, it's more of a personal data / security thing? I would expect most incoming to already be scrubbed, it looks like `pict-rs` does for lemmy and I _think_ imgur did unless they changed something

cleaning only local images does make sense, but then I'll either go back into the 2nd point where reusing existing image is less effective, or having to duplicate the `findOrCreateFromPath` into `findOrCreateFromUpload` just to add the cleaner command, and I'd like to *not* duplicate the code

also choosing to clean remote images may not be entirely pointless: it reduces amount of image copies that contains sensitive information -- so ideally there would be only one copy with sensitive information unintentionally added on origin instance/server, instead of who knows how many instances that the same image was federated out and mirrored. if anything maybe make it an opt in options to enable it

src/Service/ImageManager.php
```diff
@@ -128,6 +130,13 @@ public function download(string $url): ?string
         return $tempFile;
     }
 
+    public function cleanExif(string $filePath)
+    {
+        if ($this->exifCleaner->isEnabled()) {
```
e-five256, line 135: do you think it'd be cheaper if we also checked whether there still was exif data (was already scrubbed) with something like `exif_read_data()` and early out?

asdfzdfj: I don't know much about that function or how much it can read to give an accurate answer

also it doesn't help that image metadata stuff [is quite a damn mess](https://exiftool.org/TagNames/), in particular there's [XMP tags](https://exiftool.org/TagNames/XMP.html) which seems to be a different beasts of its own and I don't know if `exif_read_data()` could read these

e-five256: Thinking about it my main concern was spending cpu cycles doing unneeded work (potentially, depending on how quick the tool is to detect it has nothing to do, but in my experience exec'ing out still has a cost). I agree that, as an instance owner, it might be nicer to have everything you cache scrubbed. Since this is opt-in, at least for the moment, instance owners that don't want to pay the cost of scrubbing all external images and choose not to enable it, I imagine

.env.example
```diff
@@ -72,6 +72,12 @@ OAUTH_KEYCLOAK_VERSION=
 # If true, sign ins and sign ups will only be possible through the OAuth providers configured above
 SSO_ONLY_MODE=
 
+# image exif cleaning options
+# available value: none, sanitize, scrub
+EXIF_CLEAN_MODE=sanitize
```
melroy89, line 77: keep in mind, to also set some defaults. To avoid regression during updating for server owners who do not yet have this option in the config.

asdfzdfj: I'll see what I can do

asdfzdfj:
updates:
- refactored/rearranged the `findOrCreateFrom{Path,Upload}` functions (reviewed separately at #287, get that merged first)
- added option to specify different cleaning modes for uploaded and external images,
  defaults to `sanitize` for uploaded images and `none` (do nothing) for external images
- set defaults for configuration such that it shouldn't crash and still functions if config env vars are not set

src/Utils/ExifCleaner.php
```diff
@@ -0,0 +1,102 @@
...
+    private readonly ?string $exiftoolPath;
+    private readonly ?string $exiftool;
+    private readonly int $timeout;
+
+    public function __construct(
+        private readonly ContainerBagInterface $params,
+        private readonly LoggerInterface $logger,
+    ) {
+        $this->exiftoolPath = $params->get('exif_exiftool_path');
+        $this->timeout = $params->get('exif_exiftool_timeout') ?? self::EXIFTOOL_TIMEOUT_SECONDS;
```
e-five256, line 40: getting a 500 with no `.env` changes:

```
Cannot assign string to property App\Utils\ExifCleaner::$timeout of type int
```

asdfzdfj: noted. changed to use `%env(int:default::EXIF_EXIFTOOL_TIMEOUT)%`, think I'll need to run it for some time to make sure that it actually fixes it

e-five256 (approved):
lgtm to me, ran with both sanitize and scrub on local uploads and it correctly removed gps data in the first test and all the shutter/focal data in the second test

one thing to note is that uploading an image that already was uploaded pre-cleaned seemed to just reuse the same image and not clean it, which I think is expected, I imagine it detects it before running the logic which makes sense, just a note that reusing past uploaded images won't change the metadata after this

src/Utils/ExifCleaner.php
e-five256: well fwiw I pulled it in with no changes on my end and it fixed the 500 for me

PR #265, opened 2023-11-14
hide nsfw magazine icons for logged out users, blur on mag sub/block pages

e-five256:
hide nsfw magazine icons for logged out users, this is consistent with how we currently handle entries. we discussed adding an admin option to allow enabling nsfw for logged out users, but that would require a bunch more work, so seemed like it could be, once again, another followup

blur magazine subscriptions (in user page or user settings page) and magazine blocks (in user settings)

this also adds the `18+` danger badge to subscription and block pages. opinions welcome if this is bad. The information isn't really hidden if you just go to the magazine, but if users make their subs public and there are nsfw mags in there, it's obvious now

another nsfw followup for #254

in my previous PR I had said the sub/block components didn't have nsfw info. Then I slept on it and realized, wait, how the heck would it have icon, there's no way that's in the sub/block table surely. I just misread the components, they have the info

<details>
<summary>magazine icon being hidden while logged out images</summary>

image
image
image
image
image

</details>

<details>
<summary>magazine sub and block icon being blurred out for user while logged in</summary>

image
image

</details>

### DIFF
--- a/assets/styles/components/_magazine.scss
+++ b/assets/styles/components/_magazine.scss
@@ -171,6 +171,14 @@
     color: var(--kbin-meta-text-color);
     font-size: .85rem;
   }
+
+  .stretched-link {
+    small {
+      &.badge.danger {
+        color: var(--kbin-danger-color);
+      }
+    }
+  }
 }
 
 td {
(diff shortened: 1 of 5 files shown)


nobodyatroot (approved):
Thank you for continuing to improve this!

PR #266, opened 2023-11-15
Fix frontend message enumeration when purged users threads are encountered

nobodyatroot:
The frontend message center will throw a 500 if it tries to reference messages of recipient users that have been purged from the database. The entire DM messaging system uses 3 different tables in the database, so ultimately I think the _"more robust"_ solution is to replace/restructure the entire thing especially since it doesn't support federation right now; however, this will fix the 500 and allow the user to see the remainder of their other messages.

**Bug replication steps prior to PR**: create a user, message the user, then purge the user... then try to visit the message center.

### DIFF
--- a/templates/messages/front.html.twig
+++ b/templates/messages/front.html.twig
@@ -14,15 +14,26 @@
 
 {% block body %}
     <h1>{{ 'messages'|trans }}</h1>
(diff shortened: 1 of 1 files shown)


e-five256 (approved):
thanks for the repro steps

PR #267, opened 2023-11-15
Fix image cache resolve 302

melroy89:
- Keep 302 (== default) to avoid breaking dev env or Docker setup.
- Explain you can set this value to 301 if you are behind Nginx (assuming it's correctly configured)
  - Update docs
- Reorder liip config
- Add some defaults. Incl. 90 quality on `default_filter_set_settings` level, so no need to repeat it in `filter_sets`.
- Update liip package
- `composer update`

Before (302, not getting cached):

image

After (301, getting cached by the browser, Firefox):

image

**EDIT:** Tested live on https://kbin.melroy.org for more than 24 hours in the past already with `301` behind nginx, without any issues :)


melroy89:
We might have a 'tiny' issue that the image might not resolve under Docker when only using Caddy, since with the 301 the browser will cache the image response and not trigger the controller anymore. In Nginx this is solved by using [this by-pass image cache directive](https://symfony.com/bundles/LiipImagineBundle/current/optimizations/avoid-redirects.html) (see below) above the image cache directive itself:

```nginx
# bypass thumbs cache image files
location ~ ^/media/cache/resolve {
  expires 1M;
  access_log off;
  add_header Cache-Control "public";
  try_files $uri $uri/ /index.php?$query_string;
}
```

But we now need to-do the same magic under Caddy I think?

image

melroy89:
I will split this PR, so I will move out all documentation related stuff to a separate PR.

melroy89:
I restored to default back to 302 to avoid issues with devenv or Docker or Caddy. However, I did add this setting now to the configuration + inline comment + admin guide documentation.

nobodyatroot (approved):

nobodyatroot (approved):
No issues from what I can tell, testing live on kbin.run and everything appears normal.

PR #268, opened 2023-11-15
Improve Admin, Docker, Contrib, FAQ docs

melroy89:
- Match the location image directive expire with the solve location expire.
- Give Docker containers better names + use a dedicated bridge network (which is a good practice)
- Improve Docker guideline with back-up and restore
- Small FAQ improvement note

### DIFF
--- a/docker/compose.yml
+++ b/docker/compose.yml
@@ -10,7 +10,10 @@ services:
     # Or remove the build, context, dockerfile and image lines above and
     # use the pre-build image from ghcr.io (you can also a tag eg.: v1.0.0 instead of: latest):
     #image: "ghcr.io/mbinorg/mbin:latest"
+    container_name: mbin-www
     restart: unless-stopped
+    networks:
+      - mbin_external_network
     command: caddy run --config /etc/caddy/Caddyfile --adapter caddyfile
     healthcheck:
       test: ["CMD-SHELL", "curl -f http://localhost:2019/metrics || exit 1"]
@@ -37,7 +40,10 @@ services:
     # Or remove the build, context, dockerfile and image lines above and
     # use the pre-build image from ghcr.io (you can also a tag eg.: v1.0.0 instead of: latest):
     #image: "ghcr.io/mbinorg/mbin:latest"
+    container_name: mbin-php
     restart: unless-stopped
+    networks:
+      - mbin_external_network
     command: php-fpm
     healthcheck:
       test:
@@ -47,6 +53,8 @@ services:
         ]
     volumes:
       - ./storage/media:/var/www/mbin/public/media
+    # If you want to change configs locally, without rebuilding the image use:
+    #  - ../config:/var/www/mbin/config
     env_file:
       - .env
     depends_on:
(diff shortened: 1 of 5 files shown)


docs/docker_deployment_guide.md
```diff
@@ -306,9 +306,16 @@ docker compose exec redis redis-cli
...
+```bash
+docker compose exec -it db pg_dump -U kbin kbin > dump.sql
+```
+
+Restore:
+
 ```bash
-docker exec -it container_id pg_dump -U kbin mbin > dump.sql
-docker compose exec -T database psql -U kbin mbin < dump.sql
+cat dump.sql | docker compose exec -i db psql -U kbin kbin
```
e-five256, line 318: hmm, this gets me `the input device is not a TTY`. I tried without `-i` as well

docker/compose.yml
```diff
@@ -139,9 +169,14 @@ services:
...
+  #    - mbin_external_network
   #  environment:
   #    - SMARTHOST_ADDRESS=mail.mysmtp.com
   #    - SMARTHOST_PORT=587
   #    - SMARTHOST_USER=myuser
   #    - SMARTHOST_PASSWORD=secret
   #    - SMARTHOST_ALIASES=*.mysmtp.com
+
+networks:
+  mbin_external_network:
```
e-five256, line 182: I'm not super familiar with docker networks. Based on the docs, it was using the default one, now this one, is there any concern of being unable to communicate? I guess it still exposes the same ports, it was only if people had custom containers they also ran, they'll have to move them over to the network as well?

docs/docker_deployment_guide.md
melroy89: ow maybe it should have been `-T`

docker/compose.yml
melroy89: it has no other negative effects, I also kept it an external network (there is also a internal option). So it's the same as the default bridge, but it's advised to have a dedicated network instead of sharing the default bridge interface. So the benefit would be that another docker container can't access your mbin docker containers. Which also makes it safer.

docs/docker_deployment_guide.md
melroy89: I trhink it will change it back.

e-five256 (approved):

docs/docker_deployment_guide.md
```diff
@@ -306,9 +306,16 @@ docker compose exec redis redis-cli
 
 ## Backup and restore
 
+Backup:
+
+```bash
+docker compose exec -it db pg_dump -U kbin kbin > dump.sql
```
BentiGorlich, line 312: is kbin still supposed to be here?

docs/docker_deployment_guide.md
```diff
@@ -306,9 +306,16 @@ docker compose exec redis redis-cli
...
+```
+
+Restore:
+
 ```bash
-docker exec -it container_id pg_dump -U kbin mbin > dump.sql
-docker compose exec -T database psql -U kbin mbin < dump.sql
+docker compose exec -T db psql -U kbin kbin < dump.sql
```
BentiGorlich, line 318: is kbin still supposed to be here?

melroy89: Yes for backwards compatibility the default username and database name are kbin.

docs/docker_deployment_guide.md
melroy89: Yes for backwards compatibility the default username and database name are kbin. Users from kbin still have the username/db name kbin.

PR #269, opened 2023-11-16
Add admin panel option to define a default theme for site

e-five256:
This adds an admin drop down which lets admins pick from the existing themes to set as a default for users who have not chosen a theme via the sidebar

image

By default, with no env or admin setting defined, ~this will switch people to see the tokyo night theme~ this will continue to use auto detection for light/dark themes, but via css rather than js

It also removes the javascript theme swapping based on system or user agent settings. I found the theme swapping, especially when javascript is slow to load, very jarring when visiting mbin instances, but it is possible there are people who find it useful (just recently I saw a post about someone who swaps between light and dark system settings throughout the day based on time). 

There's an alternative middle ground perhaps where the JS is kept, but admins can define which theme gets used in light/dark mode, with perhaps a checkbox to just use the same theme in both, potentially.

---

this also moves the light theme into its own file and moves up some root vars. the root vars themselves are untouched, just moved around to better describe which ones are shared between all themes and which ones were specific to the previously default, light theme. My thinking is that we can next begin to style the root vars in a new theme, which will end up being the new default mbin theme, potentially


nobodyatroot (dismissed):
Awesome! Nice having the extra flexibility with this option.

melroy89 (dismissed):

e-five256:
Sorry about the changes, wasn't sure what I wanted to do was possible, I should've probably converted to draft

@BentiGorlich brought up a desire to not see the auto detection be removed, so I believe I have maybe found a solution to all of our desires, maybe. I brought back auto detection based on user settings, it currently is only for light/dark themes and solarized light/solarized dark.

image

I moved the default back to `default` for the light/dark theme

So right now I'm theme "ligh/dark auto detect", and you can see my system setting is dark and the dark theme is highlighted as active

image

It's a bit hard to show just through images, but trust me that all I did was click "light" in my system settings, nothing else

image

the theme instantly changes to light, the light theme is shown as active.

if I select a theme manually everything is still normal, it writes a cookie and it uses that theme above all else. the admin is also free to change to any other theme instead, even for example forcing solarized dark and ignoring the auto detection

still no flash bang issue, since it's done via css and not js the detection doesn't have to wait for the entire DOM to load

templates/layout/_options_theme.html.twig
```diff
@@ -5,36 +5,36 @@
        title="{{ 'kbin' }} {{ 'theme'|trans }}"
        aria-label="{{ 'kbin' }} {{ 'theme'|trans }}"
        rel="nofollow">
-        <div class="{{ html_classes('theme', 'kbin', constant('App\\Controller\\User\\ThemeSettingsController::KBIN') == theme ? 'active' : '' ) }}"></div>
```
e-five256, line 8: I mentioned in the commit but these active classes didn't seem to do anything, the outline to highlight the theme is selected comes from `assets/styles/components/_sidebar.scss` which is solely based off css selectors like `.theme--dark .theme.dark`, as that is enough to tell which theme is active. There weren't any styles added by the `.active` as far as I could see

e-five256:
It does slightly increase the `app.css` file size. The code doesn't have to duplicate the vars, but this was the only way I could think of to order the `@media` checks :thinking: from my understanding it basically duplicates then entries in each part. There might be a way to write it another way which works more optimally that I'm not thinking of

.env.example
```diff
@@ -31,6 +31,8 @@ KBIN_META_KEYWORDS="mbin, content agregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 
+MBIN_DEFAULT_THEME=default
```
melroy89, line 34: Let's just add it to line 33 directly.

.env.example_docker
```diff
@@ -31,6 +31,8 @@ KBIN_META_KEYWORDS="mbin, content agregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 
+MBIN_DEFAULT_THEME=default
```
melroy89, line 34: Sam here, move it to line 33

.env.example
e-five256: done

.env.example_docker
e-five256: done

nobodyatroot (approved):

PR #283, opened 2023-11-20
cleanup unused image assets

e-five256:
cleanup missed cardano assets, mac metadata file, unused assets

noticed we had the cardano image asset still around, also think some of these other assets are unused, and don't think `.ds_store` files are meant to be checked in. went through the rest and they all appear to be from much older templates, the renders being on about landing pages that no longer exist, the entry placeholder now being done via font awesome icons, etc


nobodyatroot (approved):
Be gone with ya!

PR #285, opened 2023-11-20
Remove user facing ranking page, leftover crypto clean-up.

nobodyatroot:
PR removes the user ranking page that has always been disabled, does **NOT** touch anything with the backend ranking system used to score entries for SORT_HOT functionality.

image

image

Also remove leftover crypto column and code from `Entry` entity.


nobodyatroot:
> I noticed there are also these two files https://github.com/MbinOrg/mbin/blob/main/src/Entity/Contracts/RankingInterface.php https://github.com/MbinOrg/mbin/blob/main/src/Entity/Traits/RankingTrait.php are these both unrelated to the ranking page itself? you did mention it appears related to how it does sorting, just coincidentally named?

as discussed via matrix, i think this ranking page was meant to display or portray the entity data in a different way (or similar way?) on how it's being used by the backend to generate the sort_hot and related functions. i can't tell what the intent was since this page has always been disabled and of unknown origin (to me, anyway), but i made sure to not touch anything related to the sort behavior.

e-five256 (approved):

PR #296, opened 2023-11-24
fetch previews only once per page load

e-five256:
this is a simple change where, after clicking the preview button once, further clicking will only hide and show the content, with no further AJAX calls to fetch the content again. If it fails, it will act as it did before showing a click to retry prompt (clicking preview will hide and show this prompt, but clicking on the prompt itself works to refetch the content)

@asdfzdfj since you touched the preview stuff recently, want to check with you that this makes sense

edit: tested this with asdfzdfj's example post of linked youtube with 2 image inlines and it appears to work, fetched each preview only once and correctly hid/made visible the previews after that

### DIFF
--- a/assets/controllers/preview_controller.js
+++ b/assets/controllers/preview_controller.js
@@ -47,9 +47,8 @@ export default class extends Controller {
         event.preventDefault();
 
         if (this.containerTarget.hasChildNodes()) {
-            this.containerTarget.replaceChildren();
-            this.containerTarget.classList.add('hidden');
-            return
+            this.containerTarget.classList.toggle('hidden');
+            return;
         }
 
         try {
--- a/assets/controllers/preview_controller.js
+++ b/assets/controllers/preview_controller.js
@@ -75,7 +75,7 @@ export default class extends Controller {
                         data-action="preview#retry"
                         data-preview-url-param="${event.params.url}"
                         data-preview-ratio-param="${event.params.ratio}">
-                            Failed to load. Click to retry.
+                            Failed to load. Click here to retry.
                     </a>
                 </div>`
             this.containerTarget.innerHTML = failedHtml;


asdfzdfj (commented):
given a bit of test drive, seems to be working fine

come to think of it, maybe the wording of the retry link label should be adjusted: "Click *here* to retry." since hiding and reopening the preview now no longer refresh the preview content and only that link would trigger refresh

https://github.com/MbinOrg/mbin/blob/824905b34b2ea46c00ef40b0b423cec96feffe8e/assets/controllers/preview_controller.js#L73-L80

e-five256:
It'd be nice if it could be a loc token but I didn't see a quick example of that via JS. maybe it needs to load a twig template instead?

asdfzdfj:
> It'd be nice if it could be a loc token but I didn't see a quick example of that via JS. maybe it needs to load a twig template instead?

some ways I could think of:
- using [symfony ux translator](https://symfony.com/bundles/ux-translator/current/index.html) to expose some translations to js and then do it in the controller
- template translation in twig and pass it to preview controller by controller values, something like:
```twig
<div data-controller="preview"
     data-preview-retry-label-value="{{ 'preview_click_to_retry'|trans }}">...</div>
```
```js
export default class extends Controller {
    static values = {
        retryLabel: String,
    };
    // ...
    async show(event) {
        // ...
        let failedHtml = `<div><a>${this.retryLabelValue}</a></div>`
        // ...
    }
}
```

e-five256:
This seems to work except for inline embeds. I'm stuck on adding it to https://github.com/MbinOrg/mbin/blob/main/src/Markdown/CommonMark/EmbedElement.php#L15-L19

I have `'data-preview-retry-label-value' => $this->translator->trans('preview_click_to_retry'),` but it doesn't seem to add any attributes, even tried just adding `'data-foo' => 'bar',` and still nothing.

Then again, maybe it doesn't effect past content if that markdown is parsed only once... hmm no even posting new content doesn't add any attributes to that html :confused:

e-five256:
Maybe I should just change the hardcoded wording here, leave it alone for now, and open a separate PR for translating it to take a look at?

e-five256:
Alright not saying I'm giving up but pulling in an experimental bundle for this also seems like it'd require a lot of testing, but it's possible we want to do that anyways, we probably have more hardcoded copy elsewhere

asdfzdfj:
> Maybe I should just change the hardcoded wording here, leave it alone for now, and open a separate PR for translating it to take a look at?

yeah I think that'd be good enough for now (the changes I pulled from kbin side is also hardcoded too)

> Alright not saying I'm giving up but pulling in an experimental bundle for this also seems like it'd require a lot of testing, but it's possible we want to do that anyways, we probably have more hardcoded copy elsewhere

I also found another bundle: https://github.com/willdurand/BazingaJsTranslationBundle if you want to play around

asdfzdfj (approved):

Issue #258, opened 2023-11-14
Rename Suspend to Muted


melroy89:
**Describe the bug**

The current suspend feature isn't really suspend and is therefor confusing. The idea is to rename "Suspend" to "Muted".

**On which Mbin instance did you find the bug?**
All domains are effected, since it's part of the current source code.

**To Reproduce**
1. Go to a user as an admin user
2. See the sidebar for the "suspend account" feature/buttons:

**Expected behavior**

I expect to have a clear understanding of what the feature does on basis of a single word (without creating confusion), without additional description needed.

**Screenshots**

![image](https://github.com/MbinOrg/mbin/assets/628926/6796ffdb-8ae4-4324-8f65-49e49f3bd794)m.

**Mbin:**
 - Version v1.0.0

**Additional context**

**Additional idea:** Furthermore, introducing a dedicated database column (boolean) `muted` on the `user` table can make a big difference in performance. No need anymore to filter on visibility (text) column maybe?

PR #212, opened 2023-11-06
Kbin sync 7 + lot of rework, refactoring / Mbin changes

melroy89:
- Finish Magazine deletion (Add TODO, since this can be done much easier and simpler when adding cascading to postgresql)
- Improve moderator role
  - Auto-formatting on YAML files according to your `.editorconfig`
- Improved cross-posted threads by grouping them
- Improve User profile visibility options  + account deletion process changed (but not yet how Mbin wants it)
- Keep our code most consistent eg. we use `VisibilityInterface::VISIBILITY_SOFT_DELETED;` instead of `self::`
- Rename Kbin class to Mbin class
- Add description to migration file (which is not present in Kbin)
- We move suspend/unsuspend feature to admin panel section
- Show to the user that the account is suspended
- Show avatar/cover (user box) above all the profile pages. Which make sense to me.
- Don't allow the user to post a thread, comment or new links, images or magazines, etc.
- Small code improvements / refactoring
- Fix people menu active class activation

Fixes: #194 

Kbin PRs:
- https://codeberg.org/Kbin/kbin-core/pulls/1207
- https://codeberg.org/Kbin/kbin-core/pulls/1209
- https://codeberg.org/Kbin/kbin-core/pulls/1210
- https://codeberg.org/Kbin/kbin-core/pulls/1211
- https://codeberg.org/Kbin/kbin-core/pulls/1213
- https://codeberg.org/Kbin/kbin-core/pulls/1214

PR #297, opened 2023-11-24
Rename suspend to muted

melroy89:
- [x] Rename (un)Suspend to (un)Mute
- [x] Introduce a new user column `muted`
- [x] Update database queries to use the new muted column, which should also improve performance. Don't misuse 'visibility' string for this. 

I'm trying to replicate the "Limit" feature of Mastodon now: https://docs.joinmastodon.org/admin/moderation/#limit-user. But instead of calling it Limit, we voted in the Matrix chat about the new name, which was "Mute"/"muted".

Fixes: https://github.com/MbinOrg/mbin/issues/258


templates/entry/single.html.twig
```diff
@@ -37,7 +37,7 @@
         {{ component('entries_cross', {entry: entry}) }}
         {% include 'layout/_flash.html.twig' %}
         {% include 'entry/comment/_options.html.twig' %}
-        {% include('user/_visibility_info.html.twig') %}
+        {% include('user/_muted_info.html.twig') %}
         
         {% if user is defined and user and user.visibility is same as 'visible' and (user_settings.comment_reply_position == constant('App\\Controller\\User\\ThemeSettingsController::TOP')) %}
```
e-five256, line 42: all of these likely need to change to check muted instead now

I started to collect a list but it's easily dozens of places in https://github.com/MbinOrg/mbin/pull/212/files#diff-4474cfa4c5acf2dcb955841f0e57b248d4ee61e695467130752cf785660433a0

basically everywhere `user.visibility is same as 'visible'` appears in twig templates

melroy89:
I won't do it. I will partially revert https://github.com/MbinOrg/mbin/pull/212/

PR #300, opened 2023-11-25
do not automatically show media previews of nsfw posts

e-five256:
- do not automatically show media previews of nsfw entry / post / entry comment / post comment
- rewrite templates to use `with` and use defaults on cookie getters rather than null check

the majority of this change is: `'show-preview': SHOW_PREVIEW is same as 'true' and not [entry|post|comment].isAdult`
someone mentioned that they wished to see this change once, and I agree with it, you can still preview NSFW media manually, but this doesn't automatically open it on you

I guess perhaps it could also be a user setting, if people do want to browse nsfw mags for instance perhaps they want all to open automatically, this has always been an issue I have had where I feel like there's complete consensus of NSFW vs NSFL yet it always never is implemented (small rant)

the rest is I updated the four components to use `with` and `set` vars similar to what @asdfzdfj did for the subscription sidebar. That looked far more readable to me instead of long constants in the middle of the component. I also added defaults for each setting, so for example `SHOW_MAGAZINE_ICONS` defaults to `true` if not set so the code doesn't need to do the first half of what it was: `not SHOW_MAGAZINE_ICONS or SHOW MAGAZINE_ICONS is true`

some examples if they make sense, this is just me opening the page with show media preview on for both entries and posts, not clicking anything

- entry
image
- entry comments
image
- post and post comments
image


nobodyatroot (approved):

e-five256:
need to double check this after #301

also feeling less certain, after I opened the PR I edited the comment in about user setting, perhaps worth getting more feedback from people on if this would ruin their experience for browsing nsfw mags

the problem I'm having is I want to add dozens of user and admin settings but each of those being a column seems like not a great table structure. that might just be me, my view on table management is less columns and no nullable fields, if it's nullable it should be in another table and FK'd instead, but that's just one way of doing things

templates/components/post_comment.html.twig
```diff
@@ -207,7 +212,9 @@
             comment: comment,
             level: level,
             showNested: true,
-            view: app.request.cookies.get(constant('App\\Controller\\User\\ThemeSettingsController::POST_COMMENTS_VIEW'), 'tree')
+            view: VIEW_STYLE
```
e-five256, line 215: i'm mostly certain this isn't used. I see no visible way to change the view style when viewing microblogs, and I tried writing the cookie out manually as value tree/chat/classic and none of them changed the view of 1 single indent for replies. possibly this can all be removed

e-five256: going to leave it alone just in case. I do believe it can't be changed like entry comment can, but there are different styles depending on if it was a threaded entry that somehow ended up in the post tab or an actual microblog. might need to look at it more later to see, it might depend more on the source/type than user setting, but just won't touch it for now

e-five256:
since it didn't seem like there were strong opinions on this, just going to mark it as ready, easy enough to revert if people end up  having problems with it or we want to come back and add an additional user setting to allowing nsfw auto preview

nobodyatroot (approved):

templates/components/post_comment.html.twig
e-five256: Oh, I was indeed wrong about this, a way to change the view style does appear on the single post view. It became more apparent with a friendica thread that ended up in the microblog section

PR #306, opened 2023-11-26
hide more drop down menu after clicking moderate

e-five256:
this is a small JS hack to make it so clicking moderate in the drop down hides the drop down, sleeps for 500ms, and then makes it non-hidden again (but it will be hidden because the mouse is no longer hovering over it)

@asdfzdfj if this is too crazy let me know and I can just close this, haha, I tried all sorts of things like `blur()` and `z-index` and everything. I don't want to _break_ the dropdown but since it's all CSS driven there isn't really a way to get CSS to stop CSSing, ya know, or rather if you're hovering over something with a hover state it just wants to stay that way. maybe I could make clicking the button takeover your mouse and throw it far away (this is a joke)

---

I want to do the same to report but looking at the report panel, it looks like I want to redo the UX for that as well to make it clearer: what you're reporting, who you're reporting to, why you're reporting it (predefined options). which may or may not be related to the issue we have to get remote reporting working

### DIFF
--- a/assets/controllers/subject_controller.js
+++ b/assets/controllers/subject_controller.js
@@ -211,10 +211,25 @@ export default class extends Controller {
         }
     }
 
+    /**
+     * hideDropdownMenu is a helper to close drop down menus via a small
+     * hack of hiding it temporarily to cancel :hover states and bringing
+     * it back after. Hiding and showing too quickly results in the :hover
+     * state remaining in effect
+     */
+    async hideDropdownMenu(element) {
+      let dropdownMenu = this.element.querySelector('.dropdown .dropdown__menu');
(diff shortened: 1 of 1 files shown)


e-five256:
that blog brings up good points and something I also thought about how we handle nsfw content (hover to unblur)

e-five256:
:( I thought I had tested mobileish with chromium device toolbar but this does make it way worse on mobile. I didn't expect it to act so differently than manually clicking more on desktop, but perhaps on desktop it moves to hover state once you go over menu items which it won't be doing on mobile. Makes me want to close this and just do the change to click menus but... looking at that blog post on mobile the menu doesn't work for me >.>. Not sure if that's because of codepen or something else, though

PR #308, opened 2023-11-27
remote magazines - remove ability to apply for moderator or owner

e-five256:
- remove ability to apply for moderator and apply for ownership of remote magazines
- do not return remote magazines in the find abanadoned call
- sync apply for moderator button filters to the moderator sidebar
- add missing reject text on moderator requests page

followup to #235

### DIFF
--- a/templates/magazine/moderators.html.twig
+++ b/templates/magazine/moderators.html.twig
@@ -15,7 +15,8 @@
 {% block body %}
     <h1>{{ 'moderators'|trans }}</h1>
     <div class="flex">
-        {% if app.user and not magazine.userIsModerator(app.user) and app.user.visibility is same as 'visible' %}
+        {% if app.user and app.user.visibility is same as 'visible' %}
+        {% if magazine.apId is same as null and not magazine.userIsModerator(app.user) %}
             <form action="{{ path('magazine_moderator_request', {name: magazine.name}) }}"
                   method="POST"
                   class="float-end mb-2"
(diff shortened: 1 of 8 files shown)


templates/magazine/moderators.html.twig
```diff
@@ -49,12 +50,14 @@
...
             </form>
         {% endif %}
+        {% endif %}
     </div>
-    {% include 'magazine/_moderators_list.html.twig' %}
-    {% if(moderators.haveToPaginate is defined and moderators.haveToPaginate) %}
-        {{ pagerfanta(moderators, null, {'pageParameter':'[p]'}) }}
-    {% endif %}
-    {% if not moderators|length %}
+    {% if moderators|length %}
```
e-five256, line 55: found this while testing, list was being included even when no moderators, which is how we display kbin mags now similar to lemmy
before
image
after
image

templates/magazine/panel/moderator_requests.html.twig
```diff
@@ -54,7 +54,7 @@
                                     <button type="submit"
                                             class="btn btn__secondary"
                                             title="{{ 'reject'|trans }}">
-                                        <i class="fa-solid fa-ban"></i>
+                                        <i class="fa-solid fa-ban"></i> {{ 'reject'|trans }}
```
nobodyatroot, line 57: Is the `{{ 'reject'|trans }}` doing something on this line?

e-five256: it adds the label after the button to make it clearer what it's for.

oh, whoops, it should also be in a `<span>`; my bad, one second

before
image
after
image

nobodyatroot (approved):
Looks good to me after a quick functional test on kbin.run.

PR #313, opened 2023-11-28
changed textarea backtick wrapping to only act on selections

asdfzdfj:
*NOTE: this change is largely due to my personal opinion. if it also works out for you then great. if not, feel free to leave some suggestions*

- changed the backtick button in comment box to only add enclosing backticks on selected text, as I find the current behavior (just giving a pair of possibly empty backticks whenever) utterly irritating especially when trying to type a fenced code block
- activates the selection wrapping for single/double quote, and tilde
- activates the selection wrapping for underscore and asterisk, but it will cycle between 0,1,2 wrapping characters
  like this: `word` -> `*word*` -> `**word**` -> `word` -> ...

### DIFF
--- a/assets/controllers/rich_textarea_controller.js
+++ b/assets/controllers/rich_textarea_controller.js
@@ -3,29 +3,37 @@ import {Controller} from '@hotwired/stimulus';
 /* stimulusFetch: 'lazy' */
 export default class extends Controller {
     connect() {
-        this.element.addEventListener('keydown',
-            this.handleInput.bind(this));
+        this.element.addEventListener('keydown', this.handleInput.bind(this));
     }
 
+    // map: allowed enclosure key -> max repeats
+    enclosureKeys = {
+        '`': 1, '~': 1, '"': 1, "'": 1,
+        '*': 2, '_': 2,
+    };
+
     handleInput (event) {
+        let hasSelection = this.element.selectionStart != this.element.selectionEnd;
+        let key = event.key;
+
         // ctrl + enter to submit form
-        if (event.ctrlKey && event.key === "Enter") {
+        if (event.ctrlKey && key === "Enter") {
             this.element.form.submit();
         }
 
         // ctrl + b to toggle bold
-        else if (event.ctrlKey && event.key === "b") {
+        else if (event.ctrlKey && key === "b") {
             this.toggleFormattingEnclosure('**');
         }
 
         // ctrl + i to toggle italic
-        else if (event.ctrlKey && event.key === "i") {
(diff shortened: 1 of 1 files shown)


assets/controllers/rich_textarea_controller.js
```diff
@@ -8,6 +8,8 @@ export default class extends Controller {
     }
 
     handleInput (event) {
+        let wrapKeys = ['`', '*', '_', '"', '\''];
```
e-five256, line 11: should `~` be added to this for strikethrough?

assets/controllers/rich_textarea_controller.js
```diff
@@ -23,9 +25,12 @@ export default class extends Controller {
...
 
-        // grave to toggle inline code
-        else if (event.key === "`") {
-            this.toggleFormattingEnclosure('`');
+        // toggle wrapping on selection texts
+        else if (
+            this.element.selectionStart != this.element.selectionEnd
+            && wrapKeys.includes(event.key)
+        ) {
+            this.toggleFormattingEnclosure(event.key);
```
e-five256, line 33: both `*` and `_` evaluate to italic. this might be too crazy, but would it be possible to get doing `*` to add `**` on both sides instead so it adds bold? or should we not assume that was the users intent?

e-five256: or maybe it should just be additive always rather than this toggle add/remove? this is very nitpicky so I'm also fine with it as is if you think either aren't worthwhile

assets/controllers/rich_textarea_controller.js
asdfzdfj: yeah I think I could add that

e-five256 (dismissed):
I'm also fine with going with this as-is, it's nice to be able to type \`\`\` again

assets/controllers/rich_textarea_controller.js
asdfzdfj: added

assets/controllers/rich_textarea_controller.js
asdfzdfj: also added the cycle thing, it's still rudimentary but it should work and not get in the way too much

e-five256 (approved):
in firefox, this seems to be working fine, in my chromium browser, selecting via keyboard is fine, selecting via click and dragging is fine, but if I double click on a word for some reason it deletes 1 character to the left as well and changes the selection area. I don't think that's a reason to hold this up as it seems like an improvement overall, and it's possible chromium doesn't correctly mimic google chrome anyways

PR #316, opened 2023-11-29
2fa - handle form errors with visible error and log

### LINKED ISSUE #305: errors on 2fa setup in some cases do not generate visible details
**Describe the bug**

when setting up 2fa, some errors appear to not generate any visible error message for the user, instead the page is simply reloaded. looking at the network the http response code is 422

likely possible all <= 1.1.0, but repro'd locally in docker

**To Reproduce**

1. Setup an account with an invalid email address, for example just `invalid`
2. Navigate to https://\<domain\>/settings/2fa
3. Attempt to setup 2fa with the QR code, entering a valid TOTP code
4. Click "Add to my account"
5. Observe page simply reloads, no indication of a failure besides the HTTP status code 422 in network console

**Expected behavior**

An error message indicating why adding 2fa to the account failed

**Screenshots**
image

 - OS: LMDE 6
 - Browser: Firefox
 - Version: 120

**Additional context**

```
[Symfony\Component\Validator\ConstraintViolation](file:///var/www/mbin/vendor/symfony/validator/ConstraintViolation.php#L19) {#2123 ▼
  -message: "This value is not a valid email address."
  -messageTemplate: "This value is not a valid email address."
  -parameters: [▶]
  -plural: null
  -root: [Symfony\Component\Form\Form](file:///var/www/mbin/vendor/symfony/form/Form.php#L72) {[#2013 ▶](http://127.0.0.1/_profiler/68d6e9?panel=validator#sf-dump-1997113593-ref22013)}
  -propertyPath: "data.email"
```

My hope is looking into this will show more errors that aren't being bubbled up to the user, as there have been reports of the page reloading with no messaging, it's possible there are multiple reasons it can happen, but wanted to report the one I could repro

e-five256:
this handles the situation when the form is considered invalid for any reason by showing an error message to the user and logging to the log file.

image

log example:

```
[2023-11-29T03:44:18.462966+00:00] app.WARNING: 2fa error occurred user "admin" submitting the form "ERROR: This value is not a valid email address.
" {"username":"admin","errors":{"Symfony\\Component\\Form\\FormErrorIterator":"ERROR: This value is not a valid email address.
"}} []
```

fixes #305

"fixes" as in now displays an error message and logs the output. I think there's still the open question as to _why_ users hit this. I couldn't quite figure this out myself, as the error above shows it says email is not valid, which is true, but I have no idea what that has to do with 2fa. The form interface being submitted, `UserTwoFactorType`, does not appear to contain email address. Might need to bring in someone with more php / experience with the DTO relationships here

### DIFF
--- a/src/Controller/User/Profile/User2FAController.php
+++ b/src/Controller/User/Profile/User2FAController.php
@@ -188,7 +188,10 @@ class User2FAController extends AbstractController
         }
 
         if (!$form->isValid()) {
-            $this->logger->warning('2fa an error occurred submitting the form "{errors}"', ['errors' => $form->getErrors()]);
+            $this->logger->warning('2fa error occurred user "{username}" submitting the form "{errors}"', [
+                'username' => $dto->username,
+                'errors' => $form->getErrors(),
+            ]);
             $form->get('totpCode')->addError(new FormError($this->translator->trans('2fa.setup_error')));
 
             return $form;
--- a/src/Controller/User/Profile/User2FAController.php
+++ b/src/Controller/User/Profile/User2FAController.php
@@ -15,6 +15,7 @@ use Endroid\QrCode\Encoding\Encoding;
 use Endroid\QrCode\ErrorCorrectionLevel\ErrorCorrectionLevelHigh;
 use Endroid\QrCode\RoundBlockSizeMode\RoundBlockSizeModeMargin;
 use Endroid\QrCode\Writer\PngWriter;
+use Psr\Log\LoggerInterface;
 use Scheb\TwoFactorBundle\Security\TwoFactor\Provider\Totp\TotpAuthenticatorInterface;
 use Symfony\Bundle\SecurityBundle\Security;
 use Symfony\Component\Form\FormError;
@@ -38,6 +39,7 @@ class User2FAController extends AbstractController
         private readonly TotpAuthenticatorInterface $totpAuthenticator,
         private readonly TranslatorInterface $translator,
         private readonly Security $security,
+        private readonly LoggerInterface $logger,
     ) {
     }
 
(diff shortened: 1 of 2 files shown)


src/Controller/User/Profile/User2FAController.php
```diff
@@ -174,27 +176,38 @@ private function handleForm(
...
+        if (!$form->isSubmitted()) {
+            return $form;
+        }
+
+        if ($form->has('totpCode')
                 && !$this->setupHasValidCode($dto->totpSecret, $form->get('totpCode')->getData())) {
             $form->get('totpCode')->addError(new FormError($this->translator->trans('2fa.code_invalid')));
-        }
 
-        if ($form->isSubmitted() && $form->isValid()) {
```
e-five256, line 183: this was the key, as `$form->isSubmitted() && $form->isValid()` the form was submitted, but not valid, in which case it just dropped all the way to the end returning the form with no other logic handling

src/Controller/User/Profile/User2FAController.php
```diff
@@ -174,27 +176,38 @@ private function handleForm(
...
 
-                $this->addFlash('success', 'flash_account_settings_changed');
+            return $form;
+        }
 
-                return $this->redirectToRoute('app_login');
-            }
+        $this->manager->edit($this->getUser(), $dto);
 
+        if (!$dto->totpSecret) {
```
e-five256, line 202: I couldn't quite figure out in which cases this was true, maybe in DB errors or something if it can't edit the user? I left it as is with the redirect for now

nobodyatroot (approved):
Makes sense for manually created test users and/or  a user that has had their account's email validation bypassed by someone with db access.

Looks like the form handling is causing the DTO/Symfony assert to check for a valid email address.

`src/Controller/User/Profile/User2FAController.php`:
```
$dto = $this->manager->createDto($user);
$dto->totpSecret = $totpSecret;

// QR code generation needs a user with a code. Add one for that then immediately remove
// to stop side effects when persisting.
$user->setTotpSecret($totpSecret);
$qrCodeContent = $this->totpAuthenticator->getQRContent($user);
$user->setTotpSecret(null);

$form = $this->handleForm($this->createForm(UserTwoFactorType::class, $dto), $dto, $request);

private function handleForm(
        FormInterface $form,
        UserDto $dto,
        Request $request
    ): FormInterface|Response {
 ```

`src/DTO/UserDto.php`:

```
use Symfony\Component\Validator\Constraints as Assert;
use Symfony\Component\Validator\Context\ExecutionContextInterface;

#[Unique(User::class, errorPath: 'email', fields: ['email'], idFields: ['id'])]
#[Unique(User::class, errorPath: 'username', fields: ['username'], idFields: ['id'])]
class UserDto implements UserDtoInterface
{
    public const MAX_USERNAME_LENGTH = 30;
    public const MAX_ABOUT_LENGTH = 512;

    #[Assert\NotBlank]
    #[Assert\Length(min: 2, max: self::MAX_USERNAME_LENGTH)]
    #[Assert\Regex(pattern: RegPatterns::USERNAME, match: true)]
    public ?string $username = null;
    #[Assert\NotBlank]
    #[Assert\Email]
    public ?string $email = null;
    ```

PR #319, opened 2023-11-30
add form error to 2fa form

e-five256:
this removes the error log and instead just puts it on the form

at first I was hesitant because:
- I'm not sure all of what error values this will display, though I doubt symfony would JS inject an error on us, no idea if they'll make any sense. like the one in the image makes no sense given only the context of the form
- it's not translated, but that's true of a lot of our symfony form errors. for instance if you try to submit a thread with no body it will say body is required but only in english
- not sure this is the best way to add form errors, there seems to be about a couple dozen that have changed over 7 years. tried all sorts of `foreach` and `->getMessage()`. Most failed except this

image

related #305

### DIFF
--- a/src/Controller/User/Profile/User2FAController.php
+++ b/src/Controller/User/Profile/User2FAController.php
@@ -188,11 +188,8 @@ class User2FAController extends AbstractController
         }
 
         if (!$form->isValid()) {
-            $this->logger->warning('2fa error occurred user "{username}" submitting the form "{errors}"', [
-                'username' => $dto->username,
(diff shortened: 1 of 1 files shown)


nobodyatroot:
Given the rare occurrence of this happening, I wouldn't consider it being a show stopper. Hopefully it gets the point across to the admin that there's something wrong with the user's email address, but I'm not opposed to a follow up PR should a better methodology come about to address Symfony's weird form error handling...which sounds like it's going to have to be fixed elsewhere.

PR #320, opened 2023-11-30
Changes to report links and magazine report fix

e-five256:
- change report links to go to pending route
- move pending to far left and all to far right
- remove non-functional filter
- fix duplicate where clause

I found the default page landing on all reports to be fairly useless, generally I think you're going to care about pending reports rather than all reports ever made. Instead of changing the default in routing (which I think would change say `/reports` to be pending and `/reports/all` to exist which might break history), I changed the links that go to reports to go to `/pending` instead. I also rearranged the menu items to be: Pending, Accepted, Rejected, All

This also fixes a 500 error when viewing magazine reports for accepted/pending/rejected that I noticed while making this change

image

### DIFF
--- a/templates/magazine/panel/_options.html.twig
+++ b/templates/magazine/panel/_options.html.twig
@@ -1,3 +1,4 @@
+{%- set STATUS_PENDING = constant('App\\Entity\\Report::STATUS_PENDING') -%}
 <aside class="options options--top options-activity" id="options">
     <div class="options__title">
         <h2>{{ 'magazine_panel'|trans }}</h2>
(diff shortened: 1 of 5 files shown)


templates/magazine/panel/_options.html.twig
```diff
@@ -64,33 +65,4 @@
             </a>
         </li>
     </menu>
-    {% if is_route_name('magazine_panel_reports') %}
```
e-five256, line 67: this was unfunctional as far as I could tell. I didn't even notice it had existed before, but clicking the links does nothing

image

nobodyatroot (approved):
Verified on kbin.run, looks good to me.

PR #323, opened 2023-12-02
Remove suspend feature

melroy89:
In Matrix we discussed about moving/rewriting this feature to limit/mute feature, but also that isn't ideal and makes it over-complicated. Thus in favor of: https://github.com/MbinOrg/mbin/pull/297

We came to the conclusion removing this suspend feature is the best option for now. And focus on moderation (federation) is more important. 

The introduced suspend feature was a hack anyway.


nobodyatroot:
I've never used this feature and I find it useless (IMO) based on the way it's implemented... so I have no issues with removing it, but I'd like to hear what other people think.

e-five256:
I honestly think the PR you closed #297 wasn't a bad start, I just had confusion around the direction we wanted to go for moderation actions. Seeing as how Jerry just had to use the suspend feature as ban wasn't performing the way he desired (stopping posts from federating in (hopefully #326 fixes? but has been difficult for me to test locally) / still spam), we might not want to completely remove this, or rather completely remove a way for an entire account to be marked as trashed

I certainly don't mind trying tackle getting admin feedback on what they'd like to see, I think we're sorely missing options that entire mod teams would need to function rather than just single admin instances (not just actions on account but things like batch operations, for instance)

e-five256:
Filling out something like this I think might help:

 current ->  goal?

| Action | Ban | Mute | Suspend? | Freeze? | Sensitive? |
|--------|:---:|:----:|:--------:|:-------:|:----------:|
| Can login | :x: | :ballot_box_with_check: | ? | ? | ? |
| Can post | :x: | :x: -> :ballot_box_with_check: ? | ? | ? | ? |
| Can reply | :x: | :ballot_box_with_check: | ? | ? | ? |
| New posts federate in | :ballot_box_with_check: -> :x: ? | ? | ? | ? | ? |
| New posts federate out | :x: | ? | ? | ? | ? |
| Posts visible by: | - | - | - | - | - |
| - Unauthenticated | :ballot_box_with_check: | :x: | ? | ? | ? |
| - Authenticated | :ballot_box_with_check: | :x: | ? | ? | ? |
| - Followers | :ballot_box_with_check: | :x: -> :ballot_box_with_check: ? | ? | ? | ? |
| - Magazine mods | :ballot_box_with_check: | :x: -> :ballot_box_with_check: ? | ? | ? | ? |
| - Global mods | :ballot_box_with_check: | :x: -> :ballot_box_with_check: ? | ? | ? | ? |
| - Admins | :ballot_box_with_check: | :x: -> :ballot_box_with_check: ? | ? | ? | ? |

melroy89:
I convert to draft, since we first need to think about the long term goal. And come to some conclusions?

melroy89:
Uh.. should we keep the feature I guess? Should we rename it?

PR #325, opened 2023-12-03
Add missing deleted by translations

melroy89:
Add missing "deleted by" translation strings.

Tested and working now: https://kbin.melroy.org/m/testing/p/23213/Test-post-on-mbin

### DIFF
--- a/templates/components/entry.html.twig
+++ b/templates/components/entry.html.twig
@@ -49,9 +49,9 @@
                         </h2>
                     {% endif %}
                 {% elseif(entry.visibility is same as 'trashed') %}
-                    <p class="text-muted"><i>{{ 'deleted_by_moderator'|trans }}</i></p>
+                    <p class="text-muted font-italic">&lsqb;<i>{{ 'deleted_by_moderator'|trans }}</i>&rsqb;</p>
                 {% elseif(entry.visibility is same as 'soft_deleted') %}
-                    <p class="text-muted"><i>{{ 'deleted_by_author'|trans }}</i></p>
+                    <p class="text-muted font-italic">&lsqb;<i>{{ 'deleted_by_author'|trans }}</i>&rsqb;</p>
                 {% endif %}
             </header>
             {% if entry.visibility in ['visible', 'private'] or (entry.visibility is same as 'trashed' and this.canSeeTrashed) %}
--- a/templates/components/entry.html.twig
+++ b/templates/components/entry.html.twig
@@ -49,9 +49,9 @@
                         </h2>
                     {% endif %}
                 {% elseif(entry.visibility is same as 'trashed') %}
-                    <p class="text-muted">{{ 'deleted_by_moderator' }}</p>
+                    <p class="text-muted">{{ 'deleted_by_moderator'|trans }}</p>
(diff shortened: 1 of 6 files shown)


translations/messages.en.yaml
```diff
@@ -778,3 +778,5 @@ user_badge_moderator: Mod
 user_badge_bot: Bot
 announcement: Announcement
 keywords: Keywords
+deleted_by_moderator: Thread, post or comment is deleted by the moderator.
```
e-five256, line 781: I do think we should translate these but I think it should remain more technical than a sentence. reddit/lemmy use `[deleted by moderator]` `[deleted by author]`, I think? Writing a sentence makes it seem no different than a message someone sent

templates/components/entry.html.twig
```diff
@@ -49,9 +49,9 @@
                         </h2>
                     {% endif %}
                 {% elseif(entry.visibility is same as 'trashed') %}
-                    <p class="text-muted">{{ 'deleted_by_moderator' }}</p>
+                    <p class="text-muted"><i>{{ 'deleted_by_moderator'|trans }}</i></p>
```
e-five256, line 52: Not sure on `<i>` tags here, I thought `<i>` and `<b>` had both been deprecated out of html, but it looks like https://developer.mozilla.org/en-US/docs/Web/HTML/Element/em#i_vs._em `<i>` can still be used for "technical terms" but it's hard to say what that exactly means, perhaps if this is an error code or something, it would count

melroy89: I didn't like the text without italic so I used: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/i

translations/messages.en.yaml
melroy89: I can add it between brackets? It's up to us how we want to handle this message. Currently vintprox was actually confused about this now. Which I understand, hence the reason of this PR.

melroy89: agreed?

e-five256: I'd agree from an accessibility point of view it could be viewed as better compared to reddit / lemmy

templates/components/entry.html.twig
```diff
@@ -49,9 +49,9 @@
                         </h2>
                     {% endif %}
                 {% elseif(entry.visibility is same as 'trashed') %}
-                    <p class="text-muted">{{ 'deleted_by_moderator' }}</p>
+                    <p class="text-muted font-italic">&lsqb;<i>{{ 'deleted_by_moderator'|trans }}</i>&rsqb;</p>
```
e-five256, line 52: `font-italic` might be leftover here? Looks like is isn't a class in any `scss` file so might be able to remove it, as it's not on any of the others

translations/messages.en.yaml
```diff
@@ -778,3 +778,5 @@ user_badge_moderator: Mod
 user_badge_bot: Bot
 announcement: Announcement
 keywords: Keywords
+deleted_by_moderator: Thread, post or comment is deleted by the moderator
+deleted_by_author: Thread, post or comment is deleted by the author
```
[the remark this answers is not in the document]
melroy89: Agreed!

e-five256 (approved):

PR #328, opened 2023-12-04
change AP as:Hashtag text property key to 'name'

asdfzdfj:
changed the property of `as:Hashtag` object that contains tag's label from `tag` to `name` to be more consistent with other software

compared to how other software (specifically, mastodon and akkoma) does tags in AP representation, they use `name` property to put the tag's label in but mbin/kbin was using `tag` property for some reason, and from what I could see from git blame, it switched between `name` and `tag` a few times

### DIFF
--- a/src/Service/ActivityPub/Wrapper/TagsWrapper.php
+++ b/src/Service/ActivityPub/Wrapper/TagsWrapper.php
@@ -21,7 +21,7 @@ class TagsWrapper
                 ['name' => $tag],
                 UrlGeneratorInterface::ABSOLUTE_URL
             ),
-            'tag' => '#'.$tag,
+            'name' => '#'.$tag,
         ], $tags);
     }
 }


nobodyatroot:
Is there any concern or impact to Lemmy compatibility with this change or is it irrelevant/not used there?

asdfzdfj:
> Is there any concern or impact to Lemmy compatibility with this change or is it irrelevant/not used there?

doesn't look like lemmy is using hashtags in any way, so this shouldn't cause problems for them. this one is more on the microblog side

generally, `as:Hashtag` tend to look like this:
```jsonc
{
  "@context": [/* snipped */],
  "type": "Note", 
  "tag": [
    {
      "type": "Hashtag",
      "href": "https://instance.tld/tags/art",
      "name": "#art"
    }
  ]
}
```

but for some reason, mbin outputs what looks like this, which doesn't line up with other places:
```jsonc
{
  "@context": [/* snipped */],
  "type": "Note", 
  "tag": [
    {
      "type": "Hashtag",
      "href": "https://mbin.instance/tags/art",
      /* v- should be "name" here*/
      "tag": "#art"
    }
  ]
}
```

nobodyatroot (approved):

PR #330, opened 2023-12-04
change inline js `on*=` actions to stimulus controllers/actions

asdfzdfj:
moved inline js into stimulus controllers and actions

- replaced `<select>` location change
- replaced "are you sure?" confirm action
- replaced "reload to apply" button in appearance setting

please try to make use of stimulus for js needs rather than adding inline js afther this


nobodyatroot (approved):
Makes sense to me

PR #333, opened 2023-12-05
using blurhash for entry thumbnail blurred background

asdfzdfj:
quick and dirty thumbnail filler background image using blurhash

- can be generated on twig side using `blurhash_image` component, feeding blurhash as input props
- the image is embedded in `` tag using `data:image/png;base64`, with default resolution of 20x20 and stretched with css for display, replacing .image-filler css background-image + blur filter

look something like this:
image

### DIFF
--- a/assets/styles/components/_comment.scss
+++ b/assets/styles/components/_comment.scss
@@ -171,7 +171,6 @@ $comment-margin-sm: .3rem;
     figure img {
       margin: .5rem 0;
       max-width: 100%;
-      border: var(--kbin-avatar-border);
     }
   }
 
--- a/assets/styles/components/_comment.scss
+++ b/assets/styles/components/_comment.scss
@@ -141,7 +141,7 @@ $comment-margin-sm: .3rem;
       z-index: 4;
 
       & > a.active,
-      & > li button.active, {
+      & > li button.active {
         text-decoration: underline;
       }
 
@@ -171,6 +171,7 @@ $comment-margin-sm: .3rem;
     figure img {
       margin: .5rem 0;
       max-width: 100%;
+      border: var(--kbin-avatar-border);
     }
   }
 
--- a/templates/components/entry.html.twig
+++ b/templates/components/entry.html.twig
@@ -87,7 +87,9 @@
                     {% if entry.type is same as 'link' or entry.type is same as 'video' %}
                         <figure>
                             <div class="image-filler" aria-hidden="true">
-                                {{ component('blurhash_image', {blurhash: entry.image.blurhash}) }}
+                                {% if entry.image.blurhash %}
+                                    {{ component('blurhash_image', {blurhash: entry.image.blurhash}) }}
+                                {% endif %}
                             </div>
                             <a href="{{ is_route_name('entry_single') ? entry.url : entry_url(entry) }}"
                                rel="{{ get_rel(is_route_name('entry_single') ? entry.url : entry_url(entry)) }}">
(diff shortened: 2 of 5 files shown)


templates/components/entry.html.twig
```diff
@@ -81,9 +81,9 @@
                 {% if entry.image %}
                     {% if entry.type is same as 'link' or entry.type is same as 'video' %}
                         <figure>
-                            <div class="image-filler"
-                                 style="background-image: url({{ asset(entry.image.filePath) |imagine_filter('entry_thumb') }})"
-                                 aria-hidden="true"></div>
+                            <div class="image-filler" aria-hidden="true">
+                                {{ component('blurhash_image', {blurhash: entry.image.blurhash}) }}
```
e-five256, line 85: this might need a fallback? I'm getting `Error rendering "blurhash_image" component: Expected argument of type "string", "null" given at property path "blurhash".`

asdfzdfj: noted. opted to add a guard around blurhash filler component and provides backgroud color for `image-filler` div

it should look something like this when there's no blurhash image:

image

image

e-five256: I thought it was going to be anything generated by fixtures but that seems incorrect, many fixtures have it. Looking at my db, the only thing I'm noticing is all of the heights are very tiny

<details>
<summary>db query</summary>

```
# select width, height, blurhash from image where blurhash is null;
 width | height | blurhash 
-------+--------+----------
   300 |     13 | 
   300 |     11 | 
   300 |     11 | 
   300 |      8 | 
   300 |     11 | 
   300 |      1 | 
   300 |      6 | 
   300 |      4 | 
   300 |     12 | 
   300 |     13 | 
   300 |      5 | 
   300 |      7 | 
   300 |      4 | 
   300 |      5 | 
   300 |     13 | 
   300 |      6 | 
   300 |     12 | 
   300 |      5 | 
   300 |     11 | 
   300 |     11 | 
   300 |     11 | 
   300 |      3 | 
   300 |      6 | 
   300 |      7 | 
   300 |      4 | 
   300 |     11 | 
   300 |      2 | 
   300 |      1 | 
   300 |      3 | 
   300 |      4 | 
   300 |      8 | 
   300 |      7 | 
   300 |     13 | 
   300 |     13 | 
   300 |      9 | 
   300 |      4 | 
   300 |      7 | 
   300 |     12 | 
   300 |     12 | 
   300 |      2 | 
   300 |      6 | 
   300 |      6 | 
   300 |      7 | 
   300 |      6 | 
(44 rows)

```

</details>

assets/styles/components/_comment.scss
```diff
@@ -171,6 +171,7 @@ $comment-margin-sm: .3rem;
     figure img {
       margin: .5rem 0;
       max-width: 100%;
+      border: var(--kbin-avatar-border);
```
e-five256, line 174: I'm slightly curious what situations you weren't seeing this in. It looks like it goes away at breakpoint < sm but looking at your instance that still might be true

e-five256 (dismissed):

assets/styles/components/_comment.scss
asdfzdfj: oops perhaps this one shouldn't be here, I think I should remove it

but basically this applies to image attached in comments, especially in microblog section, as I find it a bit odd that the image in top level post get to have this border but the following replies don't 

<details><summary>some example from fedia.io</summary>
<p>

https://fedia.io/m/random/p/239118/WTF

top level post:
image

comments:
image

</p>
</details>

e-five256 (approved):

PR #337, opened 2023-12-06
Add a small margin between top level entry comments

e-five256:
Adds a very small margin above top level entry comments after the first one (as the first one already has spacing from the reply box / sort order. It also adds the current view style class to the `<section>` tag as doing this in chat view makes no sense (top parents are together at the top with their replies all at the bottom completely unlinked)

this was first suggested here https://codeberg.org/Kbin/kbin-core/issues/1165

I've been running with this in tampermonkey ever since, and I think it helps improve readability / flow and is a small enough change that hopefully wouldn't upset too many people

when comparing the screenshots, look at the spacing right above top parent 2 & 3

also does a bunch of cleanup of constants, using the correct values of the cookies to compare against just in case anyone changes the constant for true in the controller to something else, matching what @asdfzdfj did for the sidebar settings originally

before tree
image

after tree
image

<details>
<summary>classic</summary>

before classic
image

after classic
image

</details>

### DIFF
--- a/templates/post/single.html.twig
+++ b/templates/post/single.html.twig
@@ -39,19 +39,19 @@
         {% include 'layout/_flash.html.twig' %}
 
         {% if user is defined and user and user.visibility is same as 'visible' and (user_settings.comment_reply_position == constant('App\\Controller\\User\\ThemeSettingsController::TOP')) %}
-        <section id="comment-add" class="section">
-            {% include 'post/comment/_form_comment.html.twig' %}
-        </section>
+            <div id="comment-add" class="section">
+                {% include 'post/comment/_form_comment.html.twig' %}
+            </div>
         {% endif %}
 
-        <section id="comments">
+        <div id="comments">
             {% include 'post/comment/_list.html.twig' %}
-        </section>
+        </div>
 
         {% if user is defined and user and user.visibility is same as 'visible' and (user_settings.comment_reply_position == constant('App\\Controller\\User\\ThemeSettingsController::BOTTOM')) %}
-        <section id="comment-add" class="section">
-            {% include 'post/comment/_form_comment.html.twig' %}
-        </section>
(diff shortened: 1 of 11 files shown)


asdfzdfj:
perhaps consider adding this for microblog post single view too?

e-five256:
looking at doing that made me realize there's a bug here https://github.com/MbinOrg/mbin/blob/main/templates/post/comment/_list.html.twig#L13-L14

so will be spending some time trying to pull apart what the styling should look like

e-five256:
so this became quite a lot of changes for something I was hoping was simple, almost tempted to just take out the `.scss` change and change what this PR is for, but I think everything should probably still look the same except the post single page which now gets that spacing

this is the post _list page which is unchanged, it had a gap between different posts and still does
image

this is looking at a specific post, which does gain the gap in tree/classic views
image

might be hard to make out but the post in the middle of the first image here did not have a gap on the `/microblog` page but does at the bottom of the `/p/` single post page in the 2nd image

templates/post/single.html.twig
```diff
@@ -39,19 +39,19 @@
...
-        <section id="comment-add" class="section">
-            {% include 'post/comment/_form_comment.html.twig' %}
-        </section>
+            <div id="comment-add" class="section">
+                {% include 'post/comment/_form_comment.html.twig' %}
+            </div>
         {% endif %}
 
-        <section id="comments">
+        <div id="comments">
```
e-five256, line 47: I changed the DOM around in places to make post and entry match closer, removing differences where they didn't seem to make sense. I think the main difference at this point is when viewing a specific entry or post, the one being viewed is in a `<blockquote>` for post and `<article>` for entry, but DOM like comments below that are nearly synced up now, hoping to make styling easier/consistent and make it easier to unify these things

nobodyatroot (approved):
Looks good to me

PR #339, opened 2023-12-10
Hide NSFW posts on user overview page when logged out or hide NSFW enabled

e-five256:
This filters out nsfw content where the user is the **author** of the entry, post, or comment. It does not filter out non-nsfw marked replies to nsfw content (wanted to keep this small for now to keep it simple)

Logged out:
image

Logged in with a user with hide nsfw checked:
image

Logged in with a user with hide nsfw unchecked:
image


nobodyatroot (approved):

nobodyatroot (approved):
Updated with the latest commit to fix the `outbox` route, everything seems to be working fine for the past 10 hours or so... no abnormal log or message activity.

PR #341, opened 2023-12-11
add job for certificate update to bare metal instructions

TheVillageGuy:
Please check if I am applying the (for me) new document formatting correctly

### DIFF
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -701,7 +701,6 @@ sudo cronjob -e
 43 5 * * * sudo certbot renew --pre-hook "sudo systemctl stop nginx" --post-hook "sudo systemctl start nginx"
 ```
 
-
 ### Additional Mbin configuration files
 
 These are additional configuration YAML file changes in the `config` directory.
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -692,7 +692,7 @@ sudo certbot certonly --standalone -d domain.tld -d www.domain.tld
 sudo certbot --nginx -d domain.tld -d www.domain.tld
 ```
 
-For Debian/Ubuntu systemes, running nginx on port 80, set up a daily job to update your certificates (at 5h43m)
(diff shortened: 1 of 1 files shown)


melroy89 (changes requested):
Rejected. As I explain [on stackoverflow here](https://stackoverflow.com/questions/41535546/how-do-i-schedule-the-lets-encrypt-certbot-to-automatically-renew-my-certificat/54508220#54508220), when you install certbot there are already preinstalled systemd timer installed for certbot renew. 

Please, do not introduce a new cronjob, this is not required. If you need to create a cronjob you definitely did something wrong with your local setup.

TheVillageGuy:
> Rejected. As I explain [on stackoverflow here](https://stackoverflow.com/questions/41535546/how-do-i-schedule-the-lets-encrypt-certbot-to-automatically-renew-my-certificat/54508220#54508220), when you install certbot there are already preinstalled systemd timer installed for certbot renew.
> 
> Please, do not introduce a new cronjob, this is not required. If you need to create a cronjob you definitely did something wrong with your local setup.

the timers are installed but they fail. 
```
Mon 2023-12-11 18:05:00 UTC 42min left  -                           -             snap.certbot.renew.timer     snap.certbot.renew.service
```

I have seen other people having this issue with debian (not sure about ubuntu) and most of them have solved it by opening port 80 (temporarily shutting down nginx)

which is also what it says in the logs;

```
certbot.errors.PluginError: Could not bind TCP port 80 because it is already in use by another process on this system (such as a web server). Please stop the program in question and then try again.
```

melroy89:
Also, when using `certbot --nginx` you will use the Nginx installer, which also takes care about Nginx reload.

For example, since I use `certbot --nginx` I see all the confs located at: `/etc/letsencrypt/renewal/` and these conf files are aware that the certificate is deployed to a Nginx service:

```
authenticator = nginx
installer = nginx
server = https://acme-v02.api.letsencrypt.org/directory
key_type = ecdsa
```

Thus again, no need for a manual cronjob at all.

PR #342, opened 2023-12-11
using blurhash for obscuring sensitive image

asdfzdfj:
replaced blur filter usage over sensitive image to obscure the image with blurhash image, using hidden checkbox + css to control visibility and buttons to switch between show and hide sensitive image

for image attachment in post and comments, hide the sensitive image by putting blurhash image over along with the show button

for entry thumbnail, it works a bit differently:
- the sensitive image is hidden by hiding the main image and show existing blurhash background image instead
- due to space constraints, the show/hide button is icons only, additionally when in mobile layout and compact view, the hide button will not be shown after the image is revealed

currently looks something like this:

| location | hidden | revealed |
|---|---|---|
| post + comments | hidden | revealed |
| entry thumbnails | hidden | revealed |
| entry thumbnail (mobile+compact) | hidden | revealed |

### DIFF
--- a/templates/components/_figure_entry.html.twig
+++ b/templates/components/_figure_entry.html.twig
@@ -1,7 +1,8 @@
 {# this fragment is only meant to be used in entry component #}
 {% with {
     sensitiveId: 'sensitive-check-'~entry.image.id,
-    isSingle: is_route_name('entry_single')
+    isSingle: is_route_name('entry_single'),
+    route: isSingle ? singleRoute : entry_url(entry)
 } %}
 <figure>
     <div class="image-filler" aria-hidden="true">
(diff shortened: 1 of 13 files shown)


melroy89 (dismissed):

templates/components/_figure_entry.html.twig
```diff
@@ -0,0 +1,43 @@
+{# this fragment is only meant to be used in entry component #}
```
melroy89, line 1: this is just a comment right.. ? or a todo?

melroy89:
Hopefully this is the last branch that is outside of this repo, since it really makes it much harder to test if I want to switch branches.

melroy89 (approved):

PR #340, opened 2023-12-10
Add current sort and time options to navbar links

e-five256:
This changes the Threads and Microblog navbar link to preserve the currently active `sortBy` and `time` settings

| On | Current thread link | Current microblog link | New thread link | New microblog link |
|--------|--------|--------|--------|--------|
| `/` or `/microblog` | `/all` | `/microblog` | `/all` | `/microblog` |
| `/newest` or `/microblog/newest` | `/all` | `/microblog` | `/all/newest` | `/microblog/newest` |
| `/newest/1m` or `/microblog/newest/1m` | `/all` | `/microblog` | `/all/newest/1m` | `/microblog/newest/1m` |
| `/sub` or `/sub/microblog` | `/sub` | `/sub/microblog` | `/sub` | `/sub/microblog` |
| `/sub/newest` or `/sub/microblog/newest` | `/sub` | `/sub/microblog` | `/sub/newest` | `/sub/microblog/newest` |
| `/sub/newest/1m` or `/sub/microblog/newest/1m` | `/sub` | `/sub/microblog` | `/sub/newest/1m` | `/sub/microblog/newest/1m` |

Etc for `/mod` and `/fav`. At the moment it's not preserving `type` option as that is specific to threads and not shared by microblogs, but if desired I could add it. I'm a bit skeptical of the usefulness of filtering by "type" (which are: links, thread, photos, videos)

PR #343, opened 2023-12-12
fix comment sort order being used for navbar sort

e-five256:
noticed that when you browsed to a specific entry or post page, sorting the comments was using that sort in the navbar threads / microblog link, which was unintentional. perhaps the args should be changed to differentiate sorting comments and sorting entry/posts, but for now opted to just ignore the `sortBy` option in regards to the navbar when the page you are on is a comment / single page

related #340

### DIFF
--- a/src/Twig/Runtime/NavbarExtensionRuntime.php
+++ b/src/Twig/Runtime/NavbarExtensionRuntime.php
@@ -33,22 +33,22 @@ class NavbarExtensionRuntime implements RuntimeExtensionInterface
             ]);
         }
 
-        if (str_starts_with($this->getCurrentRouteName(), 'tag')) {
+        if ($this->isRouteNameStartsWith('tag')) {
             return $this->urlGenerator->generate(
                 'tag_entries',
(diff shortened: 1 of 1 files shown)


src/Twig/Runtime/NavbarExtensionRuntime.php
```diff
@@ -109,11 +109,26 @@ private function getCurrentRouteName(): string
...
     {
+        $options = [];
+
+        // don't use sortBy or time options on comment pages
+        // for the navbar links, so sorting comments by new does not mean
+        // changing the entry and microblog views to newest
+        if (!$this->isRouteName('root')
+            && !$this->isRouteNameStartsWith('front')
+            && !$this->isRouteNameStartsWith('posts')
+            && !$this->isRouteName('magazine_posts')
```
e-five256, line 120: unrelated but think we should rename this route to `posts_magazine` to match `front_magazine` that entry has, which matches all the other subcategories, that way this check could be removed and it would match closer to the entry routing

nobodyatroot (approved):

PR #360, opened 2023-12-19
Update time filter display

e-five256:
- change time filter to show currently selected option when on larger display than mobile
- change icon from asc,desc icon to clock
- move time filter first

not sure about other people but I found the time filter very hidden and confusing
image

the middle option here is time, but it uses an ascending / descending icon and doesn't really indicate what it's for or doing

this changes the bar to look like this:
image

showing the currently selected option, moved it to be the first as it seemed more important than filtering entry types

another look:
image

showing dropdown:
image

microblog:
image

on mobile it will not show the label to save room:
image

### DIFF
--- a/assets/styles/layout/_options.scss
+++ b/assets/styles/layout/_options.scss
@@ -28,15 +28,6 @@
     }
   }
 
- &__layout{
-    li{
-      button{
-        border-bottom-left-radius: 0px!important;
-        border-bottom-right-radius: 0px!important;
-      }
-    }
-  }
-
   .scroll {
     position: static;
     align-self: center;
@@ -48,7 +39,7 @@
     .scroll-left,
     .scroll-right{
       padding: 0.5rem;
-      cursor: pointer;   
+      cursor: pointer;
       color: var(--kbin-button-secondary-text-color);
 
       &:hover,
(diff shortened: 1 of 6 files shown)


templates/entry/_options.html.twig
```diff
@@ -112,7 +92,8 @@
             <button
                     aria-label="{{ 'filter_by_type'|trans }}"
                     title="{{ 'filter_by_type'|trans }}"><i
-                        class="fa-solid fa-filter"></i> {{ 'filter_by_type'|trans }}</button>
```
e-five256, line 115: I removed it saying "Filter by time" "Filter by type" "Change view". They were doing it a font-size 0 so it was never displayed visually. I _am_ wondering if this was for screen readers; but according to google, which may or may not be accurate as it has changed a lot over the years, setting to font size 0 isn't recommended (just as display none would also be skipped over). If it is an issue, we'd have it in lots of other places like federation / settings sidebar as those also don't have any text hidden on them

assets/styles/layout/_options.scss
```diff
@@ -28,15 +28,6 @@
     }
   }
 
- &__layout{
```
e-five256, line 31: it seemed like this was trying to do something to keep the border radius only effecting the right most option the layout before was

- `menu`
  - `li`
- `menu`
  - `li`
  - `li`

I changed this to just be one `menu` with 3 `li`s and made the css selector simpler with `:not last`, `:last` so it's always accurate no matter how many items exist

nobodyatroot (approved):

PR #369, opened 2023-12-20
Announce posts and comments

BentiGorlich:
Right now magazines hosted on mBin never announce posts or comments.

### Why is that a problem

We have instances A, B, and C and Users X (on A), Y (on B) and Z (on C). We have a magazine F on instance A. When Y posts something to F then only the instances A and B know of that post. Z will never know this post exists unless another activity referencing this post is sent to C (that would be likes or comments, but **only** from users on A, because other activities will not be forwarded to C).

This is always a problem, but especially for small communities.

### Why is announcing the solution

sending an `Announce` activity solves the problem, because it is basically forwarding. So every post created on a magazine, from local or remote users are just forwarded to every server with at least one subsciber

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -11,6 +11,7 @@ use App\Entity\Traits\ActivityPubActorTrait;
 use App\Entity\Traits\CreatedAtTrait;
 use App\Entity\Traits\VisibilityTrait;
 use App\Repository\UserRepository;
+use App\Service\ActivityPub\ApHttpClient;
 use Doctrine\Common\Collections\ArrayCollection;
 use Doctrine\Common\Collections\Collection;
 use Doctrine\Common\Collections\Criteria;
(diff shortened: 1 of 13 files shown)


BentiGorlich:
This is only a draft atm, because the forwarding for comments is not implemented, yet. I didn't get around to it and probably won't until the second week of January

BentiGorlich:
Comments will also get announced now

BentiGorlich:
I have this live on gehirneimer.de now

nobodyatroot (dismissed):
Works as expected on my test cluster with 3 federated Mbins! Nice work!

BentiGorlich:
shall I wait for someone to do a code review?

nobodyatroot:
> shall I wait for someone to do a code review?

it looks pretty straightforward and nothing obvious stands out to me, but maybe @e-five256 and/or @asdfzdfj should review in detail before merging?

asdfzdfj:
aside from that one outbox dispatching where the message class is suddenly full class name (presumably because of name collision, you could do something like `use ... as ...` to rename used identifiers), nothing particularly stands out at the moment. 

and if I found any after merge, don't worry: there will be patches (and for a healthier C4-ish contribution style that I understand and we're trying to use, it should be)

nobodyatroot (approved):

PR #379, opened 2023-12-23
adjust logging for docker setups

asdfzdfj:
I thought someone already hooked the app log to docker logs but apparently not so here goes:

- mount the log directory within `php`, `messenger` and `messenger_ap` services to host `storage/logs` directory for easier access and more persistent log storage, so it doesn't get washed away when recreating containers while trying to debug something
- adjust logging to also send logs to stderr, which will be output to docker logs, and can be viewed using `docker logs`, `docker compose logs` and the like  
  for bare metal setups, this stderr log should already be discarded by php-fpm (see php-fpm option `catch_workers_output`)
- add extra instructions and mentions in the docker guide about logging, as introduced by this change

### DIFF
--- a/config/packages/monolog.yaml
+++ b/config/packages/monolog.yaml
@@ -14,6 +14,11 @@ when@dev:
                 include_stacktraces: true
                 max_files: 10
                 channels: ['!event']
+            stderr:
+                type: stream
+                path: 'php://stderr'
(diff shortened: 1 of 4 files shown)


melroy89 (approved):
This will also make the logging persistent between restarts. Great!

PR #384, opened 2023-12-25
Accessibility updates to FontAwesome icons

e-five256:
Was reading the [mozilla docs on aria-hidden](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-hidden) and noticed the specific example use case they showed was for fontawesome

that led me to https://fontawesome.com/v4/accessibility/
which basically seems to indicate:
- all fontawesome icons should have the `aria-hidden="true"` attribute
- if an element is purely decorative, nothing else is needed
- if the element is the only interactive element there is, it should be in a container with at least `aria-label="<descriptive label>"` as a parent

to give a specific example

```html
<i class="fa-solid fa-ban" aria-hidden="true"></i> <span>Ban</span>
```

This has `aria-hidden="true"` and nothing further as there is descriptive text after it, the icon is purely decoration

```html
<button aria-label="Ban"
    <i class="fa-solid fa-ban" aria-hidden="true"></i>
</button>
```

this has no descriptive text and only shows the icon as the interactive element, so gains the `aria-label="<descriptive label>"` in the parent to indicate what it does

(also, based on https://fontawesome.com/docs/web/dig-deeper/accessibility#when-is-auto-accessibility-available I don't _think_ mbin has auto accessibility. As far as I could tell it was saying the rendered HTML output would gain `aria-hidden="true"` even if not included but that is not true as far as I can tell. I hope I didn't waste a bunch of time...)

also went through and added `title` to some locations especially where there is no descriptive text for what clicking a button does


assets/styles/components/_header.scss
```diff
@@ -233,7 +233,7 @@
     text-decoration: none;
     height: 100%;
 
-    div {
```
e-five256, line 236: it felt very brittle that the hamburger icon in the header was styled solely by being the only `<div>` element in the header. Gave it an `id` instead as there should only be one menu on the page inside the header

templates/bundles/NelmioApiDocBundle/SwaggerUi/index.html.twig
```diff
@@ -24,7 +24,6 @@
         <div class="kbin-container">
             <nav class="head-nav"  data-action="click->kbin#toggleNav">
                 <div class="brand">
-                    <div><i class="fa-solid fa-bars"></i></div>
```
e-five256, line 27: this was a hamburger that appears on the `/api/docs` route only on mobile screen widths. clicking it did nothing. it seemed like there isn't a need for a hamburger on the api page so removed it

templates/base.html.twig
```diff
@@ -101,8 +101,8 @@
 {% include 'layout/_footer.html.twig' with {inSidebar: true} %}
 {% include 'layout/_topbar.html.twig' %}
 <div id="popover" class="popover js-popover section section--small" role="dialog"></div>
-<div id="scroll-top" data-controller="scroll-top" style="display: none;" data-action="click->scroll-top#scrollTop">
-    <i class="fa-solid fa-arrow-up"></i> <small class="hidden">0</small>
+<div id="scroll-top" data-controller="scroll-top" style="display: none;" data-action="click->scroll-top#scrollTop" aria-hidden="true">
```
e-five256, line 104: read a few search results on accessible scroll to top [such as this](https://www.digitala11y.com/scroll-to-top-does-it-need-to-be-accessible/). It seems to go between it not being needed for screenreaders and a general "everything should be accessible". It seems like the goal is to remove non-content, non-navigation from display of screenreaders so I opt'd to hide the scroll to top button from it, but can always change it if that proves incorrect

melroy89:
I cleared the caches and PHP cache, but the hamburger icon doesn't seems to load correctly on my phone?

e-five256:
Hmm, you're https://kbin.melroy.org/build/app.4809d7a1.css file looks outdated for some reason

it says `#header .brand div{cursor:pointer;display:none;font-size:.9rem}` but it should be `#header .brand #nav-toggle{cursor:pointer;display:none;font-size:.9rem}`

nobodyatroot:
> Hmm, your https://kbin.melroy.org/build/app.4809d7a1.css file looks outdated for some reason
> 
> it says `#header .brand div{cursor:pointer;display:none;font-size:.9rem}` [but it should be](https://github.com/MbinOrg/mbin/pull/384/files#diff-d4f21d78a12635721ee89faebefaa81c4a4d2d574ad154de88d20b0c29945ab9R236) `#header .brand #nav-toggle{cursor:pointer;display:none;font-size:.9rem}`

FYI, the normal update flow is resulting in the same behavior for me on my dev instances...

e-five256:
my mistake, hopefully fixed now, it was gaining styling as a `<button>`. I reverted it back to a `<div>` (although it didn't look _too_ bad as either `btn__primary` or `btn__secondary`, I can't quite tell if it's good to attract attention to the hamburger on mobile or best to make it muted, I went with keeping it how it was for now), but kept the added `id` and styling by the `id` so other `div`s could be added to the header at some point without having styling conflicts

nobodyatroot (approved):
Hamburger icon is fixed on "mobile" after last commit for me, everything else seems to work ok.

melroy89:
> Hmm, your https://kbin.melroy.org/build/app.4809d7a1.css file looks outdated for some reason
> 
> it says `#header .brand div{cursor:pointer;display:none;font-size:.9rem}` [but it should be](https://github.com/MbinOrg/mbin/pull/384/files#diff-d4f21d78a12635721ee89faebefaa81c4a4d2d574ad154de88d20b0c29945ab9R236) `#header .brand #nav-toggle{cursor:pointer;display:none;font-size:.9rem}`

I moved back to `main` branch again.

Issue #66, opened 2023-10-18
Mobile sidebar footer in light mode not readable


e-five256:
With the addition of the footer to the sidebar in #63, in the light theme it is gaining the background of the section (`kbin-section-bg`) which is light and the text color (`kbin-footer-link-color`) which is light. This leads to it being unreadable.

~~There is also an issue where, with the sidebar open, the footer in the sidebar and the footer appear, leading to seeing both of them, but I'm not sure if this is intended or not.~~ not a bug

The screenshot shows the sidebar footer in light mode.

![Screenshot from 2023-10-17 22-02-50](https://github.com/MbinOrg/mbin/assets/146029455/9f8f0228-918e-4fb7-a2ee-55281c415161)

simonrcodrington:
I think the idea with the footer in the sidebar was just so you could get to the bottom of the page on mobile in places where you had infinite scroll (where beforehand you could never get to the bottom)

In cases where there's no infinite scroll, we're gonna get a double footer here on mobile (one in the sidebar and then the original footer at the bottom) . Though, if you have the sidebar open, you shouldn't be able to see the original footer (as the sidebar should be full height on mobile)

The coloring for the theme is definitely an issue. I'll schedule it in to have a look. Most likely we'll define some specific colors for it when the footer appears in the mobile sidebar

melroy89:
@simonrcodrington Thanks in advance to picking this up.

melroy89:
@simonrcodrington  It's only in the mobile sidebar footer indeed the "normal footer" doesn't have this issue..

e-five256:
no longer an issue now that the information is in the sidebar with the same styling as the other sidebar components

PR #385, opened 2023-12-25
Move footer information into sidebar

e-five256:
This was originally suggested by @lilfade and others

Purpose:

Remove the footer from the site and add the information into the sidebar

Reasoning:

- Reaching the footer with infinite scrolling enabled was near impossible, and completely impossible on mobile ([kbin issue on the matter](https://codeberg.org/Kbin/kbin-core/issues/891))
- To combat that, @simonrcodrington added the footer to the sidebar on mobile, though we did see some issues with that #66 and potentially duplicated in some cases
- There are plenty of examples of not having a footer for social media sites, our peer mastodon for instance, does not have a footer https://mastodon.social nor activitypub social hub https://socialhub.activitypub.rocks/c/welcome/1 (admittedly this is a bit flimsy, I'm sure for every site that doesn't have one you could link one that does, lemmy for instance does but does not have infinite scrolling (unless that's locked behind having an account) (edit: perhaps you could say anything with infinite scrolling tends to not. mlmym's lemmy ui, mastodon, activity pub have infinite scrolling and no footer, lemmy ui does not have infinite scrolling and does have a footer)

Changes:

This is what the information in the sidebar currently looks like, though very much up for feedback as it's a lot of information condensed

image

<details>
<summary>bottom of the page now (it's what you'd expect, no footer)</summary>

image

</details>

<details>
<summary>image of mastodon legal sidebar for comparison</summary>

image

</details>

main callouts are:

- added current mbin version from project info service
- removed links to create new posts, those were all duplicated as they already exist in the header, same for list all magazines
- removed the © current year thing, couldn't really find any examples of anyone else doing that, not sure it provides a legal defense, but there is a terms of use already. though I remain not a lawyer
- rss feels a bit out of place, maybe it should go in the header instead? I've seen it there some times, or maybe it's fine as is, lang switch similarly might not be in the best spot


melroy89:
This also solves the endless scroll feaure dilemma and workarounds. I love it. Most social media websites do not have a footer at the bottom for a good reason.

nobodyatroot:
Should we turn infinite scroll on by default with this change? Sometimes I wonder how many users actually go through the effort to look at all the options available in the settings menu...

e-five256:
I think it's worth discussing if we want to make that the default for the software, but maybe not part of this specific change / pr. I would like to get more admin settings for all the defaults set up soonish

nobodyatroot (approved):

asdfzdfj:
the rss button could probably be moved to upper section with about/faq/rules/etc., and give it a full label ("RSS Feed" or whatever, icons optional), while keeping the interface language selector where it was, either centered or stretched to full width

e-five256:
this is what it looks like now

image

nobodyatroot (approved):

e-five256:
@asdfzdfj does the change look acceptable to you?

asdfzdfj (approved):
> @asdfzdfj does the change look acceptable to you?

good enough for me

PR #387, opened 2023-12-30
Rename "bot" user agent string and route to "agent"

e-five256:
Remove the word "bot" from the user agent string as `lemmy.world` appears to be blocking requests that have "bot" in them and replace it with "agent"

as an example, [mastodon](https://github.com/mastodon/mastodon/blob/main/lib/mastodon/version.rb#L68) does e.g. `(Mastodon/1.2.2; +http://mastodon.dev/)`

It seems they do append ` Bot` at the end specifically [for twitter cards](https://github.com/mastodon/mastodon/blob/main/app/services/fetch_link_card_service.rb#L48) but that was the only reference I noticed to changing it


melroy89:
If we merged it.. then also change the bot url to another url without the word 'bot'?

config/packages/framework.yaml
```diff
@@ -17,7 +17,7 @@ framework:
     http_client:
         default_options:
             headers:
-                'User-Agent': 'MbinBot/0.1 (+https://%kbin_domain%/bot)'
+                'User-Agent': 'Mbin/0.1 (+https://%kbin_domain%)'
```
nobodyatroot, line 20: Ooo, I see this still has the old version number in it... probably should update it to match what's in `ProjectInfoService.php`? We might have to rework how we're handling a global var to change the version everywhere it's called... could be forced to put it the `.yaml` unless there's a way to call it in from PHP somewhere.

nobodyatroot (approved):
Agent Smith approves!

PR #388, opened 2023-12-31
Update symfony doctrine recipe to 2.11

nobodyatroot:
No signs of regression during testing.


nobodyatroot:
So apparently this is only impacting my dev environment, not production... either way, update to the latest available version.

melroy89 (approved):
LGTM as well

PR #389, opened 2024-01-03
Update InlineParserMatch regex in MentionLinkParser to correctly process usernames with hyphens

nobodyatroot:
Existing `\w` in `InlineParserMatch` regex does not include hyphens (-) thereby breaking user mentions on usernames that include hyphens. PR explicitly enumerates all allowed characters in username.


e-five256 (approved):
confirmed this fixes parsing, thanks, now I can finally be mentioned

<details>
<summary>regex example</summary>

before https://regex101.com/r/I62he0/1
image

after https://regex101.com/r/TLEEoL/1
image
</details>

PR #390, opened 2024-01-03
add oauth keycloak env vars to docker example

e-five256:
when running php commands in dev I noticed I got a bunch of:

`Environment variable not found: \"OAUTH_KEYCLOAK_SECRET\".`

it looks like this section was added to `.env.example` https://github.com/MbinOrg/mbin/commit/fe322561263d3ccef121b9aabac4f9648b0270bc but doesn't appear in the docker example. I'm not quite sure if there are any gotchas attempting to use keycloak from docker setup, but this stops the log output


nobodyatroot (approved):
Yep, docker got out of sync with bare metal... again.

PR #394, opened 2024-01-04
redirect `/instances` to `/federation`

e-five256:
this simply makes a permanent redirect of `/instances` to `/federation`. lemmy users may be more familiar with the `/instances` route and we aren't currently using it, so didn't see a reason to not allow this for now. it's only an issue for people that might manually type out `example.com/instances` into their url bar as it isn't linked anywhere


nobodyatroot (approved):

e-five256:
sorry for the dismissal. at the time I didn't think much of the `page_` route but after sleeping on it I realized that means specifically "admin pages" the admin dashboard section. Thinking about whether I would want `route_name_starts_with('page')` to effect this route, I think maybe you likely wouldn't as it isn't one of the admin page options, so renamed it

melroy89 (approved):

PR #395, opened 2024-01-04
Handle exception on long embed URL length causing messages to go to failed queue

nobodyatroot:
PR "handles" this exception that's causing messages to be dumped into the failed queue... which will never succeed unless the DB schema is updated. Regardless, this PR code won't have to change since it won't cause an exception if the URL column datatype is "fixed" / adjusted:

```
Failed Message Details
======================

 ------------- -------------------------------------------------------------------------------------------------------------------------------------------------------- 
  Class         App\Message\LinkEmbedMessage                                                                                                                            
  Message Id    16855846                                                                                                                                                
  Failed at     2024-01-03 21:01:45                                                                                                                                     
  Error         An exception occurred while executing a query: SQLSTATE[22001]: String data, right truncated: 7 ERROR:  value too long for type character varying(255)  
  Error Code    7                                                                                                                                                       
  Error Class   Doctrine\DBAL\Exception\DriverException                                                                                                                 
  Transport     failed                                                                                                                                                  
 ------------- -------------------------------------------------------------------------------------------------------------------------------------------------------- 
```

Embed URLs are limited to VARCHAR(255) by current DB schema:

image

Tested on a local 3 node federated cluster, works as expected: valid links (255 chars or less) still get entries put into the db, URLs that violate the current length limit are logged to debug channel.

### DIFF
--- a/src/Repository/EmbedRepository.php
+++ b/src/Repository/EmbedRepository.php
@@ -7,6 +7,7 @@ namespace App\Repository;
 use App\Entity\Embed;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
 use Doctrine\Persistence\ManagerRegistry;
+use Psr\Log\LoggerInterface;
 
 /**
  * @method Embed|null find($id, $lockMode = null, $lockVersion = null)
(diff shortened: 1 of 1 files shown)


e-five256 (approved):
I'm a bit worried that we might forget about this and not notice why mbin is missing content in some situations, wonder if maybe it should be a warning rather than debug just in case but don't feel strongly about it. maybe it could change to a warning if we increase the length limit to something closer to url "max" like 2k

nobodyatroot:
> I'm a bit worried that we might forget about this and not notice why mbin is missing content in some situations, wonder if maybe it should be a warning rather than debug just in case but don't feel strongly about it

im fine with promoting it to warning, it doesn't happen _that_ often..usually. agreed it'd be something good to know, new commit incoming.

e-five256 (approved):

PR #396, opened 2024-01-04
somewhat fix phpunit so it could run functional tests again

asdfzdfj:
who tests the test infrastructure

not sure what happened but it seems to be a combination of:
- symfony/phpunit-bridge doesn't support phpunit 10 yet
- but we're using phpunit 10 anyway
- applied multiple config recipes and overwritten a decent chunk out of them
- dama/doctrine-test-bundle not active when running tests, leading to db error when testing due to data conflict and mismatch between tests data in db and what the tests sees and has access to

included fixes:
- applied a small phpunit config changes just enough to get it back up and running again
- included auto test database bootstrap for phpunit, activate by setting BOOTSTRAP_DB env variable when starting tests (e.g. `BOOTSTRAP_DB=1 ./bin/phpunit`)

### DIFF
--- a/phpunit.xml.dist
+++ b/phpunit.xml.dist
@@ -28,8 +28,12 @@
         <include>
             <directory suffix=".php">src</directory>
         </include>
+        <exclude>
+            <directory suffix=".php">src/DataFixtures</directory>
+        </exclude>
     </source>
 
     <extensions>
+        <bootstrap class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension" />
     </extensions>
 </phpunit>
--- a/tests/Functional/Controller/Magazine/MagazineListControllerTest.php
+++ b/tests/Functional/Controller/Magazine/MagazineListControllerTest.php
@@ -28,7 +28,7 @@ class MagazineListControllerTest extends WebTestCase
         );
     }
 
-    public function magazines(): iterable
+    public static function magazines(): iterable
     {
         return [
             [['query' => 'test'], []],
--- a/tests/bootstrap.php
+++ b/tests/bootstrap.php
@@ -2,6 +2,9 @@
 
 declare(strict_types=1);
 
+use App\Kernel;
+use Symfony\Bundle\FrameworkBundle\Console\Application;
+use Symfony\Component\Console\Input\ArrayInput;
 use Symfony\Component\Dotenv\Dotenv;
 
 require \dirname(__DIR__).'/vendor/autoload.php';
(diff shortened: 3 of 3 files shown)


melroy89:
What does make sense is to add the functional tests to github workflow.

e-five256:
I think we'd need sqlite support or something, I could be mistaken but I recall similar conversations about setting up dependencies and sqlite not supporting the specific postgres features we use. I think that can be out of scope for this, I'd just like to verify the functional tests work correctly with this change but am having trouble myself getting them to run, which probably isn't a good sign >.> making it possible in docker might be nice so system setup is a requirement, but again not putting that on asdfzdfj, just brainstorming ways to both learn myself how to run the tests and instruct others

melroy89:
> I think we'd need sqlite support or something, I could be mistaken but I recall similar conversations about setting up dependencies and sqlite not supporting the specific postgres features we use. I think that can be out of scope for this, I'd just like to verify the functional tests work correctly with this change but am having trouble myself getting them to run, which probably isn't a good sign >.> making it possible in docker might be nice so system setup isn't a requirement, but again not putting that on asdfzdfj, just brainstorming ways to both learn myself how to run the tests and instruct others

GitHub should allow use to use a PostgreSQL docker container as a dependency for a specific GitHub Action Job: https://docs.github.com/en/actions/using-containerized-services/creating-postgresql-service-containers. Allowing us to just leverage PostgreSQL and run the tests in the same matter as you might do locally on your dev system.

asdfzdfj:
> What does make sense is to add the functional tests to github workflow.

perhaps, I'm not familiar with how github workflow/ci works and it's a bit out of scope for this patch so feel free to make a patch doing that if you want to add this

still, I'd recommended to _not_ run the full tests all the time, as it's a **really expensive** operation, using quite a bit of memory and time (I had to raise my php mem limit to 1G for phpunit to not fail halfway because of running out memory, also it took about 10m to complete), at the very least I'd like this to be run before cutting a new release so regression could be fixed before actually making a release

melroy89 (approved):

phpunit.xml.dist
```diff
@@ -28,8 +28,12 @@
         <include>
             <directory suffix=".php">src</directory>
         </include>
+        <exclude>
+            <directory suffix=".php">src/DataFixtures</directory>
+        </exclude>
     </source>
 
     <extensions>
+        <bootstrap class="DAMA\DoctrineTestBundle\PHPUnit\PHPUnitExtension" />
```
melroy89, line 37: Keep in mind. I will remove this `extensions` element again in: https://github.com/MbinOrg/mbin/actions/runs/7949978285/job/21701747844

Because we are using PHPUnit 10, which doesn't have support for this `extensions` element anymore.

melroy89: IF that is an issue... Then I think we should down-grade PHPUnit test to v9 again.

asdfzdfj: from my understanding it's more like phpunit 10 expects the inner element of `<extensions>` tag to be of type `<bootstrap>` (as I put it there), rather than of type `<extension>` as it used to be.

I got that snippet straight from the [dama/doctrine-test-bundle docs](https://github.com/dmaicher/doctrine-test-bundle#using-the-bundle-with-phpunit), so it should be fine with phpunit 10

also removing this block will mean that the test bindle is not active, and then we'll be back to square one regarding getting functional tests to run

melroy89: Ah yes.. I see it now! great.

melroy89:
@asdfzdfj Follow up could be to briefly document the `BOOTSTRAP_DB=1 ./bin/phpunit` usage, in CONTRIBUTING.md?

PR #402, opened 2024-01-05
Add various null/empty checks to abort or do nothing when encountered

nobodyatroot:
Additional null/empty checks to further cleanup message pollution in failed queue. There are still a few more that need further investigation before introducing a fix, but this PR cleans up exceptions that go unhandled, e.g, trying to access a null field in a array, passing null values as function parameters without being a nullable type, etc.

Live testing on [kbin.run](kbin.run) for awhile, nothing out of the ordinary observed so far...

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -585,15 +585,9 @@ class ActivityPubManager
         $this->logger->info('updating actor at {url}', ['url' => $actorUrl]);
         $actor = $this->apHttpClient->getActorObject($actorUrl);
 
-        try {
-            // User (We don't make a distinction between bots with type Service as Lemmy does)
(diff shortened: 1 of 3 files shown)


src/Service/ActivityPubManager.php
```diff
@@ -108,8 +108,12 @@ public function createCcFromBody(string $body): array
      *
      * @return User|Magazine|null or Magazine or null on error
      */
-    public function findActorOrCreate(string $actorUrlOrHandle): null|User|Magazine
+    public function findActorOrCreate(?string $actorUrlOrHandle): null|User|Magazine
```
e-five256, line 111: slightly curious when this and `handle` were being called with null args, maybe that is more wrong than this accepting null?

e-five256 (approved):

PR #403, opened 2024-01-06
Make comment timestamp links in most places

e-five256:
In some places, you could click on the `timeago` timestamp, e.g. `18 minutes ago` and it would be an anchor link for the current comment you're on.

I use this quite often, however it's very inconsistent right now. Take [this post chain for example](https://fedia.io/m/random/p/286157/Have-you-ever-had-so-many-meetings-in-a-day). Note that top level comment replies _aren't_ links by clicking on the age of the comment, **but** replies ([like this](https://fedia.io/m/random/p/286157/Have-you-ever-had-so-many-meetings-in-a-day#post-comment-483758)) _are_. That seems very confusing, it seems like all comments on the page should have their comment age timestamp be links.

The goal of this PR is to make all comments `timeago` clickable where it seems to make sense:
- comments in a single thread view
- comments in a single microblog view
- list of all microblog posts
- search
- comments on user overview

Not clickable on:
- moderate comment pages
- edit comment pages

these pages are really only visible if you have javascript disabled or javascript fails to load (or potentially hasn't loaded yet) (or you simply navigate there manually. If you middle mouse click on moderate or edit it will open those pages). the moderate page was like that already so I kept it consistent on the edit page.

there's still an issue of the link itself not working when not on the first page of comments, but that seemed unrelated to making the experience more consistent (having links at all)

this did require me to `flushdb` on redis, the nested comment component appears to use `_cached` and no amount of browser refreshes / php cache clears made it change for me until I flushed redis


nobodyatroot (approved):

templates/components/entry_comments_nested.html.twig
```diff
@@ -3,19 +3,21 @@
         {{ component('entry_comment', {
             comment: reply,
             showNested: false,
+            dateAsUrl: true,
```
BentiGorlich, line 6: if the default is `true` why overwrite it?

e-five256:
I've tried to spend a bunch of time thinking about where they should or shouldn't be links (and even almost just made them all links and got rid of the logic completely, as I wonder if it not being links in places really gains anything).

Pages like user overview and search, as I mentioned in the description, still have that issue where they may not work if the comment is on a page 2, but from my perspective I would rather have the option to jump to them and have it work some of the time then just not be a link at all, but people might feel differently if we start to get a lot more "links don't work" issues

templates/components/entry_comments_nested.html.twig
e-five256: removed all the instances where it would match the default

nobodyatroot (approved):
Makes sense to me

PR #413, opened 2024-01-09
Fix notifications of blocked users

BentiGorlich:
User A got notifications from a blocked user B when B:
- mentioned A in a post or entry
- replied to a post/-comment or entry/-comment of A

This should of course not happen, so don't notify A in these cases if B is on the block list.


melroy89 (approved):
Agreed

PR #418, opened 2024-01-10
include remote users in followers collection

BentiGorlich:
magazines returned the wrong amount of followers via ap because remote subscribers were not included

### DIFF
--- a/src/Controller/ActivityPub/Magazine/MagazineFollowersController.php
+++ b/src/Controller/ActivityPub/Magazine/MagazineFollowersController.php
@@ -31,30 +31,15 @@ class MagazineFollowersController extends AbstractController
         }
 
         if (!$request->get('page')) {
-            $data = $this->getCollectionInfo($magazine);
+            $data = $this->collectionInfoWrapper->build('ap_magazine_followers', ['name' => $magazine->name], $magazine->subscriptionsCount);
         } else {
             $data = $this->getCollectionItems($magazine, (int) $request->get('page'));
         }
 
         $response = new JsonResponse($data);
-
         $response->headers->set('Content-Type', 'application/activity+json');
 
         return $response;
-    }
-
-    #[ArrayShape([
-        '@context' => 'string',
-        'type' => 'string',
-        'id' => 'string',
-        'first' => 'string',
-        'totalItems' => 'int',
-    ])]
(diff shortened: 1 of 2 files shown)


asdfzdfj (dismissed):
this also shows remote subscribed user when querying for detailed list, while I can't find a group to query follower lists for cross checking, it looks like querying one for user would return similar results, so I think it should be fine

BentiGorlich:
Since this is a privacy concern I think we should get rid of listing the users, hut that is another issue

e-five256:
I am slightly confused as to why https://github.com/MbinOrg/mbin/blob/7de7c4516762ee08687751915010fe189648b88a/src/Controller/ActivityPub/Magazine/MagazineFollowersController.php#L50 does this rather than using the subscription_count column value of the magazine, which seems like it would be a much less expensive call. But maybe there's a difference in the count that I'm not realizing from just reading it off-hand

BentiGorlich:
I think there is no difference. I tried it with 
```sql
SELECT name, subscriptions_count, (SELECT COUNT(*) FROM magazine_subscription WHERE magazine_id=magazine.id) as count2 FROM magazine WHERE ap_id IS NULL LIMIT 10;
```

And the 2 counts were the same. I'll change it to the less expensive call and remove the output of the followers, as it is not used by anything and just a data leak

e-five256:
> remove the output of the followers

Is this still in progress or part of other work? Just wanted to check because it looks like this PR still has the exposed subscriber list. That's interesting, I didn't realize that about kbin/Mbin. Comparing it to lemmy it looks like they don't expose that information e.g. here's `lemmy.world/c/world`

```json
{
  "id": "https://lemmy.world/c/world/followers",
  "type": "Collection",
  "totalItems": 34492,
  "items": []
}
```

BentiGorlich:
Yeah it still includes the subscriber list. I didn't want to put that in this PR. I am aware of the problem. I just fixed that one can pull the local subscribers to a remote magazine, cause that just didn't make sense.
Yeah yeah the subscriber list is paginated. It's AP feature.

e-five256 (approved):
this looks good to me. I think there's broader privacy questions we've come up with from this that might need further looking into. definitely the less information on users we can return that isn't being used for AP actions, the better

PR #420, opened 2024-01-10
announce likes

BentiGorlich:
if a local magazine receives a like from a remote user it will announce it to all the followers of the magazine

I started a thread about it if anyone wants more context: https://gehirneimer.de/m/Mdev@kbin.run/t/151312/Your-thoughts-on-federating-likes-upvotes


nobodyatroot (dismissed):
Looks good to me, nothing obvious in the code stands out either. Testing live for the past day or so, no issues.

melroy89 (approved):

PR #425, opened 2024-01-11
add mention to Mdev@kbin.run magazine in README

asdfzdfj:
add a mention to [Mdev@kbin.run](https://kbin.run/m/Mdev) magazine in README for discussions about mbin within fediverse

although unofficial, it's the kind of place that means nothing if no people knows about it, especially when we don't have the luxury of flagship instance like [kbinMeta@kbin.social](https://kbin.social/m/kbinMeta) have, and having to spread it essentially by word of mouth hinders its usefulness

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,7 +50,7 @@ For developers:
 
 - [Official repository on GitHub](https://github.com/MbinOrg/mbin)
 - [Matrix Space for discussions](https://matrix.to/#/#mbin:melroy.org)
-- [Unofficial magazine for discussions within fediverse](https://kbin.run/m/Mdev)
+- [Unofficial magazine for discussions within the fediverse](https://kbin.run/m/Mdev)
 - [Translations](https://hosted.weblate.org/engage/mbin/)
 - [Contribution guidelines](CONTRIBUTING.md) - please read first, including before opening an issue!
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -50,6 +50,7 @@ For developers:
 
 - [Official repository on GitHub](https://github.com/MbinOrg/mbin)
 - [Matrix Space for discussions](https://matrix.to/#/#mbin:melroy.org)
+- [Unofficial magazine for discussions within fediverse](https://kbin.run/m/Mdev)
```
BentiGorlich, line 53: `within the fediverse` would be better I think

BentiGorlich (changes requested):
I don't know if the "For developers" section is the best place to put it. Though I did not find a better suited place for it either...

README.md
asdfzdfj: update the wording as suggested. I'm not that good at English

asdfzdfj:
> I don't know if the "For developers" section is the best place to put it. Though I did not find a better suited place for it either...

if we're treating it as mbin equivalent of kbinMeta (which seems to be the case) then it would be rather counterintuitive if it's mainly focused for developers rather than the users. and for that I think it can stay in the contributing section.

BentiGorlich:
Right now it is in the "For developers" section. What I meant is basically I don't know if it is at the right place at the moment

asdfzdfj:
wait I'm quite sure I put this under the "Contributing" section?

BentiGorlich (approved):
My bad. Was weird appearance bug from GitHub, sorry

PR #426, opened 2024-01-11
Introducing /api/info - general info API

melroy89:
Rationale: Improve Mobile client integration with Mbin, same PR should actually be merged into /kbin, allowing mobile developers to use the same end-point.

- Introduce a new info API (`/api/info`), inspired by NodeInfo and reuse our internal `ProjectInfoService`.
- Introduce a dedicated Schema file
- Remove 401 token error response messages for API's that are not behind oauth authentication (thus never give this error).


nobodyatroot (approved):

melroy89:
Example: https://kbin.melroy.org/api/info

PR #429, opened 2024-01-14
Add support for Zitadel authentication

thepaperpilot:
This PR adds Zitadel, a self hostable identity provider, support for OAuth logins. It works similarly to the keycloak one, but with zitadel URIs instead. It has a few more supporting files due to there not being an existing package for zitadel in the oauth bundle mbin uses.


src/Security/ZitadelAuthenticator.php
```diff
@@ -0,0 +1,125 @@
...
+                $dto = (new UserDto())->create(
+                    $username,
+                    $zitadelUser->getEmail()
+                );
+
+                $dto->plainPassword = bin2hex(random_bytes(20));
+                $dto->ip = $this->ipResolver->resolve();
+
+                $user = $this->userManager->create($dto, false);
+                $user->oauthKeycloakId = $zitadelUser->getId();
```
e-five256, line 98: should this be:

```suggestion
                $user->oauthZitadelId = $zitadelUser->getId();
```

src/Security/ZitadelAuthenticator.php
```diff
@@ -0,0 +1,125 @@
...
+use App\DTO\UserDto;
+use App\Entity\User;
+use App\Repository\UserRepository;
+use App\Service\IpResolver;
+use App\Service\UserManager;
+use App\Utils\Slugger;
+use Doctrine\ORM\EntityManagerInterface;
+use KnpU\OAuth2ClientBundle\Client\ClientRegistry;
+use KnpU\OAuth2ClientBundle\Security\Authenticator\OAuth2Authenticator;
+use thepaperpilot\OAuth2\Client\Provider\ZitadelResourceOwner;
```
e-five256, line 16: I sort of would've expected a composer vendor change for this, but I can't say for sure I know how either keycloak or zitadel work. Keycloak seems to pull in https://github.com/stevenmaguire/oauth2-keycloak/blob/master/src/Provider/KeycloakResourceOwner.php and adds it as a composer dependency

thepaperpilot: Pushing a commit for PR feedback soon, but just wanted to point out that was actually the wrong path, from when I was still figuring stuff out. It'll be updated to App\Provider\ZitadelResourceOwner, as the file is within the repository itself.

asdfzdfj (commented):
It'd be appreciated if you could provide some description for the PR

src/Security/ZitadelAuthenticator.php
```diff
@@ -0,0 +1,125 @@
...
+
+                if ($user) {
+                    $user->oauthZitadelId = $zitadelUser->getId();
+
+                    $this->entityManager->flush();
+
+                    return $user;
+                }
+
+                $username = $slugger->slug($zitadelUser->toArray()['username']);
```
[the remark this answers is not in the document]
thepaperpilot: On my instance the `username` definitely exists, but `preferred_username` included the domain which isn't what I wanted (e.g. I'd get `thepaperpilot_at_domain.tld`)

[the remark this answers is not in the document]
thepaperpilot: On [this page](https://zitadel.com/docs/apis/openidoauth/claims) in the docs it lists what fields are provided under which requests, and it has a row indicating it's provided when requested in the introspect request if the scope is specifically requested (the third column), which it is as part of the `profile` scope.
image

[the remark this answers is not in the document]
asdfzdfj: hmm ok, it might as well be zitadel misconfiguration on my part. for the `username` containing the `-at-` part is probably because zitadel likes usernames in that format, but it looks like when creating a user via admin, you can just put in user name without `@domain.tld` part and it'll accept it just fine, and that's what I did when creating a test user for testing this patch, guess I'll just let you know that this can happen for possible investigation 

for the record, this is what I get from dumping the `$zitadelUser->toArray()` just before getting username

```php
$username = $slugger->slug(dump($zitadelUser->toArray())['username'])
```

```php
 array:9 [▼
  "sub" => "249646190008795139"
  "name" => "ZITADEL Admin"
  "given_name" => "ZITADEL"
  "family_name" => "Admin"
  "locale" => "en"
  "updated_at" => 1705296726
  "preferred_username" => "tama@zitadel.localhost"
  "email" => "tama@zitadel.localhost"
  "email_verified" => true
]
```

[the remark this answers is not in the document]
thepaperpilot: I found out you were right and my code was actually preventing new users from authenticating

thepaperpilot:
Before this gets merged, I'd like to update the authenticator to update the user's username and avatar etc.

e-five256:
there's a button to convert PRs to drafts on the sidebar, we've been using that to mark/unmark things as still in progress / ready to merge, definitely recommend it to avoid us merging too early accidentally

image

thepaperpilot:
Sorry for the delay - it now copies over the avatar though.

.env.example
```diff
@@ -70,6 +70,9 @@ OAUTH_KEYCLOAK_SECRET=
 OAUTH_KEYCLOAK_URI=
 OAUTH_KEYCLOAK_REALM=
 OAUTH_KEYCLOAK_VERSION=
+OAUTH_ZITADEL_ID=
```
melroy89, line 73: does Mbin keeps working without adding this to the config?

thepaperpilot: It should, but since that file is supposed to show the possible keys I think its fine

BentiGorlich:
@thepaperpilot what is the current status? Are you finished, and only waiting for a review?

thepaperpilot:
> @thepaperpilot what is the current status? Are you finished, and only waiting for a review?

That is correct

BentiGorlich:
Ok, Then I will put that on my todo list :+1:

src/Security/ZitadelAuthenticator.php
```diff
@@ -0,0 +1,163 @@
...
+        if ($tempFile) {
+            $image = $this->imageRepository->findOrCreateFromPath($tempFile);
+            if ($image) {
+                $this->entityManager->persist($image);
+                $this->entityManager->flush();
+            }
+        }
+
+        return $image ?? null;
+    }
```
BentiGorlich, line 148: Could you refactor this, so that it is not duplicated in 3 authenticators (Zitadel, Google and Facebook)

templates/components/login_socials.html.twig
```diff
@@ -16,4 +16,8 @@
         <a href="{{ path('oauth_keycloak_connect') }}" class="btn btn__secondary"><i class="fa-solid fa-lock" aria-hidden="true"></i>
             Keycloak</a>
     {% endif %}
+    {% if this.zitadelEnabled %}
+        <a href="{{ path('oauth_zitadel_connect') }}" class="btn btn__secondary"><i class="fa-solid fa-lock" aria-hidden="true"></i>
+            Zitadel</a>
```
BentiGorlich, line 21: Could we get the icon of the zitadel org here? If it is too much effort that is fine. Maybe the Zitadel Logo itself (if they have a monochrome svg) would suffice?

BentiGorlich (commented):
If both remarks are too much work this would be good to go. 
It works and has the same standard as the other authenticators. Good job @thepaperpilot

templates/components/login_socials.html.twig
thepaperpilot: I don't think that's plausible. I checked my own instance and the url is `https://auth.incremental.social/ui/login/resources/dynamic?orgId=231964622825979915&default-policy=true&filename=policy/label/logo-dark-251434450057953312`. I have no idea how I'd get that org id though. That said, I'd urge against using the zitadel logo itself since typically instances are going to be using custom branding. In fact, I'm thinking perhaps the best thing to do would be adding 2 more env vars for the display name and logo to use for zitadel auth

.env.example
melroy89: so what I mean, is most current admin / server owners will forget to add this configuration. So Mbin should still work fine, even if `OAUTH_ZITADEL_ID` is undefined (not present in the `.env` file).

templates/components/login_socials.html.twig
melroy89: Of course it's possible. But currently we seem to use font-awesome icons (`fa`).  So now it's showing the `fa-lock` icon. 

However, font awesome doesn't have an icon / logo for Zitadel it seems. The icon seems to be come kind of triangle. so maybe `fa-caret-left` would do it for now?

image

BentiGorlich: > That said, I'd urge against using the zitadel logo itself since typically instances are going to be using custom branding. In fact, I'm thinking perhaps the best thing to do would be adding 2 more env vars for the display name and logo to use for zitadel auth

I am fine with either just keeping the lock icon or the 2 env vars with logo and name. :shrug:

melroy89: But from the user perspective it makes sense that you know you are going to Zitadel. Right?

I don't care if Zitadel then support custom logins /logos or alike. But I digress

src/Security/ZitadelAuthenticator.php
e-five256: I don't mind taking a shot at that when I go in to add the disable sso registration stuff as a follow up

templates/components/login_socials.html.twig
e-five256: I think for a first pass as is this is fine, thepaperpilot is the main user of it right now and likely already can write bespoke stuff if they haven't already. can always make it more customizable in a followup

e-five256 (approved):
seemed good to me, had a bit of trouble with zitadel itself in my own dev env, but obviously thepaperpilot is running this and I assume not seeing issues

templates/components/login_socials.html.twig
BentiGorlich: Sounds good to me

BentiGorlich (approved):

PR #433, opened 2024-01-14
Minor install doc updates

andrewmoise:
Minor installation updates, pretty self explanatory:

* s/domain.tdl/domain.tld/
* Better examples for remote SMTP servers
* ~~Install certbot from apt, and before nginx so nginx doesn't have failures trying to load the certificates~~ Store certbot via snap and explain why
* ~~Blank out postgres version since there's no sensible default~~
* ~~Load opcache extension in addition to configuring the options~~
* Don't have conflicting ssl_prefer_server_ciphers options (I did "off", for compatibility over security)

Added later by Melroy:

* Minor language changes or additional comment lines.

### DIFF
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -185,6 +185,7 @@ MAILER_DSN=sendmail://default
 MAILER_DSN=gmail+smtp://user%40domain.com:pass@smtp.gmail.com
 # Or remote SMTP:
 MAILER_DSN=smtp://username:password@smtpserver.tld:587?encryption=tls&auth_mode=log
+# Or remote SMTP with SSL on port 465
 MAILER_DSN=smtp://username:password@smtpserver.tld:465?encryption=ssl&auth_mode=log
 ```
 
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -185,7 +185,7 @@ MAILER_DSN=sendmail://default
 MAILER_DSN=gmail+smtp://user%40domain.com:pass@smtp.gmail.com
 # Or remote SMTP with TLS on port 587:
 MAILER_DSN=smtp://username:password@smtpserver.tld:587?encryption=tls&auth_mode=log
-# Or remote SMTP with SSL on port 465
+# Or remote SMTP with SSL on port 465:
 MAILER_DSN=smtp://username:password@smtpserver.tld:465?encryption=ssl&auth_mode=log
 ```
 
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -183,7 +183,7 @@ MAILER_DSN=smtp://127.0.0.1 # When you have a local SMTP server listening
 MAILER_DSN=sendmail://default
 # Or Gmail (%40 = @-sign) use:
 MAILER_DSN=gmail+smtp://user%40domain.com:pass@smtp.gmail.com
-# Or remote SMTP:
+# Or remote SMTP with TLS on port 587:
 MAILER_DSN=smtp://username:password@smtpserver.tld:587?encryption=tls&auth_mode=log
 # Or remote SMTP with SSL on port 465
 MAILER_DSN=smtp://username:password@smtpserver.tld:465?encryption=ssl&auth_mode=log
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -427,7 +427,7 @@ Make sure you have substituted all the passwords and configured the basic servic
 #### Let's Encrypt (TLS)
 
 > **Note**
-> The certbot authors recommend installing through snap as some distros' versions from apt tend to fall out of date; see https://eff-certbot.readthedocs.io/en/latest/install.html for more.
+> The Certbot authors recommend installing through snap as some distros' versions from APT tend to fall out-of-date; see https://eff-certbot.readthedocs.io/en/latest/install.html#snap-recommended for more.
 
 Install Snapd:
 
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -452,6 +452,9 @@ Follow the prompts to create TLS certificates for your domain(s). If you don't a
 
 ```bash
 sudo certbot certonly
+
+# Or if you wish not to use the standalone mode but the Nginx plugin:
+sudo certbot --nginx -d domain.tld
 ```
 
 ### NGINX
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -178,11 +178,14 @@ KBIN_STORAGE_URL=https://domain.tld/media
 POSTGRES_VERSION=14
 
 # Configure email, eg. using SMTP
-MAILER_DSN=smtp://127.0.0.1:25?encryption=ssl&auth_mode=login&username=&password=
+MAILER_DSN=smtp://127.0.0.1 # When you have a local SMTP server listening
 # But if already have Postfix configured, just use sendmail:
 MAILER_DSN=sendmail://default
 # Or Gmail (%40 = @-sign) use:
-MAILER_DSN=gmail+smtp://user%40domain.com:pass@default
+MAILER_DSN=gmail+smtp://user%40domain.com:pass@smtp.gmail.com
+# Or remote SMTP:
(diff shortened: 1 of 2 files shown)


docs/admin_guide.md
```diff
@@ -237,6 +240,8 @@ realpath_cache_ttl = 600
 Optionally also enable OPCache for improved performances with PHP:
 
 ```ini
+zend_extension=opcache
```
asdfzdfj, line 243: I may be missing something but isn't this already added/loaded when installing `php8.2-opcache` package? (from quick check in php sury that the guide recommends this should've been done already), or was there some situation where this isn't the case?

docs/admin_guide.md
```diff
@@ -421,6 +426,15 @@ npm run build # Builds frontend
...
 Make sure you have substituted all the passwords and configured the basic services.
 
+#### Let's Encrypt (TLS)
+
+Follow the prompts to create TLS certificates; if you didn't already have nginx up, you can use
+standalone mode.
+
+```bash
+sudo certbot certonly
+```
```
asdfzdfj, line 436: missing instructions? like maybe you probably need to specify `-d mbin.domain.tld` to specify the domain(s) to request certs for or something

(I can't help you fill in this as I haven't used certbot, If anyone has more info and could help fill this in that'd be appreciated)

docs/admin_guide.md
andrewmoise: On my system it wasn't enabled yet; I'm pretty sure I followed the instructions as written and when I got to this stage that line was still commented out. I can try a quick fresh install and investigate more though (e.g. maybe there's an uncommented version of the line in some other .ini file that's installed by php8.2-opcache or something) (or maybe it's a Debian vs Ubuntu difference).

docs/admin_guide.md
andrewmoise: 'certbot certonly' gives prompts asking for anything that's not specified; my feeling was that it made more sense to let the tool just prompt people for info it needed, as opposed to giving them a command line they need to figure out the meaning of and then modify. Unless there are cases where this needs to be able to run unattended based on these instructions or etc?

docs/admin_guide.md
andrewmoise: Yep - it is already enabled. Not in /etc/php/8.2/fpm/php.ini (where it is commented out) but in /etc/php/8.2/mods-available/opcache.ini. So yes, my uncommenting of that line is not needed as it was already on by installing the package, you're right.

docs/admin_guide.md
asdfzdfj: ah ok, looks like I missed the "Follow the prompts" up there, all right.
also it doesn't look like there's need/support for installing bare metal unattended at the moment.

---

my experience in this regard was from `acme.sh` corner where I specify all needed arguments up front, which can look something like this:

```
% acme.sh \
  --install-cert \
  -d domain.tld -d '*.domain.tld' \
  --dns dns_cf \
  --fullchain-file /usr/local/etc/nginx/ssl/wildcard_domain.tld.crt \
  --key-file /usr/local/etc/nginx/ssl/wildcard_domain.tld.key \
  --reloadcmd 'service nginx forcereload'
```

so when I saw/compared the new command with the old command(s) I got kinda confused and thought if I have to go look up the `certbot` arguments and have to fill it myself or something.

docs/admin_guide.md
melroy89: Please remove this line again in that case. if you use Nginx `opcache.enable=1` in PHP-FPM php.ini file should be sufficient .

docs/admin_guide.md
```diff
@@ -421,6 +426,15 @@ npm run build # Builds frontend
 
 Make sure you have substituted all the passwords and configured the basic services.
 
+#### Let's Encrypt (TLS)
+
+Follow the prompts to create TLS certificates; if you didn't already have nginx up, you can use
```
melroy89, line 431: it's called NGINX (with caps).

docs/admin_guide.md
```diff
@@ -659,39 +672,6 @@ Restart (or reload) NGINX:
 sudo systemctl restart nginx
 ```
 
-#### Let's Encrypt (TLS)
```
melroy89, line 662: Restore this section. Again Certbot shouldn't be installed from APT. The package might be too old.

melroy89 (changes requested):
See my remarks.

docs/admin_guide.md
andrewmoise: Done 👍🏻

andrewmoise:
How's this? I reinstated the section instructing to use snap, added a more accurate explanation of why snap is preferred, but still had the whole thing moved up above the NGINX section (the existing instructions say to start NGINX at a certain point, but the startup fails because we've configured it for TLS but haven't done the certbot setup yet).

melroy89 (dismissed):

melroy89:
OK I did some minor changes still. And I restored `sudo certbot --nginx -d domain.tld` for the people who have Nginx already running and want to use this flag.

docs/admin_guide.md
```diff
@@ -443,7 +479,6 @@ Edit the main NGINX config file: `sudo nano /etc/nginx/nginx.conf` with the foll
 
 ```nginx
 ssl_protocols TLSv1.2 TLSv1.3; # Requires nginx >= 1.13.0 else only use TLSv1.2
-ssl_prefer_server_ciphers on;
```
melroy89, line 446: Yea you're right you most likely want to have this set to `off` (default is off).

docs/admin_guide.md
```diff
@@ -178,11 +178,14 @@ KBIN_STORAGE_URL=https://domain.tld/media
...
 # Configure email, eg. using SMTP
-MAILER_DSN=smtp://127.0.0.1:25?encryption=ssl&auth_mode=login&username=&password=
+MAILER_DSN=smtp://127.0.0.1 # When you have a local SMTP server listening
 # But if already have Postfix configured, just use sendmail:
 MAILER_DSN=sendmail://default
 # Or Gmail (%40 = @-sign) use:
-MAILER_DSN=gmail+smtp://user%40domain.com:pass@default
+MAILER_DSN=gmail+smtp://user%40domain.com:pass@smtp.gmail.com
+# Or remote SMTP:
+MAILER_DSN=smtp://username:password@smtpserver.tld:587?encryption=tls&auth_mode=log
```
melroy89, line 187: Do not uncomment all the lines. Just, since this will fail as well. I used the first `MAILER_DSN` as default. If the user want to user another setting they need to comment that line and uncomment another line.

melroy89 (changes requested):
Just pick 1 default `MAILER_DSN` line. You now uncommented all options.

docs/admin_guide.md
```diff
@@ -178,11 +178,14 @@ KBIN_STORAGE_URL=https://domain.tld/media
...
+MAILER_DSN=smtp://username:password@smtpserver.tld:465?encryption=ssl&auth_mode=log
```
melroy89, line 188:
```suggestion
# Or remote SMTP with SSL on port 465
MAILER_DSN=smtp://username:password@smtpserver.tld:465?encryption=ssl&auth_mode=log
```

docs/admin_guide.md
```diff
@@ -178,11 +178,15 @@ KBIN_STORAGE_URL=https://domain.tld/media
...
 
 # Configure email, eg. using SMTP
-MAILER_DSN=smtp://127.0.0.1:25?encryption=ssl&auth_mode=login&username=&password=
+MAILER_DSN=smtp://127.0.0.1 # When you have a local SMTP server listening
 # But if already have Postfix configured, just use sendmail:
 MAILER_DSN=sendmail://default
 # Or Gmail (%40 = @-sign) use:
-MAILER_DSN=gmail+smtp://user%40domain.com:pass@default
+MAILER_DSN=gmail+smtp://user%40domain.com:pass@smtp.gmail.com
+# Or remote SMTP:
```
melroy89, line 186:
```suggestion
# Or remote SMTP with TLS on port 587:
```

melroy89 (approved):

melroy89 (approved):

PR #435, opened 2024-01-15
Add setting to restrict sidebar sections to local only

thepaperpilot:
When enabled by an admin, this will make the random magazines, active people, random posts, and random threads only show the local items of those types. This does not affect the sections when looking at a specific community.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -30,6 +30,7 @@ KBIN_META_DESCRIPTION="content aggregator, content voting, discussion and micro-
 KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
+KBIN_SIDEBAR_SECTIONS_LOCAL_ONLY=true
 MBIN_DEFAULT_THEME=default
 
 # Captcha (also enable in admin panel/settings)
--- a/src/Twig/Extension/SettingsExtension.php
+++ b/src/Twig/Extension/SettingsExtension.php
@@ -25,6 +25,7 @@ final class SettingsExtension extends AbstractExtension
             new TwigFunction('kbin_captcha_enabled', [SettingsExtensionRuntime::class, 'kbinCaptchaEnabled']),
             new TwigFunction('kbin_mercure_enabled', [SettingsExtensionRuntime::class, 'kbinMercureEnabled']),
             new TwigFunction('kbin_federation_page_enabled', [SettingsExtensionRuntime::class, 'kbinFederationPageEnabled']),
+            new TwigFunction('kbin_sidebar_sections_local_only', [SettingsExtensionRuntime::class, 'kbinSidebarSectionsLocalOnly']),
             new TwigFunction('mbin_current_version', [SettingsExtensionRuntime::class, 'mbinCurrentVersion']),
         ];
     }
--- a/.env.example_docker
+++ b/.env.example_docker
@@ -30,6 +30,7 @@ KBIN_META_DESCRIPTION="content aggregator, content voting, discussion and micro-
 KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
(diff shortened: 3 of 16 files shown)


e-five256:
This is very minor, but it seems like we have been going with naming things `mbin` instead of `kbin` when adding new things, but haven't quite tackled the older env vars as things like the `.env` copy usually happen only once for instances and we don't want to break already running instance owners, but it'd be nice if all the places that `kbin` was added could be replaced with `mbin`

thepaperpilot:
> This is very minor, but it seems like we have been going with naming things `mbin` instead of `kbin` when adding new things, but haven't quite tackled the older env vars as things like the `.env` copy usually happen only once for instances and we don't want to break already running instance owners, but it'd be nice if all the places that `kbin` was added could be replaced with `mbin`

To clarify, you're asking for new env vars to start with MBIN_ but leave the KBIN_ ones in tact, so we have a mix of both?

src/Twig/Extension/SettingsExtension.php
```diff
@@ -25,6 +25,7 @@ public function getFunctions(): array
             new TwigFunction('kbin_captcha_enabled', [SettingsExtensionRuntime::class, 'kbinCaptchaEnabled']),
             new TwigFunction('kbin_mercure_enabled', [SettingsExtensionRuntime::class, 'kbinMercureEnabled']),
             new TwigFunction('kbin_federation_page_enabled', [SettingsExtensionRuntime::class, 'kbinFederationPageEnabled']),
+            new TwigFunction('kbin_sidebar_sections_local_only', [SettingsExtensionRuntime::class, 'kbinSidebarSectionsLocalOnly']),
```
e-five256, line 28: very minor, but it doesn't look like this gets used in any twig template (this comment applies to [this function as well](kbinSidebarSectionsLocalOnly)). Now that I look, I don't think `kbinFederatedSearchOnlyLoggedIn()` gets called anywhere either, but that's not something related to your PR. I imagine the overhead on an unused twig function isn't high, so this isn't a big issue if true, but just wanted to mention it. These extensions expose the function to be usable in twig templates but I don't think we have a use case for it as the setting seems contain inside the backend at the moment

e-five256:
> To clarify, you're asking for new env vars to start with MBIN_ but leave the KBIN_ ones in tact, so we have a mix of both?

Yea, that's what I ended up doing when adding the default_theme one most recently https://github.com/MbinOrg/mbin/blob/2bf67ceed54c889672c4e932b9fbfd6769a78de9/.env.example#L32C1-L33

thepaperpilot:
Just a head's up, I still plan on addressing the feedback on this PR but before that I'd like to change the behavior. I'd rather the random post and threads were from local magazines, regardless of whether the user was local.

.env.example
```diff
@@ -30,6 +30,7 @@ KBIN_META_DESCRIPTION="content aggregator, content voting, discussion and micro-
 KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
+MBIN_SIDEBAR_SECTIONS_LOCAL_ONLY=true
```
e-five256, line 33:
```suggestion
MBIN_SIDEBAR_SECTIONS_LOCAL_ONLY=false
```

matching the docker example config, I think we should have this default to all as it seemed like there was more opinions of preferring it all rather than local at time of #141 though if there's strong opinions we could always recheck

e-five256:
aside from my one final comment this appears fine to me, but I'm no expert, @melroy89 @nobodyatroot not sure if you two or anyone else has opinions, I tested this locally and not having the env set doesn't cause issues (I was kind of expecting it to need `default:` in the `services.yaml` but... it seems fine?), can set the admin setting and it adds the db setting, removing it sets the value to false in db, and setting the env var as true has the admin setting checked, so all looked right to me, I just can't quite tell if the filters are correct because all my content is local so... it is just as local as it always was

e-five256 (approved):

melroy89:
This is now causing performance issues actually: https://github.com/MbinOrg/mbin/issues/1868 (especially on the random threads)

Issue #362, opened 2023-12-19
Enhance information in magazine overview


BentiGorlich:
**Is your feature request related to a problem? Please describe.**
One cannot know whether a remote magazine gets updates, because there is no way to see whether a magazine has local subscribers. Additionally it is not obvious how old the displayed information (subscriber count, description, moderators, etc.) is.

**Describe the solution you'd like**
1. show an information that the info not up to date when it is older than a day. 
2. show an information that the magazine won't get any updates because no one on the instance is subscribed

hints:
1. There is a column in the database that contains the last fetched datetime
2. When the magazine_subscriptions table does not contain a local subscription for the magazine show the warning that it will not get updates

e-five256:
I mentioned in matrix but adding to here as well:

1. seems fairly straightforward as it's just hooking up the frontend to backend data we have
2. this however seems a bit more complicated; from my understanding we lost local subscriber counts when we set the column data to use the count from the source. There are a few ways we could go about this but likely I think we'd just want to recreate a `local_follower` column or something, adding and subtracting every time a local sub/unsub or follow/unfollow happens, with a one time migration of reading the subscription table and setting the column values to current count. Reasoning being that we wouldn't want to seq scan the subscription table every time we need this information

not sure if that means this could be broken up into multiple issues, or even if not they could probably be handled in separate PRs

e-five256:
Just wanted to double check on whether this is complete, or if there is still more work to do

I see the two points listed in the OP on magazines now

![image](https://github.com/MbinOrg/mbin/assets/146029455/63c24b96-408d-4457-8fb7-1ecad5bbbdf3)

But not sure if there's remaining work or it needs some polish passes

BentiGorlich:
You're right this was done via #436

PR #436, opened 2024-01-15
First cut of showing federation staleness

andrewmoise:
First cut of a fix for #362: Add warning panels analogous to the federated magazine panel.

In addition to checking whether the way I did the added internal functionality was right, it seems worth reviewing whether we should be making the "This magazine is from a federated server" panel more subtle, now that we have this? Having the bright yellow panels only show up if there's an actual problem seems better UX wise.


andrewmoise:
Discussion in Matrix is that this should be redone a bit (in addition to fixing the CI); I'll update and repush.

melroy89:
You need to run the linting fix command. Since there is still a linting error

```
PHP runtime: 8.2.12
Loaded config default from "/__w/mbin/mbin/.php-cs-fixer.dist.php".
   1) src/Twig/Runtime/MagazineExtensionRuntime.php (blank_line_before_statement)
```

andrewmoise:
Okay, I've adjusted it as discussed. We stop showing the existing "this is a federated magazine" warning, *unless* we're on the last page of results (i.e. there probably is more on the original server). And, we show a warning on non-subscribed magazines that they're not getting updates, including a display of how long it's been since they got any updates.

translations/messages.en.yaml
```diff
@@ -57,8 +57,9 @@ cover: Cover
 related_posts: Related posts
 random_posts: Random posts
 federated_magazine_info: This magazine is from a federated server and may be incomplete.
+disconnected_magazine_info: This magazine is not receiving updates (last attempt %days% day(s) ago). Subscribe to start receiving updates.
```
BentiGorlich, line 60: I'd like to reword this to
> This magazine is not receiving updates any more (updated %days% day(s) ago). Subscribe to start receiving updates.

Or something similar.

templates/magazine/_federated_info.html.twig
```diff
@@ -1,6 +1,24 @@
...
+        {# Then show a link to original if we're at the end of content #}
+        <div class="alert alert__info">
+            <p>{{ 'federated_magazine_info'|trans }} <a
+             href="{{ magazine.apProfileId }}">{{ 'go_to_original_instance'|trans }}</a></p>
+        </div>
+    {% endif %}
+
+    {% if not magazine_has_local_subscribers(magazine) %}
+        {# Also show a warning if we're not actively receiving updates #}
+        {% set fetchedAt = magazine.apFetchedAt|date('U') %}
```
BentiGorlich, line 12: The fetch date is not really the best thing to check for the last updated timestamp. The `apFetchedAt` is the last date the actor was updated, so the title, describtion, etc. The better thing to check would be the last created entry, entrycomment, post or postcomment. As this actually is a "last received update".
For example the `apFetchedAt` date will also get updated if a user searches for this magazine.

BentiGorlich (changes requested):

templates/magazine/_federated_info.html.twig
andrewmoise: Yeah, makes sense. That was why I worded it in terms of the last attempt instead of last update, but yeah, last real update is better. I'll change it accordingly.

Question though: If someone local creates a post or comment on a federated-but-disconnected magazine, won't that create content that would get counted as within the magazine? I know Lemmy works that way. If mbin does the same, maybe it's better to check for content that's associated with the magazine but has an .apId field (or whatever) set, so we know it was an actual received-from-the-origin update.

andrewmoise: Hm... how about this? This proposal is using magazine.lastActive. So, on my instance, I actually do get federated stuff coming into even magazines that are disconnected from their home instances (I'm using hexbear to test, since I'm not on their whitelist). So even iterating over entities associated with a magazine and testing their .apId values won't work.

.lastActive seems like it does get updated by successful communication with the magazine's home instance, and it seems like checking over my server it has a roughly-appropriate value for disconnected magazines. I don't understand the issues in full, but it looks like it's easy and simple to check, and equally as accurate as anything more complex option I can see. What do you think?

config/packages/framework.yaml
```diff
@@ -17,7 +17,7 @@ framework:
     http_client:
         default_options:
             headers:
-                'User-Agent': 'Mbin/1.2.1 (+https://%kbin_domain%/agent)'
+                'User-Agent': 'Mbin/1.3.0 (+https://%kbin_domain%/agent)'
```
melroyvandenberg, line 20: not sure why this file show changes, if you rebased with the latest main branch, I would expect to just show 1.3.0.

templates/magazine/_federated_info.html.twig
BentiGorlich: I have to look when this value is actually set

andrewmoise: Sounds good. Yeah, I switched to it purely based on looking at some common cases on my instance without really understanding much behind it.

config/packages/framework.yaml
melroyvandenberg: Is now resolved.

templates/magazine/_federated_info.html.twig
```diff
@@ -1,6 +1,24 @@
...
-    <div class="alert alert__info">
-        <p>{{ 'federated_magazine_info'|trans }} <a
-                    href="{{ magazine.apProfileId }}">{{ 'go_to_original_instance'|trans }}</a></p>
-    </div>
+{% if magazine.apId %}   {# I.e. if we're federated #}
+    {% if entries is defined and entries and not entries.hasNextPage %}
+        {# Then show a link to original if we're at the end of content #}
+        <div class="alert alert__info">
+            <p>{{ 'federated_magazine_info'|trans }} <a
+             href="{{ magazine.apProfileId }}">{{ 'go_to_original_instance'|trans }}</a></p>
```
BentiGorlich, line 6: please fix the formatting to
```html
<p>
    <a></a>
</p>
```

translations/messages.en.yaml
```diff
@@ -57,8 +57,10 @@ cover: Cover
 related_posts: Related posts
 random_posts: Random posts
 federated_magazine_info: This magazine is from a federated server and may be incomplete.
+disconnected_magazine_info: This magazine is not receiving updates (last activity %days% day(s) ago). Subscribe to start receiving updates.
+always_disconnected_magazine_info: This magazine has not yet received any updates. Subscribe to start receiving updates.
 federated_user_info: This profile is from a federated server and may be incomplete.
-go_to_original_instance: Browse more on the original instance.
+go_to_original_instance: For a complete list of posts, browse on the original instance.
```
BentiGorlich, line 63: maybe: `For a complete list of posts, visit the original instance.`? I am not sure though

BentiGorlich (changes requested):
Just the 2 little things I mentioned in the comments. 

About the `lastActive` thing: the field is getting updated whenever a new `Entry`, `EntryComment`, `Post` or `PostComment` is received. So this field is probably getting updated when you post something **to** a remote magazine. I suggest that you add a field `lastRemoteUpdate` to `Magazine` and set it the same way and on the same place as `lastActive` is set, but with the condition, that the magazine has an `apId`, the post or whatever does as well and that the domain of the `apDomain` of the post is the same as the magazine. That way we know we receive updates from the host.

This involves writing a migration, so a bit more fiddly for you, sorry :sweat_smile:

templates/magazine/_federated_info.html.twig
melroy89: Fixed for you.

translations/messages.en.yaml
melroy89:
```suggestion
go_to_original_instance: For a complete list of posts, visit the original instance.
```

melroy89 (dismissed):

templates/magazine/_federated_info.html.twig
BentiGorlich: I've done the groundwork for this in #503 
When that is merged, you can just change from `lastActive` -> `lastOriginUpdate` :)

melroy89:
After PR https://github.com/MbinOrg/mbin/pull/503 is merged, we will update this PR I suppose?

andrewmoise:
Yeah, that sounds good to me, I can take a look, it should be pretty much trivial if the new column is done for me, so I can do the update then if that sounds good.

templates/magazine/_federated_info.html.twig
andrewmoise: Done

andrewmoise:
Okay, I switched over to using the new field @BentiGorlich added. I think it's good to go, with two important caveats:

One, I removed the notification of "never has received an update" and changed always_disconnected_magazine_info to just say it's not receiving updates. Two reasons: (1) Right after the transition, it'll show magazines as "never having received an update" when they had, if it just hadn't happened recently enough for the code to have changed so it had set the column in the DB. (2) To me it seems needlessly confusing; like if I've never received updates, why do I see all these posts? It requires a little in depth knowledge for it to make sense why that happens and I feel like there's no reason to confuse people. Just "we're not receiving updates" is accurate, and the main thing they need to know (although if we *can* also give them a specific day when we were last subscribed and receiving all updates, then we still tell them the day obviously).

And two, I kind of like having the ability to easily get to the origin server's version of the magazine. Maybe we want to add a less-obtrusive thing in the magazine panel or something, which restores that functionality if someone wants it, without it being a big yellow alert box.

LMK what y'all think

translations/messages.en.yaml
```diff
@@ -57,8 +57,10 @@ cover: Cover
 related_posts: Related posts
 random_posts: Random posts
 federated_magazine_info: This magazine is from a federated server and may be incomplete.
+disconnected_magazine_info: This magazine is not receiving updates (last activity %days% day(s) ago). Subscribe to start receiving updates.
+always_disconnected_magazine_info: This magazine is not receiving updates. Subscribe to start receiving updates.
 federated_user_info: This profile is from a federated server and may be incomplete.
-go_to_original_instance: Browse more on the original instance.
+go_to_original_instance: For a complete list of posts, visit the original instance.
```
melroy89, line 63: it can be posts.. but also threads, right?

src/Twig/Runtime/MagazineExtensionRuntime.php
```diff
@@ -5,13 +5,17 @@
...
+use App\Repository\MagazineSubscriptionRepository;
 use Symfony\Bundle\SecurityBundle\Security;
 use Twig\Extension\RuntimeExtensionInterface;
 
 class MagazineExtensionRuntime implements RuntimeExtensionInterface
 {
-    public function __construct(private readonly Security $security)
+    public function __construct(
+        private readonly Security $security,
+        MagazineSubscriptionRepository $magazineSubscriptionRepository)
```
BentiGorlich, line 16: this should be `private readonly` as well, shouldn't it?
Then the `$this->magazineSubscriptionRepository = $magazineSubscriptionRepository;` line can be deleted as well

translations/messages.en.yaml
BentiGorlich: I think in this case "posts" is alright. In our UI it is `Threads` and `Microblogs` so posts can count as both

BentiGorlich (changes requested):
Last one, I promise. After that it can be merged :sweat_smile:

src/Twig/Runtime/MagazineExtensionRuntime.php
andrewmoise: Sure, done

andrewmoise:
Hey, I did the one fix recommended. One thing though that showed up in my testing is that I suspect that last_origin_update might be getting updated by some things that don't require us to be subscribed. The reason I say that is that I'm not subscribed to technology@lemmy.world and don't get every post from them, but my instance still showed it updated 0 day(s) ago. I didn't dig into it in detail, just a quick observation, but might be worth digging into before applying it.

translations/messages.en.yaml
melroy89: Sure ok.

andrewmoise:
Hm, how about this?:

We could only update last_origin_update on posts or entries, not on comments, and only when we have local subscribers to the magazine. That way, it'd get a continuous flow of updates for as long as we were in good communication with the origin server, and with local subscribers, but not otherwise.

It's a little not-ideal maybe because it wouldn't show anything if we aren't in communication with an origin server (and that's not an imaginary situation - IDK how mbin works but on Lemmy you can be "subscribed" to a community on a server that's defederated from you, and people on your server can interact with you, but it doesn't show any content from the origin. It'd be nice to show a warning for that situation). IDK, maybe keeping track of good communication per-remote-instance would be a good thing to track as an alternate way of triggering the same type of warning?

But in any case for this change, if changing last_origin_update semantics sounds ok I'm happy to make the change.

BentiGorlich:
Sorry didn't get a chance to answer, yet. I'd like to merge this after I've looked at it again and deal with the `last_origin_update` in another PR. 

I think the biggest problem is, that we cannot know whether the update is received because someone is subscribed to the magazine or to a user posting to the magazine if both, the user and the magazine have the same home server. That problem is there, regardless of how we deal with it. We could of course only update `last_origin_update` when we get an announce activity from the magazine actor, that would definitely guarantee that we have contact with the magazine. Maybe thats the way to go...

andrewmoise:
Ooh. Yeah, that makes perfect sense to me. I'm not super familiar with the federation details, but from what you say, updating last_origin_update when we get announces from the magazine actor sounds perfect.

Okay, just LMK if there's anything you want me to do with it.

templates/magazine/_federated_info.html.twig
```diff
@@ -1,6 +1,25 @@
...
+                {% set currentTime = "now"|date('U') %}
+                {% set secondsDifference = currentTime - (lastOriginUpdate|date('U')) %}
+                {% set daysDifference = (secondsDifference / 86400)|round(0, 'floor') %}
+                <p>{{ 'disconnected_magazine_info'|trans({'%days%': daysDifference}) }}</p>
+            {% else %}
+                <p>{{ 'always_disconnected_magazine_info'|trans }}</p>
+            {% endif %}
+        </div>
+    {% endif %}
+{% endif %}
```
BentiGorlich, line 25: > Subscribe to start receiving updates.

Doesn't really make sense if the user isn't logged in. Maybe split the text in two so the second sentence is only displayed if the user is logged in.

BentiGorlich (changes requested):
When that is done, we can finally merge this PR. Sorry for taking so long to review it

templates/magazine/_federated_info.html.twig
andrewmoise: Okay done

andrewmoise:
All good. I think the requested changes are in now -- did you want me to modify the last_origin_update semantics though? To me it still seems like when it gets updated still isn't the perfect setup last I looked.

BentiGorlich:
Nono, I'll do it, but in a different PR. I'd appreciate it if you would create an issue though 😁

andrewmoise:
Done 👍🏻

BentiGorlich:
The Info for 
>  Subscribe to start receiving updates. 

Is not getting added to the `always_disconnected_magazine_info`

andrewmoise:
?

What do you mean by this?

BentiGorlich:
I meant that if the magazine never received any updates the info that ypu can subscribe to get updates is not being displayed

andrewmoise:
What? Last week you asked me to split it in two and only display the second sentence if the user is logged in. It's dependent on the existence of app.user, nothing to do with subscription status; you can look at the diff.

BentiGorlich:
Bildschirmfoto vom 2024-03-18 13-50-17
This is the part I mean and for this message the "subscribe to receive updates" is not getting displayed...

Yes this message should only be displayed to logged in users, but that part is missing here

andrewmoise:
Got it, I see what you meant. Fixed.

BentiGorlich (approved):

nobodyatroot (approved):

PR #438, opened 2024-01-15
Delete user despite possible errors

melroy89:
When a user delete is triggered by an instance owner the `__invoke` method of `DeleteUserHandler.php` is called. Which in turn calls several private methods to clean-up the local database as well as remote API calls to other instances. If such a remote call fails for whatever reason (down? time-out?..) the admin gets a 500 internal error. This is unwanted behavior.

Instead, we always want to remove the user, despite possible errors due to some kind of remote server. Hence try to catch the error and print a warning instead to the monolog. And then continue with the next removal step in the queue. This will no longer throw a 500 error message, and will successfully delete the user, even in time-out error cases.

Before (in `dev` mode, in prod mode the admin just sees '500 internal error'): 

image


BentiGorlich (commented):
In my opinion this code needs cleanup in general, as we should just send out a delete message with the user as the object. We have to send it to all the inboxes the user ever send any post, entry, comment, like, etc. to. 
However we need to check if we handle incoming delete messages correctly and if lemmy handles it correctly. If yes, all the calls to explicitly delete anything on remote servers can safely be ignored...

melroy89:
> In my opinion this code needs cleanup in general, as we should just send out a delete message with the user as the object. We have to send it to all the inboxes the user ever send any post, entry, comment, like, etc. to. However we need to check if we handle incoming delete messages correctly and if lemmy handles it correctly. If yes, all the calls to explicitly delete anything on remote servers can safely be ignored...

Agreed, in the mean while do we want to have this merged, so there is no issue from the perspective of the instance owners? After all, removing a user currently might result into 500 errors.

BentiGorlich:
Remote servers will end up with orphan entries if we ignore the errors and delete the user anyways. Question is: is that a bad enough thing for us to postpone fixing the error...

melroy89:
Well.. I do agree with you, we should check whether just a single delete request should maybe sufficient. If so, we might need to change the time-out manually for this action (by using a parameter for the timeout, rather than a single value). Since I can imagine such a call might take a while to process on the remote erver, unless the activitypub software will directly return with 'OK' (and push the actual deletion to a background task).

That being said, even then a remote server can have internal issues/might be offline... In those case I think we still want to continue with the deletion, right? Maybe we need to create additional queues for retrying such actions. If we do not allow to delete a user in such a scenario, we might never be able to delete a local user unless all remote servers are up-and-running, no errors during the processing and not causing anytime-outs or other issues.

BentiGorlich:
Honestly I think we do not have to validate that the content is getting deleted on the remote server. We can't do anything about it if it doesn't delete it anyways...
The good thing about our way of handling stuff like message delivery is, that one deliver message will be created for each server, so if the remote server is offline the message for this specific server fails and can be retried.
We can just continue with out deletion without worrying about the remote server

melroy89:
OK, so now I dunno what we want to do next.. I think failed HTTP requests in this case will be retried a you mentioned and I catch the errors, so if such a request goes wrong it doesn't cause an 500 error (which is really bad atm). 

We can later improve to reduce maybe the amount of remote calls if the remote AP software is executing a delete user in a better way (eg. a single delete might be enough!? More research is needed on all AP sfotware). In our internal DB we might definitely need to perform multiple delete queries, since I don't think we fixed all the PostgreSQL foreign keys with cascading.

BentiGorlich:
Honestly I am thinking that your changes are fine, because I hate this delete implementation and needs to be changed anyways... It should just be: "delete user, send a delete user message to all followers and magazines he created something on"...

I want to have the cascading deletes so badly :sweat_smile:

melroy89:
> I want to have the cascading deletes so badly 😅

Same :D

melroy89:
@BentiGorlich Can we clean-up something in here now do you think? Feel free to add additional commits.

BentiGorlich:
I've already been working to completely change the way we delete users. I would actually like to just close this PR

BentiGorlich:
I'd like to close this in favor of #522 
I couldn't test it, yet and there are features missing, like "Delete Content", but thats why it is WIP

melroy89:
Sure no problem. I will close it.

PR #441, opened 2024-01-16
prune some likely unused ActivityPhp related code

asdfzdfj:
this patch removes most of the files/constructs that was copied from landrok/ActivityPhp, overall federation functionality shouldn't change.

there's a decent chunk of code that was copied from the ActivityPhp library into kbin/mbin, but they're barely used in the codebase.

the closest I've seen it in active use is in Webfinger factory and even then it was modified to use our own service and didn't actually call out to the library


melroy89:
It seems to work fine. I still would like to run it a bit longer.

melroy89 (approved):

PR #446, opened 2024-01-17
add default check to local sidebar only env var

e-five256:
after #435 

@nobodyatroot reported getting `Environment variable not found: \"MBIN_SIDEBAR_SECTIONS_LOCAL_ONLY\"`

I'm not able to repro this currently, with no `.env` changes at all on main, clearing all my caches, but understood in the PR that it seemed like what would happen. Sadly that means I can't even check if this fixes it or not, but followed the pattern of https://github.com/MbinOrg/mbin/blob/main/config/services.yaml#L70


e-five256:
closing in favor of #448

PR #448, opened 2024-01-17
set default of local sidebar sections

e-five256:
move out from env var similar to federated search

mentioned in #446 adding a new env var comes with the price of either breaking existing instance owners or defaulting, but the defaulting in symfony always appears to have a lot of issues, especially with non-string types like `int`s and `bool`s (e.g. `Invalid env fallback in \"default:false:MBIN_SIDEBAR_SECTIONS_LOCAL_ONLY\"`)

I think it might be best if we keep new settings to having sane defaults in php, and using the db as the source of truth for when set


nobodyatroot (approved):
Verified this works as expected on a bare metal dev env.

e-five256:
@thepaperpilot headsup if you were relying on setting the env var to true locally, you'll want to get it saved and written to the DB settings table before getting this change

PR #453, opened 2024-01-21
UI filter changes

andrewmoise:
Hey, so here's a very rough draft of some UI changes -- moving controls which filter or modify the front page's view of posts into a consistent place, in the bar alongside hot/new/top/etc.

The frontend should mostly work with the exception of some translation issues. Some of the backend changes are pretty invasive (and sometimes I've broken something which I haven't made work again in a new way) -- this is just for comments or feedback, not really in shape to be a proposed set of changes yet. I've probably broken some stuff I haven't discovered yet; I just poked around with it slightly. You can see the results live at https://mbin.grits.dev/ if you like.

In the long run I'd like to make it so *all* the filter options bars work this way consistently, and make there generally be a "save settings" icon over on the right side that you can click to make them persistent.

Feedback / commentary / criticism is welcome.


melroy89:
I like it quite a lot.

andrewmoise:
Hey, that's excellent! I am flattered. Okay, cool, @BentiGorlich said he could give it some testing on his server which I think would be a really good idea given the size of the change. But yeah if he says go, I can flip it out of the draft state for if and when it seems like a good idea to pull it. I might have a couple little fixes ("Newest" -> "New" for example) in addition to anything anyone else sees.

e-five256:
I'm trying to keep in mind accessibility, but I'm not well-versed enough to know how a screenreader would deal with these options. What I was going for when making the current selection of `All time` label visible, is that it's inside the `aria-label = Filter by time`, so I was hoping that would make it more obvious what the selection is.

To give a specific example, loading the main page right now shows

```
- button aria-label Filter by time
   - span All time (current selection)
- ul of all options (current selection given active class, not sure this even emits a different voice in screenreaders)
```

With these changes the structure changes to

```
- button aria-label Filter by time (nothing indicating current selection inside the element with the label)
- ul of all options (current selection given active class, not sure this even emits a different voice in screenreaders)
```

I don't want perfect to be the enemy of good, so this isn't a request for changes, but more wanted to put my thoughts down on why I took the choices I did with the UI change originally to add the visible label of the selection

andrewmoise:
Hm, seems like it should be possible to structure things so it works for screen readers. I'm not at all familiar with what's involved, but I have some feeling that someone at sdf.org is vision-impaired and they have some level of accessibility focus; I can ask over there and see if there's someone who's knowledgeable can help with feedback on it.

Maybe for a next cut? I'd like to do this in a few rounds of changes anyway, so maybe the next round can have aria fixes and consult with someone who uses a screenreader about how exactly to make it work well.

melroy89:
> Hm, seems like it should be possible to structure things so it works for screen readers. I'm not at all familiar with what's involved, but I have some feeling that someone at sdf.org is vision-impaired and they have some level of accessibility focus; I can ask over there and see if there's someone who's knowledgeable can help with feedback on it.
> 
> Maybe for a next cut? I'd like to do this in a few rounds of changes anyway, so maybe the next round can have aria fixes and consult with someone who uses a screenreader about how exactly to make it work well.

What would be a nice follow-up. Thanks! It's indeed very hard for developers to validate against this kind of accessibility software. After all we do not own such devices, so any feedback or PRs are welcome to improve accessibility in any way or form!

assets/styles/layout/_options.scss
```diff
@@ -28,6 +28,26 @@
...
+  
+  .options__filter .dropdown:hover .dropdown__menu,
+  .options__filter .dropdown:focus-within .dropdown__menu {
+    /* Positioning for dropdown menus inside .options__main */
+    left: 0;
+    right: auto; /* Reset the right property */
+    top: 100%; /* Position it below the trigger element */
+    transform: none;
+    min-width: 10rem;
+  }
```
BentiGorlich, line 49: These 2 are the same so they can be merged

assets/styles/layout/_options.scss
```diff
@@ -65,6 +85,13 @@
 
   }
 
+  &__filter{
+    button{
+      border-bottom-left-radius: 0px!important;
+      border-bottom-right-radius: 0px!important;
```
BentiGorlich, line 91: is the `!important` really necessary? Its always a bad idea to put that in css if it can be done a different way, e.g. by specifying the selector more precisely

assets/styles/layout/_options.scss
```diff
@@ -123,6 +154,25 @@
...
+
+    i {
+      font-size: .85rem;
+    }
+
+    span {
+      font-size: .85rem;
+      margin-left: 0.5rem;
+    }
+  }
```
BentiGorlich, line 174: There already is a selector `.options __filter button` in line 88, please merge it with that

assets/styles/layout/_options.scss
```diff
@@ -113,6 +140,10 @@
     scrollbar-width: none;
   }
 
+  &__filter {
+    justify-content: start;
+  }
```
BentiGorlich, line 145: please put this in the other `__filter` line 88

assets/styles/layout/_options.scss
```diff
@@ -123,6 +154,25 @@
     text-transform: uppercase;
   }
 
+  .hide-on-mobile {
+    @include media-breakpoint-down(lg) {
+      display: none;
+    }
+  }
```
BentiGorlich, line 161: This could be a global css thing right?

src/Controller/Entry/EntryFrontController.php
```diff
@@ -17,46 +16,62 @@
...
     }
 
-    public function front(?string $sortBy, ?string $time, ?string $type, Request $request): Response
+    public function front(?string $sortBy,
+        ?string $time,
+        ?string $type,
+        ?string $filter,
+        string $federation,
+        Request $request): Response
     {
```
BentiGorlich, line 37: I'd prefer this to be single line. I don't know what our guidelines are on this though

src/Controller/Entry/EntryFrontController.php
```diff
@@ -17,46 +16,62 @@
...
         $user = $this->getUser();
+
+        if (null === $filter) {
+            if (null !== $request->query->get('filter')) {
+                $filter = $request->query->get('filter');
+            } elseif ($user) {
```
BentiGorlich, line 43: this should not be an `elseif` should it?

templates/entry/_options.html.twig
```diff
@@ -1,136 +1,227 @@
...
-    <menu class="options__view">
         <li class="dropdown">
             <button aria-label="{{ 'filter_by_time'|trans }}"
-                    title="{{ 'filter_by_time'|trans }}"><i
-                        class="fa-solid fa-clock" aria-hidden="true"></i> <span>{{ get_time_param_translated() }}</span>
+                    title="{{ 'filter_by_time'|trans }}">
+                <i aria-hidden="true" class="fa-solid fa-clock"></i>
+                {% if showFilterLabels == 'on' or (showFilterLabels == 'auto' and criteria.getOption('time') != '∞') %}
+                    <span>{{ criteria.getOption('time')|trans }}</span>
+                {% endif %}
```
BentiGorlich, line 51: The default theme settings suggest, that the default is showing the labels. However if you do not accept `null`/`empty value` in this if, the labels will not be shown unless the theme setting is set to "on" by the user explicitely

templates/entry/_options.html.twig
```diff
@@ -1,136 +1,227 @@
...
         </li>
         <li class="dropdown">
-            <button
-                    aria-label="{{ 'filter_by_type'|trans }}"
-                    title="{{ 'filter_by_type'|trans }}"><i
-                        class="fa-solid fa-filter" aria-hidden="true"></i>
+            <button aria-label="{{ 'filter_by_type'|trans }}"
+                    title="{{ 'filter_by_type'|trans }}">
+                <i aria-hidden="true" class="
+                    {% if criteria.getOption('type') == 'all' %} fa-solid fa-file
```
BentiGorlich, line 107: I now this would be longer, but for better readability I'd go with
```twig
{% if ... %}
    statement
{% elseif ... %}
    ...
{% endif %}
```

templates/entry/_options.html.twig
```diff
@@ -1,136 +1,227 @@
...
+            <li class="dropdown">
+                <button
+                        aria-label="{{ 'filter_by_subscription'|trans }}"
+                        title="{{ 'filter_by_subscription'|trans }}">
+                    <i aria-hidden="true" class="fa-solid
+                        {% if criteria.favourite %} fa-heart
+                        {% elseif criteria.subscribed %} fa-folder-plus
+                        {% elseif criteria.moderated %} fa-shield-halved
+                        {% else %} fa-earth-americas {% endif %}">
+                    </i>
```
BentiGorlich, line 156: twig if else style like I mentioned previously

templates/entry/_options.html.twig
```diff
@@ -1,136 +1,227 @@
...
+        <li class="dropdown">
+            <button
+                    aria-label="{{ 'filter_by_federation'|trans }}"
+                    title="{{ 'filter_by_federation'|trans }}">
+                <i aria-hidden="true" class="fa-solid
+                    {% if criteria.getOption('federation') == 'all' %} fa-circle-nodes
+                    {% elseif criteria.getOption('federation') == 'local' %} fa-house-chimney
+                    {% elseif criteria.getOption('federation') == 'federated' %} fa-network-wired
+                    {% else %} fa-question {% endif %}">
+                </i>
```
BentiGorlich, line 196: twig if else style like I mentioned previously

templates/layout/_header_bread.html.twig
```diff
@@ -12,14 +12,14 @@
...
     <div class="head-title">
-        <span>/</span><a href="{{ path('front_favourite') }}">fav</a>
+        <span>/</span><a href="{{ '/fav/' }}">fav</a>
     </div>
 {% elseif is_route_name_end_with('_moderated') %}
     <div class="head-title">
-        <span>/</span><a href="{{ path('front_moderated') }}">mod</a>
+        <span>/</span><a href="{{ '/mod/' }}">mod</a>
     </div>
 {% endif %}
```
BentiGorlich, line 25: this is actually not working, because you made the `all|sub|fav|mod` a query parameter. Why is it not just part of the path?

templates/layout/_options_appearance.html.twig
```diff
@@ -8,6 +8,7 @@
     <div class="settings-section">
         {{ component('settings_row_enum', {label: 'sidebar_position'|trans, settingsKey: 'KBIN_GENERAL_SIDEBAR_POSITION', values: [ {name: 'left'|trans , value: 'LEFT'}, {name: 'right'|trans , value: 'RIGHT' } ], defaultValue: 'RIGHT' } ) }}
         {{ component('settings_row_enum', {label: 'page_width'|trans, settingsKey: 'KBIN_PAGE_WIDTH', values: [ {name: 'page_width_max'|trans , value: 'MAX'}, {name: 'page_width_auto'|trans , value: 'AUTO' }, {name: 'page_width_fixed'|trans , value: 'FIXED' } ], defaultValue: 'FIXED', class: 'width-setting' } ) }}
+        {{ component('settings_row_enum', {label: 'filter_labels'|trans, settingsKey: 'KBIN_GENERAL_FILTER_LABELS', values: [ {name: 'on'|trans , value: 'ON'}, {name: 'auto'|trans , value: 'AUTO' }, {name: 'off'|trans , value: 'OFF' } ], defaultValue: 'ON' } ) }}
```
BentiGorlich, line 11: quick comment on here, even though I already mentioned this up top: `defaultValue` means nothing other than "if there is no value yet, just put it to this". So this is not actually setting the value

templates/layout/_topbar.html.twig
```diff
@@ -4,13 +4,13 @@
...
+            <a class="stretched-link" href="{{ path('front') }}">{{ 'subscribed'|trans }}</a>
         </li>
         <li class="{{ html_classes({'active': is_route_name_contains('_moderated')}) }}">
-            <a class="stretched-link" href="{{ path('front_moderated') }}">{{ 'moderated'|trans }}</a>
+            <a class="stretched-link" href="{{ path('front') }}">{{ 'moderated'|trans }}</a>
         </li>
         <li class="{{ html_classes({'active': is_route_name_contains('_favourite')}) }}">
-            <a class="stretched-link" href="{{ path('front_favourite') }}">{{ 'favourites'|trans }}</a>
+            <a class="stretched-link" href="{{ path('front') }}">{{ 'favourites'|trans }}</a>
         </li>
```
BentiGorlich, line 14: this is also not working. Neigher the "active" part nor the link

BentiGorlich (changes requested):
I really like the changes. Though some things do not work, yet
- top bar links do not work
- the bread crumbs for /sub, /mod and /fav do not work
 To fix the two points I'd suggest to just convert the "filter" query parameter to be part of the path. That would fix the code that is there and I would prefer it in general, as it would be in line with the other front parameters

assets/styles/layout/_options.scss
andrewmoise: I didn't invent this; I copied the styling from the &__view button styling. Taking away !important means that the highlights underneath when you mouse over a filter would get shaped with a curve.

Screenshot of buttons

assets/styles/layout/_options.scss
andrewmoise: Probably could be; is there a better place to put it? Or an existing thing like this that I could use instead?

src/Controller/Entry/EntryFrontController.php
andrewmoise: I think it should be -- my feeling on the logic was:

* If the caller of the function has requested a particular filter, use that (that happens if there's one at the front of the URL)
* Otherwise if there's a filter query parameter, use that
* Otherwise if there's a logged-in user, use their default
* Otherwise use 'all'

That sounds right, right? If so, maybe it'd be a little clearer as:

```
if ($filter) {
    // nothing to do
} elseif (null !== $request->query->get('filter')) {
    $filter = $request->query->get('filter');
} elseif ($user) {
    $filter = match ($user->homepage) {
                    User::HOMEPAGE_SUB => 'sub',
                    User::HOMEPAGE_MOD => 'mod',
                    User::HOMEPAGE_FAV => 'fav',
                    default => 'all',
                };
} else {
   $filter = 'all';
}
```

?

Also, the handling of $filter in general is a little messed up; see my comment on the fav/sub/mod diff.

templates/layout/_header_bread.html.twig
andrewmoise: So, under my current code, either /mod/newest *or* /newest?filter=mod will work. They both have to work and do the same thing, and the reason is moderately complicated. Here's my thinking on it; if you can see a better way to handle this. I'm all ears.

So the issue is that before my changes, we had different routes for sub/fav/mod/all. You can see them all in config/kbin_routes/front.yaml. So the implication of that is, a lot of places we could use options_url() wouldn't work -- there was basically no general way to implement the functionality of switching one of the filters to a new setting but leaving the rest unchanged. You can see this on existing instances, e.g. switch the time filter to 6h, then switch to "moderated", and the time filter will reset back to "all time."

In the existing code there was a lot of custom hackery, manually checking the route to look for strings, and things like that, but it didn't really work even for the functionality that was already there and it wasn't going to be sustainable to try to keep making it worse to keep things working with newly-visible filters. So what I did was put it all through a single codepath: We store the filter values in $criteria, and we say that any given page that might get filtered always gets put through the same route, so that we try to always change filter values with options_url(). I didn't eliminate all of the special-cased stuff but I switched a substantial amount of different filtering stuff that went through different codepaths to all go through that codepath.

But, the cost of that unification was, because the URLs are structured with the sub/mod/fav at the beginning and optional, you *can't* have a single route entry that you can use for both e.g. /sub/newest and /newest. The cleanest way I could see to do it was to have it all go through either "root" or "front", and make both of them able to resolve the full set of filters including sub/mod/fav/all. It's just that in "root" it's a query parameter and in "front" it's part of the URL.

So yeah, it's a little weird, but the weird URL structure that causes that to happen is something we came in with, and I didn't see a better way to do it, and the end result is still an improvement on what came before IMO because we're not poking around doing string searches on routes etc, and we have a lot less duplicated code (e.g. the cut and pasted duplicate routes in EntryFrontController).

The other, cleaner, option would be just to break existing URLs and change the structure. As a general rule I'm reluctant to make a URL that used to work suddenly not work or be at a different location, but if people are cool with doing that then we can do this in a much cleaner way and drop the query param.

assets/styles/layout/_options.scss
andrewmoise: Done

assets/styles/layout/_options.scss
andrewmoise: Done 👍🏻

assets/styles/layout/_options.scss
andrewmoise: Also done 👍🏻👍🏻

templates/layout/_options_appearance.html.twig
andrewmoise: Got it, makes sense. 👍🏻

templates/layout/_topbar.html.twig
andrewmoise: Oooh... okay, I completely misunderstood this whole part. I don't have my top bar turned on, and I thought this was just something else. Yeah I'd just completely broken it. I made what I think should be a fix.

assets/styles/layout/_options.scss
andrewmoise: Actually, wait - moving this breaks the options bar. My guess is that "justify-content: end" from the "& > *" block will override it if this block doesn't come later in the file.

I was mainly just following the format of where __main and __view were defined. I consolidated all the button settings as requested, and moved this one to a little less random place while still leaving it late in the file so it'll take effect.

src/Controller/Entry/EntryFrontController.php
andrewmoise: Done

templates/entry/_options.html.twig
andrewmoise: Got it. I fixed the variable set up at the top of the file to default to 'on' if we don't find a cookie.

templates/entry/_options.html.twig
andrewmoise: Done

templates/entry/_options.html.twig
andrewmoise: Also done

templates/entry/_options.html.twig
andrewmoise: Done

andrewmoise:
Okay, I pushed a new proposal; I think I fixed everything except for:

* Putting hide-on-mobile in a better place. I'm happy to move it, just tell me where.
* Fixing !important on the hover styling for the buttons -- currently it matches with the styling on the other buttons it was modeled on. I agree it's not ideal but I'm not sure if fixing everything not-ideal but preexisting that my checkin touches should be a prerequisite for getting it in, LMK what you think though.
* If you want an alternate approach for the "filter" query param awkwardness I'm open to whatever you suggest. Like I say my current implementation is a little weird, but was the best way I saw to balance the different constraints involved and get everything working while overall making things cleaner. But, I'm open to pretty much anything that seems like it would work.

templates/entry/_options.html.twig
```diff
@@ -1,136 +1,244 @@
...
+                        <a href="{{ options_url('filter', 'fav') }}" class="{{ html_classes({'active': criteria.favourite }) }}">
+                            <i class="fa-solid fa-heart" aria-hidden="true"></i> &nbsp; {{ 'favourites'|trans }}
+                        </a>
+                    </li>
+                </ul>
+            </li>
+        {% endif %}
+        <li class="dropdown">
+            <button
+                    aria-label="{{ 'filter_by_federation'|trans }}"
```
e-five256, line 201: this isn't in the translation file

translations/messages.en.yaml
```diff
@@ -24,6 +25,7 @@ commented: Commented
 change_view: Change view
 filter_by_time: Filter by time
 filter_by_type: Filter by type
+filter_by_subscription: Filter by subscription
```
e-five256, line 28: this appears unused, unless git is hiding it from me, maybe it was meant to be `filter_by_federation`?

src/Security/KbinAuthenticator.php
```diff
@@ -56,7 +56,7 @@ public function onAuthenticationSuccess(Request $request, TokenInterface $token,
         /** @var User $user */
         $user = $token->getUser();
 
-        return new RedirectResponse($this->urlGenerator->generate($user->homepage));
```
e-five256, line 59: this is an interesting change; I think it's unrelated to the UI filter changes? Looking at it, I assume it works because the front controller itself redirects based on user. it looks like `$user` is unused now, so maybe no longer needs to be fetched? Unless there's some magic PHP thing it's doing

templates/entry/_options.html.twig
```diff
@@ -1,136 +1,244 @@
...
-    </menu>
-    <menu class="options__view">
         <li class="dropdown">
             <button aria-label="{{ 'filter_by_time'|trans }}"
-                    title="{{ 'filter_by_time'|trans }}"><i
-                        class="fa-solid fa-clock" aria-hidden="true"></i> <span>{{ get_time_param_translated() }}</span>
+                    title="{{ 'filter_by_time'|trans }}">
+                <i aria-hidden="true" class="fa-solid fa-clock"></i>
+                {% if showFilterLabels == 'on' or (showFilterLabels == 'auto' and criteria.getOption('time') != '∞') %}
+                    <span>{{ criteria.getOption('time')|trans }}</span>
```
e-five256, line 50: I'm fine if you want to ditch `get_time_param_translated` completely including inside the menu selection, and delete the code for it. My intent with it was to sanitize the url paramters being sent to the translate package, but perhaps symfony does that already and it was overly cautious of me

However, a slight nitpick but I think it would be better to show `All` rather than `∞` to be consistent. @nobodyatroot and I actually discussed removing `∞` altogether as it's a bit out of place, but even if it's how the routing uses, I think the text `All` might be more clear

andrewmoise: Yeah, absolutely on both counts. I think in the long run we should be eliminating get_time_param_translated() in addition to a bunch of other similar one-off functions, and just using Criteria.getOption() as the unified way. My thinking was to leave a lot of them in place temporarily, convert the filtering display to the new way one controller at a time, and then clean up and remove the no-longer-needed one-offs at the end.

And yes, I actually tried to remove `∞` but backed it up once I ran into a fair hornet's nest of stuff. I think in the long run, it'd be good to do a whole cleanup pass and remove a lot of the inconsistency in how strings are used to represent these options values internally -- `∞` versus `all` is far from the worst of it; for a single option of the subscription-status filter we use `featured`, `fav`, `favorite`, `favourite`, or `favourites` at different places in the code, and then attempt to translate among those values at various points so that code that expects one version of the string won't break when it gets the wrong version. Honestly, it's madness. I had to sort of draw the line of what level of cleanup it made sense to do while I was doing this already-too-invasive change, but if you're open to the idea I'd like to go in at some point later and take a pass at making it all consistent internally. It would have saved me a certain amount of time and confusion on this change if it had been consistent going into it.

Changing the displayed value to `all` is easy if that's preferable though, sure, I can do that.

src/Security/KbinAuthenticator.php
andrewmoise: Oh, you're right - I just wasn't looking; I don't think we don't need to fetch $user anymore in that codepath no. I'll push the appropriate fix.

And it's... tangentially related to the UI filter code. In order to do the filtering, I had to deduplicate some of the code between what used to be different routes for the home page, and make it all one route. So, now that we only have one route, this change (a) keeps things working with the new code and (b) is a good cleanup anyway, since IMO this code shouldn't have to care about user settings or routes.

translations/messages.en.yaml
andrewmoise: Hm? It's used for mouseover text, I think.

```
$ grep -r filter_by_subscription `ls | egrep -v var\|public`
templates/entry/_options.html.twig:                        aria-label="{{ 'filter_by_subscription'|trans }}"
templates/entry/_options.html.twig:                        title="{{ 'filter_by_subscription'|trans }}">
translations/messages.en.yaml:filter_by_subscription: Filter by subscription
$
```

templates/layout/_header_bread.html.twig
andrewmoise: So... I thought a little more after writing this comment. How do people feel about just changing the URL structure so that fav/mod/sub/all is later on in the URL?

I actually feel like breaking the /mod/ and friends URLs maybe isn't too bad, and the gain we'd have by moving to only one URL scheme would maybe be worth it.

The other option would be to move towards using query params for all this stuff; it's starting to get a little bit awkward to have so many parameters just in an exactly-ordered list of parts of the URL, where a key-value approach would be more robust, apply a lot more easily as we're trying to do this for other routes, and also easier to make sense of if someone's looking at the URL.

What do you think?

templates/entry/_options.html.twig
andrewmoise: Good spot; fixed

assets/styles/layout/_options.scss
BentiGorlich: Got it :+1:

src/Controller/Entry/EntryFrontController.php
BentiGorlich: checks out :+1:

templates/layout/_header_bread.html.twig
BentiGorlich: I'd just make the fav/mod/sub/all stuff part of the URL path. It does not have to be in the first place, but it should be part of it and not a query parameter. If we want the breadcrumbs for it removed, I fine with it as well, though I'd like to keep it for fav/mod/sub. This is a very easy way to see that one is on the correct page, but I am open for discussion on this one. 

We could still keep redirects on `/sub` etc. to the new path if we'd want to

andrewmoise: Yeah, having it on the URL path in one consistent new way and having a redirect from the old (current) way sounds way better, you're right. I'll fix it to be that way + push a fix.

For breadcrumbs, let me share my thoughts in Matrix -- I think I may be causing some short-term breakage of breadcrumbs, but it's in service of going to a destination that I feel like would be way better in the long run. But if I share my overall thinking on it maybe we can come to something that's good for both the short and long term. And of course as always I may be missing or misunderstanding something.

src/Security/KbinAuthenticator.php
e-five256: My only slight concern if this code was there for a reason; I suppose it makes sure the user gets their desired homepage on login, otherwise it relies on the symfony redirect handler. Locally, I didn't see an issue, but local dev env doesn't use nginx or other rewrite handlers. I _think_ @nobodyatroot saw an issue with the redirect component not working, but I can't recall if that was an nginx thing, or cloudflare thing

assets/styles/layout/_options.scss
e-five256: The goal of this is to override the curved edges when there is more of the element to the left and right. So in the screenshot, that appears incorrect to me as it should be gaining 0px bottom left and bottom right, because it appears in the middle of that selection bar.

options on the far left edge would have 0px right only, the rightmost edge would have 0px left only, and everything in the middle gets 0px left and right.

you should be able to accomplish this with things like `:first-child` or `:not(:first-child)`, etc. I did this when rearranging the bar last time.

To try to give visual examples, the goal is that the curve highlight matches the curve of the bar itself

image
image
image

(bottom one has the curve but very subtle)

andrewmoise: Yeah, sounds good - it's currently working and I think it's using the approach you described. The screenshot was just an illustrative example for why we need !important in it for the time being until some other approach gets figured out.

src/Security/KbinAuthenticator.php
andrewmoise: Yeah, definitely - I've replaced the functionality that this was implementing (making sure that people got their desired homepage) with a different implementation, not just removed it.

We used to make that decision by sending people down one of a variety of routes, but now 'front' does all that, so it should be safe to just send everyone to 'front'.

templates/layout/_header_bread.html.twig
andrewmoise: Just pushed an update that puts everything through the /$filter/$sortBy/etc route, and uses a redirect if anyone tries to access a /$sortby/etc path.

Should I put breadcrumbs for /$filter back into the nav bar? To me that's weird if the nav bar is adjusting as you're picking new filter params way down the page, but I'm fine with whatever way.

src/Controller/Entry/EntryFrontController.php
andrewmoise: Should be mooted now, by the single-route changes

BentiGorlich:
I noticed another problem with your changes, which sadly make it a bit more complicated :sweat_smile: 

You kinda forgot about the microblog side. What I'd want you to do, is to merge the `EntryFrontController` and the `PostFrontController` in a new `FrontController`. I think it should not be that big a deal, since the filter options are exactly the same. If you need any help, just hit me up on matrix. 

It would be very nice to have this already done. On the base of your changes I wanted to work on the combined microblog + thread view. 

Additionally, the federation setting next to the theme settings is still there, that needs to be removed as well, right?

assets/styles/layout/_options.scss
andrewmoise: I put it in _breakpoints.scss with the other screen size stuff.

templates/layout/_header_bread.html.twig
andrewmoise: Just put the breadcrumbs back. The breadcrumb and top bar links probably want some testing but I poked at them some and it seem like they're working to me now.

templates/entry/_options.html.twig
andrewmoise: Latest change changes the display value to "All".

andrewmoise:
Yeah; I didn't forget about microblog, I just wanted to do the thing in stages. I was already a little uncomfortable with the depth of changes I was making, being a little unfamiliar with the codebase, and so my thinking was to do it for the front threads view, then the microblog as the next stage, then the magazine view, and then stop and reevaluate and maybe do some cleanup. That's also the reason for not taking the federation switch away quite yet; I don't want to remove it until *all* the things that currently use it are converted over to a new way.

IDK, you tell me - now that I'm more familiar, I'd be fine with keeping going with constructing the changes, and then doing the whole "add filtering to all places at once" change as one big PR and then getting testing (from review and from deploying it to real instances) on the whole thing at once. That might be less disruptive than doing it piece by piece and leaving an inconsistent UI from route to route while we're in the middle of the transition.

andrewmoise:
(Oh, and making the combined thread+microblog view, yeah holy crap that sounds awesome I love it)

(Also, my proposal was about making *just* the thread and microblog filters as one PR if you think that's better. Magazine filtering, I do want to do, but I think that should go with a later PR so this one can get testing first.)

melroy89:
@andrewmoise can we continue with this PR? What are the blocking changes before we can merge it? Do you need help?

andrewmoise:
Oh, I don't think there are any blocking changes - as far as I know everything is addressed and it's good to go, I was just waiting for people to get back to me. I think the changes are big enough that they want some careful looking over / some testing before actually going in, but I think it's good.

Maybe if we're a week away from a 1.4.0 release, this should wait until after the release, and then we can shoot for with the next release, this going in along with some further UI improvements? You tell me, but that would be my feeling.

BentiGorlich:
Oh ok, I though that it wasn't ready yet to be reviewed again, too. I'll have a look at it hopefully sometime this week. I am very excited about it. From what I've seen so far you did a great job with it :)

andrewmoise:
Ayy, thank you! Yeah, I think I said something on Matrix but never updated anything on github indicating that it was good to go. Probably github or both is better. Anyway all good.

src/Controller/Entry/EntryFrontController.php
```diff
@@ -5,213 +5,132 @@
...
-
-        $front = match ($user->homepage) {
-            User::HOMEPAGE_SUB => 'subscribed',
-            User::HOMEPAGE_MOD => 'moderated',
-            User::HOMEPAGE_FAV => 'favourite',
-            default => 'front',
-        };
-
-        return $this->$front($sortBy, $time, $type, $request);
+        return $this->front($sortBy, $time, $type, $request->query->get('subscription'), $federation, $request);
```
BentiGorlich, line 29: `content` is missing in the method declaration

src/Controller/Entry/EntryFrontController.php
```diff
@@ -5,213 +5,132 @@
...
-                        [
-                            'entries' => $listing,
-                        ]
-                    ),
-                ]
-            );
+        $user = $this->getUser(); // Fetch the user
+        $subscription = $this->subscriptionFor($user); // Determine the subscription filter based on the user
+
+        if ($magazine) {
```
BentiGorlich, line 114: `$magazine` is never defined. It is not available in this scope

src/Controller/Magazine/MagazineListController.php
```diff
@@ -29,7 +30,7 @@ public function __invoke(string $sortBy, string $view, Request $request): Respon
         $criteria = new MagazinePageView(
             $this->getPageNb($request),
             $sortBy,
-            $this->getValueOfFederationCriteria($request),
+            Criteria::AP_ALL,
```
BentiGorlich, line 33: So this cannot be filtered for federated/local content anymore?

src/Repository/Criteria.php
```diff
@@ -266,4 +297,20 @@ public function getSince(): \DateTimeImmutable
             default => throw new \LogicException(),
         };
     }
+
+    public function getOption(string $key): string
+    {
+        return match ($key) {
+            'sort' => $this->resolveSort($this->sortOption),
+            'time' => '∞' === $this->resolveTime($this->time, true) ? 'all' : $this->resolveTime($this->time, true),
```
BentiGorlich, line 305: why is `reverse` here `true`?

src/Controller/Entry/Comment/EntryCommentFrontController.php
```diff
@@ -26,7 +26,6 @@ public function front(?Magazine $magazine, ?string $sortBy, ?string $time, Reque
         $params = [];
         $criteria = new EntryCommentPageView($this->getPageNb($request));
         $criteria->showSortOption($criteria->resolveSort($sortBy ?? Criteria::SORT_DEFAULT))
-            ->setFederation('false' === $request->cookies->get(ThemeSettingsController::KBIN_FEDERATION_ENABLED, true) ? Criteria::AP_LOCAL : Criteria::AP_ALL)
```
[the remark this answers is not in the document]
andrewmoise: This is what I was asking about on Matrix some time back I think -- it's the route for /comments/ with no particular magazine attached to it, just all comments on the instance, which honestly doesn't even make a ton of sense to me as a thing that anyone would ever use. I advocated for just removing it if no one could think of a use for it, and the consensus was instead just to break federation-filtering on it instead and go back to not thinking about it.

src/Controller/Entry/Comment/EntryCommentFrontController.php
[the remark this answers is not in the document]
andrewmoise: Ah, good spot, I can do that if we do want to keep this route around.

src/Controller/Entry/Comment/EntryCommentFrontController.php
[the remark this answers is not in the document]
BentiGorlich: Ah ok then. Yeah didn't know this view existed. Maybe it is for moderation purposes?

src/Repository/Criteria.php
```diff
@@ -211,13 +208,23 @@ public function resolveTime(?string $value): ?string
             'all' => Criteria::TIME_ALL,
         ];
 
-        return $routes[$value] ?? null;
+        if ($reverse) {
```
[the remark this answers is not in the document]
andrewmoise: This is used to map time values "backwards" from Criteria:: labels into human-readable labels. It's used for changing "∞" back to "All" as requested during some previous round of review.

Frankly, the whole way this is set up needs to be refactored in my opinion. It's *maybe* fine to have one set of values internally, and one set of human-readable values that get used for translation, in some cases. In most cases I think we shouldn't even do that because it's just extra complexity. The whole thing of using a variety of different labels beyond those two options for different options in different portions of the code and having this byzantine translation layer that converts among the different ways of doing it, to me is madness.

Like with some other things I had to sort of draw the line in terms of deciding when to stop wandering around and fixing things that maybe aren't ideal, so my decision with "resolving" the time values was to just add another layer to the pile of conversion code, having in mind that a good change down the road would be to take a pass through the whole codebase and filter down to, ideally, one single representation for this kind of stuff so there's no translation layer at all.

src/Controller/Entry/EntryFrontController.php
```diff
@@ -5,213 +5,132 @@
...
+            $content_tmpl = 'entry/';
+            $content_key = 'entries';
+        } elseif ('microblog' === $content) {
+            $posts = $this->postRepository->findByCriteria($criteria);
 
-        $method = $criteria->resolveSort($sortBy);
-        $listing = $this->$method($criteria);
+            $content_tmpl = 'post/';
+            $content_key = 'posts';
+        }
```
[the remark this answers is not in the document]
andrewmoise: It should yes, I'll do that

src/Controller/Entry/Comment/EntryCommentFrontController.php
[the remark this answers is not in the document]
andrewmoise: Yeah, I thought about that, like *maybe* you want to be able to view all comments but filter it to moderated subs or something... that was part of the thinking in keeping it working. Like maybe there's some weird use case, but the chance that that use case overlaps with needing to do federation filtering is unlikely on top of the already-unlikely of someone being using it in the first place.

src/Twig/Runtime/UrlExtensionRuntime.php
```diff
@@ -252,13 +252,17 @@ public function postCommentDeleteUrl(PostComment $comment): string
         ]);
     }
 
-    public function optionsUrl(string $name, string $value, string $routeName = null): string
+    // Hopefully $additionalParams is temporary
+    public function optionsUrl(string $name, string $value, string $routeName = null, array $additionalParams = []): string
```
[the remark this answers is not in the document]
andrewmoise: I think currently $additionalParams is only used in one very specific case (header nav bar for per-magazine view), and the hope is that that use case will go away with future UI changes (unification of microblog and theads content) and so this change won't need to stay around.

There's nothing particularly wrong with needing to use optionsUrl() to add multiple new options, just it'd be cleaner to avoid adding this rarely-used codepath and I think in the near future it'll go back away again.

src/Repository/Criteria.php
andrewmoise: See above talking about conversion of '∞', this is using that functionality

src/Repository/Criteria.php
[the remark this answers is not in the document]
BentiGorlich: omg, ok yeah that is terrible. I'd say remove '∞' completely, but if that is too much work, please rename that variable or at least write a comment above it.

src/Repository/Criteria.php
BentiGorlich: please add a comment here why this is necessary. Right now it is just borderline unreadable :sweat_smile:

src/Repository/Criteria.php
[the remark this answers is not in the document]
andrewmoise: Yeah. It's bonkers. User presentation layer is one thing, but when we're saying different sections of the codebase need to represent the same thing by using strings that say "all" versus "∞" or like I talked about before, "favourite" versus "favorite" versus "favourites" versus "featured", and do conversion layers between the different sections to turn one of those values into another, to say I think it's not ideal is an understatement.

In this case we use "∞" internally to some certain extent, so unfortunately we still need the translation layer 🙁, until the day comes to go through the whole codebase and standardize on one representation (ideally the one that's defined in those constants in Criteria:: -- maybe that indicates that someone had made a start on doing that, IDK).

src/Repository/Criteria.php
andrewmoise: Hm, how's this sound?

```
    // resolveTime() converts our internal values into ones for human presenation
    // $reverse = true indicates converting back, from human values to internal ones

    // This whole approach is a mess; this translation layer is temporary until
    // we have time to take a pass through the whole codebase and convert so there's
    // no such thing as multiple alternate value strings and translation layers
    // between them. This is just a temporary measure to produce desired output
    // until the whole layer goes away.
    public function resolveTime(?string $value, bool $reverse = false): ?string

```

Also, you'll be entertained to learn that Copilot auto-suggested this comment when I typed "// So:":

```
    // So: This whole method is a mess. It's not clear what it's supposed to do, and it's not clear what the input and output are supposed to be.
```

So it agrees with you 🙂

src/Controller/Entry/EntryFrontController.php
andrewmoise: Ah, good spot, that's embarrassing. And... it's easy to fix the typo $magazine -> $name, but it led me to a whole area of microblog routing that seems like will need some careful attention to make sure is working properly under the new scheme. I'll look into it + make sure it's right but it'll take a little time.

src/Controller/Entry/EntryFrontController.php
andrewmoise: It was dead code because we didn't use the 'root' route; I eliminated ::root() and the root route completely.

src/Controller/Entry/EntryFrontController.php
```diff
@@ -228,78 +149,93 @@ public function magazine(
...
 
-        $criteria = (new EntryPageView($this->getPageNb($request)));
+        if ('threads' === $content) {
+            $criteria = new EntryPageView($this->getPageNb($request));
+            $criteria->setContent('threads');
+        } elseif ('microblog' === $content) {
+            $criteria = new PostPageView($this->getPageNb($request));
+            $criteria->setContent('microblog');
+        } else {
+            throw new LogicException('Invalid content '.$content);
```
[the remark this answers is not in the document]
andrewmoise: Hm, my fixer wants to turn it into \LogicException for some reason so that's what I checked in. Das good?

src/Controller/Entry/EntryFrontController.php
andrewmoise: Okay, I think I sorted it out; I *think* it's enough to just add the right parameters to the magazine_posts redirect and everything will work. I tested it lightly but not real intensively.

templates/layout/_header_nav.html.twig
```diff
@@ -1,13 +1,13 @@
...
+           class="{{ html_classes({'active': not is_route_name_contains('user_') and ((is_route_name_starts_with('front') and criteria and 'threads' == criteria.getOption('content')) or is_route_name_contains('_entries') or entry is defined and entry)}) }}">
             {{ 'threads'|trans }} {% if magazine is defined and magazine %}({{ magazine.entryCount }}){% endif %}
         </a>
     </li>
     <li>
-        <a href="{{ navbar_posts_url(magazine ?? null) }}"
-           class="{{ html_classes({'active': (is_route_name('posts_front') or is_route_name_contains('posts') or post is defined)  and not is_route_name_contains('user') }) }}">
+        <a href="{{ options_url('content', 'microblog', magazine is defined and magazine ? 'front_magazine' : 'front', {'name': magazine.name} ) }}"
+           class="{{ html_classes({'active': ((is_route_name_starts_with('front') and criteria and 'microblog' == criteria.getOption('content')) or is_route_name('posts_front') or is_route_name_contains('posts') or post is defined)  and not is_route_name_contains('user') }) }}">
```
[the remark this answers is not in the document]
andrewmoise: Hm, when would a magazine have a null name? We're depending on name for some other stuff so there are some other things I might have to re-figure-out if that's an allowed case.

src/Repository/Criteria.php
andrewmoise: Added

templates/layout/_header_bread.html.twig
```diff
@@ -10,16 +12,16 @@
...
+{% elseif filter_option == 'fav' or is_route_name_end_with('_favourite') %}
     <div class="head-title">
-        <span>/</span><a href="{{ path('front_favourite') }}">fav</a>
+        <span>/</span><a href="{{ '/fav/' }}">fav</a>
     </div>
-{% elseif is_route_name_end_with('_moderated') %}
+{% elseif filter_option == 'mod' or is_route_name_end_with('_moderated') %}
     <div class="head-title">
-        <span>/</span><a href="{{ path('front_moderated') }}">mod</a>
+        <span>/</span><a href="{{ '/mod/' }}">mod</a>
```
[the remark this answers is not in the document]
andrewmoise: Fixed, I just hadn't fully updated the template for some cleanups elsewhere

src/Repository/Criteria.php
melroy89: Thank you copilot for feeling bad

andrewmoise: I mean... tell you what. One other option for how to do the whole thing would be to refactor *first* -- consolidate down to just one representation / one string value for each filtering option everywhere, and *then* do the filter changes with this whole ugly translation layer already removed.

Basically make things better and then add the feature, as opposed to making things worse with adding the feature, with the vague plan at some point in the distant future of making it back better again. I don't like the idea of blowing up the scope of this PR until everything it touches can be beautiful again, but if this way of doing the translation is too complex and weird, to the point that we want to wait on a separate change to clean things up first, I can kind of understand that.

I was biased in favor of trying to get the filtering changes working + merged, but the other way works too; it just might take a little longer. Just LMK.

melroy89:
Don't forget to click "resolve conversation" when you fixed it. Do it's then easier to see if there are any things left open / Todo.

andrewmoise:
Okay, I think this is ready for review again. I gave it some testing and found some stuff which I've now fixed; I still won't say the testing is crazy in depth or comprehensive but more so than before.

Of note, I've crapped up the optionsUrl() interface a little more and it's now no longer a temporary change. So -- with this and resolveTime() both, I have done something as kind of a hack to get things working, with vague plans to go back later and do the proper fix in a separate PR. As an alternative to committing the hacks, would it maybe be better to do that cleanup *first*, and then have this PR as the second stage? If people are having objections to doing a hacky solution with the excuse that it's only temporary? I mean, that type of objection I can understand. I don't want to get caught up into a snowballing ever expanding scope of changing out stuff in order to do this one feature; I think it's up to y'all where to draw that judgement of hack to get it working vs. refactor but it takes a long time to get done. But if your conclusion is refactor first, I can get that; it makes sense to me.

(Specifically the proper fixes I'm talking about are: (1) making optionsUrl() just take an associative array instead of a single name+value but then having an optional associative array as a separate parameter (2) doing away with the whole "different string values for the same thing with a translation layer between them" so we can just remove resolveTime() + resolveType() + translateType() + etc etc etc.)

templates/layout/_header_nav.html.twig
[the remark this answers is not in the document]
andrewmoise: So, I'm not sure, but I suspect that the underlying issue behind your error was not magazine.name but one of the other server-error issues that I fixed with recent commits. Can you retest with the current version and see if it's still an issue?

andrewmoise:
> Don't forget to click "resolve conversation" when you fixed it. So it's then easier to see if there are any things left open / Todo.

Done. I didn't want to be closing people's comments for them, just to give them a chance to weigh in on whether it's *actually* fixed or not, but if it's preferable for me to do it, I've just gone back and resolved any of the recent ones where it's clearly as simple as "hey change this" "okay I changed that / mooted the issue by removing that code". Anything more complex, I left as unresolved pending someone else looking at it.

src/Twig/Runtime/UrlExtensionRuntime.php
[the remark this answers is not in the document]
andrewmoise: So, I made it worse (with the current iteration we're wanting to use optionsUrl() to clear out the ?p= parameter when we change the filters, so we go back to the first page of results for the new filter settings, so it may not be so temporary anymore). See my comment below.

BentiGorlich (changes requested):
The most important thing is that the site should not break when magazine is undefined :D
After you fixed that I will put in on my main instance and test it there.

But regardless of yet another review round: you're doing an awesome job. This PR is just quite large and there there is a lot of room for error, but I love this change and cannot wait to have it on main :)

templates/layout/_header_nav.html.twig
```diff
@@ -1,26 +1,46 @@
...
+    {% endif %}
+{% endif %}
+
 {% if header_nav is empty %}
     <li>
-        <a href="{{ navbar_threads_url(magazine ?? null) }}"
-           class="{{ html_classes({'active': not is_route_name_contains('user_') and (is_route_name_starts_with('front') or is_route_name('root') or is_route_name_contains('_entries') or entry is defined and entry)}) }}">
+        <a href="{{
+            is_route_name_starts_with('front')
+            ? options_url('content', 'threads', magazine is defined and magazine ? 'front_magazine' : 'front', {'name': magazine.name})
```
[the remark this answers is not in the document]
andrewmoise: Done. IDK why this wasn't causing any problems for me, but yes we need to check before accessing magazine.name.

BentiGorlich:
After the error on the homepage is fixed now I put it up on gehirneimer.de
Let's see what I'll find :grin:

BentiGorlich:
I didn't find any errors, but have a different branch running on gehirneimer now. If you fix the remaining annotations, then I think this is good to go

src/Controller/Entry/EntryFrontController.php
```diff
@@ -5,213 +5,129 @@
...
         $criteria->showSortOption($criteria->resolveSort($sortBy))
-            ->setFederation(
-                'false' === $request->cookies->get(
-                    ThemeSettingsController::KBIN_FEDERATION_ENABLED,
-                    true
-                ) ? Criteria::AP_LOCAL : Criteria::AP_ALL
-            )
+            ->setFederation($federation)
             ->setTime($criteria->resolveTime($time))
             ->setType($criteria->resolveType($type));
```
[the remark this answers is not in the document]
andrewmoise: Wait, are you talking about this?

```
        if ('threads' === $content) {
            $criteria = new EntryPageView($this->getPageNb($request));
            $criteria->setContent('threads');
        } elseif ('microblog' === $content) {
            $criteria = new PostPageView($this->getPageNb($request));
            $criteria->setContent('microblog');
        } else {
```

That doesn't seem like duplicated code to me; am I missing something?

src/Controller/Entry/EntryFrontController.php
```diff
@@ -5,213 +5,129 @@
...
+            $criteria->subscribed = true;
+        } elseif ('mod' === $subscription) {
+            $this->denyAccessUnlessGranted('ROLE_USER');
+            $criteria->moderated = true;
+        } elseif ('fav' === $subscription) {
+            $this->denyAccessUnlessGranted('ROLE_USER');
+            $criteria->favourite = true;
+        } elseif ($subscription && 'all' !== $subscription) {
+            throw new \LogicException('Invalid subscription filter '.$subscription);
         }
```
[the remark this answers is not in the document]
andrewmoise: So... this if statement is a consequence of Criteria handling the subscription filters with a few boolean variables instead of a single choice-among-options variable like all the other filters are. Honestly in my opinion it'd be better to clean that up into a single choice-among-options variable instead, but like I was saying for some other things I had to sort of draw the line in terms of how much other refactoring to wander around doing as part of the UI change. My thinking was more or less, as long as I've refactored some stuff to improve it overall, I'm not obligated to go around refactoring every bit of code I'm touching (e.g. in this case condensing ->subscribed and ->moderated and ->favourite all down into one variable). With the three separate variables, I'm not real sure how I could eliminate the branching if here with some duplication among the different pathways. The further-ahead solution I think would be to condense all those variables into one at which point the duplication would go away. What do you think?

(As a separate issue, I realized looking at this that we need to be setting $user on the other pathways also; I'll check in a change to do that.)

src/Controller/Entry/EntryFrontController.php
```diff
@@ -228,78 +146,93 @@ public function magazine(
...
         $criteria->showSortOption($criteria->resolveSort($sortBy))
-            ->setFederation(
-                'false' === $request->cookies->get(
-                    ThemeSettingsController::KBIN_FEDERATION_ENABLED,
-                    true
-                ) ? Criteria::AP_LOCAL : Criteria::AP_ALL
-            )
+            ->setFederation($federation)
             ->setTime($criteria->resolveTime($time))
             ->setType($criteria->resolveType($type));
```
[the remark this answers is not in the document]
andrewmoise: How is this duplication? EntryPageView and PostPageView are different here.

src/Controller/Entry/EntryFrontController.php
```diff
@@ -228,78 +146,93 @@ public function magazine(
...
+        } elseif ('mod' === $subscription) {
+            $this->denyAccessUnlessGranted('ROLE_USER');
+            $criteria->moderated = true;
+        } elseif ('fav' === $subscription) {
+            $this->denyAccessUnlessGranted('ROLE_USER');
+            $criteria->favourite = true;
+        } elseif ($subscription && 'all' !== $subscription) {
+            throw new \LogicException('Invalid subscription filter '.$subscription);
+        }
+
```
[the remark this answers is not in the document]
andrewmoise: See above

src/Controller/Entry/EntryFrontController.php
```diff
@@ -228,78 +146,93 @@ public function magazine(
...
+            $listing = $this->entryRepository->findByCriteria($criteria);
+
+            $content_tmpl = 'entry/';
+            $content_key = 'entries';
+        } elseif ('microblog' === $content) {
+            $listing = $this->postRepository->findByCriteria($criteria);
+
+            $content_tmpl = 'post/';
+            $content_key = 'posts';
+        }
```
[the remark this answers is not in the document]
andrewmoise: We already had a check for that up above nearer the beginning of the function, but sure I can add another one here for safety in case something changes / something weird happens.

src/Controller/Entry/EntryFrontController.php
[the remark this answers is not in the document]
BentiGorlich: grafik

src/Controller/Entry/EntryFrontController.php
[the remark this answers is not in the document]
BentiGorlich: No you are not obligated to refactor everything you touch. This code is just really got to move into a small method so it is not duplicated :sweat_smile:

templates/layout/_topbar.html.twig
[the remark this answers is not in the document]
andrewmoise: Hm... wait, if the filter bar can switch moderated/subscribed/etc for the currently displayed content, would we even need links that do exactly the same thing in the top bar anymore?

A lot of my thinking behind this change was to simplify the UI and not have stuff scattered around to different places... honestly from a UX perspective my thinking would be, if we would have the top bar links just change the currently-displayed filter, to just remove the top bar links the same way we removed the button that changed the subscription filter. Just make it all work in a single sensible way instead of two redundant ways. No? What do you think?

templates/post/_options.html.twig
```diff
@@ -1,94 +1,258 @@
...
+                    <a href="{{ options_url('type', 'photos', null, {'p': null}) }}" class="{{ html_classes({'active': criteria.getOption('type') == 'photos' }) }}">
+                        <i class="fa-regular fa-file-image" aria-hidden="true"></i> &nbsp; {{ 'photos'|trans }}
+                    </a>
+                </li>
+                <li>
+                    <a href="{{ options_url('type', 'videos', null, {'p': null}) }}" class="{{ html_classes({'active': criteria.getOption('type') == 'videos'}) }}">
+                        <i class="fa-regular fa-file-video" aria-hidden="true"></i> &nbsp; {{ 'videos'|trans }}
+                    </a>
+                </li>
+            </ul>
```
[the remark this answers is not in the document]
andrewmoise: Done

andrewmoise:
Hey, so I talked a bit to Benti on Matrix -- I did pull some of those common functions out of EntryFrontController::front() and EntryFrontController::magazine(), so it's now even more consolidated than the consolidation that I already did from the eight different copy-pasted functions it was before I came along.

It's good in that it's cleaner, but now it'll probably want at least another short round of testing to make sure there are no new problems with it as a result. Long story short I think we're in agreement about trying to get this in as opposed to more general fixes; I think I've addressed everything on the list above, so if there are no specific problems with this after people have given it some looking-over and testing, I think it should be ready to go in.

nobodyatroot:
Been running the latest code for a little while now and all I/we see (thanks @e-five256) is a Javascript error in the console that seems to be related to the PR:

```javascript
894.9b1cc6c2.js:2 Error connecting controller

TypeError: Cannot read properties of null (reading 'scrollWidth')
    at e.value (862.d8aa2795.js:1:2845)
    at e.value (862.d8aa2795.js:1:2488)
    at N.connect (894.9b1cc6c2.js:2:26680)
    at z.connectContextForScope (894.9b1cc6c2.js:2:29482)
    at 894.9b1cc6c2.js:2:37072
    at Array.forEach (<anonymous>)
    at G.connectModule (894.9b1cc6c2.js:2:37058)
    at G.loadDefinition (894.9b1cc6c2.js:2:35915)
    at 894.9b1cc6c2.js:2:38874
    at Array.forEach (<anonymous>)

Object
handleError @ 894.9b1cc6c2.js:2
```

```javascript
key: "handleOptionsBarScroll",
value: function() {
    var e = document.getElementById("options");
    if (e) {
        var t = e.clientWidth
          , n = e.querySelector(".options__main");
        if (n.scrollWidth > t && !n.nextElementSibling) {
            e.insertAdjacentHTML("beforeend", '<menu class="scroll"><li class="scroll-left"><i class="fa-solid fa-circle-left"></i></li><li class="scroll-right"><i class="fa-solid fa-circle-right"></i></li></menu>');
            var r = e.querySelector(".scroll-left")
              , o = e.querySelector(".scroll-right")
              , i = e.querySelector(".options__main");
            o.addEventListener("click", (function() {
                i.scrollLeft += 100
            }
            )),
            r.addEventListener("click", (function() {
                i.scrollLeft -= 100
            }
            ))
        }
    }
}
```

And the federated filter also seems to pull in local users?

image

templates/entry/_options.html.twig
```diff
@@ -1,136 +1,244 @@
...
-               class="{{ html_classes({'active': route_has_param('sortBy', 'active')}) }}">
-                {{ 'active'|trans }}
-            </a>
-        </li>
-        <li>
-            <a href="{{ options_url('sortBy', 'commented') }}"
-               class="{{ html_classes({'active': route_has_param('sortBy', 'commented')}) }}">
-                {{ 'commented'|trans }}
-            </a>
+    <menu class="options__filter">
```
BentiGorlich, line 5: each html element with the id `options` needs to have a sub element with the class `options__main` for our auto scroll logic to work. @nobodyatroot that is the cause of the javascript error I think

templates/post/_options.html.twig
```diff
@@ -1,94 +1,262 @@
...
-               class="{{ html_classes({'active': route_has_param('sortBy', 'active')}) }}">
-                {{ 'active'|trans }}
-            </a>
-        </li>
-        <li>
-            <a href="{{ options_url('sortBy', 'commented') }}"
-               class="{{ html_classes({'active': route_has_param('sortBy', 'commented')}) }}">
-                {{ 'commented'|trans }}
-            </a>
+    <menu class="options__filter">
```
BentiGorlich, line 5: same problem here, the class `options__main` needs to be added

BentiGorlich:
The federation "problem" did exist beforehand, you just couldn't only display federated content I think. I'd say this is fine for now, but if someone wanted to fix it, an if has to be added to here: https://github.com/MbinOrg/mbin/blob/c5ab3809cbaaeadf12840b53874136a651e86b28/src/Repository/EntryRepository.php#L138-L140
atm it only adds a where clause for `Criteria::AP_LOCAL`

e-five256:
> The federation "problem" did exist beforehand, you just couldn't only display federated content I think. I'd say this is fine for now, but if someone wanted to fix it, an if has to be added to here:

imo let's just add `{# #}` around the federated option for entry/post template to comment it out for now and handle implementing it in another PR, this one is already big enough

templates/post/_options.html.twig
e-five256: Looking at it, we might be fine as is as, and I could be wrong, I can't get the content to take up enough space to require a scroll bar

instead we could add

```js
         if (container) {
             const containerWidth = container.clientWidth;
             const area = container.querySelector('.options__main');
+
+            if (null === area) {
+                return;
+            }
```

which will stop the js error as it won't attempt to make the option bar a scrollbar unless it's set up with those classes (maybe we can name it better, I see @andrewmoise removed `no-scroll` class and it does sort of seem like it might've been unused, unless I was missing it, so maybe we could make a class specifically for when a scroll bar should be introduced

but I think we should make more things drop downs, the user profile which has a ton of options being one of them, which is already something KES had

src/Controller/Entry/Comment/EntryCommentFrontController.php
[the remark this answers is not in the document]
e-five256: Ah, I missed this conversation. I had brought this up before about removing this route, but I actually do end up using it swapped to local to check on local user issues, so not sure about removing it completely. It looks like it still functions routing wise, just the filters aren't quite right anymore but the url can still be modified to add them, perhaps it needs to be rethought or moved into the admin section or something

e-five256:
the only thing @nobodyatroot and I were discussing is that `def` might be unclear to users as to what it means. I think the intent was `default` (correct me if I'm wrong), and functionally what it shows is what the user selects as their "homepage" in user settings. @nobodyatroot recommended that maybe `home` would make it more clear what its relation to the user setting was. Before making any changes though, want to check with @andrewmoise and others if that made sense, as we don't want to change URLs more than we need to and mess with peoples' bookmarks

templates/post/_options.html.twig
BentiGorlich: sure we can do that as well

e-five256 (approved):

nobodyatroot (approved):
Approved with intention to follow up with PR to fix minor issues.

e-five256:
@andrewmoise I think we're all in agreement of getting these changes in, but want to make sure you're still good with it being merged as well. The only outstanding changes are all minor nitpicks (can take a look at the suggestions [here if you want](https://github.com/MbinOrg/mbin/commit/acc32e8e6182efb96ef3f5f59551c0c4b031f72b)) which, at least in my opinion shouldn't hold up merging, but everyone was hoping to get your opinion first

andrewmoise:
@e-five256 I'm gonna keep it real with you -- this whole PR has been unexpectedly extensive with a bunch of extra effort spent, and as of late Feb / early March I've pretty much moved on from wanting to do that big UX plan I was talking about originally. It seems like at this pace it'll take at least a year maybe two with at least 5-6 big PR cycles like this (probably more in practice) to get the nice polished type of UI that I had envisioned originally, and I don't want to sign up for that.

So to me the first question with any change request is not "would this be a nice idea to do" but "should we spend time on this instead of potentially some other change, and if so, does it need to go in this PR specifically." By focusing only on the first question it's easy to eat up basically unlimited time and effort, and it feels like I've been repeatedly getting pulled into issues of the first type, and in repeated review cycles instead of just a clean organized "hey here's the change" "okay here's everything that needs to change with it" (cycle maybe repeats one time) "okay done let's merge it."

For the detail of outstanding issues at this point, my feeling is that at least the null properties and the federation filter are impactful issues with easy fixes, so worth doing either in this PR or directly thereafter. But also, this whole thing is pretty much on the back burner for me at this point, so if you're comfortable with putting it in as-is then I'm good with that.

BentiGorlich:
Yeah I know I kinda screwed that up and I am sorry for that.
I think nobody thinks that you have the one to finish this, we just wanted to make surethat you're fine with your code being merged in our repo.

e-five256:
I think I've said this a few times lately but just to repeat it here, we value any contributions to the project, doesn't matter how large or small. You definitely should not feel obligated to having to spend a lifetime working on things here, everyone is free to come and go as they see fit, and once again, any time spent on anything: reviews, feedback, code, etc, are always appreciated.

Everyone had positive feedback about these changes to the filtering setup, making it easier and fixing some annoyances with how things worked. I'll make sure to get this merged soon, and then look over your comments on next steps and see what can be added / what I can add issues for, as providing a good UX experience for our users is very much a concern of mine.

PR #465, opened 2024-01-30
make search recognize `!group@instance.domain` handle format

asdfzdfj:
problem raised here: https://moist.catsweat.com/m/Mdev@kbin.run/t/133545

adjust search to recognize lemmy styled group handle `!group@instance.domain` and proceed to search without having to edit the handle before searching

also a bit of niceties: if this group handle format is used when searching, then the users will be excluded from search results

### DIFF
--- a/src/ActivityPub/ActorHandle.php
+++ b/src/ActivityPub/ActorHandle.php
@@ -6,11 +6,11 @@ namespace App\ActivityPub;
 
 class ActorHandle
 {
-    public const HANDLE_PATTERN = '/^(?P<prefix>[@!])?(?P<user>[\w\-\.]+)@(?P<host>[\w\.\-]+)(?P<port>:[\d]+)?$/';
+    public const HANDLE_PATTERN = '/^(?P<prefix>[@!])?(?P<name>[\w\-\.]+)@(?P<host>[\w\.\-]+)(?P<port>:[\d]+)?$/';
 
     public function __construct(
         public ?string $prefix = null,
-        public ?string $username = null,
+        public ?string $name = null,
         public ?string $host = null,
         public ?int $port = null,
     ) {
@@ -26,7 +26,7 @@ class ActorHandle
         if (preg_match(static::HANDLE_PATTERN, $handle, $match)) {
             $new = new static(
                 $match['prefix'] ?? null,
-                $match['user'],
+                $match['name'],
                 $match['host']
             );
             $new->setPort($match['port'] ?? null);
@@ -40,7 +40,7 @@ class ActorHandle
     public static function isHandle(string $handle)
     {
         if (preg_match(static::HANDLE_PATTERN, $handle, $matches)) {
-            return !empty($matches['user']) && !empty($matches['host']);
+            return !empty($matches['name']) && !empty($matches['host']);
         }
 
         return false;
@@ -57,9 +57,7 @@ class ActorHandle
         return !empty($this->port) ? ':'.$this->port : '';
     }
 
-    /**
-     * @param int|string|null $port port as either plain int or string formatted like ':9000'
-     */
+    /** @param int|string|null $port port as either plain int or string formatted like ':9000' */
     public function setPort(int|string|null $port)
     {
         if (\is_string($port)) {
@@ -71,14 +69,16 @@ class ActorHandle
         return $this;
     }
 
+    /** @return string the handle's domain and optionally port in the format `host[:port]` */
     public function getDomain(): string
     {
         return $this->host.$this->getPortString();
     }
 
+    /** @param ?string $domain the domain in the format `host[:port]` to set both handle's host and port */
     public function setDomain(?string $domain)
     {
-        $url = parse_url($domain);
+        $url = parse_url($domain ?? '');
 
         $this->host = $url['host'] ?? null;
         $this->port = $url['port'] ?? null;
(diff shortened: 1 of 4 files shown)


melroy89:
Yea there were quite some discussions about this. Nice to see this support in mbin!

src/ActivityPub/ActorHandle.php
```diff
@@ -0,0 +1,111 @@
...
+        return $this->host.$this->getPortString();
+    }
+
+    public function setDomain(?string $domain)
+    {
+        $url = parse_url($domain);
+
+        $this->host = $url['host'] ?? null;
+        $this->port = $url['port'] ?? null;
+
```
melroy89, line 85: Is null a valid default? Or 443?

src/ActivityPub/ActorHandle.php
```diff
@@ -0,0 +1,111 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\ActivityPub;
+
+class ActorHandle
+{
+    public const HANDLE_PATTERN = '/^(?P<prefix>[@!])?(?P<user>[\w\-\.]+)@(?P<host>[\w\.\-]+)(?P<port>:[\d]+)?$/';
```
melroy89, line 9: I hope this works well 🤗. Maybe you can add a unit test to validate this regex on various strings?

src/ActivityPub/ActorHandle.php
```diff
@@ -0,0 +1,111 @@
...
+    {
+        return $this->formatWithPrefix($this->prefix);
+    }
+
+    public static function parse(string $handle): ?static
+    {
+        if (preg_match(static::HANDLE_PATTERN, $handle, $match)) {
+            $new = new static(
+                $match['prefix'] ?? null,
+                $match['user'],
```
melroy89, line 29: It's not really a `user` right? It's a group in AP or community. Maybe change the regex `<user>`

melroy89: Ow wait you gonna use this for user? No.. you shoudn't. `!` syntax is used for finding the community or group.

src/ActivityPub/ActorHandle.php
```diff
@@ -0,0 +1,111 @@
...
+        return $this->formatWithPrefix('');
+    }
+
+    /** @return string handle in the form @user@domain */
+    public function atHandle(): string
+    {
+        return $this->formatWithPrefix('@');
+    }
+
+    /** @return string handle in the form !user@domain */
```
melroy89, line 106: This should be used for: `!group@domain`

src/ActivityPub/ActorHandle.php
asdfzdfj: I plucked this regex from the webfinger part and modify a bit to also recognize and extract additional prefix part, so it should be fine for the time being

a bit of unit tests to check this wouldn't hurt but then I'll need those samples to make it, also it's out of scope for this patch, feel free to file a patch

src/ActivityPub/ActorHandle.php
asdfzdfj: it's the username part of the handle, whether it identify a user, group or other kinds of actor that can be expressed in the format of `username@domain[:port]`

perhaps I could at least rename this to just `name` to avoid confusion and avoid implying extra connection between the handle and what kind of actor it probably points to

src/ActivityPub/ActorHandle.php
asdfzdfj: default `null` is fine, it means that there's no port number after the host part i.e. the handle it holds is just `username@domain` and not `username@domain:443`

this class is first and foremost a string manipulation tool, it does not know or care that https is port 443 and that should be the default

src/ActivityPub/ActorHandle.php
asdfzdfj: as far as the intention of this class is concerned, the `user` in the doctring example corresponds to the `user` field of the `ActorHandle` object, and doesn't mean the handle it holds should points to a specific kind of actor

perhaps the above applies: rename the field `user` -> `name` to avoid this confusion and nonexistent implications

asdfzdfj (commented):
guess I should make a bit of clarifications regarding the `ActorHandle` class:
- its main job is about _**string** processing tool_ on piece of text that supposed to be an actor identifier, be it for user or groups or something else, as long as it's represeinted in the `[prefix]username@domain[:port]` format
- the class makes no extra connotation between the handle it contained and what kind of actor this handle it probably points to, that is up to the caller, and so it was made to be able to handle both `@user@domain` and `!group@domain` alike (2 kinds of prefixed handle it recognize, also the plain unprefixed `username@domain` too)
- although the class is introduced in this patch, it's not just for this purpose. I planned to replace the existing handle string manipulation that I saw scattered around in the code base, all slightly different, with this class which should help in consistency and ease of editing

---

and maybe a bit on **my perception** of kbin/mbin and where I stand regarding lemmy interop, since it influences the implementation of this patch:
the way I saw/understands kbin/mbin is NOT "php lemmy with microblogs", and I'd rather not want to see mbin (or kbin for that matter) gets turned into "php lemmy with microblogs", either by my own hands or others.

when people look at the threadiverse corner, I want people to see that's there's lemmy, kbin, piefed, and whatever new things that's yet to come, and not lemmy, php lemmy with microblogs, python lemmy and something that's ultimately just a different implementation of lemmy.
much like how you might see mastodon, pleroma, and misskey when looking in the microblogging corner, and not mastodon, elixir mastodon and mastodon but fancy and on steroids.

melroy89 (dismissed):

src/ActivityPub/ActorHandle.php
asdfzdfj: well in any case I added a small unit tests for this, if you have any more interesting handle samples feel free to add it there

nobodyatroot (approved):

PR #467, opened 2024-01-31
add user badges to entry and post templates

e-five256:
this is an addition to the user badges shown on comments and user profile pages

these are useful for comments, but for users that may never post comments such as bots that only post entries or posts, there was very little clarity on the fact that it's a bot doing the posting without visiting their profile page

I figured while doing bots, might as well add mod/admin as well as I think it's useful information to be clear about, and is similar to how other sites show the information

before

image
image

after

image
image

### DIFF
--- a/assets/styles/components/_entry.scss
+++ b/assets/styles/components/_entry.scss
@@ -291,6 +291,13 @@
     align-items: center;
     column-gap: 0.25rem;
 
+    .user-badge {
+      border: var(--kbin-section-border);
+      padding: 0.25rem 0.5rem;
+      margin-left: .25rem;
+      border-radius: var(--kbin-rounded-edges-radius);
+    }
+
     .edited {
       font-style: italic;
     }
--- a/assets/styles/components/_entry.scss
+++ b/assets/styles/components/_entry.scss
@@ -291,13 +291,6 @@
     align-items: center;
     column-gap: 0.25rem;
 
-    .user-badge {
-      border: var(--kbin-section-border);
-      padding: 0.25rem 0.5rem;
-      margin-left: .25rem;
-      border-radius: var(--kbin-rounded-edges-radius);
-    }
-
     .edited {
       font-style: italic;
     }
--- a/templates/components/entry.html.twig
+++ b/templates/components/entry.html.twig
@@ -78,7 +78,16 @@
             {% endif %}
             <aside class="meta entry__meta">
                 <span>{{ component('user_inline', {user: entry.user, showAvatar: SHOW_USER_AVATARS is same as V_TRUE}) -}}</span>
-                <span>,</span>
+                {% if entry.user.admin() %}
+                    <span class="user-badge">{{ 'user_badge_admin'|trans }}</span>
+                {% elseif entry.user.moderator() %}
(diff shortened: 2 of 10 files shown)


nobodyatroot (dismissed):

nobodyatroot:
One interesting thing I noticed... this L4s account is a bot AND a mod of these mags and the `if` ordering just so happens to show `Mod`.... not sure if the order should be moved so that it evaluates bot first?

image

image

assets/styles/components/_entry.scss
```diff
@@ -291,6 +291,13 @@
     align-items: center;
     column-gap: 0.25rem;
 
+    .user-badge {
+      border: var(--kbin-section-border);
+      padding: 0.25rem 0.5rem;
+      margin-left: .25rem;
+      border-radius: var(--kbin-rounded-edges-radius);
+    }
```
BentiGorlich, line 299: instead of copying this section we should make it global shouldn't we?

templates/components/entry.html.twig
```diff
@@ -78,7 +78,16 @@
...
-                <span>,</span>
+                {% if entry.user.admin() %}
+                    <span class="user-badge">{{ 'user_badge_admin'|trans }}</span>
+                {% elseif entry.user.moderator() %}
+                    <span class="user-badge">{{ 'user_badge_global_moderator'|trans }}</span>
+                {% elseif entry.magazine.userIsModerator(entry.user) %}
+                    <span class="user-badge">{{ 'user_badge_moderator'|trans }}</span>
+                {% elseif (entry.user.type) == "Service" %}
+                    <span class="user-badge">{{ 'user_badge_bot'|trans }}</span>
+                {% endif %}
```
BentiGorlich, line 89: This sounds like the bot badge is not shown if the user is one of the previous. But it imo it should be shown

BentiGorlich (changes requested):

BentiGorlich:
> One interesting thing I noticed... this L4s account is a `Bot` AND a `Mod` of these mags and the `if` ordering just so happens to show `Mod` first.... not sure if the order should be rearranged so that it checks for a `Bot` first? 

Imo it should show both, not just one of them. Currently only one of them is shown and "Mod" has precedence over "bot"

templates/components/entry.html.twig
e-five256: changed

assets/styles/components/_entry.scss
e-five256: I've never seen a good place for global css, if you have any suggestions let me know. CSS is divided into components, mixins, and variables. components only have css that effect their single component. layout, similar, only have things that effect the class they're named after (e.g. `_options.scss` starts with `.options {` so only effects things under that

I could instead change this to a mixin so it isn't repeated, but that will still increase the file size by the same amount this copy paste does

e-five256:
separated out the check for bot from permission level similar to op, as a user can be both a bot and mod / global mod / admin

of course a user can also likely be a mod, global mod, and admin all at the same time but as the permission set is inclusive of the previous steps doesn't seem like that needs to be bubbled up

image

(after making the bot account a global mod to test)

image
image

BentiGorlich (approved):

nobodyatroot (approved):

PR #469, opened 2024-02-04
Make usernames case insensitive everywhere

thepaperpilot:


asdfzdfj (dismissed):
give it a shot on my instance and seems to work fine? but then again all I have is just single user instance so it's a rather inadequate data point

@nobodyatroot @BentiGorlich any second opinions just in case you were testing this out?

migrations/Version20240204141515.php
```diff
@@ -0,0 +1,31 @@
...
+    public function up(Schema $schema): void
+    {
+        $this->addSql('CREATE EXTENSION IF NOT EXISTS citext');
+        $this->addSql('ALTER TABLE "user" ALTER COLUMN username TYPE citext');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE "user" ALTER COLUMN username TYPE VARCHAR(35)');
+        $this->addSql('DROP EXTENSION citext');
```
BentiGorlich, line 29: I am not sure if we should drop the extension here. It doesn't hurt to have it and we don't know if we added it in the upper statement, maybe it already existed

BentiGorlich (changes requested):
Did you check whether doctrine thinks if needs fo change the db schema after this change? If not, you need to run `phpbin/console migrations:diff` (I think this is the correct one). If it generated anz changes we nees to tell doctrine about this change in some way

migrations/Version20240204141515.php
melroy89: but it's part of `down` right? You clean it up again since on line 22 you add it on `up`.

BentiGorlich: Well the `up` contains a `IF NOT EXISTS`, so it might have already been present on the machine. I think it could break some pg setups (if this migration is reverted) if `citext` is needed somewhere else (not in mbin), and was therefore already installed...

melroy89: Ah yes. I got it.

BentiGorlich: @thepaperpilot if you remove this line I'll approve, the rest is looking good to me

BentiGorlich (approved):

BentiGorlich:
Problem here is: I thought we already had a unique key on the lowercase usernames, but we don't... We have an index for the lower case usernames: user_username_lower_idx, but that does not make them unique...

sorry @thepaperpilot don't really know what we are able to do about it...

asdfzdfj:
personally, I think it should be fine for newly created instance (and for @thepaperpilot probably) to use `citext` username column, since there isn't much of data to conflict, and you also get to enforce the unique case insentive constraint on the column too. better to enforce this early rather than leave it to fester long enough to get what's similar to fedia's database state when it was migrating from kbin to mbin

however, we'll likely need to come up with a plan for existing deplyments' on how to deal with this new constraints is enforced from the column type change, like what originalucifer encountered in #502.

I don't think it's a bug on this patch per se, but rather we don't have enough pieces to ensure smooth migration on this column change

maybe we can try to merge this again at a late date once we also have the contingency plan for this (maybe we get to make a first potentially breaking changes with this, yay....)

PR #488, opened 2024-02-13
hotfix: allow custom style while still in 2fa

asdfzdfj:
this is a hotfix of a problem where users with 2fa enabled will get wrongly redirected to `/custom-style` custom style route after completing 2fa, most likely because this route will be loaded after the challenge page was loaded as part of the base template, and that seems to confuse 2fa system into misremembering that the custom style route is where the users came from

by allowing public access it should now be accessible while 2fa is in progress and not trip the aforementioned 2fa route redirection

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -58,6 +58,8 @@ security:
         # This makes the logout route accessible during two-factor authentication. Allows the user to
         # cancel two-factor authentication, if they need to.
         - { path: ^/logout, role: PUBLIC_ACCESS }
+        # allow custom style route access during two-factor authentication
+        # to avoid redirecting (wrongly) to this route after two-factor authentication is completed
         - { path: ^/custom-style, role: PUBLIC_ACCESS }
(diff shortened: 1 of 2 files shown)


config/packages/security.yaml
```diff
@@ -58,6 +58,7 @@ security:
         # This makes the logout route accessible during two-factor authentication. Allows the user to
         # cancel two-factor authentication, if they need to.
         - { path: ^/logout, role: PUBLIC_ACCESS }
+        - { path: ^/custom-style, role: PUBLIC_ACCESS }
```
BentiGorlich, line 61: can you put a comment here why this is needs to be publicly accessible?

asdfzdfj: alright

asdfzdfj: added

BentiGorlich (approved):

asdfzdfj:
@nobodyatroot can you confirm that this fix works?

melroy89 (approved):

PR #491, opened 2024-02-15
Using loading lazy on images

melroy89:
I apply `loading="lazy"` on all images, except on the logo and form images like 2FA images, etc. Feel free to test it.

I see a first load image reduction from about 1.6MB to only 185kB:

image

DOM content loaded in 300-400 ms.

----

It's already running live at: https://kbin.melroy.org


nobodyatroot (dismissed):
I see a 40% improvement in DOM loading time, definitely something to change.

nobodyatroot (approved):

PR #500, opened 2024-02-16
Add the correct cascading attribute to Entities

BentiGorlich:
- add `cascade: ['persist', 'remove']` where it is needed
- add `onDelete: 'CASCADE'` where it is needed

users and magazines can now be deleted with just one SQL command. However because we cache a lot of count values, they will be inaccurate until someone posts, votes, subscribes, etc. depending on the counter


melroy89 (approved):
I assume the migration is generated based on the php class changes, right?

melroy89:
Ps. This might has impact on the existing code of https://github.com/MbinOrg/mbin/pull/438, which then can be cleaned-up in some cases maybe.

BentiGorlich:
> I assume the migration is generated based on the php class changes, right?

Yes, thats why I left the comment in the migration. I did check the statements, but I didn't write them :D

Issue #457, opened 2024-01-23
Image preview via icon uses wrong source


BentiGorlich:
**Describe the bug**
When previewing an image via the little image icon the image is loaded from the original source.
In comparision when zoh open an kmage via clicking the small thumbnail inside of an entry the cached image is beeing used as the source

**On which Mbin instance did you find the bug?**
[gehirneimer.de]

**To Reproduce**
1. Go to any post with an image
2. Click on the little image icon to expand the preview
3. look at the source
4. compare it to the source thats being used when clicking on the thjmbnail of the image (not the preview)

**Expected behavior**
We are caching the image, so the cache should be used

PR #501, opened 2024-02-17
Improve handling of images

### LINKED ISSUE #498: Posts with too large images get created as articles
**Describe the bug**
When a remote instance sends a post with an attached image, but the image is too large for our server, it gets created as an article and not as an image. That means, that there is no thumbnail and that there is no preview icon in the post. Basically there is no way of seeing the attached image, other than to go to the original instance

[kbin.melroy.org, fedia.io, gehirneimer.de, rimworld.gallery]

**To Reproduce**

**Expected behavior**
If an image is too big to save it, it should eigher be converted to something smaller or it should not get created at all and instead it should just use / link to the original image

**Screenshots**

**Additional context**
rimworld.gallery uses very high quality images which should get picked up correctly...
@TheVillageGuy feel free to add more information

BentiGorlich:
- when an attached image is too large, safe a reverence (`filePath` -> `null`) instead of discarding the image.
- `DeleteImageMessage` now uses the id as parameter
- fix `sourceUrl` being overwritten to `null` by `AttachEntryEmbedHandler`
- adjust twig function `uploaded_asset` (`MediaExtensionRuntime.getPublicPath`) to use the source url if there is no file path
- fix preview button using the entry url before the stored image url
- improve the exception logging by not catching any exception in `ImageManager.store` and just rethrowing it. That removed the exception type as a possible indicator as to what went wrong

This fixes #498 and #457

### DIFF
--- a/migrations/Version20240217103834.php
+++ b/migrations/Version20240217103834.php
@@ -11,7 +11,7 @@ final class Version20240217103834 extends AbstractMigration
 {
     public function getDescription(): string
     {
-        return '';
+        return 'This will make the file_path nullable, so we can store links to remote images, which could not be cached locally';
     }
 
     public function up(Schema $schema): void
--- a/migrations/Version20240217103834.php
+++ b/migrations/Version20240217103834.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20240217103834 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE image ALTER file_path DROP NOT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE image ALTER file_path SET NOT NULL');
+    }
+}
--- a/src/MessageHandler/AttachEntryEmbedHandler.php
+++ b/src/MessageHandler/AttachEntryEmbedHandler.php
@@ -57,7 +57,9 @@ class AttachEntryEmbedHandler
 
         $entry->type = $type;
         $entry->hasEmbed = $html || $isImage;
-        $entry->image = $cover;
+        if ($cover) {
+            $entry->image = $cover;
+        }
 
         $this->entityManager->flush();
     }
--- a/src/Repository/ImageRepository.php
+++ b/src/Repository/ImageRepository.php
@@ -95,7 +95,6 @@ class ImageRepository extends ServiceEntityRepository
 
         try {
             $this->imageManager->store($source, $filePath);
-            $this->logger->debug('stored the image');
 
             return $image;
         } catch (ImageDownloadTooLargeException $e) {
--- a/src/Repository/ImageRepository.php
+++ b/src/Repository/ImageRepository.php
@@ -6,6 +6,7 @@ namespace App\Repository;
 
 use App\Entity\Image;
 use App\Event\ImagePostProcessEvent;
+use App\Exception\ImageDownloadTooLargeException;
 use App\Service\ImageManager;
 use App\Utils\ImageOrigin;
 use Doctrine\Bundle\DoctrineBundle\Repository\ServiceEntityRepository;
@@ -60,10 +61,8 @@ class ImageRepository extends ServiceEntityRepository
     /**
      * Process and store an image from source file given path.
      *
-     * @param source file path of the image
-     * @param origin where the image comes from
-     *
-     * @throws RuntimeException if image type can't be identified
+     * @param string      $source file path of the image
+     * @param ImageOrigin $origin where the image comes from
      */
     private function findOrCreateFromSource(string $source, ImageOrigin $origin): ?Image
     {
@@ -75,6 +74,8 @@ class ImageRepository extends ServiceEntityRepository
             if (file_exists($source)) {
                 unlink($source);
             }
+
+            $this->logger->debug('found image by Sha256, imageId: {id}', ['id' => $image->getId()]);
 
             return $image;
         }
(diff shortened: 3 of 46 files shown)


BentiGorlich:
main problem with this is: if there is an entry without a local image (because it was too large) we do not have a thumb_image of that. If anybody has an idea how to get one regardless, please tell me :sweat_smile:

TheVillageGuy:
> main problem with this is: if there is an entry without a local image (because it was too large) we do not have a thumb_image of that. If anybody has an idea how to get one regardless, please tell me 😅

Make the thumb before checking filesize?

BentiGorlich:
After looking into the docs and 2 PRs (https://github.com/liip/LiipImagineBundle/issues/328 and https://github.com/liip/LiipImagineBundle/issues/1464) I think it is fair to say that it will not work like that. We can't just generate a thumbnail for remote images with liip_imagine.
My go to solution would be to compress the image and safe it afterwards. We add a new column `is_compressed`, so we can display the original full res image from the source url when it is needed and we know that we don't have the original full res image at hand

melroy89:
> My go to solution would be to compress the image and safe it afterwards. We add a new column `is_compressed`, so we can display the original full res image from the source url when it is needed and we know that we don't have the original full res image at hand

I think there is currently indeed no better alternative. Especially since LiipImagineBundle isn't helping here. However, keep in mind not all servers will allow you to load image / content from other domains, depending on the CORS header settings.

BentiGorlich:
Ok, then I will create a follow-up issue for that. I think it is best to separate it from this PR as this will probably be a little bit more code

BentiGorlich:
> However, keep in mind not all servers will allow you to load image / content from other domains, depending on the CSP header settings.

I don't think that this is a big problem, because Lemmy always let the user load images from the source URL, it does not safe a cached version of the image... 
Maybe we could offer 2 preview buttons if we know we only have a compressed image, one for our copy and one for the original which may not load due to CSRF settings.

BentiGorlich:
I have it running on https://gehirneimer.de now

nobodyatroot (dismissed):
Running on kbin.run as well, no issues.

melroy89:
TheVillageGuy would be eternally grateful to you :)

TheVillageGuy:
> TheVillageGuy would be eternally grateful to you :)

Indeed, I'm looking forward to this being released. Great work @BentiGorlich

BentiGorlich:
Can someone take a look at the code as well? I am fairly confident that this is working, but I don't want it to land on main without someone else looking at the code first :D

nobodyatroot:
> Can someone take a look at the code as well? I am fairly confident that this is working, but I don't want it to land on main without someone else looking at the code first :D

Nothing stands out to me, but then again... I have no idea what I'm doing. ;-)

TheVillageGuy:
> Can someone take a look at the code as well? I am fairly confident that this is working, but I don't want it to land on main without someone else looking at the code first :D

Looks good to me. What happens if it is a null but there is no reference? Could this somehow occur, like if some other exception was triggered instead of the file being too large? How will the front end handle that?

src/MessageHandler/AttachEntryEmbedHandler.php
```diff
@@ -57,7 +57,9 @@ public function __invoke(EntryEmbedMessage $message): void
 
         $entry->type = $type;
         $entry->hasEmbed = $html || $isImage;
-        $entry->image = $cover;
+        if ($cover) {
```
melroy89, line 60: Above we already have a `!$cover &` check, this this not redundant?

src/Repository/ImageRepository.php
```diff
@@ -94,20 +95,36 @@ private function findOrCreateFromSource(string $source, ImageOrigin $origin): ?I
 
         try {
             $this->imageManager->store($source, $filePath);
+            $this->logger->debug('stored the image');
```
melroy89, line 98: you still need this debug line? If not, you might want to remove it again. Sure, it's behind debug(), but it can get messy fast.

src/Service/EntryCommentManager.php
```diff
@@ -159,7 +159,7 @@ public function purge(User $user, EntryComment $comment): void
         $this->dispatcher->dispatch(new EntryCommentBeforePurgeEvent($comment, $user));
 
         $magazine = $comment->entry->magazine;
-        $image = $comment->image?->filePath;
+        $image = $comment->image?->getId();
```
melroy89, line 162: `image?` doesn't that mean that `image` valuable can be empty? Meaning `getId()` will cause errors in some cases (null error).

src/Service/EntryManager.php
```diff
@@ -215,7 +219,7 @@ public function purge(User $user, Entry $entry): void
     {
         $this->dispatcher->dispatch(new EntryBeforePurgeEvent($entry, $user));
 
-        $image = $entry->image?->filePath;
+        $image = $entry->image?->getId();
```
melroy89, line 222: Same question as above, here `image` can be null as well, right?

src/Service/ImageManager.php
```diff
@@ -46,6 +48,10 @@ public static function isImageType(string $mediaType): bool
         return \in_array($mediaType, self::IMAGE_MIMETYPES);
     }
 
+    /**
+     * @throws \Exception                     if the file could not be found
+     * @throws ImageDownloadTooLargeException
```
melroy89, line 53: I think you don't throw `ImageDownloadTooLargeException` this anymore here..

melroy89:
> Can someone take a look at the code as well? I am fairly confident that this is working, but I don't want it to land on main without someone else looking at the code first :D

Sure I will look into your PR.

src/Service/PostManager.php
```diff
@@ -163,7 +166,7 @@ public function purge(User $user, Post $post): void
     {
         $this->dispatcher->dispatch(new PostBeforePurgeEvent($post, $user));
 
-        $image = $post->image?->filePath;
+        $image = $post->image?->getId();
```
melroy89, line 169: Same here with `image` null

migrations/Version20240217103834.php
```diff
@@ -0,0 +1,26 @@
...
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20240217103834 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
```
melroy89, line 14: Add description of what this will bring?

melroy89:
@BentiGorlich OK, reviewed ;)

src/MessageHandler/AttachEntryEmbedHandler.php
BentiGorlich: no it is not, because it is enough if one of the other two is set for it to overwrite the image even if there is not a cover

src/Repository/ImageRepository.php
BentiGorlich: true

src/Service/EntryCommentManager.php
BentiGorlich: `$image?->...` is the same as `image?.something` in c# for example. Meaning it will only call the `something` part if `image` is not null. https://wiki.php.net/rfc/nullsafe_operator. `image` can be `null`, but it is checked for that before it is used

src/Service/EntryManager.php
BentiGorlich: yes it can, but it is only used once and that is behind an `if ($image)`

src/Service/ImageManager.php
BentiGorlich: true

src/Service/PostManager.php
BentiGorlich: same answer as above

migrations/Version20240217103834.php
BentiGorlich: Will do

src/Service/EntryManager.php
melroy89: Yes ok. we are safe here.

src/Service/EntryCommentManager.php
melroy89: Yes ok. we are safe here.

src/MessageHandler/AttachEntryEmbedHandler.php
melroy89: ah you're correct, these are `&&` checks. My bad.

melroy89 (approved):

PR #503, opened 2024-02-17
Add column `last_origin_update` to magazine

BentiGorlich:
- magazines now have an additional property `last_origin_update` for when the last update came directly from the origin

this is acting as the base for #436


nobodyatroot (approved):
Looks good to me

melroy89 (approved):

PR #509, opened 2024-02-18
add remote url image detection for incoming images

asdfzdfj:
adjust incoming image detection to also try and detect if the given url leads to an image by downloading and inspecting it with symfony mime type guesser.

because sometimes a valid url to image doesn't end with image extension and the path is mostly generated (e.g. uuid filename), in such case there's no good way to check if it's actually an image besides actually inspecting the contents

this can be a relatively expensive operation and I can only hope that most of the url can be detected by preceding checks, and that this check doesn't get called too often

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -662,10 +662,15 @@ class ActivityPubManager
             return false;
         }
 
-        // attachment is either:
-        // - has `mediaType` field and is a recognized image types
-        // - image url looks like a link to image
+        /*
+         * attachment is either:
+         * - has `mediaType` field and is a recognized image types
+         * - image url looks like a link to image
+         * - detected mime type from the image file is recognized image type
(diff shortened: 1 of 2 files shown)


melroy89:
I hope bentigorlich can also check this to avoid regression on his image related pr.

BentiGorlich (changes requested):
Since you made the call possibly a lot more expensive, I'd suggest to change the `handleImages` method a bit. We only use the first image we find, so it does not make sense to iterate through all attachments and test whether it is actually an image. I mean there is `todo` to support multiple images, but I don't see that happen any time soon, so I think for now it is safe to ignore that

asdfzdfj:
noted, in that case I think I'll close this PR for now, as I was also thinking to move this download check to somewhere else, perhaps to be within the image download function, and then maybe change how an image attachment is detected instead.

PR #510, opened 2024-02-18
Update Symfony packages + recipes+ NPM

melroy89:
Before v1.4.0 release, I would like to update the Symfony package minor releases + recipes.

PR #512, opened 2024-02-20
Update remaining composer packages

melroy89:
Update some of the remaining composer packages. I don't update all doctrine packages for now (see: https://github.com/doctrine/orm/issues/11277). 

Also update php-cs-fixer in the tools directory.

Follow-up PR from: https://github.com/MbinOrg/mbin/pull/510

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -15098,16 +15098,16 @@
         },
         {
             "name": "fakerphp/faker",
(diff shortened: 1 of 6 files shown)


nobodyatroot (approved):

melroy89:
I was running this for the past day without any issues as well..

PR #514, opened 2024-02-20
Fix circular message

BentiGorlich:
fix circular `ChainActivityMessage` when the `object` is of type `string`


melroy89:
.. testing..

melroy89:
image

melroy89 (approved):
It's of course hard to tell if it will solve the recursive issues I was seeing or not. But at least it doesn't break anything AFAIK.

PR #516, opened 2024-02-21
Remove route for down votes and make it a span

BentiGorlich:
As discussed in #513 we remove the ability to view who has down voted a post in the activity tab. You can still see how many down votes there are, it just is not a link anymore that leads to a page where you can see who down voted


nobodyatroot (approved):
Looks good on my end, tested live

PR #522, opened 2024-02-22
Rework account deletion

BentiGorlich:
This PR completely reworks the account deletion process. A user has full control over the deletion now. There is a new tab in the user settings for account deletion: account deletion
In there the user can choose whether the account will be marked for deletion in 30 days or deleted immediately.

Deleting a user will now result in sending every instance with which the user has interacted at least once (through commenting, liking, etc.) a delete message for said user.

The `DeleteHandler` now takes advantage of the cascading delete system. Deleted users are no longer completely purged and just vanish from the database, instead a user row is deleted, so the cascade deletes can do their work, and then recreated with only the necessary information like the username and public and private key. We have to have a remainder of the user, so the actor id is blocked for future use.

For the scheduled delete to work we have a new symfony package and a new queue which has to be worked on, so supervisor/docker-compose changes are in order. 

Since we now have the information which users are deleted we return a tombstone for them now. Additionally we properly handle tombstone responses in the update actor code now and the `updateActor` method in the `ActivityPubManager` got fixed

Deleted users are not shown in the `People` tab anymore, nor in the admin user list. Deleted users can not login anymore, they just get a "Bad Credentials" message

The admin actions "Purge Content", "Delete Content" and "Purge Account" are gone now. If we think they were useful enough we can bring them back, but for now they are gone.

The handling of delete activities regarding users was added as well.

Lemmy compatibility:
- I modeled the activities after lemmys. I liked it better to have a tombstone actor as the object, but lemmy does not support it.
- Deleting a user will delete the user and all of their data on lemmy as well (see issue in their repo: LemmyNet/lemmy#4544)
- Deleting a user on lemmy did not federate to my instance, but that might also be just a dev instance problem, so we'll have to see, but since our delete activites match the lemmy ones it should work


melroy89:
FYI. I fixed the merge conflict for you

melroy89:
It has conflicts again, but this time I'm not resolving it.. Since I'm unsure what you are planning to do ;)

BentiGorlich:
Just an update:
- incoming delete messages are now being handled correctly for users (was not even noted as a todo, it was just not implemented at all :sweat_smile: )
- if a user deletes their account the delete message is immediately dispatched and handled
- I added the symphony scheduler and a scheduled task to delete accounts marked for deletion
- I added a delete dispatch when we update a user and get a tombstone

What needs to be done:
- I want the immediate deletion to be optional, the default behavior should imo be to schedule a delete. So adding a checkbox for that
- the user should have to insert their password in order to delete the account, that needs to be done
- testing in conjunction with Lemmy. Sp far I only tested mbin <-> mbin

docs/admin_guide.md
```diff
@@ -908,7 +908,7 @@ With the following content:
 
 ```ini
 [program:messenger]
-command=php /var/www/mbin/bin/console messenger:consume async async_ap failed --time-limit=3600
```
melroy89, line 911: owo no again an update on the messenger lol lol. Hopefully this will be the last change, if not, add now as many as you want. Creating no further impact on the setup for admins in the future ideally.

melroy89 (approved):

BentiGorlich:
@e-five256 what do you think about this:
grafik

nobodyatroot:
See matrix discussion about user count statistics, I fixed it by adding this to the user criteria before the return in `src/Service/InstanceStatsManager.php`:

```php
$user_criteria = clone $criteria;
$user_criteria->andWhere(Criteria::expr()->eq('isDeleted', false));

return [
    'users' => $this->userRepository->matching($user_criteria)->count(),

**Comment by nobodyatroot** [comment]
Also, if the user decides to not delete their account immediately (and it's still within the 30 day window), should they be allowed to reactivate their account once logging in again (or maybe after triggering a password reset)? I didn't see an option in the admin panel under deletion to restore a deleted user that wasn't deleted "immediately"...so is the account just waiting until the 30 days are up?

File: src/Controller/User/AccountDeletionController.php [hunk 2]
```diff
@@ -0,0 +1,76 @@
...
+        $this->denyAccessUnlessGranted('edit_profile', $this->getUserOrThrow());
+
+        $form = $this->createForm(UserAccountDeletionType::class);
+        $user = $this->getUserOrThrow();
+        try {
+            // Could throw an error
+            $form->handleRequest($request);
+            if ($form->isSubmitted() && $form->has('currentPassword')) {
+                if (!$this->userPasswordHasher->isPasswordValid(
+                    $this->getUser(),
```
e-five256, line 47: it looks like you already grab the user on ln 41, could this be changed to reuse that?

```suggestion
                    $user
```

e-five256, line 47: this and ln 59 both call `getUser` or `getUserOrThrow` but you did that at the top of the method, could `$user` just be reused instead?

BentiGorlich: true

BentiGorlich:
> See matrix discussion about user count statistics, I fixed it by adding this to the user criteria before the return in `src/Service/InstanceStatsManager.php`:
> 
> ```
> $user_criteria = clone $criteria;
> $user_criteria->andWhere(Criteria::expr()->eq('isDeleted', false));
> 
> return [
>     'users' => $this->userRepository->matching($user_criteria)->count(),
> ```

All the stats repos have to be adjusted, I am working on it right now

src/Controller/User/AccountDeletionController.php
```diff
@@ -0,0 +1,76 @@
...
+        private readonly LoggerInterface $logger,
+        private readonly UserPasswordHasherInterface $userPasswordHasher,
+        private readonly TranslatorInterface $translator,
+    ) {
+    }
+
+    #[IsGranted('ROLE_USER')]
+    public function __invoke(Request $request): Response
+    {
+        $this->denyAccessUnlessGranted('edit_profile', $this->getUserOrThrow());
```
e-five256, line 38: it seems like this might have lost csrf validation, should it have it?

BentiGorlich: No it didn't loose it. Symfony adds that automatically if we use their form stuff and since I converted the code to use it I don't have to include it in our code

BentiGorlich:
> Also, if the user decides to not delete their account immediately (and it's still within the 30 day window), should they be allowed to reactivate their account once logging in again (or maybe after triggering a password reset)? I didn't see an option in the admin panel under deletion to restore a deleted user that wasn't deleted "immediately"...so is the account just waiting until the 30 days are up?
> 
> And it also ties into what should happen when an admin deletes an account, pretty sure that should be an immediate delete with no recourse given to the user.... might be good to put back the danger button highlight for the delete account button.
> 
> image

Looks like this now: 
grafik

Access to non admins and non moderators is blocked if an account is scheduled for deletion

melroy89:
Maybe update / extend some user documentation or FAQ regarding user deletion and internal working with respect to this? I'm just thinking how an end-user without technical knowledge would like to learn more about this topic.

src/Command/TestCommand.php
```diff
@@ -0,0 +1,51 @@
...
+use App\Service\UserManager;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputArgument;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+use Symfony\Component\Console\Style\SymfonyStyle;
+
+#[AsCommand(
+    name: 'mbin:test',
```
melroy89, line 17: still called `mbin:test`? If you want to keep this command, please rename.

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
```diff
@@ -35,16 +40,21 @@ public function __construct(
...
+        $id = \is_array($message->payload['object']) ? $message->payload['object']['id'] : $message->payload['object'];
+        $object = $this->apActivityRepository->findByObjectId($id);
 
-        if (\is_array($message->payload['object'])) {
-            $object = $this->apActivityRepository->findByObjectId($message->payload['object']['id']);
-        } else {
-            $object = $this->apActivityRepository->findByObjectId($message->payload['object']);
+        if (!$object) {
+            $user = $this->userRepository->findOneBy(['apProfileId' => $id]);
+            if ($actor instanceof User && $user instanceof User && $actor->apDomain === $user->apDomain) {
```
melroy89, line 50: maybe also check on $user not null/empty? Yea those userRepository calls can sometimes return null.

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
```diff
@@ -35,16 +40,21 @@ public function __construct(
...
+        $actor = $this->activityPubManager->findActorOrCreate($message->payload['actor']);
+
+        $id = \is_array($message->payload['object']) ? $message->payload['object']['id'] : $message->payload['object'];
+        $object = $this->apActivityRepository->findByObjectId($id);
 
-        if (\is_array($message->payload['object'])) {
-            $object = $this->apActivityRepository->findByObjectId($message->payload['object']['id']);
-        } else {
-            $object = $this->apActivityRepository->findByObjectId($message->payload['object']);
+        if (!$object) {
```
melroy89, line 48: add check if $actor is not null/undefined/... thus, something like:

```suggestion
        if (!$object && $actor) {
```

melroy89: So also `findActorOrCreate` can return null/empty objects. Trust me.. >><<

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
```diff
@@ -53,23 +63,17 @@ public function __invoke(DeleteMessage $message): void
 
         $object = $this->entityManager->getRepository($object['type'])->find((int) $object['id']);
```
melroy89, line 65: add `if ($object)`? or something..

src/Repository/StatsContentRepository.php
```diff
@@ -177,106 +121,68 @@ public function getStats(
...
-        $stmt = $conn->prepare($sql);
-        if ($magazine) {
-            $stmt->bindValue(1, $magazine->getId());
+        if (null !== $magazine) {
+            $query->setParameter(':magId', $magazine->getId());
         }
+        $res = $query->getScalarResult();
 
-        return $stmt->executeQuery()->fetchAllAssociative();
+        return $res[0][0];
```
melroy89, line 171: check if $res contains an array and length not zero.. And then another `&&` check for the other array (multidimensional array).

src/Repository/StatsRepository.php
```diff
@@ -105,4 +105,44 @@ protected function getStartDate(array $values): array
...
+        );
+
+        if (empty($startDate)) {
+            return [
+                'entries' => [],
+                'comments' => [],
+                'posts' => [],
+                'replies' => [],
+            ];
+        }
```
melroy89, line 127: Validate $startDate before continue. Also check if `$startDate[0]` exists.

src/Service/ActivityPubManager.php
```diff
@@ -266,6 +283,13 @@ public function updateUser(string $actorUrl): ?User
         $user = $this->userRepository->findOneBy(['apProfileId' => $actorUrl]);
 
         $actor = $this->apHttpClient->getActorObject($actorUrl);
+
+        if (isset($actor['type']) && 'Tombstone' === $actor['type'] && $user instanceof User) {
```
melroy89, line 287: Also check here if `$actor` exists/defined. Same for `$user` I guess.

src/Service/UserManager.php
```diff
@@ -308,30 +311,35 @@ public function detachCover(User $user): void
...
-        $this->entityManager->flush();
+            $this->entityManager->persist($user);
+            $this->entityManager->flush();
+        } else {
+            $this->delete($user);
+        }
     }
 
-    public function revokeDeleteRequest(User $user): void
+    public function removeDeleteRequest(User $user): void
```
melroy89, line 327: Add code documentation above these methods, so other developers do not need to read all the lines in the future to understand what each of these methods do. 

Also just in general this can help onboard new developers to the project.

src/Command/RemoveAccountsMarkedForDeletion.php
```diff
@@ -0,0 +1,39 @@
...
+    ) {
+        parent::__construct();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        $users = $this->userManager->getUsersMarkedForDeletionBefore();
+        foreach ($users as $user) {
+            $output->writeln("deleting $user->username");
+            $this->bus->dispatch(new DeleteUserMessage($user->getId()));
```
melroy89, line 33: And try/catch.

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
melroy89: OK let's hope so 😂

BentiGorlich: A PHP playground just confirmed it :D

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
BentiGorlich: $actor is checked by the `instanceof` call in the next if as well, but it won't hurt to check it here as well

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
BentiGorlich: above it:
```php
        if (!$object) {
            return;
        }
```

src/Repository/StatsRepository.php
BentiGorlich: `$startDate` is checked for being empty just above. I'll add the check whether `year` and `month` exist

src/Service/ActivityPubManager.php
BentiGorlich: User is checked through `instanceof`. The other one will be added

BentiGorlich:
> Maybe update / extend some user documentation or FAQ regarding user deletion and internal working with respect to this? I'm just thinking how an end-user without technical knowledge would like to learn more about this topic.

I think the internal working for developers is quite clear after I added the documentation. For users I don't think that any of this info is helpful. I think they are fine with "I can delete my account instantly or have it be deleted in 30 days".

src/Repository/StatsRepository.php
melroy89: ah yes sorry my bad.

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
melroy89: No.. after this call .. Who says that `$object = $this->entityManager->getRepository($object['type'])->find((int) $object['id']);` is not returning empty/null data?

src/Command/TestCommand.php
melroy89: Can happen ;P

BentiGorlich:
@nobodyatroot reports from mastodon do indeed use the instance actor, but that actually gets created as a user on mbin. So the field is not empty...

So I will make the cascade do a delete instead of set null

templates/user/settings/_options.html.twig
```diff
@@ -25,6 +25,12 @@
                 {{ 'password_and_2fa'|trans }}
             </a>
         </li>
+        <li>
```
e-five256, line 28: might be worth not showing this if user is admin, and maybe any backend checks people think are necessary to make sure the admin doesn't immediately delete themselves

nobodyatroot:
Adding stats comment observation in case matrix convo gets lost:

Threads, comments, and posts stats swapped between local and federated:

Local:

image


Federated:

image

src/Repository/StatsContentRepository.php
```diff
@@ -177,106 +121,72 @@ public function getStats(
...
-            $end = new \DateTime();
+        $tableName = match ($tableName) {
+            'entry' => 'entry',
+            'entry_comment' => 'entry_comment',
+            'post' => 'post',
+            'post_comment' => 'post_comment',
+            default => throw new \InvalidArgumentException("$tableName is not a valid countable")
+        };
+
+        $federatedCond = $federated ? ' AND e.ap_id IS NULL ' : '';
```
e-five256, line 148: think this is swapped, if I'm reading it right, it says if federated is true, only count those where ap id is null; if federated is false, count all

```suggestion
        $federatedCond = $federated ? '' : ' AND e.ap_id IS NULL ';
```

src/Repository/StatsContentRepository.php
```diff
@@ -177,106 +121,72 @@ public function getStats(
...
+        $magazineCond = $magazine ? 'AND e.magazine_id = :magId' : '';
+        $SinceDateCond = $sinceDate ? 'AND e.created_at > :date' : '';
```
e-five256, line 150: was this meant to be capitalized? it being gold makes me wonder what php thinks of this. I'm used to capital being exported in golang but not familiar with php xD

templates/user/settings/_options.html.twig
BentiGorlich: Yeah you're absolutely right

docs/admin_guide.md
```diff
@@ -908,7 +908,7 @@ With the following content:
...
+command=php /var/www/mbin/bin/console messenger:consume scheduler_default async async_ap failed --time-limit=3600
```
nobodyatroot, line 911: Doesn't `config/packages/messenger.yaml` need to be updated to include this new `scheduler_default` queue?

nobodyatroot (changes requested):

docs/admin_guide.md
BentiGorlich: nope. It has actually nothing to do with the messengers besides the consume command: https://symfony.com/doc/current/scheduler.html#symfony-scheduler-basics

src/Repository/StatsContentRepository.php
nobodyatroot: confirmed this fixes the problem

BentiGorlich:
I have some thoughts about my humongous SQL query to get the inboxes of the users interactions: I think we only need to gather the remote magazine inboxes and the local subscribers of the magazines in which the user interacted with someone. That would make the query at least a little shorter. Maybe we only focus on magazines the user posted to and ignore votes completely, it's your call, but the query just looks ridiculous... (and technically there are still 4 selects missing, I did not include the local subscribers of the magazines the user favourited something :grin: )

BentiGorlich:
And deletes by moderators/admins do look exactly the same for remote instances as a user deleting their own account, but I think that is fine for now and can be fixed in a future PR

config/packages/monolog.yaml
```diff
@@ -7,13 +7,13 @@ when@dev:
         handlers:
             main:
                 type: rotating_file
-                path: '%kernel.logs_dir%/%kernel.environment%.log'
+                path: '/var/log/mbin/%kernel.environment%.log'
```
nobodyatroot, line 10: I noticed my dev environment broke because this is hardcoded to `/var/log/mbin`, any particular reason?

BentiGorlich: Its because I am stupid... Shouldn't have committed that...

nobodyatroot:
@BentiGorlich i tried to delete a spam lemmy user in a remote mag (since purge/delete content action is missing) and it crashed my instance... the queues were exploding with messages (no errors, just a massive accumulation), are we accidently trying to send out ban/delete user messages for remote users that we are not responsible for? i think it should work that we have ban/delete content authority of remote users in our own magazines, but maybe you're already doing that.... I'm not sure what happened, I've temporarily reverted back to main.

BentiGorlich:
Oh no... I gotta test that...

src/MessageHandler/ActivityPub/Inbox/DeleteHandler.php
BentiGorlich: null checks are covered by the `instanceof` underneath. though I put the result in a different variable to not reassign `$object` to a different datatype

BentiGorlich:
@nobodyatroot I just tried deleting remote users on my dev instance with the PR up and running and it behaved exactly like expected. It just creates one `DeleteUserMessage` and thats it. It just worked. I don't know what to do about it :thinking:

templates/user/settings/_options.html.twig
BentiGorlich: @melroy89 @nobodyatroot @asdfzdfj what do you think about this? Do we need to disable the form for admins? I mean it is only possible to make people admins via cli, and we can always create new admins via cli. I mean most admins would never use it anyway, so might as well disable it, but I am not sure...

nobodyatroot (approved):

nobodyatroot:
It may have been a combo of unfortunately events with the spam wave happening at the same time and I was having DNS resolution issues that I've since corrected.......

templates/user/settings/_options.html.twig
melroy89: Can we also delete an admin via CLI (if case the admin really wants to)? Then yes, you do want to not show this UI option I believe. Since this could often only end up in a bad day.

BentiGorlich: I don't think we have a command for deleting an admin, yet...

nobodyatroot (approved):
Verified recent commits work as expected in dev environment.

templates/user/settings/_options.html.twig
melroy89: Now we have. Thanks!

melroy89 (approved):
Lgtm

melroy89:
Regression again :) https://github.com/MbinOrg/mbin/pull/655

PR #533, opened 2024-02-27
Set BaseEntityAutocompleteType with Ajax Form

melroy89:
Use `ParentEntityAutocompleteType::class` as return value on the `getParent()` method, since we are using Ajax forms.

https://symfony.com/bundles/ux-autocomplete/current/index.html#usage-in-a-form-with-ajax

See also: https://github.com/symfony/ux/issues/1541


melroy89:
Tested. Works..

image

BentiGorlich (approved):

PR #535, opened 2024-02-28
Update delete image message and handler to use object ids

nobodyatroot:
Delete image handler and message was still using paths instead of IDs added in https://github.com/MbinOrg/mbin/pull/501 causing 500s to be generated when image removal was attempted. Path is still needed to remove from filesystem, so method added to image entity class to get path from image to be compatible with existing `imageManager->remove` method.

### DIFF
--- a/src/Entity/Image.php
+++ b/src/Entity/Image.php
@@ -89,6 +89,11 @@ class Image
         }
     }
 
+    public function getPath(): string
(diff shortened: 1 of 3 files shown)


melroy89:
lint :)

melroy89 (approved):

PR #540, opened 2024-02-29
Fix default theme admin setting's aria-label

thepaperpilot:


e-five256 (approved):
whoops, bad copy paste by me

PR #546, opened 2024-03-03
adjust external link icon size

e-five256:
based on recommendation from @asdfzdfj to make the external link icon slightly smaller to match the url it's next to

before

image

after

image


asdfzdfj (approved):
oh hey, thanks for making this!

honestly "my suggestion" feels more like ramblings with personal opinion, but since it's here then might as well add it

Issue #548, opened 2024-03-03
Internal server error trying to subscribe to unavailable magazine


andrewmoise:
* I have novid@chirp.social visible on my instance
* It [doesn't seem to exist](https://chirp.social/) as a real available resource
* When I click to try to subscribe to it, I get a server error:

```
{"message":"Warning: Trying to access array offset on value of type null","context":{"exception":{"class":"ErrorException","message":"Warning: Trying to access array offset on value of type null","code":0,"file":"/var/www/mbin/src/Service/ActivityPub/ApHttpClient.php:87"}},"level":300,"level_name":"WARNING","channel":"php","datetime":"2024-03-03T15:33:39.364434+00:00","extra":{}}
{"message":"Uncaught PHP Exception Symfony\\Component\\Messenger\\Exception\\HandlerFailedException: \"Handling \"App\\Message\\ActivityPub\\Outbox\\FollowMessage\" failed: App\\Service\\ActivityPub\\ApHttpClient::getInboxUrl(): Return value must be of type string, null returned\" at HandleMessageMiddleware.php line 124","context":{"exception"{"class":"Symfony\\Component\\Messenger\\Exception\\HandlerFailedException","message":"Handling \"App\\Message\\ActivityPub\\Outbox\\FollowMessage\" failed:App\\Service\\ActivityPub\\ApHttpClient::getInboxUrl(): Return value must be of type string, null returned","code":0,"file":"/var/www/mbin/vendor/symfony/messenger/Middleware/HandleMessageMiddleware.php:124","previous":{"class":"TypeError","message":"App\\Service\\ActivityPub\\ApHttpClient::getInboxUrl(): Return value must be of type string, null returned","code":0,"file":"/var/www/mbin/src/Service/ActivityPub/ApHttpClient.php:87"}}},"level":500,"level_name":"CRITICAL","channel":"request","datetime":"2024-03-03T15:33:39.365064+00:00","extra":{}}                                                                  
```

Pretty trivial but maybe a nicer error message would be better.

This is on git main as of 2024-03-02.

melroy89:
As you can see I created a PR for this.

PR #550, opened 2024-03-03
Improve AP documentation, debugging and error throwing

melroy89:
- Add more useful debugging lines
- Improve error message thrown instead of `Return value must be of type string, null returned`
- Improve code documentation

Fixed: https://github.com/MbinOrg/mbin/issues/548

### DIFF
--- a/src/Service/ActivityPub/ApHttpClient.php
+++ b/src/Service/ActivityPub/ApHttpClient.php
@@ -52,7 +52,7 @@ class ApHttpClient
     public function getActivityObject(string $url, bool $decoded = true): array|string|null
     {
         $resp = $this->cache->get('ap_'.hash('sha256', $url), function (ItemInterface $item) use ($url) {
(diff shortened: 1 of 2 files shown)


src/Service/ActivityPub/ApHttpClient.php
```diff
@@ -80,19 +80,38 @@ public function getActivityObject(string $url, bool $decoded = true): array|stri
...
+     * @return inbox return the inbox URL of the actor
+     *
+     * @throws LogicException if the AP actor object cannot be found
+     */
     public function getInboxUrl(string $apProfileId): string
     {
         $actor = $this->getActorObject($apProfileId);
-
-        return $actor['endpoints']['sharedInbox'] ?? $actor['inbox'];
+        if (!empty($actor)) {
```
melroy89, line 93: This will now check if the actor is actually not empty/not null.

PR #552, opened 2024-03-05
Add AMQP proxy to docs + use Docker images by default + Fix compose.prod.yml

melroy89:
**General:**

- Add AMQP Proxy (`amqproxy`) to our documentation guides. Both Docker & Bare metal setup.

**Docker:**

- Improve Docker documentation in general.
- Always use `env_file` in docker compose on our Docker image
- Introduce a `mbin_internal_network` in Docker.
- Extend `.dockerignore`
- Make our [Mbin ghcr.io Docker images](https://github.com/MbinOrg/mbin/pkgs/container/mbin) used by default in the Docker setup. I don't think users need to build their own images anymore, looking at the "[billing & plans](https://github.com/organizations/MbinOrg/settings/billing/summary)", we can safely say that I think it's fine to fetch images from ghcr.io by default without any problems. Thus keep using the free plan we already have.


TheVillageGuy:
I'll give this a go today and let you know if they're were any issues

melroy89:
> I'll give this a go today and let you know if they're were any issues

Docker or bare metal?

TheVillageGuy:
It's not working for me yet

docs/admin_guide.md
```diff
@@ -796,6 +797,60 @@ MESSENGER_TRANSPORT_DSN=amqp://kbin:${RABBITMQ_PASSWORD}@127.0.0.1:5672/%2f/mess
...
+
+```sh
+curl -fsSL https://packagecloud.io/cloudamqp/amqproxy/gpgkey | gpg --dearmor | sudo tee /usr/share/keyrings/amqproxy.gpg > /dev/null
+. /etc/os-release
+echo "deb [signed-by=/usr/share/keyrings/amqproxy.gpg] https://packagecloud.io/cloudamqp/amqproxy/$ID $VERSION_CODENAME main" | sudo tee /etc/apt/sources.list.d/amqproxy.list
+sudo apt update
+sudo apt install amqproxy
+```
+
+Configure AMQP Proxy by editing the `/etc/amqproxy.ini` config file:
```
TheVillageGuy, line 817: File does not exist yet

melroy89:
Ps. since this a new docker setup, and the docker image is not yet pushed to ghrc.io, be sure to comment out `image: ghrc...` for now in your compose.yml and enable:

```
    build:
      context: ../
      dockerfile: docker/Dockerfile
    image: mbin
```

During testing..

e-five256:
That's one of the things that makes me not sure about the docker change. I know I've discussed doing it before, but looking at lemmy they also default to locally build images for their repo. I think if you're cloning the code repo, you probably wanted to do development, and what we're missing is entirely different repos that don't include the source code but just have things like pulling the publicly tagged and pushed images, since if you're wanting to run those you likely don't care about syncing the source code

melroy89:
Yeaa so for local development you do want to build local docker images.

However if we setup the docker configuration correctly (meaning all configs should be outside the docker container etc). The pre-build docker image should be sufficient. And in fact it should be recommended for running production (aside from the bare metal setup).

melroy89:
But some of the changes in this PR in another smaller PR: https://github.com/MbinOrg/mbin/pull/586

melroy89:
I'm taking a break.

melroy89:
yea, but my Tinnitus is killing me now due to the dike work, because they are continuously pounding sheet piles ("damwanden") into the ground near me.

PR #554, opened 2024-03-05
Use title in "related magazines" panel

BentiGorlich:
- change the inline magazine component to use the title of the magazine if the full name is not needed
- set full name to false for related magazines

| Before | After |
| ------ | ----- |
| grafik | grafik |


melroy89:
sorry maybe a before & after screenshot could help here.

melroy89 (approved):
Thanks! This makes it easier to read indeed. And also more consistent.

asdfzdfj (approved):
giving it a go on my instance and it seems to be working fine

PR #556, opened 2024-03-06
add figure badges to entry thumbs and for gif image

asdfzdfj:
add `GIF` badge to attached image preview when image src ends with `.gif` to notify viewer and maybe hopefully invite them to check out the full image

image

image

added because the thumbnail filter with standard gd can't handle animated gifs at the moment (and doesn't look like it ever will), so it will appear static although it could be an animated image

the detection is based on filename extension, so for local image it should be fairly accurate since filename is normalized with extension generated based on its type, but possibly hit and miss for remote image though


melroy89 (approved):
Approved for now, but see my comment.

melroy89 (approved):

asdfzdfj:
> As you noticed earlier in the matrix chat some images do not always end with `.gif` despite being it a GIF image. I think it's fine for now, but you could also check the image type AKA MIME type.

I'm well aware of that, but that should only be a problem for *remote* images that's produced by #501 (i.e. having the `Image` entity but not having a `filePath` with already normalized extension from its mime type set), if you want accurate detection of these images then it'll likely need something to load the remote image to check its type everytime, either server side check while templating or a client-side js that adds the badge after the fact, both of which I'm not too keen to add it just for this.

also the badges are only added for image figures that's being templated server side using the `Image` entity, the ones from embed preview links is another story and isn't covered by this (besides, if it's animated gif then you'll see it, there's no animation-less thumbnailer in the way).

PR #559, opened 2024-03-06
Fix clipping / z-index on visibility private posts

e-five256:
remove inline styles from private posts and comments

saw this in the wild, seems pretty rare but for posts and post/entry comments with a private visibility, they end up with this hardcoded block. It probably needs more work, private is untranslated for instance, and of course there's the bigger can of worms on using visibility at all. but for now, wanted to make the styling look not broken, ideally. I tried to test out with the scenario I saw it in, which was post, but couldn't repro how exactly it looks for post comment or entry comment (the comments just disappear, unlike post which shows up with the word "Private"), but I assumed they might be acting similar

before:

(note in first image the comment clips into the post it's replying to)
image
image

---

after

image

### DIFF
--- a/templates/components/entry_comment.html.twig
+++ b/templates/components/entry_comment.html.twig
@@ -8,8 +8,7 @@
 
 {% if not app.user or (app.user and not app.user.isBlocked(comment.user)) %}
     {% if comment.visibility is same as 'private' and (not app.user or not app.user.isFollower(comment.user)) %}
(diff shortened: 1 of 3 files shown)


melroy89 (approved):
This actually annoyed me since day 1 ;P Thanks for finding and fixing the issue

e-five256:
Might have given you false hope it'd fix issues with z index elsewhere but believe this is only for private posts which I don't think I've seen except once from mastodon. Will try to take a look at the issue you mentioned, though

PR #565, opened 2024-03-08
Fix profile option being marked active when viewing other profiles

e-five256:
Right now the `Profile` header option gets marked `active` on any user overview page, not just your own. This fixes it so it will only be marked as active when you're viewing your own profile overview.

Before (note looking at `user2` as `user1`)

image

---

After `user2`

image

After `user1`

image

### DIFF
--- a/templates/layout/_header.html.twig
+++ b/templates/layout/_header.html.twig
@@ -150,7 +150,7 @@
                     <ul class="dropdown__menu">
                         <li>
                             <a href="{{ path('user_overview', {username: app.user.username}) }}"
-                               class="{{ html_classes({'active': is_route_name_contains('user_overview')}) }}">
+                               class="{{ html_classes({'active': is_route_name_contains('user_overview') and user is same as app.user}) }}">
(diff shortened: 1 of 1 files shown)


templates/layout/_header.html.twig
```diff
@@ -150,7 +150,7 @@
                     <ul class="dropdown__menu">
                         <li>
                             <a href="{{ path('user_overview', {username: app.user.username}) }}"
-                               class="{{ html_classes({'active': is_route_name_contains('user_overview')}) }}">
+                               class="{{ html_classes({'active': is_route_name_contains('user_overview') and user is same as app.user}) }}">
```
e-five256, line 153: I swear I browsed and posted things and everything... if this ever 500s and says "user is undefined"... maybe I should add `and user is defined` _just_ in case... but it didn't seem needed even after using my site a bunch... maybe I should double check in app_env prod just to make sure, wouldn't want this to cause any issues

nobodyatroot (approved):

PR #567, opened 2024-03-09
Update docker_deployment_guide.md

TheVillageGuy:
Removed `messenger_ap` as it's not in the file


melroy89 (approved):
Correct. This was forgotten to be updated after recent messenger changes.

PR #573, opened 2024-03-11
Fix circular like/announce message

BentiGorlich:
Under certain circumstances there could be a loop where instance A notifies instance B that a user liked something, B announces it to A which triggered A to notify B, etc. etc. (not sure that this is exactly what happened, but something like it) 

Anyway, the solution is easy and should have been done ages ago: do nothing if an activity from a local user is announced. No server should ever announce something to our server that originates from our server, but every software has bugs (ours is a nest if you ask me) and therefore we should account for that

### DIFF
--- a/src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
+++ b/src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
@@ -101,6 +101,16 @@ readonly class ActivityHandler
             }
             if (\is_array($payload['object'])) {
                 $payload = $payload['object'];
+                $actor = $payload['actor'] ?? $payload['attributedTo'] ?? null;
+                if ($actor) {
+                    $user = $this->manager->findActorOrCreate($actor);
+                    if ($user instanceof User && null === $user->apId) {
+                        // don't do anything if we get an announce activity for something a local user did
+                        $this->logger->warning('ignoring this message because it announces an activity from a local user');
+
+                        return;
+                    }
+                }
             }
         }
 


src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
```diff
@@ -101,6 +101,16 @@ private function handle(?array $payload)
             }
             if (\is_array($payload['object'])) {
                 $payload = $payload['object'];
+                $actor = $payload['actor'] ?? $payload['attributedTo'] ?? null;
```
melroy89, line 104: It's interesting that it's almost equal to line 96. Can a magazine also be part of `attributedTo`?

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
```diff
@@ -101,6 +101,16 @@ private function handle(?array $payload)
...
+                if ($actor) {
+                    $user = $this->manager->findActorOrCreate($actor);
```
melroy89, line 106: It's interesting that it's almost equal to line 96. Can a magazine also be part of attributedTo?

melroy89: ow nevermind, the actor is within the object payload.

melroy89:
Seems fine. Should I run it some ours on my instance. Just to test.

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
```diff
@@ -101,6 +101,16 @@ private function handle(?array $payload)
...
+                    if ($user instanceof User && null === $user->apId) {
```
melroy89, line 107: What about other local announcements..?

BentiGorlich: What do you mean?

melroy89: I mean, could there be local magazines, posts, ... also triggering "Announce" messages?

BentiGorlich: This only covers incoming announces. It is in the activity handler, so it is **the** poijt of incoming activites from our inbox. 

However I just realised that it will probably break boosts, as they are annpjnce activites as well. I think I already broke boosts in some earlier change,but I'll have to investigate this

BentiGorlich: Ok, so I checked, I broke boosting with
- the `AnnounceWrapper` changes in c13bf815
- and the `Outbox/AnnounceHandler` changes in 9eddb84f

So I think we can just ignore that part and I'll make another issue about it

melroy89: OK hopefully you could indeed fix the boost regression soon 👍🏽 .

melroy89 (approved):

BentiGorlich:
I'll put this code on my instance to test it first, since @nobodyatroot said that it introduced some weird behaviour on their instance...

BentiGorlich:
I jad it runningforthe day and I didn't see any weird behaviour. How about you @nobodyatroot

nobodyatroot:
> I jad it runningforthe day and I didn't see any weird behaviour. How about you @nobodyatroot

My high load observation probably wasn't directly caused by this PR and rather the favourite duplicate key issue... so I'm fine with this.

nobodyatroot (approved):

melroy89:
> rather the favourite duplicate key issue.

Or the high queue size?

PR #580, opened 2024-03-13
Add migration for unique indexes on favourite

### LINKED ISSUE #578: [SQL] Duplicate Favourite Rows
**Describe the bug**
Multiple sql rows for the same user + [entry/entry_comment/post/post_comment] combination

all of them

**To Reproduce**
1. `sudo -u kbin psql`
2. run 
```sql
SELECT entry_comment_id, user_id, COUNT (*) as cnt FROM favourite 
WHERE entry_comment_id IS NOT NULL 
GROUP BY entry_comment_id, user_id 
HAVING COUNT(*) > 1 
ORDER BY cnt DESC LIMIT 1000;
```
3. get a lot of results

**Expected behavior**
This query should yield no results, as the each user can only favourite each entry, entry_comment, post or post_comment exactly once and the code expects this in certain cases which leads to a lot of messenger errors

**What should be done**
write an sql migration which consists of:
- queries to remove duplicates, I found [this post](https://www.postgresqltutorial.com/postgresql-tutorial/how-to-delete-duplicate-rows-in-postgresql/) which outlines how to do that
- create 4 unique indices for these pairs in the favourite table:
  - user_id, entry_id
  - user_id, entry_comment_id
  - user_id, post_id
  - user_id, post_comment_id

This will be a huge migration on production servers, as the favourite table is quite large

Just a note: the 4 vote tables already have an unique index, so there is nothing to do for that

BentiGorlich:
Delete duplicate favorites and introduce 4 new indices. 

Fixes #578

### DIFF
--- a/migrations/Version20240313222328.php
+++ b/migrations/Version20240313222328.php
@@ -1,0 +1,36 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20240313222328 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'This migration removes all the duplicates from the favourite table and creates 4 unique indexes for each combination of user_id with [entry|entry_comment|post|post_comment]_id';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('DELETE FROM favourite a USING favourite b WHERE a.id > b.id AND a.entry_id = b.entry_id AND a.user_id = b.user_id');
+        $this->addSql('DELETE FROM favourite a USING favourite b WHERE a.id > b.id AND a.entry_comment_id = b.entry_comment_id AND a.user_id = b.user_id');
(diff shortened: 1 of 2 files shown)


nobodyatroot (dismissed):
Migration went smoothly, no issues encountered since I stopped the messengers first.

nobodyatroot (approved):

melroy89 (approved):

melroy89:
Did you validate a db migration diff check? So there is no diff created between doctrine and the db? I ask this since you first created the migration and then the doctrine. Instead of the other way around.

BentiGorlich:
I just derped on pushing the changes. I did generate it through doctrine 😅

e-five256:
I wanted to double check the instructions for this migration, as I was unaware of the need to stop messengers during certain migrations. Is that standard for a specific change? Are there instructions on how to do so? Obviously as you've already stated we'll want to make this known in release notes, and just letting peeps know in the matrix general is probably a good idea

BentiGorlich:
The reason for that is that you might wind up in a deadlock in postgres if you don't stop the messengers (happened to me once with a longer running migration). That is only for long running migrstions though. With this one specifically the migration might fail due to unique violations. Because it is so long running you might get newly created duplicates which then block the creation of the indeces

PR #587, opened 2024-03-15
Update crosspost component, include entry title

e-five256:
- Changes "crossposted" to "Related"
  - Not completely sure on this as it makes it suggest it _is_ related when really it's just a match on either url or text, so not sure if it could be misleading if the entries aren't _actually_ related, but might be overthinking it
- Adds the entry title to give more context
  - can be a bit long. was thinking of truncating it, but not sure how people felt. definitely can see it takes more space on mobile with long titles
- Adds 18+ badge if marked nsfw
  - OC/Language/User mod/admin/etc badges aren't on this, but felt like it deserved the nsfw badge just as a headsup before you click. The rest you can see if you open the entry, but a linked nsfw one might be surprising, just my opinion
- Adds the missing translations

---

before

image
image

---

after

image
image

<details>
<summary>more images of 1-5 related entries.
entry _list view does not stack/group like this, only showing a single row, this is entry comment view.
this is because the entry list view doesn't have an overall group that parents the related entries like entry comment view does (the `entries_cross.html.twig` is only used in entry comment view, not entry list)</summary>

image

---

image

---

image

---

image

---

image

</details>


templates/components/entry_cross.html.twig
```diff
@@ -7,7 +7,7 @@
         data-action="notifications:Notification@window->subject#notification">
     <aside class="meta entry__meta">
         {% if entry.visibility in ['visible', 'private'] or (entry.visibility is same as 'trashed' and this.canSeeTrashed) %}
-            <a href="{{ entry_url(entry) }}">{{ 'crossposted'|trans }}</a> {{ 'by'|trans }}
+            {{ 'Related'|trans }}: {% if entry.isAdult %}<small class="danger">18+</small>{% endif %} <a href="{{ entry_url(entry) }}">{{ entry.title }}</a> {{ 'by'|trans }}
```
e-five256, line 10: I think maybe I should try using the twig filter `truncate` on the entry title to a smaller length, seems simple enough and might keep it from going crazy number of lines on mobile

might also want padding though, it looks a bit more busy on mobile

translations/messages.en.yaml
```diff
@@ -797,3 +797,5 @@ all_time: All time
 show: Show
 hide: Hide
 edited: edited
+Related: Related
+by: by
```
e-five256, line 801: I wonder if this is correct, or maybe should be changed to something else, as I imagine translating "by" with no context to be difficult. I was thinking I could instead do a single string like `Related: %title% by %user%' to give it more context, but there aren't too many cases of that

BentiGorlich: I mean you could just change the key to `related_by: by`

e-five256:
After looking at it a bit on @nobodyatroot I felt it needed a bit more of a rework. The titles are super verbose when long so want to truncate them. Just having a bit of trouble getting the CSS right as CSS ellipsis isn't allowed in inline blocks and putting it in an inline block makes the position a bit finicky, where it'll look fine in chromium but not so much firefox, example

image

also while I was staring at this, I came to the realization the template defaulted null cookies to false, so if you've ever wondered why it doesn't show user avatars or magazine icons some times, it's because we had changed that null default to true and didn't do that on the cross component, so I took care of that as well

templates/components/entry_cross.html.twig
e-five256: we don't have the twig extension for truncate it turns out, and thinking about it I don't _actually_ want to chop the text off, overflow hidden/ellipsis is superior as it would be better for screenreaders

BentiGorlich:
Can't you just use `max-width` with `text-overflow`?
I always thought the absence of avatars was on purpose to make the component less intrusive :sweat_smile:

BentiGorlich:
Bildschirmfoto vom 2024-03-16 11-08-14
I've put the whole title in a div so the "by author..." part is on the next line. The `max-width` needs to be optimized, but the rest could stay that way. Question is "are we fine with it always taking 2 rows". Previously it could fit on one row...

But honestly I think the best bet is to truncate it in php, I think that is the better option

e-five256:
I mentioned in the above comment but I'm not a fan of _actually_ truncating the text because that is inaccessible to screenreaders. When I went around looking at other websites, they all do overflow/ellipsis

I think including the title at all is very verbose when it's long and reposted a bunch

image

There's a ton of repetitive information here that I don't think is useful.

I guess it'd be nice if the top post's entry title could be passed in, and if it matches maybe omit then but if it differ show it, but it's difficult to do logic in these components because they're cached

e-five256:
I'm thinking about two things:

1. making it a grid area so multiple entries can appear side by side if there is room, that way it takes less vertical space on large resolutions (with a max width on the title as you mentioned)
2. removing the inline previews. I'm not sure what the point is. if it's a link, that means every single inline preview matches, is there really a use case where you want to expand 3 of the same image or youtube video in the same window?

e-five256:
been trying to work out what exactly I had in my mind and this is what it currently looks like. I think it looks good, but it's possible people think it looks too busy.

<details>
<summary>images</summary>

image
image
image
image
image

</details>

Compared to before obviously everything was on its own line so it takes up a lot less space now, and obviously is responsive so changes to single line on mobile for instance. It's a bit tricky, the front controller doesn't look any different than it did in my earlier screenshots because on the front controller they're not linked in any way but completely separate entries that just display differently, whereas in the single controller the related entries are actually part of the template

nobodyatroot (approved):
I like it! Hopefully we can also find a better way in the future to reduce some of the clutter from the entry front view when lots of cross posts are linked.

PR #588, opened 2024-03-15
Adding the correct indexes for the user table

BentiGorlich:
This migration drastically improves the performance of our SQL query because the basic user fetch now takes < 1ms from ~50-60ms on gehirneimer.de

We already have indexes on the user table, however those are all lower(...) ones and we basically never query for `x = lower(...)` so these are basically never used. We do have unique indexes which do the same job, only for default queries, however the one for `ap_profile_id` was missing and the other ones cryptically named. This PR fixes that


e-five256 (approved):
this worked for me, didn't see any issues, and doing a migration diff only had me drop the zitadel stuff since I had swapped branches from it, so I think the doctrine generation matches the schema

PR #595, opened 2024-03-17
Rework messenger config

BentiGorlich:
- keep the "messenger" queue as "old" so existing setup keep working and have a smooth transition
- change the queue layout. We now have
  - "async": which is actually the queue name. It handles all instance specific work like sending the confirmation mail and sending notifications
  - "receive": basically only `Inbox/ActivityMessage`s will be routed here. I mainly did it so that you can see the processing status of your queue. Because these messages just get converted to other messages if they are part of the same queue one can not distinguish them and the queue appears to only grow larger. By having them in a different queue you can watch the messengers work on this
  - "inbox": every inbound activity like create messages, like messages, etc.
  - "outbox": every outbound activity
  - "deliver": actually delivering outbound activity to remote instances
  - "resolve": atm only two messages get routed to resolve: `ChainActivityMessage` and `UpdateActorMessage`
  - "old": an old queue for backwards compatibility
- The retry count was lowered to 4 and the max_delay adjusted to a flat value
- The `CreateActorMessage` and handler were deleted, because they were not used anywhere
- a lot of messages did not implement an interface for routing them. I added one to each message. Not implementing an interface which we route to an async queue results in executing the message synchronously. If that is desired in a message we should comment it on there in the future


melroy89 (changes requested):
Block merge, until further notice. As requested by Bentigorlich himself.

docs/admin_guide.md
```diff
@@ -908,7 +908,7 @@ With the following content:
 
 ```ini
 [program:messenger]
-command=php /var/www/mbin/bin/console messenger:consume async async_ap failed --time-limit=3600
+command=php /var/www/mbin/bin/console messenger:consume old async outbox deliver inbox receive resolve failed --time-limit=3600
```
melroy89, line 911: This means that `old` is the most important queue, then async and outbox. And finally inbox, receive, resolve and failed and queues with less prio.

We could still create multiple different messenger:consume commands, as long as you do not use the same messenger/queue names across multiple `messenger:consume` you should be good. 

For example n supervisor we could create a `messenger:consume` for example: "async + old + inbox + ...". And other messenger:consume for let's say "deliver + outbox + receive.".  Again just an example. This could help us prioritize the work better and do jobs in parallel if needed.

config/packages/messenger.yaml
```diff
@@ -8,19 +8,92 @@ framework:
...
+                options:
+                    queues:
+                        async: ~
+                    exchange:
+                        name: async
                 retry_strategy:
-                    max_retries: 5
+                    max_retries: 4
                     delay: 600000
                     multiplier: 10
```
melroy89, line 19: I propose the reduce the multiplier. delay is `600000` (=== 600 seconds). Multiplying this with a multiplayer of 10 means, 600000 ms, then 6000000 ms, then 60000000 ms.. and then 600000000 ms. Often you see a multiplier of a factor 2 or 3 in symfony messenger configurations. But not 10. 

600000000 is a delay of 7 days! This is too long IMO. I want to get rid of large queues faster, if it pills up, it will take days or weeks to get rid of. 

My proposal: reduce this multiplier down to 3 maybe 4 max.. Also decrease the `max_delay` option accordingly.

docs/admin_guide.md
BentiGorlich: why I want to stay at one consume command with all the queues in it:
- prioritization is taken care of. That means that **all** your messengers will prioritize the queues the way we tell them
- simplicity: you don't have multiple supervisor configs to worry about or multiple docker containers
- better worker utilization: if you have 3 consume commands all working on different queues then you have events where 
  - 2/3 of your messengers are idling while the third one is struggling to keep up
  - all of them are running at once and you defined more workers than your system should handle because usually they are not running at the same time and now your server is slow

config/packages/messenger.yaml
BentiGorlich: I deliberately chose a week as the max delay, because: servers have down time and sometimes they have a lot of downtime. A longer max delay provides better stability for the fediverse (yeah we are not a big part of it, but so what). I was fine to remove the fifth retry, because trying for 2 weeks is a bit much, I agree. I am also fine with trying more often so if the server comes back up after a day we do not need to carry the message around for another 6 days. 
So my proposal would be to decrease the max delay and increase the retry amount, so that in the end we do try for 7 days, once per day or something like that

BentiGorlich:
I have it running on both servers now and the simple fact that you can now see at which position in our processing chain problems occur is just really great

config/packages/messenger.yaml
melroy89: I still might not resolve the issue when a messenger never will be successfully processed, no matter if the server is back online. 

OK, I like the direction so we at least do it faster to potentially empty the queue faster. That is a good start, I would go for max delay of 5 days (432000000). Decrease the normal delay to 5min and like you mentioned increase the retries to 6x. After some calculations, I propose:

```yaml
                retry_strategy:
                    max_retries: 6
                    delay: 300000
                    multiplier: 4
                    max_delay: 432000000
```

Explanation:

- 300000 ms = 5 min.
- Max reties of 6 (so 6 times *  delay * multiplier)
- And multiplier of 4. Meaning: 5 min, 20 min, ~1 hour, ~5 hours, 1 day, 3.5 days with a max delay of 5 days.

BentiGorlich:
```yaml
                retry_strategy:
                    max_retries: 8
                    delay: 400000
                    multiplier: 6
                    max_delay: 86400000
```
This way we have these delays:
1. ~7 minutes
2. 40 minutes
3. 4 hours
4. 24 hours
5. 24 hours
6. 24 hours
7. 24 hours
8. 24 hours

So we have an even number as a cap which is exactly 24 hours and we retry for a total of 5 days  4 hours and ~47 minutes. The more frequent retry means that messages can quit the queue faster however in step 4-8 they will be in the same delay queue because they are all delayed by the same amount in that step

docs/admin_guide.md
```diff
@@ -908,7 +908,7 @@ With the following content:
 
 ```ini
 [program:messenger]
-command=php /var/www/mbin/bin/console messenger:consume async async_ap failed --time-limit=3600
+command=php /var/www/mbin/bin/console messenger:consume old async outbox deliver inbox resolve receive failed --time-limit=3600
 user=www-data
```
melroy89, line 912: Down below (or above) briefly explain the queues in a simple table, with DSN +  the meaning / usage of each queue. Let me help you:


```md
| Queue name | DSN        | Description |
|------------|------------|-------------|
| old        | RabbitMQ   | Some text.. |
| async      | RabbitMQ   |             |
| outbox     | RabbitMQ   |             |
| deliver    | RabbitMQ   |             |
| inbox      | RabbitMQ   |             |
| resolve    | RabbitMQ   |             |
| receive    | RabbitMQ   |             |
| failed     | PostgreSQL |             |
| dead       | PostgreSQL |             |
```

BentiGorlich: I extended the documentation in the admin guide. Should be enough, shouldn't it?

melroy89: Yea that should work :) I did like my table tho hahahaha

docs/admin_guide.md
```diff
@@ -961,7 +971,7 @@ With the following content:
...
 numprocs=4
```
melroy89, line 976: What about increasing this number.. Could be a good idea right? I ask this, because again I don't want to change this configuration for admins. So maybe increasing this number to like 6 is maybe not a bad idea?

BentiGorlich: true, 6 sounds like a useful number. We should then also adjust the docker config to run 6 replicas of the messenger container by default

melroy89 (dismissed):
Works!

docs/admin_guide.md
melroy89: I changed it to 6 now. Docker currently still only uses 1 process it seems.

melroy89: So we either need to setup supervisord inside the docker image as well.. Or.. you could leverage Docker Swarm (deploy) feature, like this: 

https://github.com/pavelsur07/womjoy-store/blob/8d820cfedcc902c883f38819395ef43a588d6b7b/docker-compose-production-ru.yml#L351

BentiGorlich: @nobodyatroot mentioned something about docker replicas a while ago, so no docker swarm (I think)

melroy89: ow my bad, As you can see in the link I posted above I do mean [replicas](https://docs.docker.com/compose/compose-file/deploy/#replicas), I thought it was a swarm thingy, but apparently not. Apparently in the normal docker compose spec 2.19.1.

BentiGorlich:
Ok, I think this PR is now final and could be merged, if we get #522 in main

melroy89:
> Ok, I think this PR is now final and could be merged, if we get #522 in main

Agreed

melroy89 (dismissed):

nobodyatroot (approved):
Looks good to me, no issues in production after making sure I updated the supervisor configuration.

melroy89:
Regression fixed in: https://github.com/MbinOrg/mbin/pull/645

melroy89:
Another regression wrt doctrine transport explained in PR: https://github.com/MbinOrg/mbin/pull/1874

PR #597, opened 2024-03-17
Rewrite the `ChainActivityHandler`

BentiGorlich:
- remove all redispatches of the `ChainActivityMessage`, just get all dependencies one by one in a recursive call
- adjust the 3 main sources for it: `LikeHandler`, `DislikeHandler` and `AnnounceHandler` (basically like, dislike and boost something that is not in our db)
- adjust the `ApActivityRepository` for better readability and remove the union select with 4 individual ones

For testing:
- you need 2 instances
- create a magazine and subscribe to it from both instances
- post something there
- at the other instance the post should be created automatically, purge it via the "moderate" panel
- from the instance where you didn't purge it: like/dislike/boost/comment on it
- The post should get recreated on the other instance

I tried it with 20 nested comments and purged the entire post. After liking the first comment everything got recreated

I really hope that this fixes one of our worst circular message offenders


src/MessageHandler/ActivityPub/Inbox/ChainActivityHandler.php
```diff
@@ -35,121 +33,83 @@ public function __construct(
...
 
-        if ($dislike) {
-            $this->bus->dispatch(new DislikeMessage($dislike));
+            if (\array_key_exists('inReplyTo', $object) && null !== $object['inReplyTo']) {
+                $parentUrl = \is_string($object['inReplyTo']) ? $object['inReplyTo'] : $object['inReplyTo']['id'];
+                $meta = $this->repository->findByObjectId($parentUrl);
+                if (!$meta) {
+                    $this->retrieveObject($parentUrl);
+                }
+                $meta = $this->repository->findByObjectId($parentUrl);
```
melroy89, line 91: Is this intentionally? You are doing the same call twice now?

src/Repository/ApActivityRepository.php
```diff
@@ -27,12 +29,55 @@ public function __construct(ManagerRegistry $registry, private SettingsManager $
...
+        }
+
+        $conn = $this->_em->getConnection();
+        $tables = [
+            ['table' => 'entry', 'class' => Entry::class],
+            ['table' => 'entry_comment', 'class' => EntryComment::class],
+            ['table' => 'post', 'class' => Post::class],
+            ['table' => 'post_comment', 'class' => PostComment::class],
+        ];
+        foreach ($tables as $table) {
```
melroy89, line 50: Maybe a single query instead of looping over multiple tables could be faster/more efficient for postgresql.

src/MessageHandler/ActivityPub/Inbox/ChainActivityHandler.php
BentiGorlich: yes it is intentionally. It checks whether the object exists, if it doesn't is calls a method that should create it and afterwards it checks if it actually did create it. Maybe I could make `retrieveObject` return the object instead then I wouldn't need the second db call...

src/Repository/ApActivityRepository.php
BentiGorlich: It previously was a union. I don't know whether it would be faster, but I don't think that this is slower, since it could happen that we only execute one or two of the queries if it finds the object.

asdfzdfj:
might have to give this a try later

also for a second opinion, I also have [my own take](https://github.com/asdfzdfj/mbin/commit/d106da854657cbe25a3584503a51c81f26624d18) on reworking chain activity handler if anyone's interested, albeit it was made with a different rationale behind them, do let me know if you want PR of that patch

BentiGorlich:
> might have to give this a try later
> 
> also for a second opinion, I also have [my own take](https://github.com/asdfzdfj/mbin/commit/d106da854657cbe25a3584503a51c81f26624d18) on reworking chain activity handler if anyone's interested, albeit it was made with a different rationale behind them, do let me know if you want PR of that patch

I think I actually like my approach better, because it is more straight forward in what it does. But if you think differently I am happy to discuss how we want to change the handler (I mean it, I do not think that my code is just better :sweat_smile: )
The main difference in our code is that you build the chain first and then create db entries for them vs my approach to just do it directly after fetching the chain objects

Your code definitely has one giant problem though (if I am not mistaken): current `ChainActivityMessages` in delay queues can not be parsed as your new version of the `ChainActivityMessage` which is probably (maybe) the cause of the `Invalid AMQP Data` error which completely nukes the messengers (constant restarting without consuming the message). That's why I didn't want to change the message signature.

src/Repository/ApActivityRepository.php
e-five256: It may have a bit to do with how often it has to establish a connection, as that can be one of the more expensive things compared to a query. Not sure if we have connection pooling to postgres already; if not we might want to put pgbouncer on the list just like amqproxy

asdfzdfj:
> Your code definitely has one giant problem though (if I am not mistaken): current `ChainActivityMessages` in delay queues can not be parsed as your new version of the `ChainActivityMessage` which is probably (maybe) the cause of the `Invalid AMQP Data` error which completely nukes the messengers (constant restarting without consuming the message). That's why I didn't want to change the message signature.

you do raise an interesing point in terms of ensuring operation continuity, and yeah my patch does fall short on that part, but I think that could be work around: changed the name for the new `ChainActivityMessage` (`BuildChainActivityMessage`?), shim the existing `ChainActivityMessage` to redispatch itself into a new one, while also adjust the dispatching in like/announce handler to dispatch the new one, and then "deprecates" the old `ChainActivityMessage` after a release or two

though I have to say that I don't like the prospect of changing message signature nuking messenger systems very much if that's the case

BentiGorlich:
> you do raise an interesing point in terms of ensuring operation continuity, and yeah my patch does fall short on that part, but I think that could be work around: changed the name for the new ChainActivityMessage (BuildChainActivityMessage?), shim the existing ChainActivityMessage to redispatch itself into a new one, while also adjust the dispatching in like/announce handler to dispatch the new one, and then "deprecates" the old ChainActivityMessage after a release or two

yeah that would work I think

> though I have to say that I don't like the prospect of changing message signature nuking messenger systems very much if that's the case

Yeah we definitely need to test that. If that is correct we really have to keep it in mind (especially me, since I probably nuked it multiple times in the past with those changes)

> as for this part I'm afraid that it'll devolve into mostly value differences, as I feel like "straight forward" for you and me is quite different here

Probably true. However I feel like the job of the ChainActivityHandler is super simple: fetch the missing parents of x activity and redispatch x afterwards.

My main goal was to get rid of the self dispatch, as I think it is just to easy to mess that up, be that in the initial implementation or in changes coming in the future. 
I think my implementation achieves that. Yes it is doing a recursive method call, but imo that's way better and more failure resistant than doing it via a redispatch. I don't think that the maybe long runtime is a problem here, though that is quite difficult to say. For example you could have a nested comment with 20 parents coming before it, everyone with a unique image that needs to be fetched and none of them are from the same instance. That would probably take a while... But I prefer a long running message over a self dispatching one.

Again you are probably right a lot of it is preference, so maybe I should shut up and let melroy, e-five, debounced and others decide on what implementation they prefer :sweat_smile:

src/Repository/ApActivityRepository.php
melroy89: we have a connection pool (see `config/packages/doctrine.yaml`), but still:

```
    doctrine:
        orm:
            auto_generate_proxy_classes: false
            proxy_dir: '%kernel.build_dir%/doctrine/orm/Proxies'
            query_cache_driver:
                type: pool
                pool: doctrine.system_cache_pool
            result_cache_driver:
                type: pool
                pool: doctrine.result_cache_pool

    framework:
        cache:
            pools:
                doctrine.result_cache_pool:
                    adapter: cache.app
                doctrine.system_cache_pool:
                    adapter: cache.system
```

melroy89: But maybe we are missing `default_doctrine_dbal_provider: 'doctrine.dbal.default_connection'`? https://symfony.com/doc/current/cache.html

melroy89: We might want to add a PDO cache adapter: `default_pdo_provider` (shortcut)

BentiGorlich: but the connection is created outside of the foreach...

melroy89: I just answer the question of @e-five256 . But I will close this review comment. If you want to do multiple queries in a for loop. OK.

nobodyatroot (approved):
So far so good, everything seems to be working fine.

src/MessageHandler/ActivityPub/Inbox/ChainActivityHandler.php
melroy89: `retrieve` it's in the name yes 👍🏽

melroy89: Shouldn't be blocking for now.

melroy89 (approved):

PR #602, opened 2024-03-18
Only do an update actor if it is actually necessary

BentiGorlich:
This has an enourmous impact. In the Activity Handler we dispatch an `UpdateActorMessage` every time an Activity comes on for a user which has not been fetched for an hour or two. If the server is down for a bit we end up with an enourmus amount of these messages. Now we didn't check if we actually need to update an actor in the Handler for the message. This PR adds that

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -265,6 +265,10 @@ class ActivityPubManager
         $this->logger->info('updating user {name}', ['name' => $actorUrl]);
         $user = $this->userRepository->findOneBy(['apProfileId' => $actorUrl]);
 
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):

e-five256:
is `apFetchedAt` ever `null` coming into this? like on first fetch?

BentiGorlich:
> is `apFetchedAt` ever `null` coming into this? like on first fetch?

could be :thinking:

BentiGorlich:
would still be working though

BentiGorlich:
grafik

e-five256 (approved):

PR #603, opened 2024-03-18
Remove password and 2fa options for sso controlled users

### LINKED ISSUE #576: Improve sso-only mode
We are discussing of implementing an Mbin instance in our company to manage our communities and would like to disallow the use of user/password registration and authentication, using only an Azure OIDC provider.

A configuration + env variable allowing to not display the login and registration forms but only the configured OIDC buttons.
Also I believe hiding the MFA and password change tabs of the profile would be a good idea

Leaving the form available, but this would lead to errors from our users

**Additional context**
We would be able to propose a PR if the feature request is accepted

e-five256:
Remove the password and 2fa setting options for users that have oauth configured

If the user attempts to access the password / 2fa pages manually, gets a 403 access denied response

2fa remove/disable is still allowed in code, so admins should still be able to remove 2fa if for some reason it was or gets enabled, but enabling is not allowed

---

Fixes #576


src/Entity/User.php
```diff
@@ -763,6 +763,11 @@ public function removeOAuth2UserConsent(OAuth2UserConsent $oAuth2UserConsent): s
         return $this;
     }
 
+    public function isSsoControlled(): bool
+    {
+        return $this->oauthGithubId || $this->oauthGoogleId || $this->oauthFacebookId || $this->oauthKeycloakId || $this->oauthZitadelId;
```
e-five256, line 768: this is obviously prone to getting out of date if new identity providers are added. Not sure if there's a better way, or maybe can think of one later. We could add a column on user that they are sso controlled, but that means we have to be sure to set it to true every time a new provider is added which is also possible to mess up, but perhaps less likely

nobodyatroot:
Does it also make sense to remove the email tab since it requires a password to change?

image

e-five256:
Dang, I clearly got it wrong in the issue. I assume that means it's impossible to update an SSO controlled users email, as obviously there's no current password for them to use.

I suppose I'll keep the page, but remove the form to update it, so you can see your current email value. If it ever changes upstream (downstream?) it won't update here, but at least users will be able to see that and potentially ask an admin to fix it until we have a way to resync with the SSO provider somehow

e-five256:
It now looks like this

nobodyatroot:
So...another problem (maybe), if an SSO user wants to delete their account, should it be allowed?

image

e-five256:
Is your comment about deleting / tombstoning the account, or the fact that it requires current password (and thus isn't currently possible)?

nobodyatroot:
> Is your comment about deleting / tombstoning the account, or the fact that it requires current password (and thus isn't currently possible)?

The latter, without a password its not possible for an SSO user to delete their account now.

nobodyatroot (approved):
SSO user delete seems like it's going to require a much bigger PR to address, so I'll approve this one as is since it's out of scope from the original intent.

PR #604, opened 2024-03-18
Add admin option to disable SSO registrations

### LINKED ISSUE #518: Third party login methods allow for new accounts when registration is disabled
**Describe the bug**

Right now third party authentication methods go through a few steps, mainly checking if a user already exists and if not, creating one.

[In between those two steps](https://github.com/MbinOrg/mbin/blob/225a50b9dba9957b7518e250c227df1da6bb815a/src/Security/FacebookAuthenticator.php#L79), they should check if registration is disabled on the site and if so, return an error. This ideally allows users that already have registered to still login, while blocking new users

All

**To Reproduce**
1. Disable registration
2. Attempt to login with facebook/google on a new email address
3. Observe that a new account is created

**Expected behavior**

Registrations being disabled blocking new accounts from being created

e-five256:
Add admin option to disable SSO registrations while still allowing SSO logins

image

Right now as described in the linked issue, disabling site registration does not stop new account creation via SSO methods. This adds a new overall admin setting to disable all sso options new account registration, while allowing logins to still function for existing users

the authenticators have 3 stages:
1. sso id already existed, login
2. email already in use, remap to sso
3. new account completely

I chose to implement the check between options 2 and 3, as if the user was already able to make a native account with the email, it seems like there's no harm in allowing to remap to the SSO option, but I could move it up between 1 and 2 instead if people think that is more correct

suggest hiding whitespace changes when looking at the diff

---

when attempting to register with an account that does not exist, users get redirected to login and shown this message

image

---

Fixes #518


src/Security/KeycloakAuthenticator.php
```diff
@@ -75,11 +78,16 @@ public function authenticate(Request $request): Passport
                 if ($user) {
                     $user->oauthKeycloakId = $keycloakUser->getId();
 
+                    $this->entityManager->persist($user);
```
e-five256, line 81: this was missing from keycloak but appears in all the other authenticators, so seemed like it might be needed

nobodyatroot:
This will remove (or at least deter) a vector currently being used by spammers to bypass our email verification and captcha checks while giving legit users the ability to still use SSO after normal registration... a much needed feature for those of us with SSO enabled.

e-five256:
@BentiGorlich @thepaperpilot since you both were involved with the discussion on the issue wanted to make sure this made sense to both of you for how it was implemented

and ofc don't worry about getting this in the zitadel stuff, I'll make sure to do that if it comes before or after

nobodyatroot (approved):
Functional review using dev environment with Github provider works as expected.

BentiGorlich:
Does this new setting default to true or will it have to be enabled for instances that have some SSO providers enabled?

Also @nobodyatroot can you explain what you mean exactly?

nobodyatroot:
@BentiGorlich When register with SSO is enabled (or just having Google/Facebook/Github SSO enabled before this PR), there is no email verification or captcha check... I've been hammered multiple times with spammers signing up using this feature because it bypasses these spam "checks", I would turn SSO off but I have legit users using Google to signin.

BentiGorlich:
Ah ok. Yeah I have noticed that most spam accounts on mastodon use gmail addresses. I would have expected that Hoogle would do a better job at preventing spam accounts

nobodyatroot:
> Ah ok. Yeah I have noticed that most spam accounts on mastodon use gmail addresses. I would have expected that Hoogle would do a better job at preventing spam accounts

Me too, but I guess if they're not actively using the Gmail account to spam emails, then they don't care....

e-five256:
> Does this new setting default to true or will it have to be enabled for instances that have some SSO providers enabled?

Yep! https://github.com/MbinOrg/mbin/blob/bfa4ea91e8d140c00c050061521d1911a945535d/src/Service/SettingsManager.php#L69

should be no unexpected changes for admins pulling this in. I could see it be a little confusing for admins that don't use SSO at all, since it saying sso registration enabled (which it is by default) when there is none configured has no effect.

There isn't a great way to check whether SSO is enabled without checking every single provider, but I could add that and hide the admin option if none are set, or just change the option to have a description that says only has an effect when having an sso provider enabled

BentiGorlich:
I think it is fine as is, but a description doesn't burt nobody 😁

PR #606, opened 2024-03-19
Remove empty panel from admin users pages when there is no pagination

e-five256:
Removes the empty block that appears at the bottom of the page here

image

if you don't have enough users to paginate

now when empty it will look like this

image

which I think is ok for now, as an empty table should still be fairly obvious that it's 0 results

I think long term, looking at it we might want to make a user search that can be accessed here to search by criteria like name

### DIFF
--- a/assets/controllers/subject_controller.js
+++ b/assets/controllers/subject_controller.js
@@ -218,7 +218,7 @@ export default class extends Controller {
     async showModPanel(event) {
         event.preventDefault();
 
-        let container = this.element.querySelector('.moderate-inline')
+        let container = this.element.querySelector('.moderate-inline');
         if (null !== container) {
             // moderate panel was already added to this post, toggle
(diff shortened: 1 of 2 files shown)


assets/controllers/subject_controller.js
```diff
@@ -218,7 +218,7 @@ export default class extends Controller {
     async showModPanel(event) {
         event.preventDefault();
 
-        let container = this.element.querySelector('.moderate-inline')
+        let container = this.element.querySelector('.moderate-inline');
```
e-five256, line 221: sneaking in this ultra high priority change that I may or may not have forgotten about

nobodyatroot (approved):

assets/controllers/subject_controller.js
melroy89: ULTRA HIGH PRIO! 🚨

PR #607, opened 2024-03-19
Only unsubscribe or unfollow on block if already doing so

e-five256:
When blocking a magazine or user, an unsubscribe or unfollow event is emitted first regardless of the status of the user. This ends up causing it to be impossible to block magazine/users if the unfollow AP event does not happen successfully, specifically against a remote (say it is down or denying AP events)

This change simply makes it only do the unsubscribe/unfollow if the user is currently subscribed or following. It may be able to still happen if the user is following and the remote goes down, but this seemed like an ok first step. Perhaps a real or better solution would be putting this in a `try catch` and ignoring (certain) errors


nobodyatroot (approved):
Live tested on kbin.run, @e-five256 is happy

PR #608, opened 2024-03-19
Add sso registration disabled check to zitadel

e-five256:
this merges the changes from #604 into #429 

also noticed the persist on existing email flow missing similar to keycloak so added it in


nobodyatroot (approved):
I don't use Zitadel, but I reviewed the code to ensure consistency with https://github.com/MbinOrg/mbin/pull/604.

PR #610, opened 2024-03-21
Admin option - Restrict magazine creation to admins/global moderators

e-five256:
Adds a new admin panel option, disabled by default, that will restrict creation of new **local** magazines to admins and global moderators. Magazines added through remotes should be unaffected.

image

- [x] Admin panel option
- When enabled:
- [x] Hide option in header for regular users
- [x] Hide option in sidebar when logged out
- [x] Deny access to route for regular users
- [x] Deny access in API for regular users
- [x] Only effect local magazines


TheVillageGuy:
I have this 'hardcoded' on my instance now, so I am in favor

e-five256:
Not sure if marking it as ready for review was a mistake, I still have work to do as indicated in the description / todo list, hopefully I can find some time this weekend to finish this up

nobodyatroot (approved):
I tested in dev with each user role using this new option both enabled and disabled, PR functions as expected. Also confirmed you're checking for `apid == null` in the magazine creation, so it shouldn't impact remote magazine creation.

PR #611, opened 2024-03-21
Add missing search method to domain repository

nobodyatroot:
Implementation of search method in domain repository was missing, PR adds it.


nobodyatroot:
I saw this was getting called from the API in production causing 500 errors.

melroy89:
> I saw this was getting called from the API in production causing 500 errors.

Nothing new there hahaha..

PR #616, opened 2024-03-22
Open remote mag/user links in new window, add mag link to sidebar

e-five256:
- Makes the "Go to original instance" links open in a new window
- Adds `noopener noreferrer nofollow` to them all
- Adds a sidebar icon next to magazine name for remote magazines to go to the remote magazine
  - Now that there isn't a banner for remote magazines when they are getting updates, there isn't an easy way to jump to the original magazine if you want to view it. Compare with lemmy that makes the entire magazine name the link to the original location. If people prefer that, I could go with that instead

the only visual difference is the popout icons and federation icon

---

remote magazines

image
image

---

remote users

image
image


BentiGorlich (approved):
I really like it :+1:

PR #618, opened 2024-03-23
Update user popover for remote users

e-five256:
changes loc message to "View on remote instance"

remote user profile page

image

remote user popup

image

remote user OP entry sidebar

image

remote user OP post sidebar

image

---

remote magazine

image

remote magazine sidebar

image

---

<details>
<summary>OLD</summary>

- update user popover for remote users
- new loc token for specifying opening original profiles or magazines~
  - changed the magazine loc token that I just added in #616  as I feel it was also unclear as to the function, but wasn't quite convinced to add a new one, but since we might be adding one for opening profiles, felt we should add one to open for magazines. not sure how clear "open original profile/magazine" is

based on conversation from matrix, the user popover text is unclear as to what its function is. A suggestion was to remove the text and just have the icon next to the username like it now appears for remote magazines

now is the chance for any comments. different text? smaller popout icons of the alert banners? remove the alert banner completely from user profiles? add the remote icon next to usernames in the user overview box as well? feel free to give feedback

image
image

</details>

### DIFF
--- a/assets/styles/layout/_alerts.scss
+++ b/assets/styles/layout/_alerts.scss
@@ -8,8 +8,13 @@
     margin: 0;
   }
 
-  a{
+  a {
     font-weight: bold;
+  }
+
+  i {
+    font-size: 0.8rem;
+    vertical-align: middle;
   }
 
   &__info {
@@ -18,7 +23,7 @@
     color: var(--kbin-alert-info-text-color);
 
     a {
-      color: var(--kbin-alert-info-link-color);    
+      color: var(--kbin-alert-info-link-color);
     }
   }
 
--- a/assets/styles/components/_magazine.scss
+++ b/assets/styles/components/_magazine.scss
@@ -25,6 +25,10 @@
 
   &__name {
     margin-top: 0;
+
+    i {
+      font-size: 0.7rem;
+    }
   }
 
   &__description,
--- a/assets/styles/components/_user.scss
+++ b/assets/styles/components/_user.scss
@@ -39,6 +39,12 @@
 
     opacity: .75;
   }
+
+  &__name {
+    i {
+      font-size: 0.7rem;
+    }
+  }
 }
 
 .user-inline {
--- a/templates/components/magazine_box.html.twig
+++ b/templates/components/magazine_box.html.twig
@@ -23,7 +23,7 @@
             <p class="magazine__name">
                 <span>{{ ('@'~magazine.name)|username(true) }} {% if magazine.isAdult %}<small class="badge danger">18+</small>{% endif %}</span>
                 {% if magazine.apId %}
-                    <a href="{{ magazine.apProfileId }}" rel="noopener noreferrer nofollow" target="_blank" title="{{ 'go_to_original_instance'|trans }}" aria-label="{{ 'go_to_original_instance'|trans }}">
+                    <a href="{{ magazine.apProfileId }}" rel="noopener noreferrer nofollow" target="_blank" title="{{ 'open_original_magazine'|trans }}" aria-label="{{ 'open_original_magazine'|trans }}">
                     <i class="fa-solid fa-circle-nodes" aria-hidden="true"></i></a>
                 {% endif %}
             </p>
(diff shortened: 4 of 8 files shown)


asdfzdfj:
my quick 2c:

I think the external link icon (box with pointing out arrow) is a more suitable icon for opening original actor (user/mags) profile, if anyone's looking to view the original actor profile, the external link icon answer that question better
   (granted, I don't have any supporting evidence other than the fact that akkoma also does it, one might argue that the circle nodes was chosen because lemmy used the fediverse icon for it)
<details><summary>how akkoma does it</summary>

image

</details>

also it looks like sharkey/misskey and whatever newer mastodon that infosec.exchange is running opts to put the alert banner near the top and just use a text link to original profile without icon, so perhaps it's possible to only use icon in a space constrained places (user/mag box) and just use the plain text when more space is available (like the alert banner, or even the user pop over, as it was in a line of its own)
this could also be a point for always show the remote alert banner, even when they're subscribed
<details><summary>how those two does it</summary>

(sharkey)
image

(infosec.exchange)
image

</details>

---

the problem/personal gripe that I brought up in matrix actually has more to do with the translation change, where it works nicely in alert banner for content listing, but not as well in the user popout since you're more likely to be interested in the *actor profile itself* rather than what they put out if you decided to click the link from that context
  and for that, either use a different translation keys and tailor them for the context, or revert to the old, more generic translation or similar, like "View more on the original instance.", sharkey and new mastodon also seems to go this way, as seen in the above example

e-five256:
I wouldn't expect the popout icon to have anything to do with bringing me to the original instance (without the text before it saying it would)... We need a fediverse arrow icon...

---

Darn, did they change mastodon recently, or was it just that subtle always. The _intent_ was to mimic their message when you run out of posts to see of external users, like here:

image

So the banner was changed to show on the last page of pagination, but sadly doesn't seem to appear when using infinite scroll, which hopefully can be fixed

---

I'm fine with reverting the key and making a new one, I checked at it was translated in 2 other languages so far, but I guess ideally those can be reverted as well

asdfzdfj:
when fediverse arrow icon doesn't exist, my answer as to what's the next best thing is the external link icon or similar, for they are first and foremost a link to external resources

still, fair enough, come to think of it, it should be fine to use the circle node for linking to external actor, provided that its use is consistent enough that the users can learn what it means in the context (much like how I could learn that's what the fediverse icon does in lemmy), or, y'know, just use the text or lead with them

tangent: for some odd reason sharkey have another "view on remote" button in the users' 3-dot menu and it also used arrow pointing out icon (apparently it's [share from phosphors icon](https://phosphoricons.com/?q=%22share%22)), this menu seems to only showed up when you're logged in though

<details><summary>screenshot</summary>

image

</details> 

---

well it looks like the alert banner on infosec.exchange is more likely to come from glich-soc rather than upstream mastodon, just rechecked on mastodon.social and it doesn't seems to be there, **apologies for the confusion caused**
(I heard that mastodon just merged a good deal of ui improvements in their latest release and I thought that part comes from this change and not glich-soc's, because some people definitely thought that those ui changes comes from glich-soc although it's not)

<details><summary>screenshot from mastodon.social</summary>

image

</details> 

---

I don't think I have more objection for the translations

in any case I think I'll stop here and let others to chime in, make changes as you see fit

e-five256:
I trust your opinion far more than my own; I'll swap the icons to external link and try to reword the localization. I don't really want to revert it, as I feel the previous one was overly verbose itself. Maybe I will try to unify the two I added into "Browse on the original instance" or whatever akkoma/mastodon are saying

BentiGorlich:
I am fine with whatever you guys decide. 😁

e-five256:
OP updated with an attempt to match feedback. If I was off with anything, just let me know

BentiGorlich:
I think the previous icon looks better, but the "new" icon better communicates what happens when you click on it...

BentiGorlich:
I think my main gripe with the new icon is that it's just tad too big I think

e-five256:
made them smaller

BentiGorlich (approved):

asdfzdfj (approved):

assets/styles/layout/_alerts.scss
```diff
@@ -8,17 +8,22 @@
...
   }
 
-  a{
+  a {
     font-weight: bold;
   }
 
+  i {
+    font-size: 0.8rem;
+    vertical-align: middle;
```
e-five256, line 17: bleh, just noticed this doesn't work in firefox. appears to work fine in chromium browsers. not a big deal, probably, but might want to try to adjust it later

PR #621, opened 2024-03-24
fix cryptic strpos error in LikeHandler

asdfzdfj:
occasionally this cryptic error turned up in failed messages:

    Handling "App\Message\ActivityPub\Inbox\LikeMessage" failed: strpos(): Argument #1 ($haystack) must be of type string, null given

turns out this comes from null guard in Undo/Like handler where it checks actor and entity to not be null, but entity is resolved from activity which *could* be null when the target object isn't known to us, and trying to resolve the entity with that would results in the cryptic error above

moved the empty check in Undo/Like handler from `$entity` to `$activity` since entity is resolved from activity and when the activity has any values then the target object entity exists and can be resolved

also applied this fix to the DislikeHandler as well


BentiGorlich:
I think this is actually solved in #597 :thinking:

src/MessageHandler/ActivityPub/Inbox/LikeHandler.php
```diff
@@ -55,8 +55,10 @@ public function __invoke(LikeMessage $message): void
                 }
 
                 $actor = $this->activityPubManager->findActorOrCreate($message->payload['actor']);
-                // Check if actor and entity aren't empty
-                if (!empty($actor) && !empty($entity)) {
+                // Check if actor and activity aren't empty
+                // the entity can be resolved from activity if it has values
+                if (!empty($actor) && !empty($activity)) {
```
melroy89, line 60: `$activity` is never set?


Also `$entity` is fetch above already on line 52.

asdfzdfj:
superseded by #597

src/MessageHandler/ActivityPub/Inbox/LikeHandler.php
asdfzdfj: in the old way the flow was like (lookup) -> `$activity` -> `$entity`, the null check was at entity although it's already nullable at activity, and due to the way the data from `$activity` is used to resolve the target entity (the `$this->entityManager->getRepository($activity['type'])->find((int) $activity['id'])` idiom) tha'll result in aforementioned error at the `getRepository()` part because `$activity` is null

with #597 that lookup idiom is wrapped and this change no longer applies, as it'll always try to create the entity objects.
still, is it really needed to fetch/create potentially the entire object chain just to undo a (dis)like that most likely will not even exists in our record?

PR #627, opened 2024-03-25
Fix error pollution on null $user object

melroy89:
Add missing `$user` check on possible empty/null object. Causing errors:

```
{"message":"Warning: Attempt to read property \"apFetchedAt\" on null","context":{"exception":{"class":"ErrorException","message":"Warning: Attempt to read property \"apFetchedAt\" on null","code":0,"file":"/var/www/kbin.melroy.org/html/src/Service/ActivityPubManager.php:133"}},"level":300,"level_name":"WARNING","channel":"php","datetime":"2024-03-25T00:47:01.732142+01:00","extra":{}}
```

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -130,7 +130,7 @@ class ActivityPubManager
         if (false === filter_var($actorUrl, FILTER_VALIDATE_URL)) {
             if (!substr_count(ltrim($actorUrl, '@'), '@')) {
                 $user = $this->userRepository->findOneBy(['username' => ltrim($actorUrl, '@')]);
-                if (!empty($user)) {
+                if ($user instanceof User) {
                     if ($user->apFetchedAt->modify('+1 hour') < (new \DateTime())) {
                         $this->bus->dispatch(new UpdateActorMessage($user->apProfileId));
                     }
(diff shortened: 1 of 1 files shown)


nobodyatroot (dismissed):
I don't see the same condition on my instance (yet), but the change makes sense.

src/Service/ActivityPubManager.php
```diff
@@ -130,11 +130,13 @@ public function findActorOrCreate(?string $actorUrlOrHandle): null|User|Magazine
...
-                    $this->bus->dispatch(new UpdateActorMessage($user->apProfileId));
-                }
+                if ($user instanceof User) {
+                    if ($user->apFetchedAt->modify('+1 hour') < (new \DateTime())) {
+                        $this->bus->dispatch(new UpdateActorMessage($user->apProfileId));
+                    }
 
-                return $user;
+                    return $user;
+                }
```
melroy89, line 139: Should we do a return null? Should we do a `createUser`?

BentiGorlich: If you found the user object you should return the user, like you did

BentiGorlich (approved):

src/Service/ActivityPubManager.php
BentiGorlich: And if you mean "should we create the user in the else or return null" my answer would be neither. If the actor doesn't exist it already is created in this method, because we verified that it is a user handle and if we can't find the user locally we do a webfinger request to get the actual URL which we then fetch

melroy89: OK. So this might still return null. Then I think a simple check I added should then be sufficient.

PR #635, opened 2024-03-27
Fix list display in magazine description

BentiGorlich:
The magazine description used weird list styles which were additionally cut off if it was an ordered list. 

| Before | After |
| ------ | ----- |
| Bildschirmfoto vom 2024-03-27 10-20-15 | Bildschirmfoto vom 2024-03-27 10-20-06 |

### DIFF
--- a/assets/styles/components/_magazine.scss
+++ b/assets/styles/components/_magazine.scss
@@ -34,19 +34,17 @@
   &__description,
   &__rules {
     ul {
-      padding-left: .5rem;
+      padding-left: .75rem;
 
       li {
-        list-style: none;
         margin-bottom: .5rem;
       }
     }
 
     ol {
-      padding-left: 0.5rem;
+      padding-left: 0.75rem;
 
       li {
-        list-style: circle;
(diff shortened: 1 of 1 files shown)


e-five256:
I think they were cut off on purpose to make a `>` looking shape rather than the circle?

BentiGorlich:
I think an ordered list should still be an ordered list, especially because it is often used for rules which refer to each other like "unless it violates 2."
So my point is, that I don't get why you would mess with the `list-style` so an ordered list looks like a normal list and a normal list just doesn't look like a list. Doesn't make sense to me, even if it was intentional :sweat_smile:

e-five256 (approved):

BentiGorlich:
maybe more feedback is needed?
@asdfzdfj @melroy89 @nobodyatroot

nobodyatroot (approved):
Yea, seems like a hack job before if it was intentional... I'm fine with making it "proper".

melroy89 (approved):
lgtm

PR #636, opened 2024-03-27
Fix width class missing if `KBIN_PAGE_WIDTH` has not been set

BentiGorlich:
When the page width has not been explicitly set the `width--fixed` class was missing from the kbin contianer. That lead to the subscription sidebar not having a max width leading to this display:
grafik


e-five256:
It'd also be doable by removing `width--fixed` [here](https://github.com/MbinOrg/mbin/blob/b68b2dd8dc45451d003d1844aa534840e39cdf8d/assets/styles/components/_sidebar-subscriptions.scss#L5) as that's how the main container is styled (the lack of class [means not the other other two](https://github.com/MbinOrg/mbin/blob/b68b2dd8dc45451d003d1844aa534840e39cdf8d/assets/styles/layout/_layout.scss#L11-L12)) but this also works

melroy89:
Ow so that was happening to me randomly!

PR #638, opened 2024-03-28
restore css to limit figure preview size

asdfzdfj:
added back css that limits the size of attached figure image to not exceeds 600x500px, the same size used when generating the thumbnails

back when making the new sensitive figure, there was a bit of problem in keeping the image size limited without destroying the aspect ratio or causes image stretching, so that was left out and relied on generated thumbnails for limiting the figure size, however with the later introduced large image handling where the link to original image may be used for the figure instead, which can make the rendered figure bigger than what it should be

### DIFF
--- a/assets/styles/components/_figure_image.scss
+++ b/assets/styles/components/_figure_image.scss
@@ -15,6 +15,8 @@
     display: block;
     width: 100%;
     height: 100%;
+    max-width: 600px;
+    max-height: 500px;
   }
 }
 


melroy89 (approved):

TheVillageGuy:
I would appreciate 600x600 so square images will fit perfectly

asdfzdfj:
> I would appreciate 600x600 so square images will fit perfectly

you should still get the square image, albeit at 500x500px because it hit the height limit first

TheVillageGuy:
My thumbnails are generated at 600x600, it all looks fine.

TheVillageGuy:
Screenshot_20240329-104426

PR #639, opened 2024-03-28
AttachEntryEmbedHandler add missing image source url

asdfzdfj:
it seems like #501 strikes again, occasionally leaving entries with broken thumbnail when it was attached using AttachEntryEmbedHandler

fix the AttachEntryEmbedHandler to properly attach a link to the created cover image entity if the resulting entity doesn't have their filePath set, most likely because the image is too big

### DIFF
--- a/src/MessageHandler/AttachEntryEmbedHandler.php
+++ b/src/MessageHandler/AttachEntryEmbedHandler.php
@@ -45,38 +45,41 @@ class AttachEntryEmbedHandler
             return;
         }
 
-        $html = $embed->html;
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):
Good fix, I/we (@e-five256) have been seeing this occasionally.

e-five256:
I'm not seeing most thumbnails anymore, after it sounds like @nobodyatroot sync'd these changed onto kbin.run

| kbin.run | fedia |
| ------ | ----- |
| image | image |

PR #641, opened 2024-03-28
Bye Dragonfly, Hello KeyDB

melroy89:
Dragonfly was misbehaving from day 1 actually.  I already try to fix some configurations via PR (in vain): https://github.com/MbinOrg/mbin/pull/634, but I left Dragonfly. If users want to switch, try KeyDB instead...

Feel free to experiment with Dragonfly if you wish, and add it back to the documentation if you're sure it works properly. But I'm out.

This is how my CPU IO Wait looked like:

image
image

### DIFF
--- a/docs/admin_guide.md
+++ b/docs/admin_guide.md
@@ -307,7 +307,7 @@ composer clear-cache
 
 ### Caching
 
-You can choose between either Redis or Dragonfly.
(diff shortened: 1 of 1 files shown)


melroy89:
Maybe I will so add https://github.com/valkey-io/valkey

nobodyatroot (approved):

PR #642, opened 2024-03-28
Make social login more prominently visible part 2

TheVillageGuy:


e-five256:
I can definitely see your point from your screenshots that the UX of registration could be improved.

From that, I think what I'd like in the end is:

- admin option to specify whether SSO options on both login and registration page are first or last
- buttons to take up the same width as the form, rather than being super small and only the size of the text content
- add color to the buttons if possible. For example, make the facebook social button blue, assuming that isn't trademarked/registered
- Turn the login and reset password links into buttons rather than text links, matching register
- add horizontal bars between the sections to denote flow changes: namely between SSO, native+register, login+reset

<details>
<summary>Taking a look at a random example</summary>

One thing I was comparing ours to is this one. Although they do sort of hide the sign in link, I think it makes the sections and options clearer and most likely looks better on mobile than what we have

image


</details>

templates/user/register.html.twig
```diff
@@ -21,6 +21,10 @@
                 {% for flash_error in app.flashes('verify_email_error') %}
                     <div class="alert alert__danger">{{ flash_error }}</div>
                 {% endfor %}
+                {% if mbin_sso_registrations_enabled() %}
+                    <h4 class="text-muted" style="text-align: center; margin:0 !important;">{{ 'Register using' }}</h4>
```
BentiGorlich, line 25: To use the translation system you can change it to
```suggestion
                    <h4 class="text-muted" style="text-align: center; margin:0 !important;">{{ 'register_using_sso'|trans }}</h4>
```
Then you have to create an entry in `messages.en.yml` with the key `register_using_sso`

BentiGorlich (commented):
When you add the translation, I think it is good to go :)

TheVillageGuy:
I don't have a dev server set up currently so won't be making changes. As
soon as I do I'll finish it and do some other things as well

PR #643, opened 2024-03-28
Make social login more prominently visible part 1

TheVillageGuy:
Melroy was here: You could also update this description here instead of a new comment below ;)

### DIFF
--- a/assets/styles/components/_login.scss
+++ b/assets/styles/components/_login.scss
@@ -24,7 +24,7 @@
       display: flex;
       gap: 2rem;
       justify-content: center;
-      margin-top: 3rem;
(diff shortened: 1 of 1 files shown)


TheVillageGuy:
Before: 
image
After:
image

e-five256:
These images don't seem to correspond to the changes, did you mean to do only https://github.com/MbinOrg/mbin/pull/642 ?

TheVillageGuy:
The screenshots show the result of both this and #642

melroy89:
Since you already combine those two PRs. I believe they belong together. Then also create a single pr with these changes together, I don't think two separate PRs are necessary.

PR #648, opened 2024-03-31
Add page to view specific entry comments

### LINKED ISSUE #406: go to comment doesn't work / Notification link is unaware of pagination
**Describe the bug**
A copy of the [bug on kbin](https://codeberg.org/Kbin/kbin-core/issues/370)

> When there is a reply to a comment of mine, I clicked on the notification. But the link goes to page 1.. While the reply is on page 2.
>
> Kbin notification link is: https://kbin.melroy.org/m/RedditMigration@kbin.social/t/9/23k-New-Lemmy-and-Kbin-Accounts-Created-in-Last-Hour#entry-comment-203 (which is incorrect).
>
> Since the comment in on the 2nd page. The notification link should have been: https://kbin.melroy.org/m/RedditMigration@kbin.social/t/9/23k-New-Lemmy-and-Kbin-Accounts-Created-in-Last-Hour?p=2#entry-comment-203 (notice the ?p2)

Possibly opened by @melroy89 himself.

fedia.io

**To Reproduce**
1. Find a post with more than one comment page
2. Go to comment page 2+
3. Open link to any comment on that page in a new tab

**Expected behavior**

The link should open the post with the comment itself and subcomments (just like on reddit and lemmy).

**Screenshots**

On lemmy  https://hexbear.net/comment/3753319
image

On mbin https://fedia.io/m/meta@lemm.ee/t/215813/Hexbear-federation-megathread#entry-comment-2565198 should show 
image
 but instead it just shows 
image

 - OS: NixOS 23.11
 - Browser Firefox
 - Version 121

**Additional context**

Lemmy and Reddit have already solved the problem and probably mbin should follow suit.

e-five256:
This creates a new view that is specifically for viewing an entry comment.

Problems attempting to be solved:
- Viewing an entry comment right now requires a login
  - This is because the page that it loads (`EntryCreateController`) requires `#[IsGranted('ROLE_USER')]`
- Go to entry comment does not work with pagination
  - For notifications / modlog / reports, the url generated is just the entry url with a url fragment of the entry comment. If the entry comment is not on the first page, it does not work
  - This doesn't technically _fix_ that problem so much as the aforementioned areas now link specifically to a view of that specific comment

The page loaded here shows an entire comment chain from root, including all its replies. It proved difficult to trim off comments from either direction as everything under a root is cached html so any changes would reflect how it's cached (there's more details of my issues collapsed at the bottom of the PR)

---

- [x] viewing specific comments no longer requires login
- [x] notification / modlog / report link the specific comment rather than overall thread
- [x] show parent / replies in view
- [x] ensure AP for entry / entry comment views is unaffected

:x:  only clear notifications related to comment, not entire entry (out of scope imo)

---

image

<details>
<summary>other theme examples</summary>

outline is less noticeable in light themes but the browser view still does jump down to the comment using url fragment so that helps (I'm manually scrolling up for screenshots)

| theme | |
|-|-|
| dark | image |
| light | image |
| solarized light | image |
| solarized dark | image |

</details>

| notifications | clicking link loads... |
|-|-|
| image | image |

| report panel | clicking link loads... |
|-|-|
| image | image |

| mod log | clicking link loads... |
|-|-|
| image | image |

---

<details>
<summary>challenges faced</summary>
Open issues:

I've been stuck on one problem for a very long time:

I imagine people would prefer this to show replies in at least one direction if not both. Specifically, show comments that replied to this or the parent that it was replying to.

I was having a lot of trouble with that, and doing the always show timestamp PR helped me understand why, the replies to a top level entry are all pulled in via `entry_comments_nested` component which is _heavily_ cached by the root parent id. If I tried to show the nested replies, instead it would duplicate the entire thread from top to bottom including the entry comment you were looking at...

And if you tried to trim comments, if you loaded the comment view first and went to the full thread, those comments would be missing since it was cached together.

If you tried to change the cache key so everything was cached differently, there would still be oddities (level of comment chain for instance)

---

As commented, viewing an entry comment marks all notifications for that entry as viewed. This isn't different than how it's always been, but is more noticeable now that you might see less of the comments.
</details>

---

Fixes #406


templates/entry/comment/view.html.twig
```diff
@@ -0,0 +1,56 @@
...
+{% block stylesheets %}
+    {{ parent() }}
+
+    <style>
+        #entry-comment-{{comment.id}} {
+            border-top: var(--kbin-alert-danger-border);
+            border-bottom: var(--kbin-alert-danger-border);
+            border-right: var(--kbin-alert-danger-border);
+        }
+    </style>
```
e-five256, line 22: this is a bit hacky where; I can't modify the css in the nested comments (like adding `highlighted` to the comment class list) as it's cached html, so instead this sets the css for the comment that is being viewed to style it differently. It could also be done via js but this doesn't seem _that_ bad,  but I could be wrong

nobodyatroot (approved):
I don't see a good reason to hold this up, speak now or forever hold your peace @asdfzdfj @BentiGorlich @melroy89

PR #650, opened 2024-04-01
AttachEntryEmbedHandler add missing image source url (take 2)

asdfzdfj:
let's try this again, apparently the old take on this patch broke link cover fetching, so this one only contains just the fix and nothing else

it seems like #501 strikes again, occasionally leaving entries with broken thumbnail when it was attached using AttachEntryEmbedHandler

fix the AttachEntryEmbedHandler to properly attach a link to the created cover image entity if the resulting entity doesn't have their filePath set, most likely because the image is too big

### DIFF
--- a/src/MessageHandler/AttachEntryEmbedHandler.php
+++ b/src/MessageHandler/AttachEntryEmbedHandler.php
@@ -67,16 +67,27 @@ class AttachEntryEmbedHandler
     private function fetchCover(Entry $entry, Embed $embed): ?Image
     {
         if (!$entry->image) {
-            $tempFile = null;
-            if ($embed->image) {
-                $tempFile = $this->fetchImage($embed->image);
-            } elseif ($embed->isImageUrl()) {
-                $tempFile = $this->fetchImage($entry->url);
-            }
+            if ($imageUrl = $this->getCoverUrl($entry, $embed)) {
+                if ($tempFile = $this->fetchImage($imageUrl)) {
+                    $image = $this->imageRepository->findOrCreateFromPath($tempFile);
+                    if ($image && !$image->filePath) {
+                        $image->sourceUrl = $imageUrl;
+                    }
 
-            if ($tempFile) {
-                return $this->imageRepository->findOrCreateFromPath($tempFile);
+                    return $image;
+                }
             }
+        }
+
+        return null;
+    }
+
+    private function getCoverUrl(Entry $entry, Embed $embed): ?string
+    {
+        if ($embed->image) {
+            return $embed->image;
+        } elseif ($embed->isImageUrl()) {
+            return $entry->url;
         }
 
         return null;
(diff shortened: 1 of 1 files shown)


BentiGorlich:
My 2c: I think the source URL should **never** be null, it should always be set regardless of the local save. We want to eventually be able to clear older images that can be refetched when linked content is accessed. For that to work we need to have the source urls of all images (sure we can always exclude images without a source url from a clean)

src/MessageHandler/AttachEntryEmbedHandler.php
```diff
@@ -67,16 +67,27 @@ public function __invoke(EntryEmbedMessage $message): void
...
-            if ($embed->image) {
-                $tempFile = $this->fetchImage($embed->image);
-            } elseif ($embed->isImageUrl()) {
-                $tempFile = $this->fetchImage($entry->url);
+            if ($imageUrl = $this->getCoverUrl($entry, $embed)) {
+                if ($tempFile = $this->fetchImage($imageUrl)) {
+                    $image = $this->imageRepository->findOrCreateFromPath($tempFile);
+                    if ($image && !$image->filePath) {
+                        $image->sourceUrl = $imageUrl;
+                    }
```
BentiGorlich, line 75: I'd suggest to always write the source url and I am not sure, but I think a `persist` call is needed
```suggestion
                    if ($image) {
                        $image->sourceUrl = $imageUrl;
                        $this->entityManager->persist($image);
                    }
```

asdfzdfj: - it looks like `Entity::$image` is already cascade persisted (https://github.com/MbinOrg/mbin/blob/12ecacfc4d12e597792f0ea2d7316ffcb9425619/src/Entity/Entry.php#L80-L82) so I don't think an extra `persist()` call is needed (and if it is, we should already see its detrimental effects long before the current occasional thumbnail went missing problem)
- always recording source url seems doable, I'll add that, though keep in mind that it might overwrite the previous url if the returned image entity is already known and have a different sourceUrl set, or perhaps worse, unintentionally "downgrading" an uploaded image because the both sides used/reposted the same image, and further down the line the user might not be too pleased to see that their supposedly local uploaded image got swept away from remote image cleaning routine

BentiGorlich: Maybe just setit when there is not one set yet?

asdfzdfj: added, see second commit

BentiGorlich (approved):
From a pure code review it LGTM. However I did not test it. If you're confident that this won't break anything we can skip the test part :)

BentiGorlich:
Ok sounds good. As far as I am concerned you can merge it then

e-five256:
Think he might be taken up with the tag stuff, so I'll take a look right now

e-five256 (approved):
lgtm

PR #653, opened 2024-04-01
Remove obsolete package-versions-deprecated

melroy89:
During the new docker setup, I notice we still use this very old `package-versions-deprecated` in your composer.json. We don't need it afaik. Thus I removed it.

This package is EOL, we don't use it. Thus why do we still have it?


asdfzdfj (approved):

nobodyatroot:
Merging for @melroy89 while he's on sabbatical.

PR #655, opened 2024-04-02
Fix error on local user comments on local users

e-five256:
When fetching the AP for a local user replying to a local user, a 500 was being thrown:

```
Error:
Call to a member function modify() on null

  at src/Service/ActivityPubManager.php:134
  at App\Service\ActivityPubManager-&gt;findActorOrCreate('@user1')
     (src/Service/ActivityPub/Wrapper/MentionsWrapper.php:29)
  at App\Service\ActivityPub\Wrapper\MentionsWrapper-&gt;build(array('@user1'), 'remote entry comment')
     (src/Factory/ActivityPub/EntryCommentNoteFactory.php:71)
  at App\Factory\ActivityPub\EntryCommentNoteFactory-&gt;create(object(EntryComment), true)
     (src/Controller/ActivityPub/EntryCommentController.php:41)
  at App\Controller\ActivityPub\EntryCommentController-&gt;__invoke(object(Magazine), object(Entry), object(EntryComment), object(Request))
     (vendor/symfony/http-kernel/HttpKernel.php:181)
  at Symfony\Component\HttpKernel\HttpKernel-&gt;handleRaw(object(Request), 1)
     (vendor/symfony/http-kernel/HttpKernel.php:76)
  at Symfony\Component\HttpKernel\HttpKernel-&gt;handle(object(Request), 1, true)
     (vendor/symfony/http-kernel/Kernel.php:197)
  at Symfony\Component\HttpKernel\Kernel-&gt;handle(object(Request))
     (vendor/symfony/runtime/Runner/Symfony/HttpKernelRunner.php:35)
  at Symfony\Component\Runtime\Runner\Symfony\HttpKernelRunner-&gt;run()
     (vendor/autoload_runtime.php:29)
  at require_once('/var/www/mbin/vendor/autoload_runtime.php')
     (public/index.php:7)
```

this line https://github.com/MbinOrg/mbin/blob/823c83ca9f92fa4da42f6aba861012cfb97b8f5f/src/Service/ActivityPubManager.php#L134

was being run on local users, and modify was throwing an error when it was null. this check makes it only emit the dispatch if the user is remote

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -131,7 +131,7 @@ class ActivityPubManager
             if (!substr_count(ltrim($actorUrl, '@'), '@')) {
                 $user = $this->userRepository->findOneBy(['username' => ltrim($actorUrl, '@')]);
                 if ($user instanceof User) {
-                    if ($user->apFetchedAt->modify('+1 hour') < (new \DateTime())) {
+                    if (isset($user->apId) && $user->apFetchedAt->modify('+1 hour') < (new \DateTime())) {
                         $this->bus->dispatch(new UpdateActorMessage($user->apProfileId));
                     }
 


melroy89:
Could you also check on line: https://github.com/MbinOrg/mbin/blob/57a816d082006b2385536a298ff270bb476529b3/src/Service/ActivityPubManager.php#L167 ?

melroy89 (dismissed):

e-five256:
This method is really difficult to read...

I think by this point https://github.com/MbinOrg/mbin/blob/823c83ca9f92fa4da42f6aba861012cfb97b8f5f/src/Service/ActivityPubManager.php#L145-L148

that checks if the user is local and if so, return, meaning anything after it knows the user isn't local

so https://github.com/MbinOrg/mbin/blob/823c83ca9f92fa4da42f6aba861012cfb97b8f5f/src/Service/ActivityPubManager.php#L167

is fine because it fires off if the user hasn't been fetched at all or not recently, and we know they have an apfetchedat because if they were local it would've returned

but above all those checks were this was happening I don't think if we know whether the user is local or remote yet. I am assuming calling https://github.com/MbinOrg/mbin/blob/823c83ca9f92fa4da42f6aba861012cfb97b8f5f/src/Service/ActivityPubManager.php#L135 on remote users is pointless, but I could be mistaken

e-five256:
I think you're right that it should have the same check for having never been fetched, at the very least to match the way it's done later. Assuming my reading of the code was correct, which is highly possibly not, the only difference here is we don't know if it's a local or remote user returned yet, but later we do so don't need the check...

I could really use some comments on figuring out what this method is doing. Like I think I understand the, if actor url is local, lookup local user and return. Next it does the remote check and updates if necessary. But what is this one I'm change the line in. What is it even for? It _also_ looks up the user. Why was it needed compared to the two below it. There's three different `$this->userRepository->findOneBy` but only two of them make sense to me :thinking:

e-five256:
It now matches more closely to https://github.com/MbinOrg/mbin/blob/823c83ca9f92fa4da42f6aba861012cfb97b8f5f/src/Service/ActivityPubManager.php#L666-L668 which makes me feel slightly more confident

BentiGorlich:
There are 3 `findOneBy` calls because a usercan bespecified like
- @user@domain -> username
- https://domain.tld/u/user (remote url) -> apProfileId
- or https://localhost/u/user (local url) -> remove url and -> username

I think you can remove one of the uodate actor dispatches because fhe user will always be local for it (the local url call)

BentiGorlich:
Nevermjnd on the remove part.
LGTM, though I want to lokk at it in my IDE go be sure 😁

BentiGorlich (approved):

PR #663, opened 2024-04-05
add editorconfig for more filetypes

asdfzdfj:
add some `.editorconfig` configs that should be applicable for all file types in the project, and indent config for more file types used in the project, specifically php, twig, js, and (s)css files

the settings are based on usage in existing files, which should help in ensuring more consistent indent styles for (new) files in the project forward

many editors/IDEs should be able to pick up and apply these settings automatically

~~totally not adding this just so I don't have fiddling with `:setl ts=...` on every file I touched~~

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,3 +1,26 @@
+root = true
+
+[*]
+end_of_line = LF
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[*.php]
+indent_style = space
+indent_size = 4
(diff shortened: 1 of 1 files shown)


e-five256 (approved):
seems good to me. also realized I had to install a package to use the editorconfig in my ide so thanks for helping me notice that

I do wish the number was consist between css/js/php :sob: but it's probably better to just keep things how they are

PR #664, opened 2024-04-05
initial eslint integration and ruleset config

asdfzdfj:
add eslint and some initial rules for linting js code and enforce some styles to keep it somewhat consistent

php code already has php-cs-fixer to deal with style guides, but it's quite a wild west over here on the js side of things

for the time being, the configured rules are made to somewhat mirrors the coding style that's current enforced by php-cs-fixer for php codes, there maybe some subtle rules that it doesn't match up but for the most part it should results in a more similar styles between php and js code

the basic lint command can be invoked by `npm run lint`, or use `npx eslint` if more control over eslint is needed,
note that unlike php-cs-fixer, some rule violations can't be automatically fixed and requires manual intervention

also note that I chose not to add it to ci yet, due to the existing code fails the lint in various ways, many of which requires manual intervention or even potential rewrite to make it conformant, and I'd like to get those sorted out before start enforcing the new style forward


.eslintrc.json
```diff
@@ -0,0 +1,43 @@
...
+    },
+    "plugins": [
+        "@stylistic"
+    ],
+    "extends": "eslint:recommended",
+    "parserOptions": {
+        "ecmaVersion": "latest",
+        "sourceType": "module"
+    },
+    "ignorePatterns": [
```
e-five256, line 15: I'm getting

```
Error: ESLint configuration in .eslintrc.json is invalid:
	- Unexpected top-level property "ignorePatterns".
```

I'm sure it's likely I just didn't configure eslint correctly

```
$ eslint --version
v6.4.0
```

asdfzdfj: I'm using eslint `8.57.0`

actually, I already added needed eslint into `package.json`, you should be able to just `npm i` and then `npx eslint` to start using it, where did you even dig up that [ancient](https://eslint.org/blog/2019/09/eslint-v6.4.0-released/) eslint 6.4.0 from?

```
mbin % git checkout new/eslint-basic-integration
Switched to branch 'new/eslint-basic-integration'
mbin % npm i

added 98 packages, and audited 853 packages in 3s

133 packages are looking for funding
  run `npm fund` for details

found 0 vulnerabilities
mbin % npx eslint --version
v8.57.0
```

also eslint 9?? apparently released like 2 days ago???, should I upgrade this to eslint 9 too while I'm at it? looks like there's some config format and organizational changes so it needs a bit more change than just update the installed version

e-five256: I guess that's a problem I have, as `npm i` (i thought you had typo'd, didn't realize it was a real command, haha) gets me

```
npm ERR! enoent ENOENT: no such file or directory, open '/git/MbinOrg/mbin/vendor/symfony/stimulus-bundle/assets/package.json'
```

e-five256: I guess I'll probably be fighting with composer/symfony as neither want to install on my system for some reason. That's why I kind of always hope for container solutions or binaries but, that's going off topic, I guess I'll just mark this as resolved since it's a me problem

BentiGorlich (commented):
I think we need documentation of its existence and how to run the fix command (`npx eslint . --fix` or add another script to package.json)

asdfzdfj:
> I think we need documentation of its existence and how to run the fix command (`npx eslint . --fix` or add another script to package.json)

yeah I should add that, and I could use some idea for the eslint fix command (`npm run lint-fix`?)

asdfzdfj:
added a bit of docs and `lint-fix` autofix script

CONTRIBUTING.md
```diff
@@ -37,6 +37,15 @@ Install PHP-CS-Fixer first: `composer -d tools install`
 
 Then run the following command trying to auto-fix the issues: `./tools/vendor/bin/php-cs-fixer fix`
 
+For JavaScript code inside `assets/` directory, we provided ESLint setup and configuration for linting.
```
BentiGorlich, line 40:
```suggestion
For JavaScript code inside the `assets/` directory, we use an ESLint setup and configuration for linting.
```

BentiGorlich (commented):
Just the small text correction

CONTRIBUTING.md
asdfzdfj: applied the first one, the second change make it feel weird so I ended up rewording the entire half

BentiGorlich (approved):

Issue #575, opened 2024-03-12
Implement OIDC provider for Microsoft Azure


WebVoid:
**Is your feature request related to a problem? Please describe.**
We are discussing of implementing an Mbin instance in our company to manage our communities and would like to centralize authentication using our Azure tenant

**Describe the solution you'd like**
An Azure OIDC provider with the right env variables to configure it

**Describe alternatives you've considered**
-

**Additional context**
We would be able to implement such solution if accepted

e-five256:
@WebVoid Just want to make sure we're good to close this as complete with your work having been merged

e-five256:
I'm going to assume this is all set with #665, feel free to reopen or make a new issue if you're seeing any problems

PR #665, opened 2024-04-05
feat: implement microsoft azure oidc connector

WebVoid:
Implement an OIDC connector for Azure AD.
Supports general microsoft authentication with `common` tenant, and tenant-specific authentication for closed instances.

References #575

### DIFF
--- a/migrations/Version20240405134821.php
+++ b/migrations/Version20240405134821.php
@@ -24,7 +24,6 @@ final class Version20240405134821 extends AbstractMigration
 
     public function down(Schema $schema): void
     {
-        $this->addSql('CREATE SCHEMA public');
         $this->addSql('ALTER TABLE "user" DROP oauth_azure_id');
     }
 }
--- a/migrations/Version20240405134821.php
+++ b/migrations/Version20240405134821.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20240405134821 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return '';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE "user" ADD oauth_azure_id VARCHAR(255) DEFAULT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('CREATE SCHEMA public');
+        $this->addSql('ALTER TABLE "user" DROP oauth_azure_id');
+    }
+}
--- a/src/Security/AzureAuthenticator.php
+++ b/src/Security/AzureAuthenticator.php
@@ -1,0 +1,140 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Security;
+
+use App\DTO\UserDto;
+use App\Entity\User;
+use App\Repository\UserRepository;
+use App\Service\IpResolver;
+use App\Service\SettingsManager;
+use App\Service\UserManager;
+use App\Utils\Slugger;
+use Doctrine\ORM\EntityManagerInterface;
+use KnpU\OAuth2ClientBundle\Client\ClientRegistry;
+use KnpU\OAuth2ClientBundle\Security\Authenticator\OAuth2Authenticator;
+use TheNetworg\OAuth2\Client\Provider\AzureResourceOwner;
+use Symfony\Component\HttpFoundation\RedirectResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpFoundation\Response;
+use Symfony\Component\Routing\RouterInterface;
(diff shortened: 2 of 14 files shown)


e-five256:
This build failed due to linting, you can find instructions on how to run it here: https://github.com/MbinOrg/mbin?tab=readme-ov-file#linting

But, spoilers, it just wants the oauth import at the end

```diff
--- a/src/Security/AzureAuthenticator.php
+++ b/src/Security/AzureAuthenticator.php
@@ -14,7 +14,6 @@ use App\Utils\Slugger;
 use Doctrine\ORM\EntityManagerInterface;
 use KnpU\OAuth2ClientBundle\Client\ClientRegistry;
 use KnpU\OAuth2ClientBundle\Security\Authenticator\OAuth2Authenticator;
-use TheNetworg\OAuth2\Client\Provider\AzureResourceOwner;
 use Symfony\Component\HttpFoundation\RedirectResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
@@ -26,6 +25,7 @@ use Symfony\Component\Security\Http\Authenticator\Passport\Badge\RememberMeBadge
 use Symfony\Component\Security\Http\Authenticator\Passport\Badge\UserBadge;
 use Symfony\Component\Security\Http\Authenticator\Passport\Passport;
 use Symfony\Component\Security\Http\Authenticator\Passport\SelfValidatingPassport;
+use TheNetworg\OAuth2\Client\Provider\AzureResourceOwner;
```

WebVoid:
Yep, I will fix it as soon as I got some minutes for it ^^

migrations/Version20240405134821.php
```diff
@@ -0,0 +1,30 @@
...
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE "user" ADD oauth_azure_id VARCHAR(255) DEFAULT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('CREATE SCHEMA public');
```
nobodyatroot, line 27: I'm not sure why `doctrine:diff` is adding this line to the `down` method since the `public` schema already exists... any ideas?

WebVoid: I created the migration by copy pasting the one where another oauth property was introduced, I asked myself why this line was there, but I left it as I couldn't find any plausible answer

nobodyatroot: Very strange... please remove this line when you get a chance, I ran `bin/console doctrine:migrations:diff` locally between a fresh copy of main and your branch and it does not generate this... now I'm curious which migration you copied from, lol. Thank you!

nobodyatroot (approved):
I don't have a way to test that the oauth actually works, I'll trust that @WebVoid has tested integration with Azure, but it looks good to me... does not break our main build.

WebVoid:
Yes, I developed by iteratively testing it, it works for both common and custom tenant.

src/Security/AzureAuthenticator.php
```diff
@@ -0,0 +1,140 @@
...
+
+                $existingUser = $this->entityManager->getRepository(User::class)->findOneBy(
+                    ['oauthAzureId' => $azureUser->getUpn()]
+                );
+
+                if ($existingUser) {
+                    return $existingUser;
+                }
+
+                $user = $this->userRepository->findOneBy(['email' => $azureUser->getUpn()]);
```
e-five256, line 76: I'm not quite sure on this line, I guess I'm not familiar with what `Upn` is. Reading up the [docs I could find on it](https://learn.microsoft.com/en-us/entra/identity/hybrid/connect/plan-connect-userprincipalname)

This line is looking for an already created native account with an email value that is the same as an azure `Upn`. Whether that is ever possible, I'm not sure, but it's also possible it doesn't harm anything. It seems like maybe Upn _can_  be an email, so maybe this is alright? I just wanted to check, it doesn't seem like the [AzureResourceOwner](https://github.com/TheNetworg/oauth2-azure/blob/master/src/Provider/AzureResourceOwner.php) has a field for email or anything so I guess this is the most right it could be

e-five256: My comment might've seemed a bit rambling so my actual question is just, does this part seem ok to you, @WebVoid ?

Let's say your Upn is `bob`, the only way you'd ever get this to run is most likely creating an account via CLI with the email `bob` (since you wouldn't be able to via registration on the site as it's not a valid email), and then it would enter this if, find that user, and assign their `oauthAzureId` to `bob`. Probably fine?

WebVoid: As for the tests I have done, the upn was the email of the person authenticating. I remember trying to look for a better field as an identifier but couldn't find any. I could var_dump the object returned by azure and post it here if you wanted to check if you can see a better solution.

e-five256: I don't think that's necessary, that answers my question, thanks!

e-five256 (approved):

e-five256:
lgtm, just would like to double check with you @WebVoid that you're good with us merging this PR

WebVoid:
I am, I will provide support if anything goes wrong ^^

e-five256:
Thanks!

PR #662, opened 2024-04-05
Upgrade to Symfony 7.x, package updates and removal of deprecated features

nobodyatroot:
I followed this guide: https://symfony.com/doc/current/setup/upgrade_major.html

A few packages needed to get bumped to their latest version in order to support Symfony 7.x, so I had to modify a couple minor things here and there because either the naming changed or a feature is no longer supported.

~~**The biggest issue**: the ViewCounterBundle (https://github.com/tchoulom/ViewCounterBundle, https://packagist.org/packages/tchoulom/view-counter-bundle) does not support Symfony 7.x, but I was questioning whether this feature/stat is needed at all... it seems like unnecessary db overhead, but maybe someone uses it? Either way, we can't upgrade past 6.4 without replacement/removal or wait until the maintainer pushes a 7.x compatible version.~~

~~**EDIT**: It was brought to my attention (thanks @e-five256) that kbin has also removed the view counter & stats, so maybe not too unreasonable? Again, seems like extra overhead for little use.~~

View counter issue resolved in https://github.com/MbinOrg/mbin/pull/666

PR #666, opened 2024-04-05
Removal of view counters and associated statistics

nobodyatroot:
Removal of view counters and associated statistics.

It seems like this is an unnecessary/unused feature in the context of a federated infrastructure since the view counter only gets updated on local views, which on a large instance will cause extra db overhead.

This PR also complements https://github.com/MbinOrg/mbin/pull/662 (which will get updated to only address the Symfony 7.x modifications once this PR is merged) since the ViewCounterBundle (https://github.com/tchoulom/ViewCounterBundle, https://packagist.org/packages/tchoulom/view-counter-bundle) does not support Symfony 7.x.


jwr1:
@e-five256, I don't think that should affect Interstellar at all :+1:. We don't use `/api/stats/views` or `/api/stats/magazine/{magazine_id}/views` or any `views` keys from other routes. Thanks for checking.

e-five256 (approved):
lgtm

I could see this being useful for people that want to track engagement of their posts, but as you say that only really works if 1. it's a centralized system or 2. it's supported by activitypub. Neither of those things are true here, instead we kinda just put a bunch of overhead on things for little gain.

It sounded like from what Benti was saying there might be an AP field for tracking views, in which case we could start parsing and sending that, but similar to other things we should probably remove this and reimplement it when there is desire / someone that wants to champion that feature

Not sure if you want to give others one last chance of speaking up before merging, up to you

BentiGorlich:
I'm in favor of removing it.

Regarding AP. There is a view activity which peertube does use and rely on. So it is part of AP and could be done. Though it is a huge privacy risk / invasion to use it for just viewing posts (imo), because evedything in AP is just public. So you would havd to have it offby default or ask theuser upon registration and at this point the system becomes unreliable and therefore not really usable imo. So even if it could be implemented, I don't see a use case for it that would work well and protecg the userprivacy...

PR #670, opened 2024-04-06
ui adjustments

e-five256:
based on followup to #453

- null check in JS controller to fix exception, this is needed for the tag rework as well
- minor css change for the rounded edges on filter
- rename def to home, think this makes it clearer, but we can always change it later if people give more feedback
- pass criteria into component in domain controller similar to entry controller (fixes 500 when viewing more from domain in dev mode, prod mode worked but criteria wasn't usable)
- remove federated only drop down option, will implement in the future
- strip pagination from header nav links, matching the regular drop down filters
  - if you were on say page 100 of entries and clicked microblog at the top and there was only 1 page, you'd get a 404, this just resets it as it doesn't seem to make sense to keep pagination when switching between threads/microblogs

### DIFF
--- a/templates/domain/_options.html.twig
+++ b/templates/domain/_options.html.twig
@@ -1,0 +1,176 @@
+{% set showFilterLabels = app.request.cookies.get('kbin_general_filter_labels')|default('on') %}
+
+<aside class="options options--top" id="options">
+    <div></div>
+    <menu class="options__filter">
+        <li class="dropdown">
+            <button aria-label="{{ 'sort_by'|trans }}"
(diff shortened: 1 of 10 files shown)


nobodyatroot (approved):

templates/domain/_options.html.twig
e-five256, line 1: this duplicates the `entry/_options` template. domain can't set federated / subscription filters. rather than add more `if`s to a template, I thought it would be easier to just separate them for now

I realized this morning maybe we could add those filters so you could see what you favorite on a domain, but after spending a few hours couldn't get it working, so maybe this can be a future thing and we make the component use the same options again later

PR #673, opened 2024-04-06
add route requirements for filter options

e-five256:
this adds filter options similar to what was done for `sortBy` to all other filters

right now if you go to say `/all/hot/4h` it will load but the time selection is invalid and it won't affect anything. This instead makes that route 404 as it is not a valid time option. This seems like where we'd want to stop invalid options first before they get to us and we have to constantly filter out bad data (it does duplicate the values from `Criteria`, but hopefully it isn't too bad to be in two main places

### DIFF
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -60,7 +60,11 @@ parameters:
     html5_validation: true
 
     front_sort_options: top|hot|active|newest|oldest|commented
-    front_time_options: 6h|12h|1d|1w|1m|1y|all|∞
+    front_time_options: 3h|6h|12h|1d|1w|1m|1y|all|∞
+    front_type_options: article|articles|link|links|video|videos|photo|photos|image|images|all
+    front_subscription_options: sub|fav|mod|all|home
+    front_federation_options: local|all
+    front_content_options: threads|microblog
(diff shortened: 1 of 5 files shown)


config/services.yaml
```diff
@@ -60,7 +60,11 @@ parameters:
     html5_validation: true
 
     front_sort_options: top|hot|active|newest|oldest|commented
```
e-five256, line 62: `commented` isn't valid for post/entry single so maybe worth breaking out to a new var

config/services.yaml
```diff
@@ -59,8 +59,16 @@ parameters:
 
     html5_validation: true
 
-    front_sort_options: top|hot|active|newest|oldest|commented
-    front_time_options: 6h|12h|1d|1w|1m|1y|all|∞
+    front_sort_options: top|hot|active|newest|oldest|commented # TODO remove fallback after tag rework
```
e-five256, line 62: decided to keep this for now as I don't want to touch the tag route file while the tag rework is in progress, we can change it to `default_sort_options` (and optionally add any other filters) later

nobodyatroot (approved):

Issue #577, opened 2024-03-12
Restrict the instance to authenticated users only


WebVoid:
**Is your feature request related to a problem? Please describe.**
We are discussing of implementing an Mbin instance in our company to manage our communities and would like to deny access to any content to unauthenticated users, kind of like a "private" instance.

**Describe the solution you'd like**
When "private" (or whatever) mode is enabled when trying to reach the instance, auto redirect to the login page, allowing only authenticated people to see any content.

**Describe alternatives you've considered**
Protect the whole instance with an OAuth2 proxy, but it would double the need for authentication, once for the proxy and once in the app.

**Additional context**
We would be able to propose a PR if the feature request was accepted.

BentiGorlich:
I honestly don't know how much effort this would be to implement. I sadly do not know symfony all that well in that regard. However I think it is a feature that at the very least doesn't hurt anybody and that can be useful for folks like you. So if you want to implement it, I am all for it :D

WebVoid:
I will take a look at it, I did some symfony some ages ago, I believe the security layer could cover the entirety of the use-case.

melroy89:
This is now resolved with the PR #678 ?

WebVoid:
It is, I will close it, thanks for the reminder

PR #678, opened 2024-04-07
feat: implement private mode

WebVoid:
Implement a private mode, with associated settings and environment variables.
When private mode is enabled, unauthenticated users cannot view any content of the instance, they MUST be authenticate.

References #577 

## How to test

- Start an instance with the `KBIN_PRIVATE_INSTANCE` env variable set to false, you should be able to navigate everything in your instance, wether you are authenticated or not
- Restart your instance with the `KBIN_PRIVATE_INSTANCE` env variable set to true, you should be redirected to the login form, whatever you are trying to do. **Registration, password reset, email activation, consent should still be available even in private mode**.
- An entry should be available on the admin settings page, you can also play with it


nobodyatroot:
We use Weblate (https://hosted.weblate.org/projects/mbin/mbin/) to 2-way sync all the translations, so you just have to add your string to `translations/messages.en.yaml' and the others will follow once someone on Weblate gets the time to translate into the specific language, so please revert all of them except what you added to `messages.en.yaml`. Thanks!

src/Service/SettingsManager.php
```diff
@@ -64,6 +65,7 @@ public function __construct(
                 $this->find($results, 'KBIN_MERCURE_ENABLED', FILTER_VALIDATE_BOOLEAN) ?? true,
                 $this->find($results, 'KBIN_FEDERATION_PAGE_ENABLED', FILTER_VALIDATE_BOOLEAN) ?? $this->kbinFederationPageEnabled,
                 $this->find($results, 'KBIN_ADMIN_ONLY_OAUTH_CLIENTS', FILTER_VALIDATE_BOOLEAN) ?? $this->kbinAdminOnlyOauthClients,
+                $this->find($results, 'KBIN_PRIVATE_INSTANCE', FILTER_VALIDATE_BOOLEAN) ?? $this->kbinPrivateInstance,
```
e-five256, line 68: We've stopped setting env vars in place of defaulting here, this is because new env vars added to this (oauth is fine because they don't actually get fetched unless enabled) break existing server owners

Can see https://github.com/MbinOrg/mbin/pull/448 for reference of removing it and defaulting. So here I would expect the right side to be `?? false` and then the `.example`/docker exmaple, `config/services.yaml` to be removed. The rest should be fine, as it all calls into here and will just default false

Also, a nitpick, but we try to add new settings as `MBIN` rather than `KBIN`. We can't rename the old ones as it would break everyone running an instance

WebVoid: I renamed the variable and removed the need to set it, tested it and it doesn't break the backend not setting the env variable

BentiGorlich:
Did you intend federation to work when an instance is in private mode? Because I did not see any ActivityPub related route white listed, so the way the code is now federation won't work in private mode

WebVoid:
I did not look into it but that's a very good point, I am not really aware of how federation works right now, I might have to look into it.

Still, my guess it that when you want your instance to be "private", you might not want federation to be enabled, as it might be counterintuitive to restrict your contents to authenticated users only but let other instances source themselves on your contents...

A discussion here might be appropriate I think.

BentiGorlich:
We have the AP routes which are only called if the requester specifies the accept header to `application/activity+json`

WebVoid:
Yes, I'm reading the `config/kbin_routes/activity_pub.yaml` as we speak, I think it might be pertinent to add them to the public routes, I will do the edit right this instant

WebVoid:
I did the fix, still something is bothering me: even with federation disabled, I'm able to request data from my users with an unauthenticated curl request with the `/u/{username}` endpoint as long as I have an `Accept: application/activity+json` header set..... is it supposed to do that ?

BentiGorlich:
Yes it is supposed to do that. AP is a very public protocol and we did not yet implement the "authorized fetch", which mastodon uses to force instances to authenticate themselves as a specific domain (via HTTP signatures)

WebVoid:
This is something we don't not want in a "private" instance, if contents are restricted to authenticated users but anyone knowing the API endpoints can retrieve them, it could represent a security breach for us...

There is, IMO and with my limited knowledge, three paths ahead:
- Keep the endpoints restricted with the same rules as the other ones, which means private instance == no unauthenticated calls to AP endpoints
- Disable these endpoints if federation is disabled, which can be done easily with a `Voter` (the same way I implemented private instance mode)
- Let these endpoints accessible even if private mode is enabled, which kind of kills the feature, and we (my company) won't be able to use MBin for our needs, as we really need an instance where everything is kept private...

e-five256:
I can certainly see this requiring some thought so maybe we need to take some time to consider the options, but remember that you're also free to run this any way you see fit / with any customizations you want (and many of the instance owners here do as such).

I may have mistakenly thought your goal was to still federate with the wider fediverse, but lock down usage of the website to users. If it's instead to lock down a central site, you could always rip out those pieces for your own instance.

There may also be other options here too, I hadn't gotten a chance to speak on the issue itself, but I was curious if something in front of the site could do what you desired (nginx / haproxy / envoy proxy) ahead of requests getting to the instance, but I could be wrong, I'm not super familiar with proxies myself.

WebVoid:
We indeed considered infrastructure options, but our users are not supposed to be on a VPN, also we could use something like oauth2-proxy, that we use on another auth-less application, but it would require users to double-login to interact with contents, which is not very Ux-friendly. 

I can see that this requires some deeper thinking and I would be happy to join the discussion. We do not desire to maintain our own repository of mbin aside from the main repo. 

The idea of the feature was kind of like on a private forum where you are required to login to view ans interact with content, and it effectively requires that said content cannot be accessed by any "backdoor"

BentiGorlich:
I mean we already have a feature disabling federation. I am not certain that it deacticates the endpoints but you could look into that. If it does not do it, then you could just add it, because imo it should do that 😁

WebVoid:
I agree, personnally when I have an option to disable Fediverse, what I understand of it is that I won't be able to fetch contents from other instances but other instances won't be able to fetch content from mine, not just the first.

I can totally look into that, in which case I believe pertinent to implement the voter in another PR for commit history clarity?

You tell me. Either I review the code from this PR one last time and it is accepted "as-is" and I implement the AP Voter in another PR, either I add it here and it is merged with this PR.

BentiGorlich:
I think it is fine to check/fix the deactivation of federation in this PR. After all, like you said, if you have "private" mode enabled and deactivate federation it should not be possible to fetch anything via ActivityPub...

WebVoid:
Perfect, I will implement this today

e-five256:
I'm just trying to figure out a good way to continue testing in dev with this change. For context, I run locally with federation disabled during development but also rely on the AP urls to work so I can make sure nothing breaks from my changes. Maybe I can just enable federation when I need to test them or try to isolate my dev env from the internet (though I have no idea how I would do that), just don't want my IP randomly sending payloads accidentally.

.env.example
```diff
@@ -58,6 +58,9 @@ S3_VERSION=
 # Only let admins generated oauth clients
 KBIN_ADMIN_ONLY_OAUTH_CLIENTS=false
 
+# Let only authenticated users access contents
+MBIN_PRIVATE_INSTANCE=false
```
e-five256, line 62: Oh I thought you had removed these in https://github.com/MbinOrg/mbin/pull/678/commits/504a5bf16a5dc3b5f7d6d90d063d66201dce5701 but missed that you didn't. I don't think it should be needed here or in `.env.example_docker` anymore, the removal of https://github.com/MbinOrg/mbin/pull/678/commits/504a5bf16a5dc3b5f7d6d90d063d66201dce5701#diff-c47b03734cd2b4337b345eeba955637ba790d51fd98979f6d366d4acbdb104e0L163 means nothing looks for these anymore

WebVoid: true, I missed it too, I will remove them

e-five256:
The login/register/reset password pages, which are still accessible in private mode, show the sidebar which will show local posts/accounts

image

I imagine you might want to make the private instance setting available via a [twig extension](https://github.com/MbinOrg/mbin/blob/dfda72f96e68581043c7475cdd6c65a731a776e8/src/Twig/Extension/SettingsExtension.php#L15) and no render those components if logged out and private instance is true

e-five256:
I did check out `/api/docs` as it seemed to work in that I couldn't make, even the unauthenticated calls, when private instance was enabled, but do double check it yourself as well. Though, not sure if you plan to use oauth at all, as I'm not sure how apps would react etc to these settings, it might be outside your use case

WebVoid:
I will check out all of this later today

WebVoid:
My installation started giving me this error:
Fatal error: Type of App\Entity\Client::$identifier must not be defined (as in class League\Bundle\OAuth2ServerBundle\Model\AbstractClient)

Do you have any idea ? I did not change any code since the last time...

WebVoid:
> the privacy policy / terms of service aren't accessible, but you have to consent to them to register. I realize this might be not in your use case if you're only allowing azure but, wonder if those two routes should be allowed access as well

They should still be accessible, I will fix that



> The login/register/reset password pages, which are still accessible in private mode, show the sidebar which will show local posts/accounts
> 
> image
> 
> I imagine you might want to make the private instance setting available via a [twig extension](https://github.com/MbinOrg/mbin/blob/dfda72f96e68581043c7475cdd6c65a731a776e8/src/Twig/Extension/SettingsExtension.php#L15) and not render those components if logged out (`not app.user`) and private instance is true

forgot it, will be done



> I did check out `/api/docs` as it seemed to work in that I couldn't make, even the unauthenticated, calls when private instance was enabled, but do double check it yourself as well. Though, not sure if you plan to use oauth at all, as I'm not sure how apps would react etc to these settings, it might be outside your use case

I don't think I get exactly what you mean by the second sentence, and if I understand the first one correctly, I would effectively expect private mode to block unauthenticated requests to the APIs.

nobodyatroot:
> My installation started giving me this error: Fatal error: Type of App\Entity\Client::$identifier must not be defined (as in class League\Bundle\OAuth2ServerBundle\Model\AbstractClient)
> 
> Do you have any idea ? I did not change any code since the last time...

Did you happen to run the `post-upgrade` script following a recent pull of main? https://docs.joinmbin.org/admin/running-mbin/upgrades


I upgraded the project to Symfony 7 and it sounds related to a change I had to make here: https://github.com/MbinOrg/mbin/pull/662/files#diff-29f6e1f439aea58278fbddae61fa6d9d61f03af0d2b666496d1e4c736a7b670e

WebVoid:
I will try that, thanks

> > My installation started giving me this error: Fatal error: Type of App\Entity\Client::$identifier must not be defined (as in class League\Bundle\OAuth2ServerBundle\Model\AbstractClient)
> > Do you have any idea ? I did not change any code since the last time...
> 
> Did you happen to run the `post-upgrade` script following a recent pull of main? https://docs.joinmbin.org/admin/running-mbin/upgrades
> 
> I upgraded the project to Symfony 7 and it sounds related to a change I had to make here: https://github.com/MbinOrg/mbin/pull/662/files#diff-29f6e1f439aea58278fbddae61fa6d9d61f03af0d2b666496d1e4c736a7b670e

e-five256:
> I don't think I get exactly what you mean by the second sentence, and if I understand the first one correctly, I would effectively expect private mode to block unauthenticated requests to the APIs.

You can probably just ignore my last statement. I just wanted to mention to you the `/api/docs` route so you'd be familiar with it and check it out. If you don't have OAuth configured, it doesn't even load/do anything, so isn't a risk, but if you do configure it, it allows mobile apps. I have no idea how said mobile apps would _react_ to a private instance with federation disabled, so it's possible this isn't a concern for you as long as it's disabled. Like I said, just wanted to make you aware it existed

WebVoid:
Hey, following the errors I had the last time, I resetted my whole environment, and now when I run my **initial** `doctrine:migrations:migrate`, I got this error popping out:
```
In Connection.php line 101:
                                                                                                                                                                      
  Unknown option found: [queues, exchange]. Allowed options are [table_name, queue_name, redeliver_timeout, auto_setup, check_delayed_interval, get_notify_timeout].  
```

I tried running it from a rebased main branch and it still fails with this error, I can't see why....

e-five256:
Hmm, I can't seem to repro that on `main`, and was able to run this branch fine locally, is it alright if I merge main into this branch, or do you want to do it? maybe there were some oddities after the merge, but that error seems new to me

WebVoid:
Fun fact, I resetted and installed my whole environment from scratch, I still had the error, switched to a fresh VM, installed everything from scratch and now it works. Anyway, I will start working on this feature again.

WebVoid:
@e-five256 I have FINALLY fixed my env and made the changes you pointed out =D

e-five256:
Is it alright if I merge main into your branch and fixup the conflicts (that I probably caused with a change to the base template)?

e-five256 (approved):
Functional testing seems to check out to me for what this does. I think it _might_ be worth exposing the `/contact` page as well to enable a way to message the admin but, won't extend this PR any further, would make a followup if that makes sense

For your use case, both federation disabled and private true would be set, and for most others the opposite would be true. It does make me wonder if we should require federation disabled for private true. If that turns out to be the case, that might happen in a future update

I do also plan to rearrange the admin panel a bit. I'm adding a new option now for SSO only mode so it can be done via the admin panel rather than env var only. Once I do I might try to organize like options together under new headers, and I think I plan to put federation disabled and private instance into a "danger zone" similar to account deletion, to call out that they're pretty dangerous switches to flip

nobodyatroot (approved):

WebVoid:
Checks out for me

PR #680, opened 2024-04-07
Fix circular ChainActivityMessage again again

### LINKED ISSUE #668: ChainActivityMessage infinite queue issue
**Describe the bug**

In some situations, a `ChainActivityMessage` may come in that loops infinitely, causing queue processing to stall

fedia.io

1.5.2

BentiGorlich:
when we get a like message which has url A as the "object" (or as the "id" of the "object") and the returned json has a different url B as the "id", then we went to a circular message, because the ChainActivityHandler would create the object with id B, but Like-, Dislike- and AnnounceHandler would search for an object with id A.

If that occurs now, a warning will be logged and the Like-, Dislike- and AnnounceMessage that will be dispatched by the ChainActivityHandler will be adjusted to use the id that is actually in the json response from the url

Closes #668

### DIFF
--- a/src/MessageHandler/ActivityPub/Inbox/AnnounceHandler.php
+++ b/src/MessageHandler/ActivityPub/Inbox/AnnounceHandler.php
@@ -11,6 +11,7 @@ use App\Message\ActivityPub\Inbox\ChainActivityMessage;
 use App\Service\ActivityPubManager;
 use App\Service\VoteManager;
 use Doctrine\ORM\EntityManagerInterface;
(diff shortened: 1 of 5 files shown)


nobodyatroot (approved):
I don't have an active circular message situation to test this on, but it makes sense if this is another cause.

BentiGorlich:
You could test it by publishing a message via rabbit mq management

nobodyatroot:
> You could test it by publishing a message via rabbit mq management

Ok, yep... thanks! lgtm

PR #687, opened 2024-04-08
front controller path/routing adjustment

asdfzdfj:
the new filter ui changes also introduces a couple more filtering parameters in the form of controller path params, for threads/entries listing it looks fine and relatively staightforward, but for microblog posts listing, what used to be a simple `/microblog` is now something like `/all/hot/∞/all/all/microblog`, which hinders the user experience on url readability and memorability compared to the previous scheme, for those who still care

this changes tries to adjust the path/routing of these front controller especially for the microblog part, to make it more simpler and could omit some default filter parameters if they aren't set, similar to the previous path scheme

note that this patch main focus is just the path routing, there's probably more to adjust/polish in front controller and related areas but those would likely be a separate follow up patch

---

some of the front path routing that I have in mind, based on on previous scheme:
- `/{subscription}/{content}/{sortBy}/{time}/{type}/{federation}` -- `/sub/microblog/newest/1d`
- `/{subscription}/{sortBy}/{time}/{type}/{federation}` -- `/mod/commented/1w` `(*)`
- `/{content}/{sortBy}/{time}/{type}/{federation}` -- `/microblog/newest`
- `/{sortBy}/{time}/{type}/{federation}` -- `/top/6h` `(*)`
- `/m/{magazineName}/{content}/{sortBy}/{time}/{type}/{federation}` -- `/m/playground/microblog/active`
- `/m/{magazineName}/{sortBy}/{time}/{type}/{federation}` -- `/m/random/newest` `(*)`

path components `sortBy` and after shouldn't showed up if using default filter values

paths marked with `(*)` could be later used for combined views default/short entry point when the time comes? right now they should be alias for when `content` param is `threads` i.e. threads/entries front view

`type` and `federation` may be able to be pulled out into query params, esp. `type` where it currently only make sense in the context of threads listing, as we don't do such categorization for microblog posts

### DIFF
--- a/config/kbin_routes/front.yaml
+++ b/config/kbin_routes/front.yaml
@@ -1,45 +1,55 @@
 front:
   controller: App\Controller\Entry\EntryFrontController::front
-  defaults: { subscription: home, sortBy: hot, time: '∞', type: all, federation: all, content: threads }
-  path: /{subscription}/{sortBy}/{time}/{type}/{federation}/{content}
+  defaults: &front_defaults { subscription: home, content: threads, sortBy: hot, time: '∞', federation: all }
+  path: /{subscription}/{content}/{sortBy}/{time}/{federation}
   methods: [GET]
-  requirements:
+  requirements: &front_requirement
     subscription: "%default_subscription_options%"
     sortBy: "%default_sort_options%"
     time: "%default_time_options%"
-    type: "%default_type_options%"
     federation: "%default_federation_options%"
     content: "%default_content_options%"
 
-front_redirect:
-  controller: App\Controller\Entry\EntryFrontController::front_redirect
-  defaults: { sortBy: hot, time: '∞', type: all, federation: all, content: threads }
-  path: /{sortBy}/{time}/{type}/{federation}/{content}
+front_sub:
+  controller: App\Controller\Entry\EntryFrontController::front
+  defaults: *front_defaults
+  path: /{subscription}/{sortBy}/{time}/{federation}
   methods: [GET]
-  requirements:
-    sortBy: "%default_sort_options%"
-    time: "%default_time_options%"
-    type: "%default_type_options%"
-    federation: "%default_federation_options%"
-    content: "%default_content_options%"
+  requirements: *front_requirement
+
+front_content:
+  controller: App\Controller\Entry\EntryFrontController::front
+  defaults: *front_defaults
+  path: /{content}/{sortBy}/{time}/{federation}
+  methods: [GET]
+  requirements: *front_requirement
+
+front_short:
+  controller: App\Controller\Entry\EntryFrontController::front
+  defaults: *front_defaults
+  path: /{sortBy}/{time}/{federation}
+  methods: [GET]
+  requirements: *front_requirement
 
 front_magazine:
   controller: App\Controller\Entry\EntryFrontController::magazine
-  defaults: { sortBy: hot, time: '∞', type: all, federation: all, content: threads }
-  path: /m/{name}/{sortBy}/{time}/{type}/{federation}/{content}
+  defaults: &front_magazine_defaults { content: threads, sortBy: hot, time: '∞', federation: all }
+  path: /m/{name}/{content}/{sortBy}/{time}/{federation}
   methods: [GET]
-  requirements:
-    sortBy: "%default_sort_options%"
-    time: "%default_time_options%"
-    type: "%default_type_options%"
-    federation: "%default_federation_options%"
-    content: "%default_content_options%"
+  requirements: *front_requirement
+
+front_magazine_short:
+  controller: App\Controller\Entry\EntryFrontController::magazine
+  defaults: &front_magazine_defaults
+  path: /m/{name}/{sortBy}/{time}/{federation}
+  methods: [GET]
+  requirements: *front_requirement
 
 # Microblog compatibility stuff, redirects from the old routes' URLs
 
 posts_front:
-  controller: App\Controller\Entry\EntryFrontController::front_redirect
-  defaults: { sortBy: hot, time: '∞', type: all, federation: all, content: microblog }
+  controller: App\Controller\Entry\EntryFrontController::frontRedirect
+  defaults: { sortBy: hot, time: '∞', federation: all, content: microblog }
   path: /microblog/{sortBy}/{time}
   methods: [ GET ]
   requirements:
@@ -47,8 +57,8 @@ posts_front:
     time: "%default_time_options%"
 
 posts_subscribed:
-  controller: App\Controller\Entry\EntryFrontController::front_redirect
-  defaults: { sortBy: hot, time: '∞', type: all, federation: all, content: microblog, subscription: 'sub' }
+  controller: App\Controller\Entry\EntryFrontController::frontRedirect
+  defaults: { sortBy: hot, time: '∞', federation: all, content: microblog, subscription: 'sub' }
   path: /sub/microblog/{sortBy}/{time}
   methods: [ GET ]
   requirements:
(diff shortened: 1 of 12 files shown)


e-five256:
I agree with your that `type` should be pulled out because it's exclusive to `threads` so it being in the url for `microblogs` and not doing anything is confusing

asdfzdfj:
already sends this to e-five but I figured it should be here too:

---

as for how things are currently:

- managed to get some `front_*` routes that should match all the url patterns I described, and thanks to route param requirements you added it should be able to match unambiguously
- currently live on my instance and so far, I'm quite happy about it (duh), could use some help to check that I didn't break stuff anywhere else and that the url naviaged/generated made sense _as a user_ (frankly I count this entire thing as more of a ux problem)

something that could use help/input:

- the new front controller introduces one front controller function/endpoint `EntryFrontController::front` that does the main job, and it also introduces `EntryFrontController::front_redirect` which seems to exist for compat reasons? so now I'm not sure on how to best map routes to controller endpoint
    - like should all of them mapped to main `front` but with default params acting as controller function partials or something
    - or only the full route (with all params) should get the main `front` endpoint, and then shim the rest with `front_redirect` endpoint
- there's also the route naming itself
    - from what I've seen the route name often encodes some context into it and there exist some naming convention in it, but with the new front controller names that I've given currently doesn't do that job too well
    - a more practical one would be to migrate the other front routes (post, magazine) into this front controller (incl. path generation, the route name context thing), but somehow this feels like more of an answer to "unifying front controller" question rather than "new front controller paths look like shit" question

asdfzdfj:
current state as of writing, if that means anything:
- trying the "all front route use main front controller" methid, will see how this goes but seems to work fine
- made a specialized `front_options_url` for main front filtering options, where it can "upgrade" the front route used in url generation, otherwise when generating url that uses `{subs}` or `{content}` in front route that doesn't have them defned will result in it being appended as query params (e.g. `/microblog/newest?subscription=sub` rather than `/sub/microblog/newest`)
- pulling out `{type}` is trickier than expected: deleting `{type}` from path params seem to make it not working consistenly, and looks like the `type` in defaults would also need to be removed otherwise pager link generation would lost the `?type=...` somehow, do both and type filtering seem to break completely, more test is needed, and thus `{type}` path param is kept for the time being

e-five256:
Took this for a spin and it looks good to me, do you still have any outstanding work?

I was going to say if the redirect stuff made things more difficult, we could consider removing it, but it seems it's sort of been reworked into allowing short form URLs, so maybe useful to keep around? (e.g. `/home/threads/hot/1d` can be specified as `/hot/1d` which is how it used to be). I think there will definitely be cases of 404s for both `<1.5.3` and `>1.5.3` instances but as long as we try to communicate it I'm sure people can figure out the new URLs

config/kbin_routes/front.yaml
```diff
@@ -1,85 +1,94 @@
...
+  defaults: *front_defaults
+  path: /{subscription}/{sortBy}/{time}/{federation}
   methods: [GET]
-  requirements:
-    sortBy: "%default_sort_options%"
-    time: "%default_time_options%"
-    type: "%default_type_options%"
-    federation: "%default_federation_options%"
-    content: "%default_content_options%"
+  requirements: *front_requirement
```
e-five256, line 18: I appreciate how you find these things that can be refactored into less copy paste

src/Repository/Criteria.php
```diff
@@ -230,22 +230,13 @@ public function resolveTime(?string $value, bool $reverse = false): ?string
...
-            'videos' => Entry::ENTRY_TYPE_VIDEO,
-            'photo' => Entry::ENTRY_TYPE_IMAGE,
-            'photos' => Entry::ENTRY_TYPE_IMAGE,
-            'image' => Entry::ENTRY_TYPE_IMAGE,
-            'images' => Entry::ENTRY_TYPE_IMAGE,
-        ];
-
-        return $routes[$value] ?? 'all';
+        return match ($value) {
+            'article', 'articles' => Entry::ENTRY_TYPE_ARTICLE,
```
e-five256, line 234: same here re: appreciate the refactoring

config/kbin_routes/front.yaml
asdfzdfj: thankfully this is one of the simpler ones, yaml anchors can be quite confusing at time, especially when you also add merge keys to the mix

technically this could go a step further by using merge keys referencing first full front route definition then overwrite `path` with different routes (iirc the old front routing before ui filter patch actually does this)

src/Repository/Criteria.php
asdfzdfj: using array map to do lookup is totally fine (and it's quite the oldest trick in the book), but in here I though that using `match` express the intention better, especially when mapping multiple inputs to a single result and also handling defaults (and the fact that we now have `match` at disposal)

asdfzdfj:
> Took this for a spin and it looks good to me, do you still have any outstanding work?

I don't think I have any, the main problem of path path has been adjusted to look somewhat like the previous ones, `{type}` has been pluked out, keeping `{federation}` for the time being as I could see it being a common filter option

I think this should be ready for review/inclusion

> I was going to say if the redirect stuff made things more difficult, we could consider removing it, but it seems it's sort of been reworked into allowing short form URLs, so maybe useful to keep around? (e.g. `/home/threads/hot/1d` can be specified as `/hot/1d` which is how it used to be). I think there will definitely be cases of 404s for both `<1.5.3` and `>1.5.3` instances but as long as we try to communicate it I'm sure people can figure out the new URLs

I didn't rework the redirect stuff so much that I just split them in two controllers that also encodes/indicates the context it should be used in (rather than having to note that `$name` is meant to be magazine name and then do a branching on that), also I've seen them used in shimming old routes so I think I'll keep it for that for now

e-five256:
One very minor thing, maybe not worth holding this up, but it looks like `?type=` is kept in the header nav bar when set and selecting microblog.

For example on `http://127.0.0.1:8666/m/testing/threads/hot/1y?type=links` if I hover over Microblog I get the link `http://127.0.0.1:8666/m/testing/microblog/hot/1y?type=links`

I was able to remove it with

```diff
diff --git a/templates/layout/_header_nav.html.twig b/templates/layout/_header_nav.html.twig
index 2e1a5441..2e1bfc4b 100644
--- a/templates/layout/_header_nav.html.twig
+++ b/templates/layout/_header_nav.html.twig
@@ -40,13 +40,13 @@
         <a href="
             {% if magazine is defined and magazine %}
                 {% if is_route_name_starts_with('front') %}
-                    {{ options_url('content', 'microblog', 'front_magazine', {'name': magazine.name,'p': null}) }}
+                    {{ options_url('content', 'microblog', 'front_magazine', {'name': magazine.name,'p': null,'type':null}) }}
                 {% else %}
                     {{ navbar_posts_url(magazine) }}
                 {% endif %}
             {% else %}
                 {% if is_route_name_starts_with('front') %}
-                    {{ options_url('content', 'microblog', 'front', {'p': null}) }}
+                    {{ options_url('content', 'microblog', 'front', {'p': null,'type':null}) }}
                 {% else %}
                     {{ navbar_posts_url(null) }}
                 {% endif %}
```

Or maybe there's a better way in `optionsUrl` / `frontOptionsUrl` that would be preferred.

asdfzdfj:
> One very minor thing, maybe not worth holding this up, but it looks like `?type=` is kept in the header nav bar when set and selecting microblog.

good point, added the suggested patch, that does bother me a little

to be honest, I kinda glossed over the navbar part, though that's likely because I also have another plan to try and push the condition in templating front link (the block you edited) down to the `navbar_*_url`  twig functions so I didn't look or touch it much in here, and planned to make it a different/followed up patch of sorts

e-five256 (approved):

PR #688, opened 2024-04-08
disable trim trailing space on markdown files

asdfzdfj:
adjust .editorconfig to disable auto disable trimming trailing space on markdown (`*.md`) files

in markdown you could insert a line break by adding 2 or more trailing spaces and then a new line,
but the recently adjusted .editorconfig is set to always trim trailing whitespaces a tthe end of the line which interferes with this markdown syntax

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -30,6 +30,9 @@ indent_style = space
 indent_size = 4
 quote_type = single
 
+[*.md]
+trim_trailing_whitespace = false
+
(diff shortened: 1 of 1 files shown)


e-five256:
Hmm, but my editor trims trailing whitespace every time I save any file. That sort of sounds like I would break markdown files any time I save them. I could disable it but... I like it... Guess I should disable it?

e-five256 (approved):

PR #694, opened 2024-04-08
Implement rules federation, fix markdown federation

### LINKED ISSUE #415: [Bug] Rules are not federated
**Describe the bug**
The rules of a remote mbin magazine do not get pulled in to other instances

[gehirneimer.de, kbin.run]

**To Reproduce**
1. Go to <gehirneimer.de/m/mdev@kbin.run>
2. Open <kbin.run/m/mdev>
3. Look at the missing rules on gehirneimer.de

**Expected behavior**
It should get synced

**Screenshots**
Gehirneimer: 
grafik

Kbin.run:
grafik

BentiGorlich:
- the summary of magazines includes the markdown source now
- the summary of magazines includes the rules now
- when updating a magazine the markdown source is used if available and if the last header includes "rules" it will be put in the rules field
- the summary actually works now. Before it would output the markdown as the summary and we would assume that the summary field is an html field, so we would try to convert html -> markdown, except that the html we tried to convert already was markdown, so it just didn't work properly

Closes #415


src/Factory/ActivityPub/GroupFactory.php
```diff
@@ -13,13 +15,25 @@ class GroupFactory
...
         private readonly ImageManager $imageManager
     ) {
     }
 
     public function create(Magazine $magazine): array
     {
+        $markdownSummary = $magazine->description;
+
+        if (!empty($magazine->rules)) {
+            $markdownSummary .= "\r\n\r\n### Rules\r\n\r\n".$magazine->rules;
```
e-five256, line 29: my concern here is that this isn't localized, and of course it's difficult to know what language the magazine owner wanted

BentiGorlich:
I observed the problem especially while looking at https://gehirneimer.de/m/FloatingIsFun@fedia.io vs https://fedia.io/m/FloatingIsFun
Which just looks hilariously bad...

src/Factory/ActivityPub/GroupFactory.php
BentiGorlich: You're right it is not. The "cleanest" way would be to introduce a `rules` property to the AP object, but that wouldn't be understood by lemmy, so it is kind of useless this way. I thought about completely removing the field, because, well it can't be displayed via AP...
We could use the language of the owner, but that is not that good eigher. We could introduce a language field in the magazine that defaults to english, but what good is that if it is only used for the rules... And how do we parse it back into our rules field...
It just complicates everything, but this way it at least federates somewhat :D

BentiGorlich: I don't like it that much either, but it is understood by our software, so a user with a German interface will see it correctly under the "Regeln" header, because we put it in a different section. For users of other software it will always be a "Rules" header within the summary.

e-five256: What if it was just a `---` line instead? On our side it would say rules and outside would just be an ordered list? that doesn't seem to bad to me, but that's just my opinion

BentiGorlich: It's ambigous. A lemmy magazine could have it just as a separator for whatever reason...

BentiGorlich: we could maybe do a double line. That would look funny on other software though :thinking:

BentiGorlich: Does anyone have a good idea how to handle this? Or is hardcoding "Rules" the best we have? :sweat_smile: 

I thought about it yesterday and I really don't want to remove that field, it could be really good for moderation improvements, like referencing a rules in a report and so on... So maybe we should introduce a magazine language field?

e-five256: My opinion is we should fully deprecate the rules field as I don't see it being useful considering it isn't part of the AP spec

Incoming remote group descriptions should stay as they are, kept as a description without us trying to regex/explode/parse them into our view of things, as we are messing with assuming how remotes parties expected their description to be read

The rules form text area should be hidden if no rules are present on the magazine, thus discouraging new magazines from setting any

Outgoing could be hardcoded to adding `Rules` for now and appending it to description. We could eventually add a notification that if you haven't moved rules into description yet it will be done automatically in a future release, at that point we migrate rules into the description column and drop it

BentiGorlich: If we decide that the rules field does not provide enough value to keep it, I totally agree with the plan you have.

Here are my thoughts why the rules field could be very valuable:
If you are familiar with mastodon, then you know that each server has a set of rules which a user must agree to before they can register on a server. On mastodon these rules are not federated, as they are server specific. In the threadiverse however magazine rules are not server specific, so there should be a way to federate them. At the moment we do that by just adding them to the summary. So my vision is a mastodon server rule like implementation for magazines, where you 
- have a list of rules which are stored as an array
- can specify which rule was broken when reporting somebody
- have to agree to a magazine's rules before being able to post anything

This would of course need to be written down as an fep of sorts and it would be hard to keep it compatible with Lemmy if they don't implement it, but I think it is doable, even it it is harder and I think it would be valuable for moderation (though I don't know, I am not a moderator of a big community)

src/Service/ActivityPubManager.php
```diff
@@ -789,4 +794,44 @@ public function getEntityObject(string|array $apObject, array $fullPayload, call
...
+            if (str_contains(strtolower($lastHeader['header']), 'rules')) {
+                $descLines = \array_slice($lines, 0, $lastHeader['line']);
+                $ruleLines = \array_slice($lines, $lastHeader['line'] + 1, \sizeof($lines) - $lastHeader['line']);
+                $description = join("\n", $descLines);
+                $rules = join("\n", $ruleLines);
+            }
+        }
+
+        return $rules;
+    }
```
e-five256, line 836: I just don't think this is a good idea vs parsing it as a regular description like we've been doing, it seems way too prone to failure. For instance a random remote magazine might have a quote in its description like `## “Know the rules well, so you can break them effectively.”  - Dalai Lama XIV` we suddenly parse that as the rules. Also, it doesn't work for non-english, so is flaky

BentiGorlich: Before thinking to hard about this what do you think about my ideas for the rules field? Because if you think it is too complicated and not really worth it then we will just deprecate the field (if everyone agrees)

e-five256: I think it's fine to keep it if you see value

My opinions on keeping it would be:
- Add it as "Rules" as you have now on outgoing
- Remove this extract rules from remote, just keep it all in description for now, even from other mbin instances
- Make an FEP for adding a rules field ordered list to group collections, if that's what you believe in

TheVillageGuy:
I'm sorry for being a bit late to the party, but isn't this related to a feature request I made a while ago asking for local and external description/rules. I'm having a bit of a hard time grasping what exactly you're handling here so can't say whether I agree or not

BentiGorlich:
If I remove the dev / AP specific jargon, I did the following:
- add the rules to the description so they get federated to other instances (e-five correctly criticized that I add a non translated "Rules" header)
- when getting the info of a remote magazine, we "parse" the description of it to extract rules into our "Rules" field. That is error prone and only works for magazines with an English description (e-five explained why it is error prone)

The "parsing" will most likely be removed

nobodyatroot:
i vote for whatever gives us the most/best compatibility across the fediverse :-)

src/Service/ActivityPubManager.php
```diff
@@ -441,8 +442,10 @@ public function updateMagazine(string $actorUrl): ?Magazine
 
         if (isset($actor['endpoints']['sharedInbox']) || isset($actor['inbox'])) {
             if (isset($actor['summary'])) {
-                $converter = new HtmlConverter(['strip_tags' => true]);
-                $magazine->description = stripslashes($converter->convert($actor['summary']));
+                $magazine->description = $this->extractMarkdownSummary($actor);
+                if (!empty($rules)) {
+                    $magazine->rules = $rules;
+                }
```
e-five256, line 448: I might be missing it, is `$rules` still set here?

BentiGorlich: no that part should be removed

e-five256: everything else in this PR seems fine to me, it's just removing these three lines I think otherwise it'll throw an undefined variable exception, unless I'm misreading it, difficult for me to be sure locally since I'm not sure how I could fake an AP event

BentiGorlich: I don't think it will lead to exceptions because `empty` is null safe afaik, though yes ot should be removed.
If you want to, feel free to fix it and then merge it. I will not be able to 'til wednesday

e-five256 (approved):

PR #699, opened 2024-04-10
switch magazine list to show magazine display names

e-five256:
this simply changes the magazine table list view to display the magazine titles rather than addresses

|before|after|
|-|-|
|image|image|

### DIFF
--- a/templates/magazine/_list.html.twig
+++ b/templates/magazine/_list.html.twig
@@ -63,7 +63,7 @@
                     <tbody>
                     {% for magazine in magazines %}
                         <tr>
-                            <td>{{ component('magazine_inline', { magazine: magazine, stretchedLink: true, showAvatar: true, fullName: true}) }}</td>
+                            <td>{{ component('magazine_inline', { magazine: magazine, stretchedLink: true, showAvatar: true}) }} {% if magazine.isAdult %}<small class="badge danger">18+</small>{% endif %}</td>
                             <td>{{ magazine.entryCount }}</td>
                             <td>{{ magazine.entryCommentCount }}</td>
                             <td>{{ magazine.postCount + magazine.postCommentCount }}</td>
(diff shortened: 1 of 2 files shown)


nobodyatroot (commented):
What happens with a reallllly long name? I've noticed some that are just ridiculous, lol.

e-five256:
image

nobodyatroot:
> I was already testing that, so it looks like this. If you think I should limit column max width let me know
> 
> image

It probably doesn't hurt to limit it... no telling what crazy-ass names we'll encounter.

nobodyatroot (approved):
lgtm

templates/magazine/_list.html.twig
```diff
@@ -63,7 +63,7 @@
                     <tbody>
                     {% for magazine in magazines %}
                         <tr>
-                            <td>{{ component('magazine_inline', { magazine: magazine, stretchedLink: true, showAvatar: true, fullName: true}) }}</td>
+                            <td>{{ component('magazine_inline', { magazine: magazine, stretchedLink: true, showAvatar: true}) }} {% if magazine.isAdult %}<small class="badge danger">18+</small>{% endif %}</td>
```
e-five256, line 66: I thought about adding the adult badge to magazine inline when full name was false (it only appears when full name is true) but since at the moment it's just this component, didn't seem worth it adding something that would be unset/false in every single other instance, so instead just appended it here

e-five256:
just so you're aware it looks a bit different on mobile (that's why there were two places to change) but no table there and the name gets the entire width of the screen so a lot more room to view on

image

nobodyatroot:
> just so you're aware it looks a bit different on mobile (that's why there were two places to change) but no table there and the name gets the entire width of the screen so a lot more room to view on
> 
> image
> 
> that's a bit of a fake mobile though, it'd look more like
> 
> image

that seems reasonable

PR #700, opened 2024-04-10
limit magazine list column width to first column

e-five256:
@nobodyatroot noticed that https://github.com/MbinOrg/mbin/pull/699 made it look a bit bad for mags you were subscribed to with large subscriber numbers. the intent was really to limit the column size on magazine names and not the other columns, so this restricts it to just the first column

|before|after|
|-|-|
|image|image|

No matter what it's possible for the actions to go multi line though, even before that was true with the ultimate narrowest settings (fixed - sub bar separate)

### DIFF
--- a/assets/styles/components/_magazine.scss
+++ b/assets/styles/components/_magazine.scss
@@ -240,8 +240,8 @@ td {
     display: block;
   }
 
-  td {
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):
Perfect!

PR #701, opened 2024-04-10
Remove kbin migration, code bases are far too diverged to be a safe recommendation anymore

nobodyatroot:
I think it's far too risky to recommend migrating from kbin to mbin anymore as there's a high probability of data loss with how the DB schema/migrations have diverged... that is, unless someone wants to take on owning the migration path (_not it_).


BentiGorlich (approved):

e-five256:
Just to comment on this, while I agree it can't be done with more thought now that there's been more commits, it still can be done (as someone just proved so), there just ~might be more involved manual steps (making the db schemas match was the main one), so joining the matrix and asking is probably a good step

PR #702, opened 2024-04-11
reduce frequency of `lastoriginupdate` db writes

e-five256:
potentially limit the amount of times we update magazine entries to only update if it has been an hour since the last update, as before it was updating every single event

### DIFF
--- a/src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
+++ b/src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
@@ -93,14 +93,17 @@ readonly class ActivityHandler
         }
 
         if ('Announce' === $payload['type']) {
-            $actor = $this->manager->findActorOrCreate($payload['actor']);
-            if ($actor instanceof Magazine) {
-                $actor->lastOriginUpdate = new \DateTime();
-                $this->entityManager->persist($actor);
-                $this->entityManager->flush();
-            }
             // we check for an array here, because boosts are announces with an url (string) as the object
             if (\is_array($payload['object'])) {
+                $actorObject = $this->manager->findActorOrCreate($payload['actor']);
+                if ($actorObject instanceof Magazine && $actorObject->lastOriginUpdate < (new \DateTime())->modify('-3 hours')) {
+                    if (isset($payload['object']['type']) && 'Create' === $payload['object']['type']) {
+                        $actorObject->lastOriginUpdate = new \DateTime();
+                        $this->entityManager->persist($actorObject);
+                        $this->entityManager->flush();
+                    }
+                }
+
                 $payload = $payload['object'];
                 $actor = $payload['actor'] ?? $payload['attributedTo'] ?? null;
                 if ($actor) {


nobodyatroot:
Output of https://wiki.postgresql.org/wiki/Lock_Monitoring#.D0.A1ombination_of_blocked_and_blocking_activity on kbin.run before applying this patch:

image

image

BentiGorlich:
What do you think about instead of updating with every announce (which includes like and is probably the biggest problem) to only update with announced create activities?

BentiGorlich:
`apfetchedat` means something totally different though. That is the date we last fetched an update from the remote server. This date means when did the remote server last send us something. And nothing is attached to the date. I would like that to be as accurate as poosible and likes do not really count, so I am willing to remove them, but not creates...

e-five256:
When your queues fall behind and you try to catch up, _every single event_ locks the magazine row to update this. It most likely causes deadlocks as debounced has shown. You just don't need this value to be accurate to the nanosecond of every event

nobodyatroot (dismissed):
Agreed that we should not be hitting the db everytime considering what this "feature" is doing. 1 minute resolution is plenty (and probably still overkill).

e-five256:
@BentiGorlich want to make sure this seems acceptable to you after the change

BentiGorlich:
I've been thinking about it and came to the conclusion that a ~3h timespan would be a good value. Atm this galue is only used to display magazine stalenes and only if no one is subscribed to the magazine... However I think we should still limit it to create activities, because even if it is just a redis hit when getting all the likes, it still involves latency. If the redis server is located on a differenf machine (who knows) that latency is super noticable... Let me know what you think.

I'd accept it the way it is now as well as it is obviously an improvement

nobodyatroot:
> I've been thinking about it and came to the conclusion that a ~3h timespan would be a good value. Atm this galue is only used to display magazine stalenes and only if no one is subscribed to the magazine... However I think we should still limit it to create activities, because even if it is just a redis hit when getting all the likes, it still involves latency. If the redis server is located on a differenf machine (who knows) that latency is super noticable... Let me know what you think.
> 
> I'd accept it the way it is now as well as it is obviously an improvement

i can reapprove if this is better, that is ~3h timespan @e-five256

e-five256:
I'll work on making the changes suggested

e-five256:
definitely super look over me here as this is outside my wheelhouse, kind of guessing based on grabbing payloads from past logs, as obviously this is difficult to test locally

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
```diff
@@ -113,6 +107,13 @@ private function handle(?array $payload)
                     }
                 }
             }
+        } elseif ('Create' === $payload['type']) {
+            $actor = $this->manager->findActorOrCreate($payload['actor']);
+            if ($actor instanceof Magazine && $actor->lastOriginUpdate < (new \DateTime())->modify('-3 hours')) {
+                $actor->lastOriginUpdate = new \DateTime();
+                $this->entityManager->persist($actor);
+                $this->entityManager->flush();
+            }
```
BentiGorlich, line 116: What actually needs checking is the announced type, so payloads like this will trigger the code:
```json
{
  "type": "Announce",
  "object": {
    "type": "Create",
    ...
  },
  ...
}
```
So I'd say move it back to where it was before and change the condition to:
```php
if ('Announce' === $payload['type']) {
    if (\is_array($payload['object'])) {
        if ('Create' === $payload['object']['type']) {
```

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
```diff
@@ -113,6 +107,13 @@ private function handle(?array $payload)
                     }
                 }
             }
+        } elseif ('Create' === $payload['type']) {
+            $actor = $this->manager->findActorOrCreate($payload['actor']);
+            if ($actor instanceof Magazine && $actor->lastOriginUpdate < (new \DateTime())->modify('-3 hours')) {
```
BentiGorlich, line 112:
```suggestion
            if ($actor instanceof Magazine && (!$actor->lastOriginUpdate || $actor->lastOriginUpdate < (new \DateTime())->modify('-3 hours'))) {
```

BentiGorlich (changes requested):

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
e-five256: modifying datetime rather than the field actually seems to protect us from exceptions, I think you had commented on it before when I had asked, as `null < now - 3 hours` will be `true`. I was contemplating changing all the `!var || var->modify comparison datetime` to `var comparison datetime->modify`

BentiGorlich:
True, then you can ignore that comment

Am 15. April 2024 14:17:43 MESZ schrieb e-five ***@***.***>:

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
e-five256: I would've thought that means it'd only update on comments that come in from external instances, but maybe I'm misunderstanding AP here. I assumed create = comes from owner instance, announce = comes from another instance, and owner is announcing it happened. Or maybe owner instances both create and announce create their own posts?

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
BentiGorlich: True, then you can ignore that comment

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
BentiGorlich: Yeah a create from a magazine and an announced create from a magazine should refresh the field. Until now the plain create did not do that, but it should, you're right

e-five256: Ok, that might actually explain some things I had seen. Looking at this one magazine, it says "This magazine is not receiving updates (last activity 29 day(s) ago). " but has posts from today, yesterday etc, continuing to get posts, but they're all posts from that instance the magazine is from

So if I understand correctly, I need both sections, what I have here which updates from the origin's creates, and then what you suggested which updates when the origin announces creates as well

BentiGorlich: Well you alsp just get the creates from the user directly which would not count towards origin uodates.

And yes both sections would be good indeed

e-five256:
I think I've handled the situations as described now where it will update the value when a create or an announced create after 3 hours

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
```diff
@@ -113,6 +117,14 @@ private function handle(?array $payload)
...
```
BentiGorlich, line 122: This will not work, because the actor of a `Create` activity is never a magazine, but always a user. A magazine only announces what a user creates. If you get a plain `Create` activity you would have to check from which instance the user is from and then get the magazine the activity is posted to and then you can adjust the origin update. 
However I think I removed that part (I initially had something that did the same) because you cannot be sure that you get create activities because you are subscribed to a magazine, because you will get create activities from users you are subscribed to, regardless of whether you're subscribed to the magazine they post it to. `Announce` activities on the other hand are only sent to people subscribed to magazines

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
```diff
@@ -94,14 +94,18 @@ private function handle(?array $payload)
...
-                $this->entityManager->flush();
-            }
+
             // we check for an array here, because boosts are announces with an url (string) as the object
             if (\is_array($payload['object'])) {
                 $payload = $payload['object'];
+
+                if ('Create' === $payload['type'] && $actor instanceof Magazine && $actor->lastOriginUpdate < (new \DateTime())->modify('-3 hours')) {
+                    // announced creates update magazine last origin timestamp when after threshold
+                    $actor->lastOriginUpdate = new \DateTime();
```
BentiGorlich, line 104: Could you rename this variable? I was confused whether this works, because this is not the actor of the `$payload` variable at this position.
```suggestion
                    $announceActor->lastOriginUpdate = new \DateTime();
```

BentiGorlich (changes requested):

src/MessageHandler/ActivityPub/Inbox/ActivityHandler.php
BentiGorlich: In this conversation I forgot the point I made in my latest comment. Sorry for the confusion on my part

e-five256:
I don't think I really have the expertise to make this change, as I'm basically taking a shot in the dark each time since I'm not familiar with AP and don't have a cluster to test what real events look like.

Not sure if this could be tackled by someone else or just closed. I think assuming @nobodyatroot isn't running it anymore since that one time since he's been running the tag PR since, maybe this isn't a problem that needs to be handled and the problem at the time was the foreign key constraint violation.

I'm just very, super overly optimized where I come from (even two DB connections are against my best designs, I'm like, forcing people to use refcursors if they need multiple data sets from a DB), whereas Mbin seems to be more like postgresql connections go brrrrrrrr

BentiGorlich:
I am happy to take this on. I am more on the side of "optimize where it is doable with reasonable effort". I agree that this should be optimized and I think I have the best understanding of the AP side in our team. And as my vacation "season" is over for now I should have a lot more time

BentiGorlich:
Ok, I made it so it updates the `lastOriginUpdate` field only if the last update was at least 3 hours ago and the incoming activity is an `Announce(Create)` with a magazine as the actor

BentiGorlich (approved):

nobodyatroot (approved):

PR #708, opened 2024-04-14
Add contributors to the README (automated)

melroy89:
Adding a table of contributors to our README in an automated way.

More info: https://github.com/marketplace/actions/contribute-list


README.md
```diff
@@ -55,8 +55,22 @@ For developers:
...
+## Collaborators
+
+<!-- readme: collaborators -start -->
+<!-- readme: collaborators -end -->
+
+## Contributors
+
+<!-- readme: contributors -start -->
+<!-- readme: contributors -end -->
+
```
BentiGorlich, line 67: I think this placement is a bit high up in the document. I don't know how it will look like, but I guess it will push everything underneath it way down, to the point where people do not expect there to be anything else. Could you provide a screenshot of how it looks? Or is it not visible until this hits main?

BentiGorlich:
Great idea to do that. It seems that I don't know anything about GitHubs features :D

asdfzdfj:
a sample screenshot would be nice (even better, a live specimen), but if it must be merged before it starts doing anything tangible then it should be fine, I think? it didn't touch the main code

asdfzdfj (approved):
guess we'll just have to wing it, but I trust that it's sufficiently safe to do so, judging from the changes.

BentiGorlich (approved):

BentiGorlich:
Let's see how it looks :)

melroy89:
Fingers crossed.

PR #710, opened 2024-04-14
use github format admonitions in docs

asdfzdfj:
change the admonitions in docs from docusaurus format to github format, so it'll show up properly when reading docs from github interface

combined with [remark-github-admonitions-to-directives][plugin] plugin on docusaurus side, this will make admonitions being rendered on both docusaurus site and the docs source file on github interface, rather than having to pick one or the other

also fix up a few older admonitions to use the new format along the way

[plugin]: https://github.com/incentro-dc/remark-github-admonitions-to-directives

### DIFF
--- a/docs/02-admin/01-installation/bare_metal.md
+++ b/docs/02-admin/01-installation/bare_metal.md
@@ -3,20 +3,14 @@
 Below is a step-by-step guide of the process for creating your own Mbin instance from the moment a new VPS/VM is created or directly on bare-metal.  
 This is a preliminary outline that will help you launch an instance for your own needs.
 
-:::note
-
-Mbin is still in development.
-
-:::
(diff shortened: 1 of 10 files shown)


melroy89 (approved):
Assuming this still builds correctly.

asdfzdfj:
> Assuming this still builds correctly.

it still builds, but until the mentioned plugin is added you'll just get a funny looking admonitions when looking on docusaurus side rather than a proper admonitions, which it supports but with different syntax

image

melroy89:
IN that case let's first add the plugin ^^

BentiGorlich:
I just added the plugin to convert the github syntax to docusaurus, so this PR can be merged now :)

BentiGorlich (approved):

PR #717, opened 2024-04-22
use only `application/activity+json` for AP fetching

asdfzdfj:
some softwares/instances doesn't handle multiple Accept header values too well, resulting in failed fetches e.g. the instance that piped link bot is on

### DIFF
--- a/src/Service/ActivityPub/ApHttpClient.php
+++ b/src/Service/ActivityPub/ApHttpClient.php
@@ -316,7 +316,7 @@ class ApHttpClient
         unset($headers['(request-target)']);
         $headers['Signature'] = $signatureHeader;
         $headers['User-Agent'] = $this->projectInfo->getUserAgent().'/'.$this->projectInfo->getVersion().' (+https://'.$this->kbinDomain.'/agent)';
-        $headers['Accept'] = 'application/activity+json, application/ld+json';
+        $headers['Accept'] = 'application/activity+json';
         $headers['Content-Type'] = 'application/activity+json';
 
         return $headers;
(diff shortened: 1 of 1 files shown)


BentiGorlich:
The question is: is there software out there that wants the `application/ld+json` type in the accept header. I think no there is not.
But it feels wrong to remove it, just because other software is not parsing the header correctly... 

I doubt that this change is hurting us in any way, so we can merge that imo, it feels wrong anyway :sweat_smile:

asdfzdfj:
I have yet to see any that uses bare `application/ld+json`,
and [_technically_](https://www.w3.org/TR/activitypub/#server-to-server-interactions), it looks like the proper value for making requests with ld+json is `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` (note the profile), with `application/activity+json` acting as a shorthand for the full value

so perhaps using just bare `application/ld+json` without specifying profile as it is now may be interpreted as malformed requests, and arguably the right thing to do is to _not_ answer back with AP objects for such requests

asdfzdfj:
>  But it feels wrong to remove it, just because other software is not parsing the header correctly...

yes, other software not parsing and handling the header correctly is also their issue that they should fix, but at the same time it also feels like we're in a "garbage in, garbage out" situation:
- the recipient software couldn't parse it correctly, doesn't know what to do, sends out some default error response back to us
- we (almost always) couldn't do anything useful with the (error) response we got back (if any) due to mismatched expectations, and so our operation is also disrupted

keeping the potentially malformed value is detrimental for both sides

BentiGorlich (approved):
If the value is technically incorrect, than we can/should remove it, so approved :+1:

melroy89:
Actually the spec says: https://www.w3.org/TR/activitypub/#retrieving-objects

> Servers MAY use HTTP content negotiation as defined in [[RFC7231](https://www.w3.org/TR/activitypub/#bib-RFC7231)] to select the type of data to return in response to a request, but MUST present the ActivityStreams object representation in response to application/ld+json; profile="https://www.w3.org/ns/activitystreams", and SHOULD also present the ActivityStreams representation in response to application/activity+json as well. The client MUST specify an Accept header with the application/ld+json; profile="https://www.w3.org/ns/activitystreams" media type in order to retrieve the activity.

PR #719, opened 2024-04-24
Replace nucleos/antispam-bundle with omines/antispam-bundle

nobodyatroot:
I've also added the necessary field options to the contact and registration forms.

https://github.com/omines/antispam-bundle


e-five256:
I took a cursory look through of both the library and this change and things seemed fine to me. Was still able to register, observed the honeypot and validation fields were added to the register and contact forms, etc. But willing to let others weigh in in case they notice anything / have any concerns / or maybe we hear back from the dev of the original package on why it was removed

asdfzdfj (approved):
from a quick look this seems to be working fine (or at least doesn't result in obvious regressions)

though from what I could piece from some remnants of the [old bundle docs on packagist](https://packagist.org/packages/nucleos/antispam-bundle), it appears to be capable of enabling some antispam measure on _all_ forms, while the new bundle [explicitly does not support such option](https://omines.github.io/antispam-bundle/#why-is-there-no-way-to-enable-a-profile-globally-for-all-forms), due to it being perceived as a potential footgun by the bundle author, and it looks like the old config does use this function, so keep that in mind, and maybe a follow up patch to restore it in the near future

still, I'll approve this for now, since this is directly impeding our ability to make any further changes because it effectively bricked our ci and merge path while waiting for this patch to be perfected

melroy89:
> I took a cursory look through of both the library and this change and things seemed fine to me. Was still able to register, observed the honeypot and validation fields were added to the register and contact forms, etc. But wanted to let others weigh in in case they notice anything / have any concerns / or maybe we hear back from the dev of the original package on why it was removed

I received a reply back from Christian Gripp. It's a sad story. But he put his repository to private indeed. And he won't make any public announcement about it. He also advices to switch to this bundle instead.

PR #726, opened 2024-04-26
Improve post-upgrade environment detection

melroy89:
Improve post-upgrade prod/dev detection using `grep -oP`, since the `bin/console about` output could often fail to various reason like missing packages. Causing unstable behavior.

For example:  (`Uncaught Error: Class "Symfony\Bundle\MakerBundle\MakerBundle" not found`) or when a package is removed upstream (`PHP Fatal error:  Uncaught Error: Class "Omines\AntiSpamBundle\AntiSpamBundle" not found in /var/www/kbin.melroy.org/html/vendor/symfony/framework-bundle/Kernel/MicroKernelTrait.php:136
Stack trace:`).

I also switch to "prod" by default in the else statement.


nobodyatroot (approved):

melroy89:
I might should have add $BIN/../ just to be sure. I will create a follow up pr.

PR #731, opened 2024-04-27
Faq vacuum messages

TheVillageGuy:
Add sql query to free up disk space after deleting messages


e-five256:
Realistically I'm not sure why the command does a `DELETE FROM table` rather than `TRUNCATE table` which [frees disk space immediately](https://www.postgresql.org/docs/13/sql-truncate.html) (and also is instant, as it doesn't actually have to scan the table it's far faster)

e-five256:
I guess one is unqualified while the other, `queue_name = dead`, is not, so perhaps it's fine to leave as is and just call out what even the `truncate` docs mention, requires a `vacuum` to mark as freed

e-five256 (dismissed):

melroy89 (dismissed):

melroy89 (approved):

PR #736, opened 2024-04-29
Add page selector to user list at the top of the list users.html.twig

TheVillageGuy:


TheVillageGuy:
I thought having it under the local/federated buttom might be better, but your solution looks better, I agree

e-five256:
I took a look at this page when I was fixing the `Empty` panel that appeared at the bottom by mistake and wanted to fix the random `local/federated` dropdown that was causing some issues, though I can't remember what off the top of my head, maybe just styling ones

To me, the problem with this page isn't that the pagination is at the bottom, it's that there is no way to filter to the results you want. Similar to the magazine list view, this should have a panel at the top for searching by name, local/federated, banned, 2fa, sorting columns, etc. However, I noticed the user repository has no `findAll` similar to the magazine repository which allows specifying criteria, so I would have to write that from scratch, which will probably take time

TheVillageGuy:
But for now this would be an ok temporary enhancement, does everybody agree?

TheVillageGuy:
After every pull I do I test basic site functionality, like uploading, logging in signing up. After doing that I delete the test account and everytime have the thought that it would be easy to have the place selector at the top. This time I made the thought reality.

I am not sure where my solution or yours is best locations, I'm ok with both

TheVillageGuy:
@e-five256 if you agree could you approve?

asdfzdfj (approved):

PR #737, opened 2024-04-29
post-upgrade detect env by reading `.env.local.php` directly

asdfzdfj:
changed env detection in bin/post-upgrade script by just sourcing the `.env.local.php` file in php and print out the `APP_ENV` directly

there exists some grep out there that doesn't understand `-P` flag, and it would make post-install script unusable in those cases but anyone who's using this script (and mbin) in the first place should have php cli readily available


melroy89 (approved):
Fine by me 😁

melroy89:
Feel free to merge it.

PR #738, opened 2024-04-29
set `doctrine.orm.controller_resolver.auto_mapping: true`

asdfzdfj:
it looks like doctrine bundle is about to deprecate controller auto mapping stuff, and it's giving out deprecation notice about this

explicitly set `doctrine.orm.controller_resolver.auto_mapping: true` as suggested by deprecation notice to keep existing behavior and remove the notice

see https://github.com/doctrine/DoctrineBundle/pull/1762 for more info

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -22,6 +22,8 @@ doctrine:
         validate_xml_mapping: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
         auto_mapping: true
+        controller_resolver:
+            auto_mapping: true
         mappings:
             App:
                 type: attribute


nobodyatroot (approved):

melroy89:
Actually if I read the documentation correctly, there is a good reason why Symfony is trying to disabled automapper by default. Maybe we should set it to false (default soon) as well as use the `MapEntity` configuration options instead.

In the next Symfony major release this automapper option will be set to `false` by default for a good reason after all.

config/packages/doctrine.yaml
```diff
@@ -22,6 +22,8 @@ doctrine:
         validate_xml_mapping: true
         naming_strategy: doctrine.orm.naming_strategy.underscore_number_aware
         auto_mapping: true
+        controller_resolver:
+            auto_mapping: true
```
melroy89, line 26: Set to `false` and use MapEntity: https://symfony.com/doc/current/doctrine.html#mapentity-options. Which is the purpose from Symfony to deprecate this auto mapping.

asdfzdfj: yes that's a better way to go at dealing with this deprecation and it should be done, but that would need a bit of more thorough check over routes, controllers and such, and that'd definitely take some time, after that's taken care of then this could be changed to `false` like suggested

so the change here is just to make sure that things just doesn't suddenly break and went up in flames when the defaults finally changed

PR #739, opened 2024-04-29
webfinger controller event pass request once

asdfzdfj:
pass the webfinger request from controller to WebfingerResponseEvent and its handlers once rather than (implicitly) pulling the request from the stack, also adjusted WebFingerParameters the same way to have it parse the params out of the given request object

just to make it clearer what's the required input, rather than magically conjuring them out of thin air (not really) on use

---

perhaps could go one step further by parsing webfinger params then pass the parsed params on dispatch instead, or perhaps it should be done as pre/post event?

### DIFF
--- a/src/Service/ActivityPub/Webfinger/WebFingerParameters.php
+++ b/src/Service/ActivityPub/Webfinger/WebFingerParameters.php
@@ -4,7 +4,7 @@ declare(strict_types=1);
 
 namespace App\Service\ActivityPub\Webfinger;
 
-use Symfony\Component\HttpFoundation\RequestStack;
+use Symfony\Component\HttpFoundation\Request;
 
 class WebFingerParameters
 {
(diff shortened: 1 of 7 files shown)


src/Service/ActivityPub/Webfinger/WebFingerParameters.php
```diff
@@ -4,24 +4,25 @@
...
 
-    public function __construct(private readonly RequestStack $requestStack)
+    public function __construct()
     {
     }
 
-    public function getParams(): array
+    /**
+     * @return array{acccount?: string, host?: string, rel?: array}
+     */
```
BentiGorlich, line 21: Maybe use `#[ArrayShape(...)]` instead of this return type? Or is that a PHPStorm only thing?

asdfzdfj: `#[ArrayShape()]` (and a bunch of others) are [phpstorm specific attributes](https://blog.jetbrains.com/phpstorm/2020/10/phpstorm-2020-3-eap-4/), and I didn't use phpstorm so using `ArrayShape` doesn't really do anything[^1], but what I'm using can pick up these doc annotations and use them in offering autocomplete/typing just fine (and imo if phpstorm is actually worth any salt then it should also be able to pick this annotation up too)

[^1]: in fact, it just result in a red diagnostics in my editor for missing attribute class, though it looks like that can be remedied by adding [this package](https://github.com/JetBrains/phpstorm-attributes)

melroy89 (approved):

PR #706, opened 2024-04-13
Restyle login/register social and action button sections

e-five256:
Based on the feedback @TheVillageGuy gave in #642 and #643

- make buttons full form width
- make social buttons larger and more grid/flex based
  - 1 social login = full width
  - mobile = full width
- make social buttons appear before actions
  - on login page social login now appears before register links
  - on register page social register now appears before login links
- add line separators that only show up if there are elements (no social login = no separator, same for register/login links)
- social buttons on reset password form have been removed, passwords on our side aren't involved in regards to SSO
- add new admin option to place social buttons above native login/registration forms
image

<details> 
<summary>Comparison screenshots</summary>

|before|after|
|-|-|
| login all social
|image|image|
| login mobile
|image|image|
| login 1 social
|image|image|
| login 0 social
|image|image|
| login social first
||image|
| login sso mode 2 socials
|image|image|
| register all social
|image|image|
| register mobile
|image|image|
| register 1 social
|image|image|
| register 0 social
|image|image|
| register social first
||image|
| register disabled
|image|image|
| reset password all social
| (reminder socials were just straight up removed in after)
|image|image|
| reset password 1 social
|image|image|
| reset password 0 social
|image|image|
| resend account activation email
|image|image|

</details>

PR #740, opened 2024-04-29
Add admin panel option to enable/disable SSO only mode

e-five256:
This add a new admin panel option to enable/disable SSO only mode, which was previously only able to be done via modifying the `.env` file

image

Because it already exists as an environment var, we should be able to use it to populate the value in the settings DTO safely (this differs from previous settings where I've said we trend away from declaring new env vars, everyone should already have this one even if it's just `SSO_ONLY_MODE=`)

So the hope is for those that run with `SSO_ONLY_MODE=true`, this should populate to the DB correctly and already default to true

~(This doesn't seem to be working currently, the code reads like it should default to the env var, and sometimes I'll have the DB setting put into my DB, but other times I don't. I have yet to see it enter the DB as `true`, so that makes this risky for people running with SSO only mode currently enabled. Not sure if there's a great way out of this situation, maybe a migration that automatically inserts it to the settings table if that's possible)~

It does seem to work, but it feels kind of flaky. Right now I don't have the setting in my DB, rebuilt as `SSO_ONLY_MODE=true` in `.env`, and it correctly shows the admin panel as checked. So I think it is working, but maybe is worth noting for people that it's worth double checking the setting was saved if this change is pulled in

---

This is partly being done as I believe admins should be able to run their instance from the UI, of course, but also to make it easier to grab this setting as I need it in multiple components for #706 as I started to run into the limit of what I could do with SASS doing things like has child/sibling selectors based on whether a form appears on the page at all, it'd be easier to just only add things based on whether the site is in SSO only or not


nobodyatroot (approved):

e-five256:
@WebVoid @thepaperpilot heads up on this new admin setting as I believe you two are running in SSO only mode. In my testing the only time it was a bit odd was when I was constantly flipping between the env var `SSO_ONLY_MODE` `false` / `true`. But then I launched with `true` and a clean DB it did correctly only set to SSO only mode with no fields in the DB, so I don't believe there should be any problems getting this change (that is, it should stay SSO only mode even without setting it in the admin panel as long as that's what the env var says). Still might be worth watching out for though, just in case I did make a mistake in my testing

PR #755, opened 2024-05-02
Disable mercure by default on new installations

nobodyatroot:
Several admins experience issues with Mercure enabled (especially those with "large" userbases), recommend it be disabled by default.


asdfzdfj (approved):
feels a bit weird that the defaults exists here but if this is what it is then I don't think I have any more objections

melroy89:
You could also disable it in the admin panel. Just saying :D

PR #756, opened 2024-05-02
Fix 2fa redirect not being triggered on login

e-five256:
The private instance voter was missing logic to detect if the user was currently awaiting 2fa validation

symptom: users were logged in without seeing the 2fa prompt

following instructions from https://github.com/scheb/2fa/issues/23 added a token check to private instance voter to wait for user to get out of 2fa validation

### DIFF
--- a/src/Security/Voter/PrivateInstanceVoter.php
+++ b/src/Security/Voter/PrivateInstanceVoter.php
@@ -6,6 +6,7 @@ namespace App\Security\Voter;
 
 use App\Entity\User;
 use App\Service\SettingsManager;
+use Scheb\TwoFactorBundle\Security\Authentication\Token\TwoFactorTokenInterface;
 use Symfony\Component\Security\Core\Authentication\Token\TokenInterface;
 use Symfony\Component\Security\Core\Authorization\Voter\Voter;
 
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):

e-five256:
@WebVoid heads up on this but think you should mostly be unaffected, as your accounts are SSO tied so don't use 2fa natively

BentiGorlich:
I just applied it to gehirneimer.de but I still have the same problem. I did the upgrade script, flushed redis and reloaded php-fpm. So I don't think that it is a caching problem

BentiGorlich:
Did you check an admin account? Because that is the one with the problems for me

e-five256:
I just launched a private browser window 5 times, landed on my default landing page, and logged in as admin, and every single time was prompted for 2fa :thinking: 

In testing this I always was able to repro it 100% (not being asked for 2fa, and then being asked for 2fa after this change)... I'm not sure why you wouldn't be experiencing it. @nobodyatroot are you seeing any issues?

nobodyatroot:
> I just launched a private browser window 5 times, opened my default landing page, and logged in as admin, and every single time was prompted for 2fa 🤔
> 
> In testing this I always was able to repro it 100% (not being asked for 2fa, and then being asked for 2fa after this change)... I'm not sure why you wouldn't be experiencing it. @nobodyatroot are you seeing any issues?

No...I wasn't even able to reproduce the original issue on my end...

BentiGorlich:
Hmm... I tried it again just now, so all the opcache buffers should be empty either way and it is still the same....

BentiGorlich:
Ok nevermind, it seems like now that I merged the code in the tag rework branch and did another upgrade it works... Strange

Issue #754, opened 2024-05-01
SimpleLogin SSO


codenyte:
**Is your feature request related to a problem? Please describe.**
The current SSO options, Google and GitHub aren't very privacy friendly. A private alternative would be great.

**Describe the solution you'd like**
[Sign in with SimpleLogin (SIWSL)](https://simplelogin.io/docs/siwsl/intro/)

**Describe alternatives you've considered**
N/A

**Additional context**
Screenshot of a forum that implements SIWSL:
![](https://github.com/MbinOrg/mbin/assets/39193783/3865aed4-0124-48dd-b3ab-0d84b7de4938)

nobodyatroot:
@codenyte i made a PR for this, wouldn't mind someone else testing it out. i checked it in dev and everything seems to work.

https://github.com/MbinOrg/mbin/pull/762

nobodyatroot:
> Thank you very much for the quick response and implementation, but I'm not quite sure how to test this. The README says:
> 
> > Developer server explained (see "Developers" section down below)
> 
> But I couldn't find a "Developers" section below. Could you give me some quick instructions on how to set this up, so I can test it?

Good catch, it should point to this: https://docs.joinmbin.org/contributing/developement_server

If you're not familiar with running a kbin/Mbin server, we can wait until https://github.com/MbinOrg/mbin/pull/654 is merged (I've been testing on kbin.run), then I can load up https://github.com/MbinOrg/mbin/pull/762

nobodyatroot:
> Thanks for implementing this! There's just one tiny thing I would change: Currently, the SIWSL option just seems to use a generic icon, maybe [this icon](https://simpleicons.org/?q=simplelogin) from [Simple Icons](https://github.com/simple-icons/simple-icons) could be added? It's licensed under CC0.

I made a PR for this: https://github.com/MbinOrg/mbin/pull/781

PR #762, opened 2024-05-03
Add SimpleLogin SSO

nobodyatroot:
Generic OAuth2 implementation of SimpleLogin, lifted bulk of implementation details from ZITADEL authenticator with the proper SimpleLogin endpoints and resource keys updated.

Closes https://github.com/MbinOrg/mbin/issues/754


src/Entity/User.php
e-five256, line 1: needs to be added down here as well https://github.com/MbinOrg/mbin/blob/6c02def44f1f2fb78ebb5d0fbe3ac780d17448df/src/Entity/User.php#L760

nobodyatroot: good catch, totally forgot about that

src/Security/SimpleLoginAuthenticator.php
```diff
@@ -0,0 +1,178 @@
...
+                    $simpleloginUser->getEmail()
+                );
+
+                $avatar = $this->getAvatar($simpleloginUser->getPictureUrl());
+
+                if ($avatar) {
+                    $dto->avatar = $this->imageFactory->createDto($avatar);
+                }
+
+                $dto->plainPassword = bin2hex(random_bytes(20));
```
melroy89, line 116: what is this?

nobodyatroot: Which line, the password? All our SSO options do this... because the user needs a password (albeit random) so the field isn't null:

image

The API does as well, but it's a little "tougher":

image

I guess we could argue if the way it's done now has enough entropy so as to not be brute forced...

e-five256:
The only odd thing I'm finding is that my username is created with the email rather than the name specified

Like if I have an email of `mbindev@fakesimplelogin.com` and a name of `efive` the user created will be `mbindev` :thinking:

src/Security/SimpleLoginAuthenticator.php
melroy89: yea indeed the password, but I understand now. Thanks!

melroy89 (approved):

e-five256 (approved):

PR #654, opened 2024-04-01
Rework the tag system

BentiGorlich:
The most integral thing I've done here is to rework the db schema for hashtags. We now have a table called `hashtag` which contains the string tag it self as well as other info (atm only `banned`) and `hashtag_link` which links posts of any kind to hashtags if they contain them. 
The migration I wrote contains sql code to transition between the new and old schema.

Changing this causes a lot of necessary adjustments. I have not tested the API methods, because I don't know how :sweat_smile: 
That should definitely be done.

The reason I did it is that the hashtag queries are one of the slowest queries we have at the moment and it is not possible to reasonably implement banning, blocking or subscribing to hashtags with the old system.

One cool thing I did in the process is to write the `NativeQueryAdapter` for pagination. It allows to put any SQL query in there and get a pagination interface for it. I replaces some of the queries in the `SearchRepository` with it.

#### 1. [new] Hashtag panel:

grafik

#### 2. [new] Hashtag Admin Panel Options:

There is now a button to ban/unban a hashtag (see screenshot above). The ban button is a danger button, the unban button is not.

#### 3. [removed] Related Hashtags:

I decided to remove the hard coded related hashtags, but keep the box they were in (see screenshot above). I tried removing the box as well, but it just looked weird. We can add the usual filter options to it in the future

#### 4. [removed] "tags" Input:

| After | Before |
| --- | --- |
| grafik | grafik |

This can be reverted, but the input always just confused me anyways, so removed it for ease of use.

### Behaviour of banned hashtags:

I went for a more temporary approach of the ban: keep posts that already exists, but block new posts from being created. 

#### Local

If a user tries to create a post with a banned hashtag in them they see this error message:
grafik

I did not add an error message that specifically said "Tag x is banned" so the user can't circumvent the ban that easily. If this doesn't make any sense to anyone we can of course change it.

#### Remote

When a post is created with a banned hashtag in the body a `TagBannedException` gets thrown. This is caught in the `CreateHandler` and the `ChainActivityHandler` and the post is discarded (these are the only two points where posts get created that I could find). So when we get a remote post which contains a banned hashtag we do not create it at all.

#### visibility

When a post contains a banned hashtag it is hidden in most places: search, timeline (thread and microblog). The one thing where they still show up is on the hashtag page itself. My idea was that the admin should still be able to see all the posts that acutally exists containing the hashtag. I considered making the hashtag side admin only if a hashtag is banned, but I didn't do it, yet. I am counting on your feedback, how we should handle it.

PR #770, opened 2024-05-05
Fix infinite scrolling of results on search page

e-five256:
infinite scrolling on the search page currently does not work, throwing the javascript exception of invalid json and showing the paginator instead. this updates the controller to handle xml requests

(this might sound familiar, it also isn't working by the same mechanic on user pages so expect that some time in the future)

based on [similar work](https://github.com/MbinOrg/mbin/pull/654/commits/e8fb274ac83dc875f92f4e7c6054b29130fe6be9) Benti did to get infinite scrolling working on tag pages


BentiGorlich:
Superseded by #1952

PR #773, opened 2024-05-11
Fixes to user filters on active user sidebar and people pages

e-five256:
This ended up being a few different things. I started making some of the changes as a separate PR but since all the code and views are very similar it was a bit difficult to test separately

---

This makes some changes to what users can be shown as Active Users in the sidebar or Local/Federated Users on the People pages in and out of magazines

Mainly noticed that suspended users still showed up in active users on the sidebar. When I went to fix this and couldn't get any of my users to show up at all, realized it was because they didn't have About descriptions set and that was still required for the sidebar and inside magazine people pages

---

- Fix user `user.visibility` not being filtered for `visible` only
  - On `PeopleFrontController` (`/people`)
  - On `MagazinePeopleFrontController` (`/m/<mag>/people`)
  - On `ActiveUserComponent` (sidebar)
- Remove user "About" requirement. This matches the change that happened to `PeopleFrontController`
  - On `MagazinePeopleFrontController` (`/m/<mag>/people`)
  - On `ActiveUserComponent` (sidebar)
- Remove user "Avatar" requirement. This matches the change that happened to `PeopleFrontController`
  - On `MagazinePeopleFrontController` (`/m/<mag>/people`)
  - **NOT** on `ActiveUserComponent` (sidebar), as the component is only avatars it remains required to show users there
- Fix `user.isDeleted` not being filtered (From #777)
  - On AP events
  - On AJAX user suggestions (not _quite_ sure this is actually used, can't find where this route is ever surfaced)
  - On `ActiveUserComponent` (sidebar)
  - On `MagazinePeopleFrontController` (`/m/<mag>/people`)
- Rename `findPeople()` -> `findUsersForMagazine()` (From #777)


e-five256:
Whoops, need to add it to `findPeople`, should've read closer, as it'd still show non-visible users inside specific magazine pages

nobodyatroot (approved):

PR #775, opened 2024-05-12
Exclude `node_modules` and `vendor` directories from PHP linter

nobodyatroot:
I noticed our PHP linter was "fixing" files in the `node_modules` and potentially `vendor` directories, I've updated the linter config to exclude files from these directories.

### DIFF
--- a/.php-cs-fixer.dist.php
+++ b/.php-cs-fixer.dist.php
@@ -8,6 +8,7 @@ $finder = (new PhpCsFixer\Finder())
         'var',
         'node_modules',
         'vendor',
+        'docker',
     ])
 ;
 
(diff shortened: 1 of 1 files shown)


.php-cs-fixer.dist.php
```diff
@@ -4,7 +4,11 @@
 
 $finder = (new PhpCsFixer\Finder())
     ->in(__DIR__)
-    ->exclude('var')
+    ->exclude([
+        'var',
+        'node_modules',
+        'vendor',
```
e-five256, line 10: Thanks for this, small suggestion

```suggestion
        'vendor',
        'docker',
```

of adding docker to this too. I could never run it and get

```
/MbinOrg/mbin/docker/storage/postgres): Failed to open directory: Permission denied
```

So I had been running it by specifying the php files I changed individually each time, should have actually looked to see that this was possible :sweat_smile:

e-five256 (approved):

PR #776, opened 2024-05-13
Do not auto-close issues because of stale

melroy89:
People in the dev community do not want to auto-close the stale issues.

I do not agree (since I don't like ending up with 10k issues), but... here is the PR anyway. 
If more people approving this PR then rejecting, then it will be merged.

### DIFF
--- a/.github/workflows/stale.yml
+++ b/.github/workflows/stale.yml
@@ -18,5 +18,5 @@ jobs:
           exempt-issue-labels: "high priority"
           days-before-issue-stale: 50
           days-before-pr-stale: 40
-          days-before-issue-close: 6
+          days-before-issue-close: -1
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):
We can revisit this at a later time... I'd be "against" this if we had more devs actively working on the project, but right now it seems like we're all overloaded with real work and not making any progress addressing Mbin issues. It's hard enough getting PRs reviewed, tested and approved right now...

melroy89:
> I'd be "against" this if we had more devs actively working on the project, but right now it seems like we're all overloaded with real work and not making any progress addressing Mbin issues. It's hard enough getting PRs reviewed, tested and approved right now...

Hence the reason to auto-close issues ;P. To focus on the real work/issues.

PR #777, opened 2024-05-13
Show users without magazine about hide deleted users

melroy89:
- Show users on the magazine people listing page even when they do not have about or an avatar
- Refactor method name to reflect the actual behavior
- Also check for user `is_deleted` database column in queries
- Small lint doc notice added.


melroy89:
Superseded by https://github.com/MbinOrg/mbin/pull/773

PR #781, opened 2024-05-14
Add available simple-icons to SimpleLogin SSO

nobodyatroot:
Add available simple-icons (https://github.com/simple-icons/simple-icons-font) to SimpleLogin SSO options.

### DIFF
--- a/assets/styles/app.scss
+++ b/assets/styles/app.scss
@@ -1,6 +1,7 @@
 @import '@fortawesome/fontawesome-free/scss/fontawesome';
 @import '@fortawesome/fontawesome-free/scss/solid';
 @import '@fortawesome/fontawesome-free/scss/brands';
+@import 'simple-icons-font/font/simple-icons';
 @import 'glightbox/dist/css/glightbox.min.css';
 @import 'mixins/animations';
 @import 'mixins/kbin';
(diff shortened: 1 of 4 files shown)


e-five256:
Likely want to consider https://github.com/simple-icons/simple-icons/blob/master/DISCLAIMER.md in this PR. Looking through things, mainly keycloak seems to specify (or linux foundation as a whole) logo usage is allowed in direct context to the links or website of the products. I'm not a lawyer, so can't say for sure. I think this was sort of brought up with zitadel as well where I'm not sure if someone would have this say keycloak, or would rename it to like "log in with our sso method"

nobodyatroot:
> Likely want to consider https://github.com/simple-icons/simple-icons/blob/master/DISCLAIMER.md in this PR. Looking through things, mainly keycloak seems to specify (or linux foundation as a whole) logo usage is allowed in direct context to the links or website of the products. I'm not a lawyer, so can't say for sure. I think this was sort of brought up with zitadel as well where I'm not sure if someone would have this say keycloak, or would rename it to like "log in with our sso method"

I removed Keycloak... not worth the potential hassle.

e-five256 (approved):
I think this could use a little restyling but not quite sure
Looking at the page, maybe the SSO options shouldn't be bold
And their icons maybe should be `1.0rem` while the text itself stays `0.85rem`
but not going to dilute this PR, I can always make another if I end up wanting to change it

PR #783, opened 2024-05-15
Implement report notifications

### LINKED ISSUE #255: Notifications for reports
Right now I can only see reports by going to the magazine panel -> reports. I don't get any indication that there is a new report

If a user reports something in a magazine I am moderating I should get a notification, with the option to get the notification via email.

None

**Additional context**
After #235 hits main: For remote magazines the local admin, which is set as the owner for remote magazines, should probably not be notified or at least they should get the option to turn it off

BentiGorlich:
- notifications when
  - a report is created -> admins, global mods, magazine mods
  - a report is accepted -> ~~reporting user and~~ reported user
  - ~~a report is rejected -> reporting user~~
- notifications are marked as read when a user can see them:
  - admin report panel: all notifications for the current user with attached reports
  - magazine report panel: all notifications for the current user with attached reports for something in the magazine
  - personal report panel: all notifications with attached reports reported by the current user
- change the personal report panel to include only the reports reported by the current user

Closes #255


e-five256:
> notification when
>- a report is rejected -> reporting user

I'll have to check this out but I can see it being a bit confusing. One of the issues it feels like we have is that people report remote content, for things like rules of the remote magazine, where really the deciding factor is the remote magazine moderators, and not the site admin. I'm not sure if it's common to reject the report in this case, or just ignore it, feels like there isn't really a good solution and then even if the content is moderated remotely, it never makes it back to Mbin, making things even more confusing

BentiGorlich:
We could add a check for open reports when we receive a delete activity and mark the report as accepted. I don't know whether that solves the problem.
However from moderating my mastodon server I gotta say that I deal with reports directly even if it is a remote userthat is reported, because youd server might just have different rules in general that apply to everything than the server of the remote magazine...

jwr1:
> notifications when
a report is accepted -> reporting user and reported user
a report is rejected -> reporting user

I can see why the reported user should know if their content is removed (so they can discuss it with the mod/admin), but why should the reporting user be notified if it's accepted or not? Let's say I see a bunch of spam posts and I report all of them, I can't think of a case where I would want to be notified of whether the posts were removed or not. Reporting is supposed to let the mod/admin know of possibly problematic content, and then they can deal with it as they see fit.

jwr1:
Also, it would be great if this works for the API as well. I'm assuming the new notification types show up in the API (`report_created`, `report_approved`, and `report_rejected`), but will the new `report_id` field be added?

BentiGorlich:
> the reporting user shouldn't have to think/worry about the content once they have reported it.

That is true, but I thought that I would want to know if it helps to report stuff... It shows that the admins/mods care about reports... 

> Also, it would be great if this works for the API as well. I'm assuming the new notification types show up in the API (report_created, report_approved, and report_rejected), but will the new report_id field be added?

Tbh I always forget the API, so yes that will definitely be added :sweat_smile:

src/Controller/Api/Notification/NotificationBaseApi.php
```diff
@@ -110,6 +111,13 @@ protected function serializeNotification(Notification $dto)
                 $ban = $dto->getSubject();
                 $toReturn['subject'] = $this->magazineFactory->createBanDto($ban);
                 break;
+            case 'report_created_notification':
+            case 'report_rejected_notification':
+            case 'report_approved_notification':
+                /** @var ReportCreatedNotification $n */
+                $n = $dto;
+                $toReturn['reportId'] = $n->report->getId();
+                break;
```
ryanjsims, line 120: The notification structure has the actual subject of the notification detailed - using the report id instead breaks this convention, and would also make it impossible for regular users to have any information about the report contained in this notification displayed by the app using the API, since retrieving reports is a moderator and admin only API endpoint.
This solution would not be sufficient if we wanted to display the reason a user's post was deleted in the notification using the API, for example

ryanjsims: What I would recommend would be to follow the convention of the other notifications and embed a subset of the report in the notification, at least the `reason`, `magazine`, `type`, and `subjectId` of the report
`subjectId` would be the post/comment/entry that was reported and removed

jwr1:
Also, what about the case where a reported post might be breaking magazine rules but not instance rules? If an admin rejects a report because they want the magazine moderator to deal with it instead (at their own discretion), won't it still notify the reporting user? Or are we supposed to just leave it in pending?

BentiGorlich:
> Also, what about the case where a reported post might be breaking magazine rules but not instance rules? If an admin rejects a report because they want the magazine moderator to deal with it instead (at their own discretion), won't it still notify the reporting user? Or are we supposed to just leave it in pending?

You're supposed to just leave it pending. Since there is just one report for everyone and no distinction between moderator and admin level...

src/Controller/Api/Notification/NotificationBaseApi.php
BentiGorlich: I'll check that

jwr1:
> You're supposed to just leave it pending. Since there is just one report for everyone and no distinction between moderator and admin level...

And then the moderator never does anything with the report, so it just sits in pending for all eternity... Maybe we need to add a "hide" button to reports, to hide them from pending but not actually approve or reject them.

jwr1:
Also, just wondering, do reports even federate to and from Lemmy correctly?

BentiGorlich:
> And then the moderator never does anything with the report, so it just sits in pending for all eternity... (Meaning I have to look at it every single time I check the reports page.) Maybe we need to add a "hide" button to reports, to hide them from pending but not actually approve or reject them.

Well maybe, but way out of scope of this PR...

> Also, just wondering, do reports even federate to and from Lemmy correctly?

Yes, they should. Report federation was introduced in #249

src/Controller/Api/Notification/NotificationBaseApi.php
BentiGorlich: Ok, here are my thoughts:

- `ReportCreatedNotification`: only admins and mods can get this, so imo its fine to only send the id of the report, but of course a dto of the report could be added for this.
- `ReportRejectedNotification`: a user should be able to see their own reports, but if not maybe add the subject of the report here, as there is no reason supplied for rejecting a report.
- `ReportApprovedNotification`: since this is going to the author of the reported content, the subject that was reported should be enough. 

Would you be fine with that?

The reason I do not just want to add a report dto to all of them is simple: you should never be informed who is reporting your content for which reason, unless the report is accepted, but even then the reason the reporting user supplied is not necessarily the one the mod removed your content for. Since we do not have a "moderators_reason" field (or something similar) it would not make much sense to just relay that information to the reported user

nobodyatroot:
so.... what is the group consensus on this @BentiGorlich @e-five256 @jwr1 @ryanjsims @melroy89? PR has been open since May...

jwr1:
Honestly, I'm still not sold on the _reporting_ user being notified. I do think the _reported_ content's user should be notified if their post/comment was removed (in case they want to appeal), and it makes sense, for obvious reasons, why admins, global mods, and mag mods should be notified on report creation.

BentiGorlich:
Yeah I have to get back on this. I am fine with what @jwr1 said. Have to change the API objects as well

nobodyatroot:
One minor thing... the user having their post being reported (and accepted) shows the "open report" link, but we throw access denied since the user is not a mod to see the report. maybe we should remove the "open report" link?

image

trying to open the report :

image

BentiGorlich:
Yeah that makes sense. I think I will just not display the link when the user does not have permissions to view it

nobodyatroot (approved):
i've already used this quite a bit, very useful for prompt mod action.

PR #793, opened 2024-05-28
Update packages

melroy89:
- Update recipe (doctrine bundle)
- Update js packages (minor version only)
- Update Symfony packages (7.0.x patch releases only)
- Update Twig
- Update pagerfanta
- Update doctrine
- And the rest... phpunit, phpstan, liip, scheb 2fa, monolog, etc.


nobodyatroot (approved):
LGTM as well

PR #801, opened 2024-06-01
Add Brazilian Portuguese and Danish language selectors

nobodyatroot:
Add Brazilian Portuguese and Danish language selectors, also update `timeago_controller.js` to reflect proper `pt_BR` usage as they do not have `pt` only translation available: https://github.com/hustcc/timeago.js/tree/master/src/lang


melroy89 (approved):
Yep.

PR #803, opened 2024-06-03
Try to fix wrong urls for images in the related panels

BentiGorlich:
This PR aims to get rid of a problem with the "related" panels on the homepage. I observed this problem on gehirneimer.de where the images in entries and posts and the avatars of users would not load in the related panels (everywhere else the images were loading just fine). Reason for that wrong source URLs. I suspect that the way the cached components work the environment can be loaded wrongly, but I am not sure of the reason.

The caching before this PR for the panels would render the HTML of the components and put that HTML in our cache (redis). 
After this PR only the IDs of the entries, posts and users are put in the cache and the components get rendered normally. 

I have this PR running on gehirneimer and it has solved the image problem


BentiGorlich:
Since I changed this a few hours ago I didn't have the problem again -> ready for review

melroy89 (approved):

PR #804, opened 2024-06-04
change sso authentication success route to front page

CocoPoops:
Changes the page which a sso user is redirected to when logged in to the front page so that it is the same as non sso users

### DIFF
--- a/src/Security/AzureAuthenticator.php
+++ b/src/Security/AzureAuthenticator.php
@@ -119,7 +119,7 @@ class AzureAuthenticator extends OAuth2Authenticator
 
     public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
     {
-        $targetUrl = $this->router->generate('user_settings_profile');
+        $targetUrl = $this->router->generate('front');
 
         return new RedirectResponse($targetUrl);
     }
--- a/src/Security/FacebookAuthenticator.php
+++ b/src/Security/FacebookAuthenticator.php
@@ -152,7 +152,7 @@ class FacebookAuthenticator extends OAuth2Authenticator
 
     public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
     {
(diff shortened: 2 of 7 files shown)


nobodyatroot:
Good catch! Maybe we should make each of them do what the `KbinAuthenticator.php` `onAuthenticationSuccess()` method does for consistency? 

https://github.com/MbinOrg/mbin/blob/635c58aaed217e38173a371772089b193ff4756d/src/Security/KbinAuthenticator.php#L49

```php
public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
    {
        if ($targetPath = $this->getTargetPath($request->getSession(), $firewallName)) {
            return new RedirectResponse($targetPath);
        } else {
            return new RedirectResponse($this->urlGenerator->generate('front'));
        }
    }

CocoPoops:
> Good catch! Maybe we should make each of them do what the `KbinAuthenticator.php` `onAuthenticationSuccess()` method does for consistency?
> 
> https://github.com/MbinOrg/mbin/blob/635c58aaed217e38173a371772089b193ff4756d/src/Security/KbinAuthenticator.php#L49-L56
> 
> Honestly I'm not sure why it was different to begin with... maybe an oversight as the parameter list appears to be the same.

i get a 500 error when logging with the function the way you suggested
i don't really know enough to know why that's the case

nobodyatroot:
> > Good catch! Maybe we should make each of them do what the `KbinAuthenticator.php` `onAuthenticationSuccess()` method does for consistency?
> > https://github.com/MbinOrg/mbin/blob/635c58aaed217e38173a371772089b193ff4756d/src/Security/KbinAuthenticator.php#L49-L56
> > 
> > Honestly I'm not sure why it was different to begin with... maybe an oversight as the parameter list appears to be the same.
> 
> i get a 500 error when logging with the function the way you suggested i don't really know enough to know why that's the case

lol, that probably explains why it's different then.

nobodyatroot (approved):

PR #806, opened 2024-06-04
Authentik SSO

CocoPoops:
Adds Support for Authentik SSO
https://goauthentik.io

Basically copied from the Zitadel implementation but with Authentik endpoints


nobodyatroot (approved):
Verified functionality with @CocoPoops, looks good to me as it's a copy of an existing SSO integration

Issue #792, opened 2024-05-25
OAUTH2 support for Privacy Portal Provider


privacyguard:
We have implemented OAUTH2 support for the [Privacy Portal](https://privacyportal.org) provider in our [mbin fork](https://github.com/privacyportal/mbin).

Is there any interest in getting this back into the main repo? Privacy Portal is a privacy focused and free OAUTH2 provider. It uses passwordless authentication and has integrated email aliases for privacy. When someone signs in using the Privacy Portal OAUTH2, mbin would receive a Privacy Alias instead of the real user email.

If you would like to test this, it's deployed on our community mbin node:
https://community.privacyportal.org/login

![mbin_screenshot](https://github.com/MbinOrg/mbin/assets/92675882/32fd1f36-bfea-427f-9f78-d65d7760225d)

BentiGorlich:
Integrating additional SSO providers is always appreciated :+1:

nobodyatroot:
Closed by https://github.com/MbinOrg/mbin/pull/818

PR #818, opened 2024-06-12
Add support for the Privacy Portal OAUTH provider

privacyguard:
This PR adds support for SSO with Privacy Portal.

**Related Issue:** https://github.com/MbinOrg/mbin/issues/792

**OAUTH Provider:** https://privacyportal.org

**Mbin Configuration:**
OAUTH_PRIVACYPORTAL_ID=<CLIENT_ID obtained from https://app.privacyportal.org >
OAUTH_PRIVACYPORTAL_SECRET=<CLIENT_SECRET obtained from https://app.privacyportal.org >

**Provider Specific Information:**
- Emails provided are Privacy Aliases which means there can never be a match with existing user emails in the DB
- Privacy Portal uses Passwordless Authentication


nobodyatroot (changes requested):
You'll need to update:

https://github.com/MbinOrg/mbin/blob/9b207e2fedc7ad9472cf52acc06a72f61fa9de3d/src/Entity/User.php#L762-L765

to include Privacy Portal

nobodyatroot (approved):
LGTM

PR #830, opened 2024-06-15
Implement message federation and redesign message view

BentiGorlich:
This PR implement federation for messages. They use the AP type `ChatMessage` which is not part of the AP standard, but is used by Lemmy and Pleroma. The main goal is interoperability with Lemmy.

- [x] message federation works between Mbin <-> Mbin
- [x] message federation works between Lemmy <-> Mbin
- [x] redesign message view
- [x] implement messaging with more than 2 participants  
      Lemmy does not support it and we do not have a way of distinguishing between the lemmy and mbin servers. The backend works for multiple participants, but I did not add an option in the UI to start a thread with multiple participants. I want it to show a warning when a lemmy user is part of a multi thread
- [x] update activities from lemmy are working (editing)

### Redesigned message view:
Overview:
| Before | After |
| ------ | ------ |
| grafik | grafik

Chat:
| Before | After |
| ------ | ----- |
| grafik | grafik |

Only the middle part scrolls and the view will scroll to the bottom when loaded


BentiGorlich:
Mastodon private mentions will be implemented separately and will probably not use our message system, but will be represented as a normal post (regardless of which type) with a visibility of `private`.

BentiGorlich:
@nobodyatroot @e-five256 @melroy89 should I implement messages with more than 2 participants or is that too much for this PR? I don't think that lemmy supports it, but implementing it shouldn't be that hard. The rest of the code should already handle infinite number of participants

nobodyatroot:
> @nobodyatroot @e-five256 @melroy89 should I implement messages with more than 2 participants or is that too much for this PR? I don't think that lemmy supports it, but implementing it shouldn't be that hard. The rest of the code should already handle infinite number of participants

that sounds like a cool Mbin specific thing as long as it doesn't break 2 participant message compatibility with lemmy. i'll leave it up to you if it's too much work for the PR since you're doing all the work ;-)

BentiGorlich:
I'll check if lemmy support breaks when there is more than one user in the `to` field. If Lemmy doesn't break, then I'll implement it, but I suspect, that Lemmy breaks :sweat_smile:

melroy89:
There are some file conflicts too bad.

migrations/Version20240626130346.php
```diff
@@ -0,0 +1,30 @@
...
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20240626130346 extends AbstractMigration
```
melroy89, line 13: Do we need 2 separate migration files in this PR? Can't it be 1?

src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
```diff
@@ -40,23 +42,20 @@ public function __construct(
     public function __invoke(CreateMessage $message): void
     {
         $this->object = $message->payload;
-        $this->logger->debug('Got a CreateMessage of type {t}', [$message->payload['type'], $message->payload]);
-
+        $this->logger->debug('Got a CreateMessage of type {t}, {m}', ['t' => $message->payload['type'], 'm' => $message->payload]);
         try {
-            if ('Note' === $this->object['type']) {
+            if (!$this->activityPubManager->isActivityPublic($message->payload)) {
```
melroy89, line 47: Do you think this is the best way of checking for a private message rather than checking on type `ChatMessage`? 😕

BentiGorlich: No its not. I should definitely just check the type. I'll change this. The original thought was the private mentions from mastodon. So I will not delete the medthod declaration, since it will be useful later on when implementing that

BentiGorlich: done

migrations/Version20240626130346.php
BentiGorlich: done

src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
```diff
@@ -95,4 +97,14 @@ private function handlePage(): void
...
     }
+
+    private function handlePrivateMessage(): void
+    {
+        $this->messageManager->createMessage($this->object);
+    }
+
+    private function handlePrivateMentions(): void
+    {
+        // TODO implement private mentions
```
melroy89, line 108: TODO left?

src/Service/ActivityPubManager.php
```diff
@@ -844,6 +844,47 @@ public function extractMarkdownSummary(array $apObject): ?string
...
+        if (isset($apObject['source']) && isset($apObject['source']['mediaType']) && isset($apObject['source']['content']) && ApObjectExtractor::MARKDOWN_TYPE === $apObject['source']['mediaType']) {
+            return $apObject['source']['content'];
+        } else {
+            $converter = new HtmlConverter(['strip_tags' => true]);
+
+            return stripslashes($converter->convert($apObject['content']));
+        }
+    }
+
+    public function isActivityPublic(array $payload): bool
```
melroy89, line 858: This method is no longer used, correct?

BentiGorlich: See: https://github.com/MbinOrg/mbin/pull/830#discussion_r1659967075

src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
BentiGorlich: Not for this PR, but for later, yes

melroy89: Sure OK. I understand. I just wanted to point it out, you never know. right?

melroy89 (dismissed):
Tested again with BentiGorlich, still works with Mbin <> Mbin.

melroy89 (approved):
We still want to show form errors, but I approve the current PR as is.

Issue #822, opened 2024-06-13
First letter of new tags is sometimes partially cut off (e.g., the letter "j")


MHLoppy:
**Describe the bug**
As per title; when "j" is the first letter while entering a new tag, the bottom and left sides are cut off, making it appear as an "i" instead. It also seems to occur similarly for other letters like "y" and "p", though I haven't carefully tested all letters. However, it can be observed that it's not (obviously, at least) happening with other letters such as "a".

**On which Mbin instance did you find the bug?**
fedia.io ( https://fedia.io/ )

**Which Mbin version was running on the instance?**
1.6.0

**To Reproduce**
1. Submit something new (as far as I know, type of submission doesn't matter?).
2. Begin typing in a new tag starting with the letter "j", but don't add it yet.
3. Part of the letter "j" is obscured.
4. (if the tag is then added, the "processed" tag does not seem to have the problem; similarly, adding a new first letter makes the full letter "j" appear)

**Expected behavior**
The full letter should be visible.

**Screenshots**
![image](https://github.com/MbinOrg/mbin/assets/12670674/d8eb9eb6-41f7-47aa-9091-b35f3c7e602c)
![image](https://github.com/MbinOrg/mbin/assets/12670674/b629f8ee-c00f-40d4-8a88-709eee804e0a)

**Desktop (please complete the following information):**
 - Windows 10 (build 19045.4291)
 - Mozilla Firefox
 - 125.0.3 (64-bit)

~~**Smartphone (please complete the following information):**~~
N/A

**Additional context**
The problem and screenshots are from just over a week ago (!), I was just late to action it here.

nobodyatroot:
I found this is related to the "autocomplete" feature that is making the tags into the nice boxes... https://symfony.com/bundles/ux-autocomplete/current/index.html#using-with-a-texttype-field

Turning off autocomplete makes this problem go away, but I'd like to understand what the CSS is doing to cause this... I think it's the border css.

PR #835, opened 2024-06-17
Remove border radius on autocomplete form to fix letters from getting cutoff at the beginning of the item

nobodyatroot:
Closes https://github.com/MbinOrg/mbin/issues/822

Before:

image

After:

image

### DIFF
--- a/assets/styles/layout/_forms.scss
+++ b/assets/styles/layout/_forms.scss
@@ -151,7 +151,7 @@ input[type=select-one] {
   display: block;
   padding: 1rem .5rem;
   width: 100%;
-  text-indent: .1rem;
+  text-indent: .1rem !important;
 }
 
 .password-preview {
--- a/assets/styles/layout/_forms.scss
+++ b/assets/styles/layout/_forms.scss
@@ -264,6 +264,7 @@ form {
       color: var(--kbin-input-text-color);
       width: auto;
       min-width: 8rem;
+      border-radius: 0 !important;
     }
 
     & > * {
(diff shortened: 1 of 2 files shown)


nobodyatroot:
I just realized it messed with the other borders... going to see if I can undo that while retaining the desired changes.

assets/styles/layout/_forms.scss
```diff
@@ -151,6 +151,7 @@ input[type=select-one] {
   display: block;
   padding: 1rem .5rem;
   width: 100%;
+  text-indent: .1rem !important;
```
BentiGorlich, line 154: What is this for?

nobodyatroot: I noticed that the text entry was still being cutoff a little without it:

before

image

after

image

this autocomplete thing pulls in it's own stylesheet that seems to really be messing things up with our css...

image

e-five256: It does seem to effect all input boxes on the site though, was that expected, or would only the auto complete ones need it?

nobodyatroot: that badges form isn't autocomplete enabled and it still needed it if you can tell by the image...i know, it's super subtle. i guess my assumption about this autocomplete css import causing problems isn't totally 100% accurate (although, maybe it is?) i dunno, seems like we have a lot of suboptimal stuff going on.

e-five256: Ah, I see, sorry, I got confused because I was like, you could just remove `!important` and remove the `0 !important` where it is but then I realized as you said it's coming from symfony so they don't really give you the option of doing that

e-five256 (approved):

PR #842, opened 2024-06-20
Update all npm packages and address dependabot security notice

nobodyatroot:
Closes:

image

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -51,13 +51,13 @@
             }
         },
         "node_modules/@babel/code-frame": {
-            "version": "7.24.6",
-            "resolved": "https://registry.npmjs.org/@babel/code-frame/-/code-frame-7.24.6.tgz",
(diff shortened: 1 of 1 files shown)


BentiGorlich:
Did you just run `npm upgrade` or `npm audit fix`?

nobodyatroot:
> Did you just run `npm upgrade` or `npm audit fix`?

In this case I ran `npm update` since we haven't done it in awhile. To target the vulnerable notice only, then yea `npm audit fix` would take care of it.

BentiGorlich:
I still get deprecation notices when running the `post-upgrade` script though :thinking:

nobodyatroot:
> I still get deprecation notices when running the `post-upgrade` script though 🤔

Afaik, `npm update` won't forcibly remove/replace/change packages that could potentially break the build, I think it only updates to latest minor versions unless there's some explicit version pinning on a package that stops it.

PR #847, opened 2024-06-20
Fix entries with not showing preview icon

BentiGorlich:
- when an entry is of type article and does not point to a url, but has an image attached to it, the preview should show the image, right now there is just no preview icon in this case

| before | after |
| ------ | ----- |
| grafik | grafik grafik
 |

### DIFF
--- a/templates/components/entry.html.twig
+++ b/templates/components/entry.html.twig
@@ -148,8 +148,8 @@
                                 </span>
                             </li>
                         {% endif %}
-                        {% if entry.hasEmbed %}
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):

e-five256:
Floating is fun was one such location where every post was shown as article so I'd have to open fedia to actually see the image

I see the icon now shows as preview:
https://gehirneimer.de/m/floatingisfun@fedia.io

but none of the previews work for me in firefox or chrome, wanted to double check if you see the same thing

PR #853, opened 2024-06-21
Improve actor fetch

BentiGorlich:
I observed an issue on my instance where moderators where removed in one update and re-added in the next one. I think the problem was just that it tried to fetch the actors from their remote instance although they were already in the db.

This PR will cause a lot less requests to remote instances when resolving actor URLs, because it will not call the remote instance if the user already exists in the db anymore.

#825 #852 CPU problems with this PR, gotta find out why


BentiGorlich:
@nobodyatroot @e-five256 do you think it might be possible that the high CPU usage this PR creates comes from us not having an index on the `ap_public_url` field?

BentiGorlich:
yes I think that was it:
grafana-screenshot

nobodyatroot (approved):
That actually makes sense based on what was happening...

src/Service/ActivityPubManager.php
```diff
@@ -157,38 +157,40 @@ public function findActorOrCreate(?string $actorUrlOrHandle): null|User|Magazine
             return $this->userRepository->findOneBy(['username' => $name]);
         }
 
+        $user = $this->userRepository->findOneBy(['apPublicUrl' => $actorUrl]);
```
nobodyatroot, line 160: It seems `findOneBy` is breaking with this change to `'apPublicUrl'` instead of `'apProfileId'`, I switched it back to `'apProfileId'` and now my microblog section is working.

src/Service/ActivityPubManager.php
```diff
@@ -157,38 +157,40 @@ public function findActorOrCreate(?string $actorUrlOrHandle): null|User|Magazine
...
+        if ($user instanceof User) {
+            $this->logger->debug('found remote user for url "{url}" in db', ['url' => $actorUrl]);
+            if ($user->apId && (!$user->apFetchedAt || $user->apFetchedAt->modify('+1 hour') < (new \DateTime()))) {
+                $this->dispatchUpdateActor($user->apProfileId);
+            }
+
+            return $user;
+        }
+        $magazine = $this->magazineRepository->findOneBy(['apPublicUrl' => $actorUrl]);
```
nobodyatroot, line 169: Same thing here, I reverted `'apPublicUrl'` to `'apProfileId'`... what was the reason for changing this?

src/Service/ActivityPubManager.php
nobodyatroot: it's also why i was getting a bunch of these errors... the mastodon users already exist, it just couldn't find them since several users (~20k) have their 'apPublicUrl' set to null, so it's trying to create them again..

```
21:00:24 WARNING   [messenger] Error thrown while handling message App\Message\ActivityPub\Inbox\ActivityMessage. Sending for retry #3 using 4800000 ms delay. Error: "Handling "App\Message\ActivityPub\Inbox\ActivityMessage" failed: An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "user_ap_id_idx"\nDETAIL:  Key (ap_id)=(AlexDafonte@red.niboe.info) already exists." ["class" => "App\Message\ActivityPub\Inbox\ActivityMessage","retryCount" => 3,"delay" => 4800000,"error" => """  Handling "App\Message\ActivityPub\Inbox\ActivityMessage" failed: An exception occurred while executing a query: SQLSTATE[23505]: Unique violation: 7 ERROR:  duplicate key value violates unique constraint "user_ap_id_idx"\n  DETAIL:  Key (ap_id)=(AlexDafonte@red.niboe.info) already exists.  ""","exception" => Symfony\Component\Messenger\Exception\HandlerFailedException { …}]

PR #870, opened 2024-06-28
findActorOrCreate was changed to search by apPublicUrl which is wrong, revert to apProfileId

BentiGorlich:
Change was made by me in #853


BentiGorlich:
Should I revert the migration adding a unique index over `apPublicUrl`?

nobodyatroot (approved):

PR #877, opened 2024-06-30
Cleanup deprecation warnings in Symfony profiler pre 1.6.1 release

nobodyatroot:
**Before:**

several (aka all...) generic/custom SSO methods are missing an explicit return type

image

and liip twig mode config

image

**After:**

image

The few remaining are from packages we use and need further investigation, but it's good practice to knock these down before any upgrades to the Symfony version.

### DIFF
--- a/config/packages/liip_imagine.yaml
+++ b/config/packages/liip_imagine.yaml
@@ -19,7 +19,7 @@ liip_imagine:
     data_loader: kbin.liip_loader
     default_image: null
     twig:
-        mode: legacy
+        mode: lazy
     default_filter_set_settings:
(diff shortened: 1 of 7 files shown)


melroy89 (approved):
Great work!

melroy89:
Ps. This might cause conflicts with people who have changed the `config/packages/liip_imagine.yaml` manually during production. Eg. use a different redirect code: `redirect_response_code: 301`

TheVillageGuy:
It's not causing any problems for me, I am curious though what the difference between legacy and lazy mode is?

PR #879, opened 2024-06-30
Add send direct message button to profile

melroy89:
Goal: Make the (now working) direct message feature better visible for the end-user, instead of looking in the info menu.

- Add an additional button to the profile, calling it "Direct message" since that covers the meaning much better
  - Button comes with an envelope icon and links to the same `<user>/message` path as the existing send message link.
- Rephrase existing "Send message" in _info.html.twig to "Send direct message" to keep it aligned. 

Example:

image

Example mobile:

image

### DIFF
--- a/translations/messages.en.yaml
+++ b/translations/messages.en.yaml
@@ -87,8 +87,7 @@ from: from
 username: Username
 email: Email
 repeat_password: Repeat password
-agree_terms:
-  Consent to %terms_link_start%Terms and Conditions%terms_link_end% and
+agree_terms: Consent to %terms_link_start%Terms and Conditions%terms_link_end% and
   %policy_link_start%Privacy Policy%policy_link_end%
 terms: Terms of service
 privacy_policy: Privacy policy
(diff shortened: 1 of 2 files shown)


translations/messages.en.yaml
```diff
@@ -292,7 +292,7 @@ comment: Comment
 post: Post
 ban_expired: Ban expired
 purge: Purge
-send_message: Send message
+send_message: Send private message
```
e-five256, line 295: I don't believe we should use the word private, as other software doesn't for the reason that it implies the messages are encrypted or secure, instead the word used is "direct message"

nobodyatroot: agree with @e-five256, there's nothing private about AP at all, lol.

e-five256:
It definitely uses a lot more room on mobile

image

It might be worth making the `<span>` with the text have `<span class="hide-on-mobile">`. That shows the text on desktop but hides it on smaller devices

image

translations/messages.en.yaml
melroy89: I see. Let's call them direct message instead.

melroy89: Fixed.

melroy89:
> It might be worth making the `<span>` with the text have `<span class="hide-on-mobile">`.

Agreed. Also done.

nobodyatroot (approved):

PR #882, opened 2024-07-02
Remove text from DM button, stops excessive text wrap on desktop view with three columns

nobodyatroot:
PR removes the text from the DM button, stops the smooshed text wrapping on desktop view with 3 column view enabled:

Before:
image

image

After:
image

image

The envelope icon should be enough to get the point across that this is to message the user... also consider we do the same thing already with the mobile view `<span class="hide-on-mobile">`. I've retained the hover over text fwiw, not that it matters much on mobile anyway.... but at least it will show the translation on desktop

### DIFF
--- a/templates/components/user_actions.html.twig
+++ b/templates/components/user_actions.html.twig
@@ -38,7 +38,7 @@
         </form>
         <a href="{{ path('messages_create', {username: user.username}) }}" title="{{ 'direct_message'|trans }}" aria-label="{{ 'direct_message'|trans }}">
             <button class="btn btn__secondary">
-                <i class="fa-solid fa-envelope" aria-hidden="true"></i> <span class="hide-on-mobile">{{ 'direct_message'|trans }}</span>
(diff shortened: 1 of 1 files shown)


BentiGorlich:
Not to add it in this PR, but what do you think about removing the text of the follow and block button as well? If we add a `title` attribute there it will still show the tooltip on hover. Though I don't know how we could make that work on mobile...

BentiGorlich (approved):

nobodyatroot:
> Not to add it in this PR, but what do you think about removing the text of the follow and block button as well? If we add a `title` attribute there it will still show the tooltip on hover. Though I don't know how we could make that work on mobile... 

@e-five256 and I talked about that and we agreed not having text next to those other buttons may not be intuitive enough on mobile without the tooltip vs the prolific envelope icon meaning"message". A better implementation might be adding a new row so the buttons have more room... but I'll let someone else with better UI skills tackle that, lol.

PR #891, opened 2024-07-10
Modify all SSO providers to redirect to the profile settings

### LINKED ISSUE #890: After signing in with SSO, redirect to user profile
When signing in using an SSO, e.g. google, discord, if the username already exists it is made by adding a random number. This can only be changed until, for instance, an upvote is made.

Google based usernames are always made based on google name + random number

Instead of bringing the user to https://my.mbin it should bring them to https://my.mbin/settings/profile so they can change their username (maybe clarify that this can only be done until they interact)

BentiGorlich:
Make all SSO providers redirect to the profile settings on successful authentication so a user can change the username if they want. When a user already exists a random number is appended to the username, for a user to be able to change their name they must not have interacted with anything and that is unlikely if we redirect them to the home page

Closes #890

### DIFF
--- a/src/Security/AuthentikAuthenticator.php
+++ b/src/Security/AuthentikAuthenticator.php
@@ -156,7 +156,7 @@ class AuthentikAuthenticator extends OAuth2Authenticator
 
     public function onAuthenticationSuccess(Request $request, TokenInterface $token, string $firewallName): ?Response
     {
-        $targetUrl = $this->router->generate('front');
+        $targetUrl = $this->router->generate('user_settings_profile');
 
         return new RedirectResponse($targetUrl);
     }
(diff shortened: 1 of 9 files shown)


nobodyatroot:
We used to have it configured this way, but @CocoPoops had a good point that even after the initial user profile is setup for an SSO user, they continue to get redirected to the profile settings page on each new login, which is not the behavior of a non-SSO enabled user. Maybe we can make it more intelligent so that after the first login it sends them to the `front` route?

BentiGorlich:
Ah that makes perfect sense... :thinking: 
I am going to make this a draft and we have to find a way to only redirect to the profile page on the initial login after registering

nobodyatroot:
closing in favor of https://github.com/MbinOrg/mbin/pull/909

PR #893, opened 2024-07-10
When receiving an "Update" activity check permissions

BentiGorlich:
Right now we do not check any permissions when receiving an "Update" (=edit) activity.
That means at the moment any user can send an "Update" activity to any post and mbin would just apply that update, regardless of whether the user should be able to do so as long as the HTTP headers are valid.

This PR implements a permission like this: a user should be able to modify a post if:
- they are from the same instance as the author (global mods and admins, and of course the author themselves)
- or they are from the same instance as the magazine (magazine mods and admins)
- or they are a moderator in the magazine the user posted to


BentiGorlich:
This is a draft right now, because I did not live-test this yet, only wrote the code. That looks good to me right now, but I still gotta test it

BentiGorlich:
Everything works as expected

nobodyatroot (approved):

PR #909, opened 2024-07-13
Redirect to user profile settings on newly created SSO users

### LINKED ISSUE #890: After signing in with SSO, redirect to user profile

CocoPoops:
Change all SSO providers to redirect to profile settings when the user is first created so that a user can change their username if their username already exists 

Closes #890


melroy89:
Maybe only redirect to profile page when their username already exists?

CocoPoops:
> Maybe only redirect to profile page when their username already exists?

i wasn't sure which way you guys would prefer as some of the authenticators add the random numbers and some check for existing users first before adding the random numbers
i assume their is a reason that some of them are that way 🤷

nobodyatroot:
> > Maybe only redirect to profile page when their username already exists?
> 
> i wasn't sure which way you guys would prefer as some of the authenticators add the random numbers and some check for existing users first before adding the random numbers i assume their is a reason that some of them are that way 🤷

the whole thing is super inconsistent, but this seems to be a reasonable way to solve the issue at hand.

nobodyatroot (approved):
tested in dev, seems to work as expected. code changes seem reasonable.

PR #910, opened 2024-07-13
Move entry comments options below the add comment box

CocoPoops:
this is a UI nit pick but it has always been weird to me that the Comments Sorting Options are so far away from the comments when logged in as the comment box isnt there when not logged in 

before
image

after 
image


melroy89 (approved):

melroy89:
Be sure to do the same on all pages with comment sections. Like posts, threads,..

Just saying.

CocoPoops:
> Be sure to do the same on all pages with comment sections. Like posts, threads,..
> 
> Just saying.

as far as i can see that was the only one that was that way as the everything else has the reply button or was already like that

unless im missing something

Microblog page 
image

Post page
image

PR #911, opened 2024-07-15
Remove link

melroy89:
/kbin is not actively developed anymore, no need to link back to kbin.


BentiGorlich:
Even if it was not actively developed anymore (I have no idea what the status is) we still share a lot of code and it is still the source of our fork. I don't think it would be a good idea to scratch the link. 
Personally I don't care that much, but I think this would be an unnecessary source of conflict

nobodyatroot:
???

i can't say i agree having the resulting drama from this ending up in our github or matrix channels...

PR #923, opened 2024-07-18
Redirect to existing thread

### LINKED ISSUE #919: Improve direct messaging from profile UI with users you have messaged before
Mbin has a very nice UI for direct messages.
IMG_5760

When I go to the profile of that user I was messaging and click the mail/message icon,
IMG_5762
I end up with this UI that does not display any of our old messages. It's as if we never messaged each other.
IMG_5761 I expected to just be put on the screen that shows the messages between that user and myself.

BentiGorlich:
If you click the "Send direct message" for someone you already have a message thread with it now redirects you to that thread instead of displaying a form which creates a new thread

Closes #919


melroy89 (approved):
Yep.. This is also what you & I discovered right? So great fix!

PR #930, opened 2024-07-19
Bump to newer version that also fixes npm install

melroy89:
Dump Nodejs version to 22.5.1, which also still should work (and contain the fix)...


melroy89:
This "should" still work :)

nobodyatroot (approved):

PR #936, opened 2024-07-25
Wrap all messenger invokes in their own transaction

BentiGorlich:
This PR aims to improve the scaling of the messengers by avoiding database dead locks

- move all previous invoke code in a `doWork` method and use the `__invoke` methods to wrap the `doWork` in a transaction using a new abstract class
- do not catch all exceptions in the `ChainActivityHandler`, but only those we want (did catch a deadlock exception as well which should not be caught)

### DIFF
--- a/src/MessageHandler/MbinMessageHandler.php
+++ b/src/MessageHandler/MbinMessageHandler.php
@@ -1,0 +1,35 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\MessageHandler;
+
+use App\Message\Contracts\MessageInterface;
+use Doctrine\DBAL\Exception;
+use Doctrine\ORM\EntityManagerInterface;
+
+abstract class MbinMessageHandler
+{
+    public function __construct(
+        private readonly EntityManagerInterface $entityManager,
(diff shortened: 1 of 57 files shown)


BentiGorlich:
nope this didn't work, still deadlocks....

src/MessageHandler/MbinMessageHandler.php
```diff
@@ -0,0 +1,35 @@
...
+     * @throws Exception
+     */
+    public function workWrapper(MessageInterface $message): void
+    {
+        $conn = $this->entityManager->getConnection();
+        if (!$conn->isConnected()) {
+            $conn->connect();
+        }
+
+        $conn->transactional(fn () => $this->doWork($message));
```
melroy89, line 29: Ah it was a bit hidden away in 58 files changed, but this is the wrapper method that is now calling `transactional`, which allows you to take over the transaction demarcation.

melroy89: This also means that no flush needs to be called anymore in the `doWork`.

PR #939, opened 2024-07-25
Add new instance table

BentiGorlich:
Add a new table with which we can track the software, version, last successful deliver and receive timestamp. 

The federation page now looks something like this:
grafik

| user panel | magazine panel |
| ---------- | -------------- |
| grafik | grafik |

The table will be populated with receives, nothing mandates an instance entry to be there, so there is not guarantee that there is an entry in the instance table for each user or magazine

There are 2 new API endpoints:
- `/api/federated` returns all instances we (knowingly) federate with
- `/api/dead` all instances considered dead by the instance

The `/api/defederated` endpoint will now return a list of `InstanceDto` like the other 2

Additionally users and magazines retrieved by the API now contain a `serverSoftware` and `serverSoftwareVersion` field


jwr1:
Will this be in the API also?

BentiGorlich:
Not yet... I literally always forget about the API 😅
I'll have a look at it tomorrow. What do you think would make sense? Just adding the general table to be queried? Or do you want it added to get the info for users (what server software they are using)? Or both?

jwr1:
I was thinking both. Like a new key (e.g. `instance`) in the user object with information about the user's instance (you could also add this for magazines if you wanted). And then probably something like a `/api/intances` route to get the whole list of instances.

BentiGorlich:
@jwr1 I added additional endpoints, see the updated PR info

jwr1:
Looks great to me! Since the change to the defederated endpoint is backward incompatible, it'll break current uses of it though; joinmbin.org, for example, lists all 10 of gehirneimer.de's defederated servers as `[object Object]`. It's not that big of a deal, since I could easily fix it on the website, but it's something to note.

BentiGorlich:
Yeah, I could make it just a v2 then... :thinking: 
I just thought that nobody is using it anyways

jwr1:
> Yeah, I could make it just a v2 then... 🤔 I just thought that nobody is using it anyways

I'd say it's fine as is. I just think it's something we should mention in the release notes.

BentiGorlich:
Should be fixed now. Actually a great call @jwr1 I created a bunch of problems with editing the banned instances which I created by modifying the InstanceDto

nobodyatroot (dismissed):

jwr1:
Wait, so does defederated v2 have an api route? I wasn't able to see it in the API docs.

BentiGorlich:
I forgot to push it :D

jwr1 (dismissed):

nobodyatroot (dismissed):

BentiGorlich:
I think there are problems with instances that only supply a 2.0 endpoint

melroy89:
I see a very very limited list to be honest. The previous list was more complete?

image


https://kbin.melroy.org/federation

melroy89:
Ow.. It seems to be slowly adding more and more...

image

BentiGorlich:
Yes this list is built with incoming activities

nobodyatroot (approved):

melroy89 (approved):

PR #950, opened 2024-07-29
Remove federated users view from admin panel users page

nobodyatroot:
I don't think I've ever used or wanted to use the federated user view in the users page in the admin panel. In kbin.run's case, it generates over 7000 pages of federated users when selected... This PR removes it.

Before:
image

After:
image

### DIFF
--- a/config/kbin_routes/admin.yaml
+++ b/config/kbin_routes/admin.yaml
@@ -1,7 +1,6 @@
 admin_users:
     controller: App\Controller\Admin\AdminUserController
-    defaults: { withFederated: false }
-    path: /admin/users/{withFederated}
+    path: /admin/users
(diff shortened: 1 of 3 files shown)


BentiGorlich:
I mean this effectively removes a feature doesn't it? Even if the feature is not used often why remove it?

nobodyatroot:
> I mean this effectively removes a feature doesn't it? Even if the feature is not used often why remove it?

I'd think the admin would want to focus on the local users of the instance, but I could be wrong I suppose. Simplifies the UI by not showing unnecessary information (and prevents accidently selecting it with a massive query in return...). Maybe just my personal preference and that's fine if others don't agree, we can close it if so.

PR #951, opened 2024-07-30
Implement "posting restricted to mods"

### LINKED ISSUE #528: Magazine setting, so only mods can create threads and microblog posts
There are a few use cases where you want to restrict the creation of threads and microblogs (**not** comments on them) to a handful of people

For such a use case there should be options in the magazine settings for: 
- restrict creation of threads to mods
- ~~and the same for microblogs~~ thought about it and I think it is better to not restrict microblog creation

**Additional context**
I want only the moderators (i.e. us devs of mbin) to be allowed to create threads in https://gehirneimer.de/m/mbinReleases

BentiGorlich:
- magazines have a new option to restrict the creation of entries to mods. This does not affect comments or microblogs
- if the posting in a magazine is restricted exceptions are thrown which are caught by the messengers and logged and show a flash message in the UI
- magazines which restrict posting have a lock icon next to their name

Closes #528

### DIFF
--- a/migrations/Version20240729174207.php
+++ b/migrations/Version20240729174207.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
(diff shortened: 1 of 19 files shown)


nobodyatroot (approved):
LGTM, tested in dev and prod

PR #954, opened 2024-07-30
Send out update activities on actor updates

BentiGorlich:
when a magazine is edited there are 2 cases:

  - remote magazine, but the editing user is a mod in that magazine: send an update activity to the source instance
  - local magazine: announce the update to all subscribers

when a user is edited send an update activity to all their followers

receive:
- user: dispatch an update actor message, completely ignoring the content of the payload
- magazine: do a bit manually, as we need a dto for the changes, so we can use the `MagazineManager::edit`. That then takes care of announcing the update if the magazine is a local one


nobodyatroot (approved):
giving a test now, but i dont expect any issues

Issue #861, opened 2024-06-24
Edit image alt text


aaronliu0130:
**Is your feature request related to a problem? Please describe.**
Currently, if I forget to add alt-text to an attached image after spending an hour mongering for äppropriate" tags, I have to delete the post and make a new one if I want to add alt-text to the photo upload.

**Describe the solution you'd like**
Allow one to edit the alt text of a photo upload after posting.

**Describe alternatives you've considered**
Gib us access to all ye mainaframas

Issue #463, opened 2024-01-29
Can't edit URLs in posted links with "edit"


andrewmoise:
**Describe the bug**

* Create a link post
* Edit the link post
* "URL" field is uneditable, and stays that way with a little spinner next to it

**On which Mbin instance did you find the bug?**

mbin.grits.dev
Version is git as of 2024-01-29

**Expected behavior**

Seems like starting out with the field editable, *without* dispatching the thing that refetches the title+description, and then having that thing dispatch as normal if you edit the field, would be better.

**Additional context**

So, I hacked it into working for me -- I feel like I must be missing something maybe about the way this is meant to work, but what I did was:

```
diff --git a/src/Form/EventListener/DisableFieldsOnEntryEdit.php b/src/Form/EventListener/DisableFieldsOnEntryEdit.php
index a5c8e212..48c85c3e 100644
--- a/src/Form/EventListener/DisableFieldsOnEntryEdit.php
+++ b/src/Form/EventListener/DisableFieldsOnEntryEdit.php
@@ -38,7 +38,7 @@ final class DisableFieldsOnEntryEdit implements EventSubscriberInterface
         if ($form->has('url')) {
             $field = $form->get('url');
             $attrs = $field->getConfig()->getOptions();
-            $attrs['disabled'] = 'disabled';
+            //$attrs['disabled'] = 'disabled';
 
             $form->remove($field->getName());
             $form->add(
diff --git a/templates/entry/edit_link.html.twig b/templates/entry/edit_link.html.twig
index c8914d0d..9bb0cb91 100644
--- a/templates/entry/edit_link.html.twig
+++ b/templates/entry/edit_link.html.twig
@@ -21,7 +21,7 @@
     <h1 hidden>{{ 'edit_link'|trans }}</h1>
     {% include 'layout/_flash.html.twig' %}
     <div id="content" class="section">
-        <div class="container">
+        <div class="container" data-controller="entry-link-create">
             {% include 'entry/_form_link.html.twig' with {edit: true} %}
         </div>
     </div>
```

The controller seemed like it wasn't hooked up, leaving the thing to spin forever without attempting to refetch anything, and the URL field started disabled, which meant even when configured there was no way for it to ever activate. I'm not claiming this is a fix (e.g. maybe changing the link is disabled on purpose for some particular reason); just including it for informational purposes.

PR #971, opened 2024-08-01
When editing an entry only use one type

### LINKED ISSUE #784: Can't edit Body field of posts with images
**Describe the bug**
No field with which to edit Body text on on posts with images

[kbin.melroy.org]

[1.5.3]

**To Reproduce**
I'll be honest, I did not check all potential ways for this error to arise. I'd imagine this occurs on any kind of post with a photo, no matter which `Add new` button was used to make it happen, but I have no proof. The reproduction steps I provide below are the ones I do have proof for, or how I made the error arise myself. 

1. Click Add new link
2. Paste a link to an image in the URL field
3. Post your post
4. Try to edit your post

Or,
1. Go to an old post you made with an image
2. Try to edit your post

In both cases, there is no field for you to edit your Body text. Even if you have Body text in your post.

**Expected behavior**
There is a visible field for me to edit Body text.

 - Device: [iPhone 13 mini]
 - OS: [iOS 17.1.2]
 - Browser [stock browser, Safari]
 - Browser Version [17.1]

BentiGorlich:
- when editing an entry only use one type for the form in which all fields are accessible
- make the URL field editable

Closes #784, #861 and #463


BentiGorlich:
Still some JS errors

nobodyatroot:
the only thing that i can think of that may still be a source of confusion is on photo entry creation, there's no option to add body text:

image

although now maybe having these multiple options is redudant since you can still upload a photo in a normal thread entry:

image

BentiGorlich:
> although now maybe having these multiple options is redudant since you can still upload a photo in a normal thread entry:

That has been the case for a while now... I was hesitant to change the way entries are created

> the only thing that i can think of that may still be a source of confusion is on photo entry creation, there's no option to add body text

That is true for the URL field as well... I think consolidating the create options should be done as well, but in a different PR

nobodyatroot (approved):

nobodyatroot:
Yep, no need to explode this PR.

TheVillageGuy:
> maybe just remove the photo entry option???

NOOO! Please don't

MHLoppy:
> > maybe just remove the photo entry option???
> 
> NOOO! Please don't

This is late, but for what it's worth I think here we're just talking about removing the option in the submission selection type UI, not the functionality itself of actually uploading an image as part of a submission (which is of course pretty important to your use case with rimworld.gallery). You would just select one of the other two options and make a submission as usual!

TheVillageGuy:
> This is late, but for what it's worth I think here we're just talking about removing the option in the submission selection type UI, not the functionality itself of actually uploading an image as part of a submission (which is of course pretty important to your use case with rimworld.gallery). You would just select one of the other two options and make a submission as usual!

As long as it's possible to set a magazine to images only, I don't want people uploading text threads or links on /m/RimWorldPorn it will be a moderation nightmare

MHLoppy:
Per-magazine submission restrictions sounds like a good feature request!

TheVillageGuy:
It can't be enforced on other instances yet though, although that could be solved by some sort of auto moderation on the home server of the magazine.

PR #983, opened 2024-08-03
Add joinmbin.org list of instance

melroy89:
See title.

### DIFF
--- a/README.md
+++ b/README.md
@@ -38,8 +38,9 @@ For developers:
 
 ## Instances
 
-- [List of instances](https://fedidb.org/software/mbin)
-- [Alternative listing of instances](https://mbin.fediverse.observer/list)
+- [List of instances](https://joinmbin.org/servers)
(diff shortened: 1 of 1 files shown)


nobodyatroot (approved):

melroy89:
Ps. is there a reason why https://fedidb.org/software/mbin is showing 31 servers and your joinmbin site "only" shows 24 servers?

jwr1:
> Ps. is there a reason why [fedidb.org/software/mbin](https://fedidb.org/software/mbin) is showing 31 servers and your joinmbin site "only" shows 24 servers?

Quite a few of the servers on fedidb are either dead or inaccessible. For example, both mbin.homelab-india.com and wiku.hu are listed on fedidb, but if you actually try to visit them, you can see there's nothing there. To count as accessible, the `/nodeinfo/2.1.json` route needs to have a `software.name` matching `mbin`.

Issue #797, opened 2024-05-28
Allow adjustment of MAX_IMAGE_BYTES


jwr1:
At the moment, the max image size is `6MB` , as defined in this line:

https://github.com/MbinOrg/mbin/blob/f2200627363f8a8b4a4fc29dabcf7bfef7f96857/src/Service/ImageManager.php#L25

It would be nice if there were a way to configure this, either through an environment variable or an admin panel option (whichever makes more sense to y'all). This should affect both incoming federated images and local user uploads. This feature could be useful for servers that would rather have a lower limit (like 1MB, 2MB, or 0.5MB) that are trying to conserve resources, or even servers that have plenty of resources that would rather allow larger image sizes (like 10MB or 15MB).

jwr1:
I still think this option would be useful.

PR #987, opened 2024-08-04
Replace MAX_IMAGE_BYTES constant in `ImageManager.php` with admin configurable environmental variable

nobodyatroot:
Closes https://github.com/MbinOrg/mbin/issues/797

Retains "sane" default of 6000000 if admins forget to set the variable in the `.env` file.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,10 @@ KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 MBIN_DEFAULT_THEME=default
 
+# Max image filesize (in bytes)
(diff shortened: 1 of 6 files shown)


BentiGorlich (commented):
Do we need a fallback a lá `?? $defaultMaxBytes`? Or is symfony always at least returning the `max_image_bytes_default` value?

melroy89 (approved):
`max_image_bytes` is set to `max_image_bytes_default` (default), so I expect no issues there.

melroy89:
This variable was wrongly introduced, and should have been called: `MBIN_MAX_IMAGE_BYTES`  instead..

Fixed: https://github.com/MbinOrg/mbin/pull/1878

PR #995, opened 2024-08-05
Add user ban and suspension lists to admin panel

nobodyatroot:
Add user inactive, suspension, and ban lists for easy admin review under the admin panel. Removes button indicators, I don't think the 2FA indicator adds any value, so I did not make a page/filter for it.

image


templates/admin/users_inactive.html.twig
```diff
@@ -0,0 +1,77 @@
...
+                </a>
+            </li>
+            <li>
+                <a href="{{ path('admin_users_banned') }}"
+                   class="{{ html_classes({'active': is_route_name('admin_users_banned')}) }}">
+                    {{ 'admin_users_banned'|trans }}
+                </a>
+            </li>
+        </menu>
+    </div>
```
BentiGorlich, line 45: you could put this in its own file and just include it

config/kbin_routes/admin.yaml
```diff
@@ -1,7 +1,25 @@
...
+admin_users_suspended:
+    controller: App\Controller\Admin\AdminUsersSuspendedController
+    defaults: { withFederated: false }
+    path: /admin/users/suspended/{withFederated}
+    methods: [GET]
+
+admin_users_banned:
+    controller: App\Controller\Admin\AdminUsersBannedController
+    defaults: { withFederated: false }
+    path: /admin/users/banned/{withFederated}
```
BentiGorlich, line 22: You could put all Controllers in a single file and just different methods. You can reference them via `App\Controller\Admin\AdminUsersBannedController::methodName`

templates/admin/users_banned.html.twig
```diff
@@ -0,0 +1,85 @@
...
+                        <td>{{ component('user_inline', {user: user}) }}</td>
+                        <td>{{ user.apId ? '-' : user.email }}</td>
+                        <td>{{ component('date', {date: user.createdAt}) }}</td>
+                        <td>{{ component('date', {date: user.lastActive}) }}</td>
+                    </tr>
+                {% endif %}
+            {% endfor %}
+            </tbody>
+        </table>
+    </div>
```
BentiGorlich, line 81: If this is the same in all 4 files you can put it in its own file as well

BentiGorlich (commented):
Just a few remarks. Gotta put it on my server

BentiGorlich:
I actually think that it would be better if this was a dropdown value which filters stuff, more like the entry filters work. That could be expanded in the future with sort options, more information, ip-address filter or something like that

BentiGorlich:
Its great either way though (already deleted like 12 users who never verified their account :D)

config/kbin_routes/admin.yaml
nobodyatroot: good suggestion!

nobodyatroot:
> I actually think that it would be better if this was a dropdown value which filters stuff, more like the entry filters work. That could be expanded in the future with sort options, more information, ip-address filter or something like that

i thought about that initially, its the crazy route names that result from multiple dropdowns that made me not do that... but its a fair point

templates/admin/users_inactive.html.twig
nobodyatroot: makes sense!

templates/admin/users_banned.html.twig
nobodyatroot: yep, also makes sense!

jwr1:
> > I actually think that it would be better if this was a dropdown value which filters stuff, more like the entry filters work. That could be expanded in the future with sort options, more information, ip-address filter or something like that
> 
> i thought about that initially, its the crazy route names that result from multiple dropdowns that made me not do that... but its a fair point

Couldn't you just use query params for that?

nobodyatroot:
i added some slight adjustments to not show the table header if there are no users matching the conditions:

image

BentiGorlich (approved):

PR #1008, opened 2024-08-08
Invalidate collection cache before getting collection for likes, dislikes and shares

BentiGorlich:
I observed that PeerTube videos do not have the correct amount of like. The first 24 hours it just stays at 0. This fixes it.

I think this is the best fix, since we only invalidate the collection cache when we received an `Update` activity indicating that the collections are not up to date anymore.


BentiGorlich:
An example can be viewed here: https://gehirneimer.de/m/thelinuxexperiment_channel@tilvids.com/t/286286/COSMIC-Desktop-ALPHA-a-unique-and-important-project-for-Linux

nobodyatroot (approved):

Issue #833, opened 2024-06-17
Add local user IP ban/visibility for instance admins


nobodyatroot:
**Is your feature request related to a problem? Please describe.**
It is time consuming for admins to scrape nginx/web logs to find the IP(s) of problematic local users, it would be nice to be able to see from the admin panel (or admin view of a user) the user's IP so that it can be banned either at the firewall or application level.

**Describe the solution you'd like**
An additional user field to see the user's IP address when an admin is viewing the user's profile (local users only)

**Describe alternatives you've considered**
An IP ban feature within Mbin so that non-privileged users (local global moderators) can also ban users without exposing the user's IP address

**Additional context**
Spam accounts bad

PR #1009, opened 2024-08-09
Allow admins to view user IP addresses in the admin panel users table

nobodyatroot:
Introduces admin ability to view local users IP addresses as resolved from `src/Service/IpResolver.php`:

image

A user's IP address is updated on:

- [ ] successful authentication
- [ ] entry creation
- [ ] entry edit
- [ ] entry comment creation
- [ ] entry comment edit
- [ ] post creation
- [ ] post edit
- [ ] post comment creation
- [ ] post comment edit

I also noticed the GitHub authenticator was missing the IP being set in the user dto which will cause the registration limiter to be ineffective, so I added it in to be consistent with the other authentication mechanisms.

Closes: https://github.com/MbinOrg/mbin/issues/833

### DIFF
--- a/migrations/Version20240808230919.php
+++ b/migrations/Version20240808230919.php
@@ -1,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+/**
+ * Auto-generated Migration: Please modify to your needs!
+ */
+final class Version20240808230919 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add user IP column';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE "user" ADD ip VARCHAR(255) DEFAULT NULL');
+    }
+
+    public function down(Schema $schema): void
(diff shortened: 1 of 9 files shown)


BentiGorlich:
I thought that we fixed the trusted proxy problem a long time ago :O

nobodyatroot:
> I thought that we fixed the trusted proxy problem a long time ago :O

not sure how since it was never configured, lol

BentiGorlich (changes requested):
I have a few things that I would like to see changed:
1. I would prefer if you'd move the trusted proxy stuff in its own PR
2. Since we are already changing the db schema to have the IP-addresses of users we should create a new table for IP addresses and one that makes the references between users and IPs. That would have a few benefits:
    - accounts for multiple devices 
    - The most important IP imo is the one used to register an account. At least that is the one I would be interested in for spam detection. 
    - since this is personal information I think we should create a scheduled job to look for IP addresses that have been used over a year ago and deleting them
    - we can use the same table for IP-blocks in the future
    - easier to lookup all users with a specific IP

I think this would be a really good addition for admins to prevent spam right upon account creation

melroy89:
Should we follow-up this PR?

melroy89:
Rebased..

melroy89:
Fixed PHP cs fixer.

Issue #780, opened 2024-05-14
can't view uploaded image in thread


Shitsurei-shimasu:
**Describe the bug**
Started a new thread at lemmyshitpost@lemmy.world.
Uploaded image thru mbin but after posting I can't view it thru mbin. I click on the icon and it just reloads the comments. I go into the sub menu and click open original URL but still doesn't load. This is on the mobile web page. It seems this happens sometimes, other times I can click thru and view image. I usually go straight to the instance to post because of this issue. Does not happen when I post at lemmy.world and view thru mbin.

**On which Mbin instance did you find the bug?**
Kbin.run

**Which Mbin version was running on the instance?**
[e.g. 1.4.0]
1.5.3

**To Reproduce**
1. Go to https://kbin.run/m/lemmyshitpost@lemmy.world/t/420733/Congrats-to-all-2024-college-graduates
2. Click on image icon
3. Image won't appear in mobile. Desktop you can click on the thumbnail to view.

**Expected behavior**
Image should load if you click on the title or image icon

**Screenshots**

**Smartphone (please complete the following information):**
 - Device: pixel 7
 - latest android
 - Browser firefox
 - Browser Version 125.3.0
![Screenshot_20240514-163818](https://github.com/MbinOrg/mbin/assets/21995163/d592529f-c38c-44ec-bfd0-83e0fcc744f2)

nobodyatroot:
@Shitsurei-shimasu i'm almost 100% sure, but can you confirm this entry was made with the "add new thread" function?

![image](https://github.com/MbinOrg/mbin/assets/35878315/a27ee6a6-18bb-4d26-88fa-001f718e399e)

Shitsurei-shimasu:
Yes I can. It is the way I post every time. I click on the add attachment
button and upload the image. This is not the first time it has happened.
Kbin has the same problem.

nobodyatroot:
@Shitsurei-shimasu i updated my original post with some ideas, but can you check your settings menu (hamburger menu -> gear icon) to see if you have "show thumbnails" setting enabled? a photo attached to a thread does not behave the same as submitting a photo entry... but i see there's no text body available to input for a photo entry for whatever reason...

Shitsurei-shimasu:
Ok that does work. I had it disabled because on mobile I prefer it to show
the full size when I scroll. I will just do the add new photo option from
now on if that works as well.

Shitsurei-shimasu:
[...]

PR #1011, opened 2024-08-09
Remove `\intval` around `MAX_IMAGE_BYTES` in `src/Service/SettingsManager.php`, breaks settings name resolution

nobodyatroot:
Image uploads are currently broken, introduced by https://github.com/MbinOrg/mbin/pull/997. PR removes it, settings can still be saved in settings page of admin panel.

image

Mentioned in https://github.com/MbinOrg/mbin/issues/780#issuecomment-2278457578 and confirmed in dev.


BentiGorlich:
Why is intval breaking it though?

BentiGorlich:
I guess intval never returns null and therefore the default is not kicking in?

BentiGorlich (approved):

nobodyatroot:
> I guess intval never returns null and therefore the default is not kicking in?

must be, Xdebug was showing 0...

BentiGorlich:
I mean it is obviously not necessary if you don't get a type error...

PR #1012, opened 2024-08-10
Add trusted proxy config to framework configuration

nobodyatroot:
The framework was not being configured with `TRUSTED_PROXIES`, I've added this configuration and `.env` variable to properly get the users actual IP as passed in the `x-forwarded-for` header, this should also resolve any rate limiter issues we could see if running behind a reverse proxy as `$request->getClientIp()` will return the `x-forwarded-for` IP if configured correctly.
https://symfony.com/doc/current/deployment/proxies.html#solution-settrustedproxies

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -32,6 +32,11 @@ KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 MBIN_DEFAULT_THEME=default
 
+# If you are running Mbin behind a reverse proxy, uncomment the line below and adjust the proxy address/range below
+# to your server's IP address if it does not already fall within the private IP spaces specified.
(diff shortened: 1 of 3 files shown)


BentiGorlich (commented):
Should the docker config have a blank `TRUSTED_PROXIES`? Or do we need a default here? The `www` container is forwarding requests to the `php` container isn't it?

nobodyatroot:
> Should the docker config have a blank `TRUSTED_PROXIES`? Or do we need a default here? The `www` container is forwarding requests to the `php` container isn't it?

Good point, we just need to swap the lines commented for docker. The container ips should fall within the private ranges I included.

BentiGorlich (approved):

PR #1013, opened 2024-08-11
Fix user soft delete and user account restore within 30 days

ghost:
User soft deletes (restorability within 30 days) were not being set with the right visibility property (aka not calling the `softDelete` method) and therefore not being displayed in the admin panel user deletion page:

After fix:
image

Users also had no way to restore their account within 30 days without contacting an admin, this PR now allows them to restore their account within 30 days by logging back into the site within the deletion window.

Added check to prevent federated users from login, should never be allowed... stops "lucky" guesses on random password generated.

I also tweaked the soft delete functions between user and magazine to be similar, but they're not exactly the same since magazine deletion is a hot mess and needs the same treatment as what @BentiGorlich did for user deletes.


src/Entity/Magazine.php
```diff
@@ -367,7 +367,7 @@ public function unsubscribe(User $user): void
 
     public function softDelete(): void
     {
-        $this->markedForDeletionAt = new \DateTime();
+        $this->markedForDeletionAt = date_add(new \DateTime(), new \DateInterval('P30D'));
```
BentiGorlich, line 370: Functional it is doing the same, but personally I like this better
```suggestion
        $this->markedForDeletionAt = new \DateTime('now + 30days');
```
But I let you be the judge :)

EDIT: I just noticed that I did it the same way in my code :D

BentiGorlich (commented):
I am not a 100% sure that it is very intuitive that logging back in will restore the account :thinking: 

Code wise I have nothing to object to.

nobodyatroot:
> I am not a 100% sure that it is very intuitive that logging back in will restore the account 🤔
> 
> Code wise I have nothing to object to.

I did update the message to tell the user they can, but I've seen a lot of sites (facebook, linkedin, etc) have this as the method to reenable accounts that the user "soft" deleted within a certain timeframe.

BentiGorlich:
Then it is just the question whether @melroy89 @e-five256 and @asdfzdfj agree with it

melroy89:
> I did update the message to tell the user they can do this or contact the admin, but I've seen a lot of sites (facebook, linkedin, etc) have this as the method to reenable/reactivate accounts that the user "soft" deleted within a certain timeframe.

This is true indeed, if you soft delete a user. Logging in will reset the delete in most other service providers like FB, etc. So it will no longer be deleted, allowing the user to safe their account even if the admin didn't respond in time.

melroy89 (approved):

PR #1021, opened 2024-08-12
Update docker + docker compose for dev and prod

LoveIsGrief:
First of all, I apologize for the size of the PR :pray:, but I think it would've been difficult to break it down.

This is a PR to rework the docker setup for dev and prod. The highlights are below, but basically the setup was not working well for development (not clear how to set it up, which files were necessary to be where, which env vars were important, updating code locally didn't reflect in the docker environment, etc.).

Now, developers should be able to spin up a development environment with minimal effort with docker installed, see their changes show up on reloads, and even use a debugger (XDebug).

The changes in the dev setup led to changes in production setup as well, though they should be minimal. Most importantly `caddy` was pulled out of the image to break the tight coupling and allow any reverse proxy with fast CGI to sit in front of mbin.

# Highlights

## Features

- Create a dev env using docker usable by noobs like me
```sh
ln -s compose.dev.yml compose.override.yml
docker compose up
```
- Support XDebug call from docker on linux with host.docker.internal

<details><summary>Details</summary>
<p>

`host.docker.internal` doesn't exist on linux because it's not running docker in a VM. Add following block to `php` service.
```yaml
   extra_hosts:
      - "host.docker.internal:host-gateway"
 ```

</p>
</details> 

- Support for arbitrary reverse proxy
- Smaller docker image (only necessary stuff is copied into image)
- Documentation for dev docker setup

## Bug fixes

- [x] get rid of `FATAL:  role "root" does not exist` in `db` service created by its healthcheck

# TODO

- [x] update docker prod documentation
- [x] mention that resetting means `sudo rm -rf docker/storage`
  - [x] move `.gitignore` out of storage otherwise resetting still creates a commit by deleting `.gitignore`
- [x] `docker/storage` needs to be writable to `mbin` user!
  - [x] `sudo chmod 777 docker/storage/*`
# Abandoned tasks for later PRs

- Use whitelist approach in .dockerignore instead of blacklist
- `MESSENGER_TRANSPORT_DSN=doctrine://default` as documented doesn't work

### CODE REVIEW

File: .dockerignore [hunk 1]
```diff
@@ -14,7 +14,7 @@
 **/compose.*.yaml
 **/compose.*.yml
 **/compose.yaml
-**/compose.yml
+compose.yml
```
LoveIsGrief, line 17: compose ymls are now in the root

compose.prod.yml
```diff
@@ -66,28 +43,24 @@ services:
...
     networks:
       - mbin_external_network
-    command: bin/console messenger:consume scheduler_default old async outbox deliver inbox resolve receive failed --time-limit=3600
+    command: bin/console -vvvv messenger:consume --all --time-limit=3600
     healthcheck:
       test: ["CMD-SHELL", "ps aux | grep 'messenger[:]consume' || exit 1"]
     volumes:
-      - ./storage/media:/var/www/mbin/public/media
-      - ./storage/logs:/var/www/mbin/var/log
+      - ./:/var/www/mbin/
```
LoveIsGrief, line 58: remove this

CocoPoops:
just a thought but maybe we should use 1 compose file with profiles instead of multiple compose files for prod and dev as i think being able to keep the compose.override file separate for admins to make changes to the compose.yml instead of editing it directly while also being able to keep up with git changes 

https://docs.docker.com/compose/profiles/

LoveIsGrief:
@CocoPoops that doesn't seem like the right feature to use, unfortunately.

> Services can be assigned to one or more profiles; unassigned services start by default, while assigned ones only start when their profile is active

Additionally, if you look at `compose.dev.yml` and `compose.prod.yml`, you'll see that the services apply different overrides. For example in `compose.dev.yml` the `php` and `messenger` services mount the CWD under `/var/www/mbin` so that local changes will be visible upon reloading the page in the browser. That isn't necessary in production because the image will have the source **and** build files within the image.

We could use a [justfile](https://github.com/casey/just) to have a `dev` and `prod` target which execute `docker compose -f compose.yml -f compose.$env.yml`, but that's out of scope for this PR.

CocoPoops:
@LoveIsGrief hopefully this explains what i meant better

For example, It would allow us to create php-prod and php-dev services in one compose.yml, along with services, and then assign them to corresponding profiles or both if no changes are between prod and dev.

This would allow administrators to override settings for either production or development services independently through the compose.override.yml file.

This flexibility is particularly useful in scenarios like mine, where I use Traefik as a reverse proxy configured via Docker labels. By organizing the Compose files this way, it would be much easier to switch between production and development environments for debugging purposes without the risk of exposing sensitive debug secrets.

CocoPoops:
I think it would also be easier that way if for some reason we need to change the compose file down the line as we should be recommending people only change things via the compose.override file

LoveIsGrief:
Could you fork my branch and make an example? 

For your scenario, you could use `include` ([see doc](https://docs.docker.com/compose/multiple-compose-files/include/)). 

**compose.override.yml**
```yaml
include:
  - compose.prod.yml

services:
  traefik: ...
```

CocoPoops:
> For your scenario, you could use `include` ([see doc](https://docs.docker.com/compose/multiple-compose-files/include/)).
> 
> **compose.override.yml**
> 
> ```yaml
> include:
>   - compose.prod.yml
> 
> services:
>   traefik: ...
> ```

ill just do this as i just tested using the compose profiles and it doesnt really seem worth it as it kinda makes it a pain to maintain


if anyone wants do that include method for overriding
i had to do this way as you cant override included compose files from the same file they are included in
> **compose.override.yml**
> 
> ```yaml
> include:
>   - compose.prod.yml
>   - compose.prod.override.yml 

> **compose.prod.override.yml**
> 
> ```yaml
> services:
>   www:
>    labels:
>      - add docker labels
>   other-services: ...

CocoPoops:
also just tried switching my instance over to test the image is working and im getting 404's on some images but not others 

tried clearing php cache and flushing the redis db but that didnt help

is there anything i would have to change in my .env to get working?

Screenshot_20240313

LoveIsGrief:
Thank you for testing!
I think I know where the issue lies. 

When making the image compatible with other reverse proxies, I copy the public/ folder to a volume shared between the php service and the reverse proxy. Uploads probably happen in the php service and end up in the php container but not on the shared volume.

I think that's the issue and I'll try to find a solution to that this weekend.

Again, thanks for testing!

melroy89:
Press "Ready for review" (and remove "WIP" from the title) if you think we should review.

LoveIsGrief:
@melroy89 I finally have it in a state where the code should be reviewable. I don't know if you guys squash-merge or not. If not, once the reviews are in, I can rebase and rework the commits to get rid of the `wip: ...` and other commits.

@CocoPoops I think I fixed the issue you were facing with missing media. Please see the [migration guide](https://github.com/LoveIsGrief/mbin/blob/main/docs/02-admin/01-installation/docker.md#migration-guide) section in the installation/docker.md docs.

melroy89:
Yes, we normally squash merge the commits. Especially in this PR that is a must haha.

compose.yml
```diff
@@ -0,0 +1,109 @@
...
+      interval: 10s
+      timeout: 5s
+      retries: 5
+
+  rabbitmq:
+    image: rabbitmq:3.13.6-management-alpine
+    container_name: mbin-rabbitmq
+    restart: unless-stopped
+    networks:
+      - mbin_external_network
```
melroy89, line 96: shouldn't rabbitmq not just be an internal network (`internal: true`) in Docker? Yes, you can have multiple networks in Docker. And in Compose you can also add multiple networks to a single service (eg. `messenger`  can both have internal and external docker networks).

compose.yml
```diff
@@ -0,0 +1,109 @@
...
+      - .env
+    healthcheck:
+      test: ["CMD", "redis-cli", "ping"]
+
+  db:
+    image: postgres:${POSTGRES_VERSION:-16}-alpine
+    container_name: mbin-db
+    restart: unless-stopped
+    networks:
+      - mbin_external_network
```
melroy89, line 82: DB & Redis can definitely be an internal network  (`internal: true`). Yes, you can have multiple networks in Docker. And in Compose you can also add multiple networks to a single service (eg. messenger can both have internal and external docker networks).

compose.prod.yml
```diff
@@ -0,0 +1,126 @@
...
+  #  networks:
+  #    - mbin_external_network
+  #  environment:
+  #    - SMARTHOST_ADDRESS=mail.mysmtp.com
+  #    - SMARTHOST_PORT=587
+  #    - SMARTHOST_USER=myuser
+  #    - SMARTHOST_PASSWORD=secret
+  #    - SMARTHOST_ALIASES=*.mysmtp.com
+
+networks:
```
melroy89, line 125: why do you repeat the networks in the `compose.prod.yml`?

This `compose.prod.yaml` file should only contain the changes you want to do on top of the compose.yml file: https://docs.docker.com/compose/production/#modify-your-compose-file-for-production. Correct?

melroy89: And later also the override: https://docs.docker.com/compose/multiple-compose-files/merge/

docker/php/Dockerfile
```diff
@@ -0,0 +1,159 @@
...
+COPY --link .env.example_docker .env
+
+# Dump-autoload and run post-install script
+RUN composer dump-autoload --classmap-authoritative --no-dev
+RUN composer run-script --no-dev post-install-cmd && \
+    chmod +x bin/console && sync
+
+
+######################################################
+FROM node:22.5.1-alpine3.20 as builder-nodejs
```
melroy89, line 86: Why not use `22.7.0-alpine3.20`?

compose.yml
LoveIsGrief: :thinking: I'm not sure what advantage that would have for the `php` and `messenger` services. If I'm not mistaken those require outgoing access to external networks in order to download stuff and communicate with other instances. The `php` service should probably be able to access direct incoming traffic. Dunno about the messenger.

The DB and redis could probably be purely internal as I don't believe they'll ever need to access external resources.

Should this be done in another PR? I feel like it could expand the scope of this PR and it might be good to have a separate PR with one concern (docker compose networking) for posterity.

compose.prod.yml
LoveIsGrief: Indeed, that was an oversight. I meant to remove them but forgot :slightly_smiling_face:

docker/php/Dockerfile
LoveIsGrief: That was unchanged from the last `Dockerfile`. Dunno why github has trouble with `git mv` 🫤 

But I can upgrade it, no problem.

LoveIsGrief: Addressed in https://github.com/MbinOrg/mbin/pull/1021/commits/3bb52831faf01406df1607e728c4b2f3593c3560

compose.prod.yml
LoveIsGrief: Addressed in https://github.com/MbinOrg/mbin/pull/1021/commits/ad9aa3eb397a75956317dcdd7f4c2d46bb470bdf

TheVillageGuy (approved):
As Bentigerolich asked for this PR and Average Dude is waiting for approval I am approving this

compose.yml
melroy89: As fair as I know only the www service (and maybe also the rabbitmq if you enabled the management interface) needs access to the external network, the rest of all the services can do internal communication between the services.

Sure, this can also be done in a follow-up PR.

compose.prod.yml
melroy89: Thx

docker/php/Dockerfile
melroy89: Thx as well.

docs/02-admin/01-installation/docker.md
```diff
@@ -1,26 +1,33 @@
...
+> If you were already using docker in production, please see the [migration guide](#migration-guide) for updates
 
 ## System Requirements
 
 - Docker Engine
 - Docker Compose V2
 
-  > If you are using Compose V1, replace `docker compose` with `docker-compose` in those commands below.
+> [!WARNING]  
+> `docker-compose` is **not** supported!
```
melroy89, line 16: This can be very confusing I believe. "Docker compose" is supported and even the `docker-compose` command is supported, if you download the [standalone docker compose](https://docs.docker.com/compose/install/standalone/)). So please revert this sentence back.

docs/02-admin/01-installation/docker.md
```diff
@@ -134,60 +127,110 @@ OAUTH_PASSPHRASE=<Your (optional) passphrase from above here>
...
+### Filesystem ACL support
+
+The filesystem ACL is disabled by default, in the `mbin` image. You can set the environment variable `ENABLE_ACL=1` to
+enable it. Remember that not all filesystems support ACL. This will cause an error if you enable filesystem ACL for such
+filesystems.
 
 ### Mbin NGINX Server Block
 
+> ![WARNING]
+> This is not up to date! PRs for a valid nginx service and config welcome!
```
melroy89, line 232: Why would this not be up to date?

docs/04-contributing/development_environment.md
```diff
@@ -0,0 +1,239 @@
...
+**Copy the compose.dev.yml**
+
+Instead of linking, copy the `compose.dev.yml` to `compose.override.yml` and make your changes.
+
+The downside of this approach is that changes and fixes to `compose.dev.yml` from version control will not reflect
+in your compose setup automatically.
+
+## How it works
+
+MBin depends on multiple services (postgreSQL, redis, a reverse proxy, PHP, ...). Instead of having one monolithic
```
melroy89, line 57:
```suggestion
Mbin depends on multiple services (PostgreSQL, Redis, a reverse proxy, PHP, ...). Instead of having one monolithic
```

docs/04-contributing/development_environment.md
```diff
@@ -0,0 +1,239 @@
...
+
+Instead of having one large compose file to cater to every need, there are multiple - the main one being `compose.yml`.
+This takes advantage of [merging][docker compose merging]. Merging is done automatically when a `compose.override.yml`
+is present, otherwise the list of files has to be passed with `docker compose -f compose.yml -f file1.yml -f file2.yml`.
+
+### php/Dockerfile
+
+[This][php-dockerfile] is the biggest and most complex `Dockerfile` in the repo.
+It's a multi-stage file that attempts to make the final image minimal. You can read it, but the most important things
+to keep in mind are
```
melroy89, line 72:
```suggestion
to keep in mind are:
```

docker/php/entrypoint.sh
```diff
@@ -7,25 +7,28 @@ if [ "${1#-}" != "$1" ]; then
 fi
 
 if [ "$1" == "php-fpm" ] || [ "$1" == "php" ] || [ "$1" == "bin/console" ]; then
+    pwd
```
melroy89, line 10: Remove `pwd`

docs/02-admin/01-installation/docker.md
```diff
@@ -248,3 +291,21 @@ server {
...
+
+```shell
+mkdir -p docker/storage/www/public
+# Copy to have the previous files as a backup
+cp -r docker/storage/media docker/storage/www/public
+```
+
+You should then be set all set.
+
+[docker compose merging]: https://docs.docker.com/compose/compose-file/13-merge/
```
melroy89, line 310: What is this line? This is invalid markdown syntax.

docs/02-admin/01-installation/docker.md
LoveIsGrief: Reverted in [0a8a3ad](https://github.com/MbinOrg/mbin/pull/1021/commits/0a8a3adec9edd4f6772cd9121bfd5922bec64ee2)

docs/04-contributing/development_environment.md
```diff
@@ -0,0 +1,239 @@
...
+
+- PHP v8.2
+- NodeJS
+- Redis
+- PostgreSQL
+- _Optionally:_ Mercure
+
+---
+
+- Increase execution time in PHP config file: `/etc/php/8.2/fpm/php.ini`:
```
melroy89, line 161: Is this PR rebased correctly? Since it should be promoting PHP 8.3.

```suggestion
- Increase execution time in PHP config file: `/etc/php/8.3/fpm/php.ini`:
```

docs/04-contributing/development_environment.md
```diff
@@ -0,0 +1,239 @@
...
+
+**NixOS**
+
+Nixos needs [iptables rules](https://discourse.nixos.org/t/docker-container-not-resolving-to-host/30259/8).
+
+# Development Server
+
+Requirements:
+
+- PHP v8.2
```
melroy89, line 153: Is this rebased correctly? Since it should be promoting php v8.3.
```suggestion
- PHP v8.3
```

docs/04-contributing/development_environment.md
```diff
@@ -0,0 +1,239 @@
...
+
+---
+
+- Increase execution time in PHP config file: `/etc/php/8.2/fpm/php.ini`:
+
+```ini
+max_execution_time = 120
+```
+
+- Restart the PHP-FPM service: `sudo systemctl restart php8.2-fpm.service`
```
melroy89, line 167:
```suggestion
- Restart the PHP-FPM service: `sudo systemctl restart php8.3-fpm.service`
```

docs/02-admin/01-installation/docker.md
LoveIsGrief: Caddy was pulled out of the `php` image, so the config is not up to date to use `php-fpm`. I haven't tested whether it still works and this config uses HTTPS, so I don't have a domain to test it with.

docs/02-admin/01-installation/docker.md
LoveIsGrief: This is actually valid. It's called a "[link reference definition][lrd]" (see the source of this comment for an example). It allows having all your links at the bottom of the document and reusing links.


[lrd]: https://spec.commonmark.org/0.31.2/#link-reference-definition

melroy89: ow like that, yes indeed that is valid.

docs/02-admin/01-installation/docker.md
melroy89: Nginx is a reverse proxy that does SSL offloading, that means in normal terms: Nginx will provide HTTPS and will remove the SSL and connect to the service (without the need of SSL) and then it will add back the SSL to the response. So, Nginx can call the www (caddy) port without the need of SSL without any issues.

compose.yml
LoveIsGrief: Created [a ticket](https://github.com/MbinOrg/mbin/issues/1073) so it won't be forgotten.

docs/02-admin/01-installation/docker.md
LoveIsGrief: Removed the warning.

docs/04-contributing/development_environment.md
```diff
@@ -0,0 +1,239 @@
+There are multiple ways to get started.
```
melroy89, line 1: You still need to start with a heading 1. Like we had before.

Also a it's not a development environment, a development environment is workspace for developers to make changes without breaking anything in a live environment. Some people even use development environment to refer to their IDE. I think it's better to just keep it "development server". Which is really what we try to setup here.

melroy89:
You say there is a feature to use: `ln -s compose.dev.yml compose.override.yml`. However, this is not documented in the documentation..?

docs/04-contributing/development_environment.md
LoveIsGrief: :smile:  this has always been a problem in every company I've been in. Does dev env mean the developer's machine or the deployed environment used by devs. There's even https://devenv.sh/ which described itself as "Fast, Declarative, Reproducible and Composable Developer Environments using Nix" which refers to the local machine.

I don't mind changing the headers, but what should they become?

```md
# Development server

## Docker

## Host machine
````

Where "host machine" is non-docker environment?

**Comment by melroy89** [comment]
Let's assume I'm a beginner. I followed:

```
# Create a symbolic link to the dev setup that merges and overrides part of compose.yml
ln -nfs compose.dev.yml compose.override.yml
# Bring up all the services and detach from the console
docker compose up -d
```

That will result into:

```
$ docker compose up -d
WARN[0000] The "POSTGRES_USER" variable is not set. Defaulting to a blank string. 
env file /media/melroy/Data/Projects/mbin_docker/.env not found: stat /media/melroy/Data/Projects/mbin_docker/.env: no such file or directory
```

**Comment by LoveIsGrief** [comment]
Found the issue @melroy89 . Fixed in https://github.com/MbinOrg/mbin/pull/1021/commits/478b5130192d4fd75e3ec7b8c6f0fefce4e80b41 . I had a `.env` the entire time :sweat_smile:

File: docs/04-contributing/development_environment.md (hunk 16 above)
Comment (melroy89) [reply docs/04-contributing/development_environment.md:1]: Development environment is literately the virtual place where you indeed develop code (hence the name). Just like integrated development environment, refers to the piece of software that help you write that code. 

So devenv is important within a company, but we aren't providing any development environment like maybe VDIs (Virtual Desktop Infrastructure) to your contributors. 

So I still believe it's a development server (it's also not a staging server), but really a local development server we try to aim to setup here. And then the software engineer can choose between either a docker setup or just "bare metal" is what we called it thus far.  "Host machine" might be an alternative for "bare metal". 
Changing the filename also means that it will break search engine indexed results (like this URL will not work anymore: https://docs.joinmbin.org/contributing/development_server).

I think the top-level heading might still then be called "Development Server" indeed, then "Docker". And well, we could go for Host machine, maybe that is confusing for most people (since Docker also runs on your host machine), we also could call it "Bare metal"? We could even considering splitting the documentation. Look how we did it for "Installation": https://docs.joinmbin.org/admin/installation/

Comment (LoveIsGrief) [reply docs/04-contributing/development_environment.md]: I finally split it in two (docker + bare metal). It looks like a good pattern to follow especially due to how long the files can become.

File: docs/04-contributing/docker.md [hunk 17]
```diff
@@ -0,0 +1,153 @@
...
+    xdebug->>php: break
+    php-->>xdebug: 
+    xdebug->>php: continue
+    php-->>xdebug: 
+    xdebug-->>-php: done
+    php-->>-rp: response
+    rp-->>-Browser: response
+```
+
+**Requirement**
```
melroy89, line 127: Instead of using bold text. Use markdown headers. This is valid for all the lines you use bold text that is actually a header.

LoveIsGrief: Updated.

docs/04-contributing/docker.md
```diff
@@ -0,0 +1,153 @@
+# Docker
```
melroy89, line 1:
```suggestion
# Docker Development Server
```

compose.yml
melroy89: I now get:

```
Error response from daemon: driver failed programming external connectivity on endpoint mbin-db (c1c3e1b40be94b691d7a32acbdcde8723b9a7402b68bbb64598c5533e354b1e5): failed to bind port 127.0.0.1:5432/tcp: Error starting userland proxy: listen tcp4 127.0.0.1:5432: bind: address already in use
```

So this external network will interfere with any host ports that run for example MySQL/MariaDB. Therefore, I reopened this conversation.

compose.prod.yml
```diff
@@ -0,0 +1,113 @@
+services:
+  www:
+    # To use a prebuilt image from ghcr.io comment out the "image" and "pull_policy" lines
+    # You can also a tag eg.: "v1.0.0" instead of "latest"
+    #image: "ghcr.io/mbinorg/mbin:latest-caddy"
+    #pull_policy: never
+    ports:
+      - 80:80
```
melroy89, line 8: Do not reuse port `80` on the host machine. 99.999% of the time this port is taken already by either Nginx or Apache2. So you should bind this port to another port like `8008` we had before (on purpose). 

_Especially_ on production you will need to add a reverse proxy anyways (eg. for SSL). Exposing port 80 like this will not work together with Nginx or Apache2, while it's actually a requirement.

JaviviJR:
I cannot run test inside php container.

`bin/phpunit -c phpunit.xml.dist`

`1018) App\Tests\Unit\Service\MentionManagerTest::testExtract with data set #6 ('Lorem https://some.instance/@... ipsum', null)
LogicException: You cannot create the client used in functional tests if the "framework.test" config is not set to true.
`

.env.test exists, also is defined APP_ENV=test in phpunit.xml.dist.

Do you have the same problem or is just on my environment?

compose.prod.yml
LoveIsGrief: That's highly dependent on the individual. I consider this to be the equivalent of the bare metal. Once the bare metal steps are run through, I assume it's possible to access mbin on http://mydomain.tld. After following the docker prod guide, the user would not (http://mydomain.tld:8008).

Someone using a reverse proxy would probably see the error from docker compose that a port cannot be allocated and modify the port. By setting it to `8008`, you're making the assumption that 0.0001% of admins have a reverse proxy. Since caddy is already the chosen reverse proxy, it has [HTTPS by default](https://caddyserver.com/docs/automatic-https), so theoretically, opening port `443`, should let caddy provision those certs.

In any case, I modified it to `8008`.

melroy89:
I stopped postgresql locally, so it now starts the docker images. Going to http://localhost:8080 gives me:

```
An exception has been thrown during the rendering of a template ("Asset manifest file "/var/www/mbin/public/build/manifest.json" does not exist. Did you forget to build the assets with npm or yarn?").
```

PR #1035, opened 2024-08-17
Change attributed to parsing to not expect a url

BentiGorlich:
Previously we extracted exactly one url from the `attributedTo` field, then fetched it and expected it to be a collection of user ids. In the case of PeerTube channels that meant, that we get a user id and expect to retrieve a moderators collection from that url, which we do not, so that does not work. Now we only handle it this way if the `attributedTo` field is a string. If it is not a string, but an array we put all contained `id`s into an array and get the moderators from that.

PeerTube `attributedTo` of a channel:
```json
{
  "attributedTo": [
    {
      "type": "Person",
      "id": "https://instance.tld/accoutns/someuser"
    }
  ]
}
```

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -533,7 +533,7 @@ class ActivityPubManager
             $magazine->apInboxUrl = $actor['endpoints']['sharedInbox'] ?? $actor['inbox'];
             $magazine->apDomain = parse_url($actor['id'], PHP_URL_HOST);
             $magazine->apFollowersUrl = $actor['followers'] ?? null;
-            $magazine->apAttributedToUrl = $this->getActorFromAttributedTo($actor['attributedTo'] ?? null, filterForPerson: false);
+            $magazine->apAttributedToUrl = \is_string($actor['attributedTo']) ? $actor['attributedTo'] : null;
             $magazine->apFeaturedUrl = $actor['featured'] ?? null;
             $magazine->apPreferredUsername = $actor['preferredUsername'] ?? null;
             $magazine->apDiscoverable = $actor['discoverable'] ?? true;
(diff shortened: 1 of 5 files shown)


melroy89 (approved):

melroy89:
@BentiGorlich I think this is causing some additional error log pollution I noticed.  

```
{"message":"Warning: Undefined array key \"attributedTo\"","context":{"exception":{"class":"ErrorException","message":"Warning: Undefined array key \"attributedTo\"","code":0,"file":"/var/www/kbin.melroy.org/html/src/Service/ActivityPubManager.php:536"}},"level":400,"level_name":"ERROR","channel":"php","datetime":"2024-08-20T16:24:42.071662+02:00","extra":{}}
{"message":"Warning: Undefined array key \"attributedTo\"","context":{"exception":{"class":"ErrorException","message":"Warning: Undefined array key \"attributedTo\"","code":0,"file":"/var/www/kbin.melroy.org/html/src/Service/ActivityPubManager.php:564"}},"level":400,"level_name":"ERROR","channel":"php","datetime":"2024-08-20T16:24:42.406210+02:00","extra":{}}
```

src/Service/ActivityPubManager.php
```diff
@@ -561,6 +561,8 @@ public function updateMagazine(string $actorUrl): ?Magazine
                     $this->handleModeratorCollection($actorUrl, $magazine);
                 } catch (InvalidArgumentException $ignored) {
                 }
+            } elseif (\is_array($actor['attributedTo'])) {
```
melroy89, line 564: Causing `undefined array key \"attributedTo\"` in come cases.

src/Service/ActivityPubManager.php
```diff
@@ -533,7 +533,7 @@ public function updateMagazine(string $actorUrl): ?Magazine
             $magazine->apInboxUrl = $actor['endpoints']['sharedInbox'] ?? $actor['inbox'];
             $magazine->apDomain = parse_url($actor['id'], PHP_URL_HOST);
             $magazine->apFollowersUrl = $actor['followers'] ?? null;
-            $magazine->apAttributedToUrl = $this->getActorFromAttributedTo($actor['attributedTo'] ?? null, filterForPerson: false);
+            $magazine->apAttributedToUrl = \is_string($actor['attributedTo']) ? $actor['attributedTo'] : null;
```
melroy89, line 536: Causing `undefined array key \"attributedTo\"` in come cases.

PR #1036, opened 2024-08-17
Only match magazine by hashtag if post was in random

BentiGorlich:
- only change the magazine of a post via hashtag matching when the post was in the random magazine
- don't use a magazines name to hashtag match, only the defined hashtags of a magazine
- re-add the announcement of posts and post comments

### DIFF
--- a/src/EventSubscriber/Post/PostCreateSubscriber.php
+++ b/src/EventSubscriber/Post/PostCreateSubscriber.php
@@ -56,14 +56,13 @@ class PostCreateSubscriber implements EventSubscriberInterface
 
     private function handleMagazine(Post $post): void
     {
+        if ('random' !== $post->magazine->name) {
+            // do not overwrite matched magazines
+            return;
+        }
         $tags = $this->tagLinkRepository->getTagsOfPost($post);
 
         foreach ($tags as $tag) {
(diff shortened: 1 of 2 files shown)


melroy89:
> * don't use a magazines name to hashtag match, only the defined hashtags of a magazine

Are you sure this breaking change will not effect the majorly of the threads, comments etc within the posted magazines? I think there was and still is a good reason to use the magazine name as well, since not all magazines will include the hashtag within the magazine name.

Let me put it differently, will a post from eg. Lemmy still land in the correct magazines?

BentiGorlich:
Hashtags are only used to match posts (meaning microblogs) to magazines. Not threads, not comments, not replies, only microblogs. It is a cool feature that mbin has, but since the magazine name was just used to match a magazine like "Kbin Run meta" which had the name "meta" (so the address was kbin.run/m/meta) got a lot of microblogs talking about #meta which is just wrong. My reasoning to remove this is just that this is not an advertised feature. If you want to match microblogs by hashtags into a magazine, just add the hashtag you want to match.

melroy89:
ah I see. Agreed. Yea this is wrong. You don't want to have meta posts in your magazine, never!

melroy89 (approved):

PR #1039, opened 2024-08-18
Return only the host as audience for remote magazines

BentiGorlich:
I noticed this a while ago, but then I thought that the magazine inbox is contained in the `To` and `Cc` fields which is retrieved by `ActivityPubManager::createInboxesFromCC` which is used most of the time when retrieving the inboxes (like in the outbox `CreateHandler`)


src/MessageHandler/ActivityPub/Outbox/CreateHandler.php
```diff
@@ -63,6 +63,7 @@ public function doWork(MessageInterface $message): void
                 ...$this->activityPubManager->createInboxesFromCC($activity, $entity->user),
                 ...$this->magazineRepository->findAudience($entity->magazine),
             ];
+            $this->logger->debug('sending create activity to {p}', ['p' => $receivers]);
```
melroy89, line 66: This needed here?

melroy89 (approved):

PR #1040, opened 2024-08-18
Fix error when updating an actor and the instance has no nodeinfo endpoint

BentiGorlich:
I noticed this kind of errors in my log:
> Error: \"Handling \"App\\Message\\ActivityPub\\UpdateActorMessage\" failed: HTTP/2 404  returned for \"https://www.threads.net/.well-known/nodeinfo\"

threads.net might not be the best example, but an update actor should not fail, because an instance does not have an nodeinfo endpoint

### DIFF
--- a/src/Service/ActivityPub/ApHttpClient.php
+++ b/src/Service/ActivityPub/ApHttpClient.php
@@ -319,7 +319,13 @@ class ApHttpClient
             $item->expiresAt(new \DateTime('+1 day'));
             try {
                 return $this->generalFetch($url, ApRequestType::NodeInfo);
-            } catch (\Exception) {
+            } catch (\Exception $e) {
+                $this->logger->warning('There was an exception fetching nodeinfo endpoints from {url}: {e} - {msg}', [
+                    'url' => $url,
+                    'e' => \get_class($e),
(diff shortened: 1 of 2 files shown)


src/Service/ActivityPub/ApHttpClient.php
```diff
@@ -313,8 +317,11 @@ public function fetchInstanceNodeInfoEndpoints(string $domain, bool $decoded = t
         $url = "https://$domain/.well-known/nodeinfo";
         $resp = $this->cache->get('nodeinfo_endpoints_'.hash('sha256', $url), function (ItemInterface $item) use ($url) {
             $item->expiresAt(new \DateTime('+1 day'));
-
-            return $this->generalFetch($url, ApRequestType::NodeInfo);
+            try {
+                return $this->generalFetch($url, ApRequestType::NodeInfo);
+            } catch (\Exception) {
+                return null;
```
melroy89, line 323: also log the error or?

BentiGorlich: yeah that'd be usefult

melroy89 (dismissed):

melroy89 (approved):
Great!

Issue #1001, opened 2024-08-06
Missing items/endpoints from api


jwr1:
Here's a non exhaustive list of missing items/endpoints from the API that would be quite useful:

- [ ] Tell if a user is an instance admin or global mod
- [ ] Tell if the author of _Content_ is a mod of the magazine it was put in
- [ ] Tell if the signed in user has moderation capabilities over _Content_
- [ ] List the magazines a user is a mod of
- [ ] List the activities of _Content_ (meaning who upvoted and boosted)

I'll add to this list once I run into more walls.

Also, whenever I say _Content_, I mean: thread/microblog/post/comment

BentiGorlich:
Could you split the list in
1. missing items
2. missing endpoints

Because I don't really know for which of these a dedicated endpoint might make sense

christianjuth:
> Related idea: Implement Lemmy-compatible API endpoints for clients to use.
> 
> This would open the possibility for already existing clients to login to an Mbin instance account (e.g. Use smartphone clients like Mlem or Voyager to login to `fedia.io` and other Mbin instances)
> 
> [Mlem has stated interest, but noted that Lemmy-compatible API would make it easier][1].  The rationale for not giving priority on implementing this on their side is Mbin & Kbin having less active users, but it's a bit of a "chicken and egg" issue.  That is to say: Mobile client support would be a precursor to more active users, but their reasoning for not working to add Mbin-specific support is based on not enough users.
> 
> [1]: https://github.com/mlemgroup/mlem/issues/120

Because Lemmy clients are starting to support PieFed, that has forced them to introduce a compatibility layer that makes it easier to support different APIs. I don't think implementing the Lemmy API exacting will be necessary.

I'm the maintainer of https://github.com/christianjuth/blorp and I'm staring to look into MBin support. While I don't need the API to match Lemmy, I do need the features of the API to roughly match Lemmy (e.g. resolve object endpoint). I also need a way to get activity pub id for all objects, regardless of if they are local or federated.

I know Voyager has shifted to using https://www.npmjs.com/package/threadiverse. If you add MBin support to that package, you should get support in Voyager pretty easily.

BentiGorlich:
I have not really reacted to this, but: our dev resources are very limited and since we are organized differently form lemmy it is not an easy task to implement their API. I know this hinders app development, but my personal priorities lie in improving mbins code and building new features. That also means that I will not be investing the time to add Mbin API support to the threadiverse package.

[...]

PR #1043, opened 2024-08-19
Add field to all content response DTOs: `canAuthUserModerate`

BentiGorlich:
- All content response DTOs (entry, entry_comment, post, post_comment) now have a field `canLoggedInUserModerate` which indicates whether the loggedin user can moderate said content

Link #1001

### DIFF
--- a/src/Controller/Api/Entry/EntriesBaseApi.php
+++ b/src/Controller/Api/Entry/EntriesBaseApi.php
@@ -45,7 +45,7 @@ class EntriesBaseApi extends BaseApi
         }
 
         if ($user = $this->getUser()) {
-            $response->canLoggedInUserModerate = $dto->getMagazine()->userIsModerator($user) || $user->isModerator() || $user->isAdmin();
+            $response->canAuthUserModerate = $dto->getMagazine()->userIsModerator($user) || $user->isModerator() || $user->isAdmin();
         }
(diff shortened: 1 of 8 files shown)


jwr1:
Just a tiny change, but would you be able to change the field name to `canAuthUserModerate` (can authenticated user moderate)? It's not too big a deal, but IMO, it sounds better than `canLoggedInUserModerate`. If you think it sounds better as it is, feel free to leave it alone though.

BentiGorlich:
done

jwr1 (changes requested):
I just tested it out, and it looks like `canAuthUserModerate` is null if it's not a first level comment. Here's an example you can look at (comments I can mod have "CAN MOD" next to them, it should be all of them though):

image

BentiGorlich:
I updated the PR :)

jwr1:
It looks like it is fixed for entry comments, but the issue still exists with post (microblog) comments.

BentiGorlich:
True, forgot about that. I will add it tomorrow

Am 19. August 2024 22:55:46 MESZ schrieb John Wesley ***@***.***>:

BentiGorlich:
Done

jwr1 (approved):
Just tried this also, looks good to me!

PR #1049, opened 2024-08-21
Add a warning message when we hit the cache in post

BentiGorlich:
- the cache in the post method is only hit when we sent something with the same ap id multiple times in 45 minutes, so log a warning when that happens


melroy89 (approved):

melroy89:
Maybe create a follow-up item? If we see a lot of these warnings. Or nothing at all, meaning we potentially can remove it..??

BentiGorlich:
Absolutely

PR #1053, opened 2024-08-22
Add missing checks for banned instance

BentiGorlich:
currently it is only checked if an instance is banned when receiving activities. Add the check to:
- CreateHandler: creating content (not so relevant, should already covered by the check in the ActivityHandler)
- ChainActivityHandler: creating content when a dependency of replies, likes, etc (probably the most relevant one)
- ActivityPubManager: when updating actors, creating users and magazines

### DIFF
--- a/src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
+++ b/src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
@@ -89,7 +89,7 @@ class CreateHandler extends MbinMessageHandler
             }
             $this->logger->info('Did not create the post, because the magazine {m} restricts posting to mods and {u} is not a mod', ['m' => $e->magazine, 'u' => $username]);
         } catch (InstanceBannedException $e) {
-            $this->logger->error('Did not create the post, because the user\'s instance is banned');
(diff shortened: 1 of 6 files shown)


src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
```diff
@@ -87,13 +88,16 @@ public function doWork(MessageInterface $message): void
                 $username = $e->actor->name;
             }
             $this->logger->info('Did not create the post, because the magazine {m} restricts posting to mods and {u} is not a mod', ['m' => $e->magazine, 'u' => $username]);
+        } catch (InstanceBannedException $e) {
+            $this->logger->error('Did not create the post, because the user\'s instance is banned');
```
melroy89, line 92: why not a `info`? It's not really an error.. Its more a notification.

BentiGorlich: true

melroy89 (approved):
It's good now 👍🏽

jwr1 (approved):
Yup, I'd say that worked.

PR #1054, opened 2024-08-22
Related links

JaviviJR:


templates/user/settings/profile.html.twig
```diff
@@ -43,6 +43,41 @@
             }}) }}
             {{ form_row(form.avatar, {label: 'avatar'}) }}
             {{ form_row(form.cover, {label: 'cover'}) }}
+
```
melroy89, line 46: Keep using `form_row()` if you want to add additional fields/rows to the form, just like above. So it stays consistent.

BentiGorlich:
Can you implement it the mastodon way where you have a name-link combination? So that each add link field has an associated name field.
grafik

JaviviJR:
> Can you implement it the mastodon way where you have a name-link combination? So that each add link field has an associated name field. grafik

Do you want that user can add just links or whatever data user wants like Mastodon?

BentiGorlich:
Limit it to links would be fine for me, but I don't think that it makes sense :thinking:

assets/controllers/form-related-links_controller.js
```diff
@@ -0,0 +1,50 @@
...
+
+        this.#addButtonDeleteLink(nodeLink);
+
+        this.relatedContainerTarget.appendChild(nodeLink);
+        this.indexValue++;
+    }
+
+    #addButtonDeleteLink(item) {
+        const removeFormButton = document.createElement('button');
+        removeFormButton.innerText = '⌫';
```
BentiGorlich, line 34: I think overwriting `innerHtml` with the font awesome trashcan icon (https://fontawesome.com/icons/trash?f=classic&s=solid) would be better. Then you could also make it a button:
```html
<div class="btn btn__secondary">
    <i class="fa-solid fa-trash"></i>
</div>
```

src/Entity/User.php
```diff
@@ -894,4 +897,14 @@ public function canUpdateUser(User $actor): bool
             return $this->apDomain === $actor->apDomain;
         }
     }
+
+    public function getRelatedLinks(): array
```
BentiGorlich, line 901: Could you add phpDoc so we know which kind of array is returned? I really liked if the returned value would be of `RelatedLink[]` so an array of objects, so we can add some validation logic so we can mark links as "checked" or something so we know that the rel=me is actually correct

melroy89:
> Can you implement it the mastodon way where you have a name-link combination? So that each add link field has an associated name field. grafik

Agreed that would be helpful, so it's basically a key/value list.

src/Form/UserRelatedDataType.php
```diff
@@ -0,0 +1,67 @@
...
+use Symfony\Component\Form\FormInterface;
+use Symfony\Component\OptionsResolver\OptionsResolver;
+
+class UserRelatedDataType extends AbstractType implements DataMapperInterface
+{
+    public function buildForm(FormBuilderInterface $builder, array $options)
+    {
+        $builder
+            ->add('label', TextType::class)
+            ->add('link', UrlType::class)
```
BentiGorlich, line 21: The url type force this value to be a url. Even if I insert something normal, it still comes out as a url. In general I think this variable/field should be called `value` instead

src/Entity/User.php
```diff
@@ -894,4 +897,14 @@ public function canUpdateUser(User $actor): bool
...
+    {
+        return $this->relatedLinks;
+    }
+
+    public function setRelatedLinks(array $relatedLinks): void
```
BentiGorlich, line 906: This one is actually not used. Maybe make the variable private and use this one?

src/Factory/UserFactory.php
```diff
@@ -4,23 +4,28 @@
...
         private readonly InstanceRepository $instanceRepository,
         private readonly Security $security,
+        private readonly DenormalizerInterface $denormalizer,
     ) {
     }
 
     public function createDto(User $user): UserDto
     {
+//        $temp = $this->denormalizer->denormalize($user->getRelatedLinks(), sprintf('%s[]', RelatedLinkDTO::class));
+
```
BentiGorlich, line 28: need to be removed

assets/controllers/form-related-links_controller.js
```diff
@@ -0,0 +1,62 @@
...
+        ));
+        rowNode.appendChild(nodeLink);
+
+        this.#addButtonDeleteLink(rowNode);
+
+        this.relatedContainerTarget.appendChild(rowNode);
+        this.indexValue++;
+    }
+
+    #addButtonDeleteLink(item) {
```
BentiGorlich, line 44: What is this hashtag for?

assets/styles/layout/_forms.scss
```diff
@@ -525,3 +525,36 @@ div.input-box {
...
+}
+
+.related-link-row button {
+    padding: 0;
+    background: none;
+    font-size: 1.5rem;
+    border: 0;
+    color: var(--kbin-input-text-color);
+
+}
```
BentiGorlich, line 560: please remove that and add the class `btn__secondary` to the trash icon buttons. Then they will look like other buttons in our UI

translations/messages.en.yaml
```diff
@@ -890,3 +890,4 @@ admin_users_suspended: Suspended
 admin_users_banned: Banned
 user_verify: Activate account
 max_image_size: Maximum file size
+related_links: Related links
```
BentiGorlich, line 893: Please keep newlines at the end of files

BentiGorlich (changes requested):
I'd love if the add button would something like this (but that is not a requirement):
grafik

I noticed that the links are not yet displayed on the profile page. Are you planning on adding that in this PR?
(The activity pub representation of the fields is also missing, but I expected to do that myself anyways :) but if you'd like to try that as well, feel free to do so)

BentiGorlich:
Oh and please change the target of the PR to `dev/new_features`

assets/controllers/form-related-links_controller.js
JaviviJR: Modified

src/Entity/User.php
JaviviJR: Discussed Discussed on element chat.
I don't know a elegant way to denormalize the json field from table and return it on get method. This'll required some dependencies inside entity that I think is not a good practice.
Instead, I created a RelatedLinkDTO class and this will be returned in UserDTO.

src/Form/UserRelatedDataType.php
JaviviJR: modified link by value

src/DTO/RelatedLinkDTO.php
```diff
@@ -0,0 +1,58 @@
+<?php
+
+namespace App\DTO;
+
+class RelatedLinkDTO
+{
+    private string $label;
+    private string $link;
```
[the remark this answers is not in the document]
JaviviJR: modified

src/DTO/RelatedLinkDTO.php
```diff
@@ -0,0 +1,58 @@
...
+    private bool $verified = false;
```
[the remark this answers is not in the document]
JaviviJR: modified

src/Entity/User.php
JaviviJR: deleted

src/Factory/UserFactory.php
JaviviJR: deleted

assets/controllers/form-related-links_controller.js
JaviviJR: Define a method private

imagen


https://railsdesigner.com/proper-stimulus-controllers/

assets/styles/layout/_forms.scss
JaviviJR: modified

melroy89:
> Oh and please change the target of the PR to `dev/new_features`

This latter might require a new pr in github.

JaviviJR:
> I'd love if the add button would something like this (but that is not a requirement): grafik
> 
> I noticed that the links are not yet displayed on the profile page. Are you planning on adding that in this PR? (The activity pub representation of the fields is also missing, but I expected to do that myself anyways :) but if you'd like to try that as well, feel free to do so)

Check it out with the new changes.
I tried to keep it centered in vertical line which separate the two fields, but I couldn't.

BentiGorlich:
First of all thanks for your time and effort :heart: 
I will have a look at it in the next days.

From your comment and screenshots I (already) have a few remarks (sorry :see_no_evil:):
> I tried to keep it centered in vertical line which separate the two fields, but I couldn't.

I think I did it by making a div with a margin left the same size as the buttons and then just put `text-align: center` in there.

> I modified userBox component to show related links in profile.

What happens when the value is not a link?

melroy89:
There is a conflict with `translations/messages.es.yaml`. Ideally only change the English language file. And translate the rest in Weblate.

assets/controllers/form-related-links_controller.js
melroy89: Today I learned... https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_properties

BentiGorlich:
could you rebase your branch on `dev/new_features` please so the github actions are triggered? We forgot to update the config for that :)

melroy89:
Also there is file conflicts xd

melroy89:
Fixed merge conflicts..

melroy89:
Fix the php-cs-fixer

BentiGorlich:
@JaviviJR did you push all your changes? The modifications of the edit profile page and the user box are not there, yet

PR #1055, opened 2024-08-23
Update _form_media.html.twig

TheVillageGuy:
A user tried uploading an image to my instance but clicked on upload file instead of add new photo. As this button does nothing, neither does the from url button, the user gave up. These buttons are useless but most of all confusing, so I propose;

Removing unused buttons on image upload dialog.

Adding width:300px to make it look good. Leaving the imageUrl, removing it may cause problems.

See what it looks like on [RimWorld.gallery](https://rimworld.gallery/m/debug/new/photo)

### DIFF
--- a/assets/styles/components/_media.scss
+++ b/assets/styles/components/_media.scss
@@ -35,7 +35,6 @@
 
     .image-input {
       max-width: 100%;
-      width: 100%;
     }
   }
 
@@ -77,10 +76,6 @@
         color: var(--kbin-button-secondary-text-hover-color);
       }
     }
-  }
-  
-  .image-form {
-    width: 300px;
   }
 }
 
--- a/assets/styles/components/_media.scss
+++ b/assets/styles/components/_media.scss
@@ -78,6 +78,10 @@
       }
     }
   }
+  
+  .image-form {
+    width: 300px;
+  }
 }
 
 .comment-add,
--- a/assets/styles/components/_media.scss
+++ b/assets/styles/components/_media.scss
@@ -77,6 +77,10 @@
       }
     }
   }
+  
+  .image-form {
+    width: 100%;
+  }
 }
 
 .comment-add,
--- a/templates/layout/_form_media.html.twig
+++ b/templates/layout/_form_media.html.twig
@@ -3,23 +3,12 @@
         <img class="image-preview" src="#">
         <button type="button" class="image-preview-clear" data-action="image-upload#clearPreview">x</button>
     </div>
-    <div>
+    <div style="width:300px;">
         {% if form.image is defined %}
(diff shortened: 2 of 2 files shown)


templates/layout/_form_media.html.twig
```diff
@@ -3,7 +3,7 @@
         <img class="image-preview" src="#">
         <button type="button" class="image-preview-clear" data-action="image-upload#clearPreview">x</button>
     </div>
-    <div>
+    <div style="width:300px;">
```
melroy89, line 6: Inline style is not recommend anymore, try to add a class. And use CSS.

TheVillageGuy: Done

melroy89 (approved):
I'm fine with this.

@BentiGorlich you too?

We can always add buttons back if we really need to. Or want to.

assets/styles/components/_media.scss
```diff
@@ -78,6 +78,10 @@
       }
     }
   }
+  
+  .image-form {
```
BentiGorlich, line 82: Did you try this on mobile? Setting a width in pixels is often causing problems on mobile...

TheVillageGuy: It works fine on mine

Screenshot_20240825-163636

TheVillageGuy: It works fine on mine

Screenshot_20240825-163636

melroy89: Depending on the DPI and display.. It could maybe result into issues indeed. You might want to use percentage (%) or vh

TheVillageGuy: Fixed is another way, see latest commit. The image-form class is now obsolete. Should I remove it from 
><div class="image-form">

BentiGorlich: I actually don't like the change. To test this you can hit f12 on your desktop when you're on the site and (if you're on Firefox) click this little phone icon: 
Bildschirmfoto vom 2024-08-27 12-25-14

With that you can test different mobile devices.

I actually don't understand which problem you were trying to solve when changing the width setting. If I apply a `width: 100%` to the `image-form` class it looks fine to me, scales very well with all display sizes...

TheVillageGuy: As to why; with no changes, it looked like this after the buttons removed: (impression)

image

which is solved by removing the  width: 100%; from .image-input

Thanks for the tip on the little camera, I always wondered why FF had no such feature, had to resort to chrome every time.

I have applied your suggestion width: 100% to image-form

BentiGorlich: Ah ok, then I misread which width you removed. The difference is only visible if you make the window very small, but I think it is fine this way :+1:

BentiGorlich (approved):

PR #1056, opened 2024-08-24
downgrade exceptions in DeleteUserHandler to warning logs

asdfzdfj:
it's clogging up dead queue and you couldn't retry them in a useful way anyway, so might as well just log them and abort the handler

also changes the `$this->user` var to explicitly pass it between functions rather than stashing it as a shared vars

### DIFF
--- a/src/MessageHandler/DeleteUserHandler.php
+++ b/src/MessageHandler/DeleteUserHandler.php
@@ -15,14 +15,11 @@ use App\Service\UserManager;
 use Doctrine\ORM\EntityManagerInterface;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Messenger\Attribute\AsMessageHandler;
-use Symfony\Component\Messenger\Exception\UnrecoverableMessageHandlingException;
 use Symfony\Component\Messenger\MessageBusInterface;
 
 #[AsMessageHandler]
 class DeleteUserHandler extends MbinMessageHandler
 {
-    private ?User $user;
-
     public function __construct(
         private readonly LoggerInterface $logger,
         private readonly ImageManager $imageManager,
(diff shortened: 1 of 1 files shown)


melroy89:
maybe just a bit off-topic: Why do we even have a dead-queue? Sorry for my ignorance.

BentiGorlich:
https://docs.joinmbin.org/admin/running-mbin/messenger
> We need the dead queue so that messages that throw a UnrecoverableMessageHandlingException, which is used to indicate that a message should not be retried and go straight to the supplied failure queue

melroy89:
> https://docs.joinmbin.org/admin/running-mbin/messenger
> 
> > We need the dead queue so that messages that throw a UnrecoverableMessageHandlingException, which is used to indicate that a message should not be retried and go straight to the supplied failure queue

Yeah, so uhm. Since we still keep the DeleteUserHandler messages, I think it's a good thing to also keep this exception and throw this error, instead of only logging them.

Then again, we do apparently need a dead queue, but I'm fine if that dead queue is just a `/dev/null` basically, instead of storing those messages into Postgresql. Which was my point of why do we need a dead queue like that. I understand that we need a dead-queue for the dead letter, but storing it postgresql isn't required IMO.

BentiGorlich:
If there is a kind of /dev/null queuetype, then yeah that wouldbe good. I was planning on adding a scheduled job that empties the dead queue, but I haven't gotten around to doing that, yet...

melroy89:
What happens when we just set `failure_transport: null`

BentiGorlich:
Then I assume it is the same as if you wouldn't set the transport at all and then the messengers would crash as soon as this exception is thrown

melroy89:
> Then I assume it is the same as if you wouldn't set the transport at all and then the messengers would crash as soon as this exception is thrown

we now really need some kind of Symfony expert to help us how if there is some DSN or something to validate if we can set it to `null` safely, and if not, how can we create a "/dev/null failed queue"..

asdfzdfj:
guess I'll consider closing this one (and keep throwing exceptions) and just make a new one that only contains the `$this->user` fix then, since those looks to be more objective problem *\*cough\* *embed thumbnails* \*cough\**, while this one may be more of a me problem

for context: I have my dev/test instance messenger config set up differently where it'll retry the failed message much faster and have the failed messages delivered to dead queue immediately, so I can quickly see what breaks, and from what I've seen is that even for a single user instance, these unretryable failed messages has slowly but surely filling up the dead queue, and when I want to monitor for what might go wrong (e.g. when making backend changes), these unretryable messages feels more like noises

another way to go about this is to perhaps evaluate the use of throwing unretryable message exception, and come to some agreement on what to do when those precondition failed whether to throw unretryable exception or to log it, abort handler, and move on

from a quick seach that I've seen, these seems to be using for the handler precondition check e.g. ensure the target entity exist

---

> If there is a kind of /dev/null queuetype, then yeah that wouldbe good. I was planning on adding a scheduled job that empties the dead queue, but I haven't gotten around to doing that, yet...

funnily enough, we might already be doing that unintentionally

last time I checked, it appears that if the message got sent to failed transport and retry again by workers consuming from the failed transport and still failed, it would not making to the dead queue, and simply went missing (which goes to my above point on different personal setup because if I screw something up that make the handlers falied I have no way to reliably see why that happen other than going to the logs which is just messy)

I don't have the specifics in hand (it's somewhere in symfony issues), but it appears that somewhere during all these symfony upgrade and changes, the messages retry behavior was reverted to what it was before: if a message was retried enough times to be sent to failure transport, it those message were to be retry and failed _again_, the message will simply be lost/shredded rather than go back to fail transport (because it was already sent there once)
apparently allowing message retry forever may cause the worker that consumes failed queue to simply stuck retrying failed messages and more or less spinlocked

PR #1062, opened 2024-08-24
Update theme-dark.scss

TheVillageGuy:
Alert-dangers are given with 0.15 alpha in some themes. With custom background they are hard to read. They can be changed with custom CSS. But for ignorant magazine admins/moderators (looks in mirror) this is easy to overlook while these messages should NOT be easy to overlook.

image

Besides that, for being alert-dangers they should be prominent. The transparency look is nice, but this is much better for catching the users' attention

image

Also, normal notifications have no alpha either, so might as well follow suit

image

### DIFF
--- a/assets/styles/mixins/theme-dark.scss
+++ b/assets/styles/mixins/theme-dark.scss
@@ -109,8 +109,8 @@
    --kbin-boosted-color: var(--kbin-upvoted-color);
 
   // alerts
-  --kbin-alert-info-bg: rgba(153,116,4,0.9);
-  --kbin-alert-info-border: 1px solid rgba(153,116,4, 1);
+  --kbin-alert-info-bg: rgba(153,116,4,0.15);
+  --kbin-alert-info-border: 1px solid rgba(153,116,4, 0.04);
   --kbin-alert-info-text-color: #997404;
   --kbin-alert-info-link-color: #997404;
 
--- a/assets/styles/mixins/theme-dark.scss
+++ b/assets/styles/mixins/theme-dark.scss
@@ -109,13 +109,13 @@
    --kbin-boosted-color: var(--kbin-upvoted-color);
 
   // alerts
-  --kbin-alert-info-bg: rgba(153,116,4,1);
-  --kbin-alert-info-border: 1px solid rgba(153,116,4,0.4);
+  --kbin-alert-info-bg: rgba(153,116,4,0.9);
+  --kbin-alert-info-border: 1px solid rgba(153,116,4, 1);
   --kbin-alert-info-text-color: #997404;
   --kbin-alert-info-link-color: #997404;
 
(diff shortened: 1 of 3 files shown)


assets/styles/mixins/theme-dark.scss
```diff
@@ -109,7 +109,7 @@
    --kbin-boosted-color: var(--kbin-upvoted-color);
 
   // alerts
-  --kbin-alert-info-bg: rgba(153,116,4,0.15);
+  --kbin-alert-info-bg: rgba(153,116,4,1);
```
melroy89, line 112:
```suggestion
  --kbin-alert-info-bg: rgba(153,116,4,0.9);
```

What about 0.9? Still looks good I think. And just ever so slightly transparent.

TheVillageGuy: I am ok with that, but it does not follow suit with the green alert not being transparent?

BentiGorlich:
I think you gotta check the alert danger for contrast ratios. I suspect that the contrast is not high enough and I have to say I find it very straining on my eyes. Please keep accessibility in mind. https://webaim.org/resources/contrastchecker/?fcolor=DC2F3F&bcolor=FFFFFF this one fails in one of the tests.

But I actually only see changes to the warning alert. Is that on purpose?

TheVillageGuy:
> I think you gotta check the alert danger for contrast ratios. I suspect that the contrast is not high enough and I have to say I find it very straining on my eyes. Please keep accessibility in mind. https://webaim.org/resources/contrastchecker/?fcolor=DC2F3F&bcolor=FFFFFF this one fails in one of the tests.

I'll look into that. What would be an acceptable ratio? Not sure if it matters but you did them the wrong way around (it seems the permalink switches them) this is it: https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=DC2F3F and it passes 5/6 tests

> But I actually only see changes to the warning alert. Is that on purpose?

The warning alert is the one that was causing problems, shall I do a full overhaul on all alerts?

BentiGorlich:
> What would be an acceptable ratio?

All tests should pass, though I do not know at which ratio they do

> Not sure if it matters but you did them the wrong way around

You're absolutely right.

> The warning alert is the one that was causing problems, shall I do a full overhaul on all alerts?

The warning dialog is not mentioned in the description, that is why I am asking. You show the danger and the success dialog bug not the warning/info alert

TheVillageGuy:
This is the current situation:

image

Only alert-warnings are modified;
Background = #AB1C28 /  0.9 alpha with color contrast verified and 1 alpha border

melroy89 (approved):

melroy89:
> All tests should pass, though I do not know at which ratio they do

@TheVillageGuy  does the tests now pass regarding the contrast ratio?

PR #1066, opened 2024-08-27
Fix messenger deadlocks caused by cache locks

BentiGorlich:
This PR aims to solve a severe problem which crashes rabbit and in the end overwhelms PostgreSQL or PHP-Fpm.

### The Problem

The problem seems to be a deadlock situation of the messenger processes. With v1.7.0 (in #936) came a change that allowed the messengers to actually run in parallel and not get immidiately locked out of the db (because of table locks). Now we have an unfortunate deadlock situation where one messenger currently has the db table locks, but another one has a lock on a cache item, which the first is waiting on, though this cache item cannot be compiled because the messenger having the lock on the cache item is waiting on the messenger process which has the table locks, a classic dead lock.

### The Consequences

Now one might ask how that overwhelms postgresql and / or php-fpm. The explanation is a bit complicated: since messengers are deadlocking each other messages don't get acknowledged to the transport - in our case this is RabbitMQ. RabbitMQ is not a fan of messages not getting acknowledged and therefore has a timeout parameter for how long it waits for an acknowledgement if that timeout limit is exceeded RabbitMQ shuts down that queue. The default for this timeout parameter is 30 minutes.

Now that RabbitMQ has shut down the queue everything trying to access this queue (be it dispatching or consuming a message) will wait indefinitely. That leads to a lot of hanging requests, each consuming a php-fpm worker and a connection to the db. So you only notice this problem on the frontend once the server either ran out of php-fpm workers or db connections.

### Restore The Damage

To get the server back up and running after users reported the crash: restart RabbitMQ, the messengers and php-fpm:
```bash
systemctl restart rabbitmq-server php8.3-fpm && supervisorctl restart messenger:*
```

If you noticed earlier that the messengers where hanging by monitoring rabbitmq and the processing rate (which plummits to 0/s) then you only have to restart the messenger processes

### The Solution

The solution is pretty simple: do not use locks for calculating cache items. Since we did not have any lock logic on top of our cache calls this PR just uses a different cache style (the PSR6 style) which the symfony cache system also supports, but does not recommend. That way the messengers cannot end up in a deadlock situation because of the cache locks.

### DIFF
--- a/src/Service/ActivityPub/ApHttpClient.php
+++ b/src/Service/ActivityPub/ApHttpClient.php
@@ -18,6 +18,7 @@ use App\Service\ProjectInfoService;
 use JetBrains\PhpStorm\ArrayShape;
 use Psr\Cache\InvalidArgumentException;
 use Psr\Log\LoggerInterface;
+use Symfony\Component\Cache\CacheItem;
 use Symfony\Component\HttpClient\CurlHttpClient;
 use Symfony\Contracts\Cache\CacheInterface;
 use Symfony\Contracts\Cache\ItemInterface;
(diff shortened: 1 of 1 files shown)


src/Service/ActivityPub/ApHttpClient.php
```diff
@@ -168,64 +199,72 @@ private function getActorCacheKey(string $apProfileId): string
...
+                if ($user = $this->userRepository->findOneByApProfileId($apProfileId)) {
+                    $user->apDeletedAt = new \DateTime();
+                    $this->userRepository->save($user, true);
+                }
+                if ($magazine = $this->magazineRepository->findOneByApProfileId($apProfileId)) {
+                    $magazine->apDeletedAt = new \DateTime();
+                    $this->magazineRepository->save($magazine, true);
+                }
+            }
+        } catch (\Exception|TransportExceptionInterface $e) {
```
melroy89, line 245: Interesting I would though that `TransportExceptionInterface` was extending from Exception, but it isn't in Symfony. Instead TransportExceptionInterface-> ExceptionInterface -> Throwable.

I think this is also a bad design pattern by Symfony IMO.

melroy89 (approved):
Thanks!

It does also runs fine on my instance now. No dead-locks anymore.

melroy89:
Most important line from the [documentation](https://symfony.com/doc/current/components/cache/cache_pools.html) is indeed: "Out of the box, using this interface provides stampede protection via locking and early expiration." Which is causing all the problems.

BentiGorlich:
Happy to hear that your problems are also fixed with this PR :)

PR #1070, opened 2024-08-30
move private vars out of some services/handlers

asdfzdfj:
refactored some services/handlers that pass extra data by means of private variables to pass them all as function arguments

functionality should remain the same, if anything it'd be more troubling if something's changed

in an ideal scenario this should not be a problem, but after what's happened with embed thumbnails, the assumption that all services/handlers will starts with a clean slate is no longer reliable

this is just from what I could trivially find at the moment

### DIFF
--- a/src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
+++ b/src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
@@ -32,9 +32,6 @@ use Symfony\Component\Messenger\MessageBusInterface;
 #[AsMessageHandler]
 class CreateHandler extends MbinMessageHandler
 {
-    private array $object;
-    private bool $stickyIt;
-
     public function __construct(
         private readonly EntityManagerInterface $entityManager,
         private readonly Note $note,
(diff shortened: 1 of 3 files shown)


melroy89 (approved):
I think this is indeed a good thing. LGTM.

@BentiGorlich Do you agree?

BentiGorlich:
I have no objections about the code changes, though I don't know whether this is truely necessary, since the properties are overwritten in each construct call. Maybe @asdfzdfj can explain why this could be a problem?

asdfzdfj:
perhaps you're right that this patch isn't strictly necessary, but here's my thoughts:

- it makes the function a bit more clear on what it needs in order to work, guess I like them more functional-y?
- try to play safe and avoiding another #262 situation, as one part of it is that some data were being written to class properties and _somehow_ it's being persisted to between handler calls even though that should not be the case  
  one possible theory I've heard is that under normal circumstances, php state variables and stuff only exists as long as the request lifetime, so everything gets initialized at the start and cleaned when the end of request just as it should/we expect, but inside a message worker, this "lifetime" seems to be far longer than individual handler calls and so it's possible to unintentionally persist some data across handler calls and have the later calls pulling out wrong data, just as it has happened in the embed thumbnail machinery (and at a glance the problematic part back then looks alright too, only for it to turn out that's not the case)

BentiGorlich (approved):

PR #1077, opened 2024-09-03
Fix the regex, so FF is no longer complaining

BentiGorlich:
Firefox complained about an improper regex, because the `-` was not escaped


melroy89 (approved):
Validated and tested, this solves the issue in Floorp at least (and most likely FF).

Issue #1071, opened 2024-08-31
Notification page: Neither the property "entry" nor one of the methods "entry()" in dev mode


melroy89:
**Describe the bug**

Frontend crash in Twig template at the notification page.

**On which Mbin instance did you find the bug?**
kbin.melroy.org

**Which Mbin version was running on the instance?**
1.7.1

**To Reproduce**
1. Run in dev mode
2. Go to Notifications

Error message:

```sh
[2024-08-31T14:15:29.877481+02:00] request.CRITICAL: Uncaught PHP Exception Twig\Error\RuntimeError: "Neither the property "entry" nor one of the methods "entry()", "getentry()"/"isentry()"/"hasentry()" or "__call()" exist and have public access in class "App\Entity\EntryCommentReport"." at _blocks.html.twig line 104 {"exception":"[object] (Twig\\Error\\RuntimeError(code: 0): Neither the property \"entry\" nor one of the methods \"entry()\", \"getentry()\"/\"isentry()\"/\"hasentry()\" or \"__call()\" exist and have public access in class \"App\\Entity\\EntryCommentReport\". at /var/www/kbin.melroy.org/html/templates/notifications/_blocks.html.twig:104)
[stacktrace]
#0 /var/www/kbin.melroy.org/html/var/cache/dev/twig/80/808539e6f82f6bfc653a06193bce03f6.php(873): Twig\\Extension\\CoreExtension::getAttribute()
#1 /var/www/kbin.melroy.org/html/vendor/twig/twig/src/Template.php(430): __TwigTemplate_788c432cb478dc9a06bcf5acf7ea54d7->block_reportlink()
#2 /var/www/kbin.melroy.org/html/var/cache/dev/twig/80/808539e6f82f6bfc653a06193bce03f6.php(966): Twig\\Template->yieldBlock()
#3 /var/www/kbin.melroy.org/html/vendor/twig/twig/src/Template.php(430): __TwigTemplate_788c432cb478dc9a06bcf5acf7ea54d7->block_report_created_notification()
#4 /var/www/kbin.melroy.org/html/var/cache/dev/twig/b0/b0f71fe94567e6b7e0f3673bbd68ed95.php(268): Twig\\Template->yieldBlock()
#5 /var/www/kbin.melroy.org/html/vendor/twig/twig/src/Template.php(430): __TwigTemplate_1043b0e468c799d438c19fd907c5fc8a->block_body()
[...]

PR #1080, opened 2024-09-03
Use is defined on twig report notification objects

melroy89:
Use `is defined`, to be sure the key exists on the object. 

Fixes: #1071

### DIFF
--- a/templates/notifications/_blocks.html.twig
+++ b/templates/notifications/_blocks.html.twig
@@ -106,7 +106,7 @@
         <a href="{{ path('entry_single', { 'magazine_name': entry.magazine.name, 'entry_id': entry.id, 'slug': entry.slug }) }}">
             {{ entry.title }}
         </a>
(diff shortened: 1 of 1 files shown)


templates/notifications/_blocks.html.twig
```diff
@@ -101,22 +101,22 @@
 {% endblock magazine_ban_notification %}
 
 {% block reportlink %}
-    {% if notification.report.entry is not same as null %}
+    {% if notification.report.entry is defined and attribute(notification.report, 'entry') is not null %}
```
BentiGorlich, line 104: Why the attribute call and not just
```suggestion
    {% if notification.report.entry is defined and notification.report.entry is not same as null %}
```

melroy89: I personally like the flexibility of attribute, but I using dot notation again.

BentiGorlich: I was not aware of that notation and it looked funky to me :D

PR #1085, opened 2024-09-04
New docker setup

melroy89:
Follow-up from: https://github.com/MbinOrg/mbin/pull/1021

- [ ] Testing
- [ ] Introduce an internal network (do not expose ports like PostgreSQL to the host) for services that can stay internal
- [ ] ...


.env.example_docker
```diff
@@ -107,6 +111,10 @@ EXIF_EXIFTOOL_PATH=
 # max execution time for exiftool in seconds, defaults to 10 seconds
 EXIF_EXIFTOOL_TIMEOUT=10
 
+###> caddy ###
+PHP_FASTCGI_HOST=php:9000
```
melroy89, line 115: This shouldn't be here.

.env.example_docker
```diff
@@ -31,6 +31,10 @@ KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 MBIN_DEFAULT_THEME=default
+MBIN_HOME=/var/www/mbin
```
melroy89, line 34: These en vars shouldn't be here as well ideally put these in the compose override yaml file.

jwr1:
Hopefully it's fine if I work on this?

jwr1:
I don't think there's a point in having the extra `MBIN_HOME`, `MBIN_SRC`, `MBIN_USER`, or `MBIN_GROUP` environment variables. It just adds unneeded complexity.

melroy89:
I just took this pr over from somebody else who left github. I believe this pr is also not great. It's plain wrong indeed.

jwr1:
Going to start from scratch...

PR #1086, opened 2024-09-04
Add trusted proxies to docs

melroy89:
See title.


BentiGorlich (commented):
Can you add some information regarding docker? Because with the www container it is a bit special isnt it?

melroy89:
> Can you add some information regarding docker? Because with the www container it is a bit special isnt it?

This is a complex question indeed.

Since we are talking about Docker, I believe that the `trusted_proxy` env var in Symfony should be: `172.16.0.0/12` (or any static IP you might give Caddy service within the docker external network). Unless you change the default IP ranges in your `/etc/docker/daemon.json` config (`default-address-pools` config and `bip` config).

We already have `php_fastcgi {$PHP_FASTCGI_HOST}` in place at: https://github.com/MbinOrg/mbin/blob/main/docker/caddy/Caddyfile. I believe that should pass the host correctly to Symfony?

I'm personally a bit confused whether we might even need set `trusted_proxies` within the Caddy config. https://caddyserver.com/docs/caddyfile/directives/reverse_proxy#defaults.

I must say I'm not running the Docker setup at the moment, so I also unable to validate these configs in a real-world scenario to be fully honest. 

Finally, we also have this rework in place, which might change everything again: https://github.com/MbinOrg/mbin/pull/1085

BentiGorlich:
So maybe a follow up issue then?

melroy89:
> So maybe a follow up issue then?

That would be easier yes.

melroy89:
Or we just add to use `172.16.0.0/12` as trusted_proxy under Docker? I fingers crossed it works for most people...

melroy89:
> Or we just add to use `172.16.0.0/12` as trusted_proxy under Docker? And fingers-crossed it works for most people...

but then again, it doesn't fit the Nginx section here. And in the other PR we are also changing the docs for Docker.

BentiGorlich (approved):

PR #1095, opened 2024-09-08
Introducing bookmarks

### LINKED ISSUE #46: Bookmark threads/posts
Similar how you can also bookmark a post in Mastodon.

Then an additional menu item in your profile drop-down menu called 'Bookmarks' need to be added. That page shall display all your saved bookmarks.

This is a feature request that is still very much much requested by the community and many of the users that are using this platform.

BentiGorlich:
In this massive commit bookmarks are added. A user can create multiple bookmarking lists and add entries, posts and comments to them. They are all shown in one view when looking at the list.

# UI

## Add to default button

Entries, posts and comments now have a button to add a bookmark to the default bookmark list:

grafik

## Add to list button

They all have the option to add them to a specific bookmark list in the dropdown menu:

grafik

## Get to the list view

To get to the new view there is a new button in the profile dropdown:

grafik

## The new list view

There is a new view for managing ones bookmark lists:

grafik

## Looking at the content in a list:

When looking at the content of a bookmark list all types can be viewed simultaneously in the same list:

grafik

# API

I already added the API, but it might not work correctly, as I didn't test it (might be possible to do so without an app to test it via test libs, but I didn't bother...)

There are some new scopes: `user:bookmark`, `user:bookmark:add`, `user:bookmark:remove`, `user:bookmark:list`, `user:bookmark:list:read`, `user:bookmark:list:edit` and `user:bookmark:list:delete`. I do not know if I forgot a few places for them to be added though

### Extras

One point that is easy to miss: the caching of the twig components `EntryCommentsNestedComponent` and `PostCommentsNestedComponent` was removed (by not using the caching template). It is not really that useful (cache gets cleared as soon as anyone votes on it) and was getting in the way of rendering. That may later be reverted

---

Closes #46


melroy89:
Ow.. it's only when you are logged in?..

melroy89:
Removing the item from the bookmark will not remove the item from the list:

image

Only after I refreshed the page manually, the item is removed.

BentiGorlich:
I fixed the minor points. 
Just to summarize yours:
1. when adding something to a specific list via dropdown the "add to default" icon will not be changed to active
2. When viewing a list and removing an item from the list it is not removed until a page refresh

Answers:
1. could be done and wouldn't be to hard, though we do it via replacing html of some anchors we have and with that it is way harder to do it, since they are 2 separate controls
2. that could of course be done, but it is not something I'd like it to do actually. What if you want to just move it to another list and because you didn't knew mbin behaved like this you remove it first and then its gone...

melroy89:
> I fixed the minor points. Just to summarize yours:
> 
>     1. when adding something to a specific list via dropdown the "add to default" icon will not be changed to active
> 
>     2. When viewing a list and removing an item from the list it is not removed until a page refresh
> 
> 
> Answers:
> 
>     1. could be done and wouldn't be to hard, though we do it via replacing html of some anchors we have and with that it is way harder to do it, since they are 2 separate controls
> 
>     2. that could of course be done, but it is not something I'd like it to do actually. What if you want to just move it to another list and because you didn't knew mbin behaved like this you remove it first and then its gone...

1. You can use JS to trigger the bookmark icon on top-level thread/comment to become active. I believe that is possible.
2. I mean when you click the "Remove from all bookmark" button (which is the bookmark icon saying when active), the bookmark will be removed. So it's not moving to another list. In this case you can also use JS to just hide the thread/comment.

BentiGorlich:
1. but I'd have to change the text and the link it goes to as well. So I'd have to implement a way to fetch this component from the server. My point is that it is not just the icon :D
2. Ok yeah that makes sense, but only on the "remove all" button

BentiGorlich:
I addressed 1. but I am hesitent to implement 2. are you fine with letting the users decide whether they want 2. when this is released @melroy89 ?

jwr1:
I think I'm with @BentiGorlich on this. I just checked, and even GitHub stars behave the same way; if you go to the star list in your profile and unstar one of the repositories, it still stays there until you actually reload the page (which works out in case you accidentally remove one of them).

melroy89:
> I think I'm with @BentiGorlich on this. I just checked, and even GitHub stars behave the same way; if you go to the star list in your profile and unstar one of the repositories, it still stays there until you actually reload the page (which works out in case you accidentally remove one of them).

Aha. I see. To be honest don't compare UX with Github, since if some interface can be improved on planet earth it would be all Microsoft products including github. ☺️

melroy89:
> are you fine with letting the users decide whether they want 2. when this is released @melroy89 ?

Sure. It's a minor UX issue I think. Mainly confusing I believe.

melroy89 (approved):
Works!

PR #1098, opened 2024-09-09
Fix update actor failing because of a serializer error

BentiGorlich:
When there was a serializer error when updating the remote instance this error would cause the update actor to fail. We now surround it in a try/catch block so this doesn't happen anymore

### DIFF
--- a/src/Service/RemoteInstanceManager.php
+++ b/src/Service/RemoteInstanceManager.php
@@ -57,11 +57,18 @@ class RemoteInstanceManager
 
             $nodeInfoRaw = $this->client->fetchInstanceNodeInfo($linkToUse->href, false);
             $this->logger->debug('got raw nodeinfo for url {url}: {raw}', ['raw' => $nodeInfoRaw, 'url' => $linkToUse]);
-            /** @var NodeInfo $nodeInfo */
-            $nodeInfo = $serializer->deserialize($nodeInfoRaw, NodeInfo::class, 'json');
-
-            $instance->software = $nodeInfo?->software?->name;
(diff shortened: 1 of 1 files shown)


src/Service/RemoteInstanceManager.php
```diff
@@ -57,11 +57,18 @@ public function updateInstance(Instance $instance, bool $force = false): bool
...
+                /** @var NodeInfo $nodeInfo */
+                $nodeInfo = $serializer->deserialize($nodeInfoRaw, NodeInfo::class, 'json');
+                $instance->software = $nodeInfo?->software?->name;
+                $instance->version = $nodeInfo?->software?->version;
+            } catch (\Error|\Exception $e) {
+                $this->logger->warning('There as an exception decoding the nodeinfo from {url}: {e} - {m}', [
+                    'url' => $instance->domain,
+                    'e' => \get_class($e),
+                    'm' => $e->getMessage(),
+                ]);
```
melroy89, line 70: no need to throw up an error?

melroy89 (approved):

src/Service/RemoteInstanceManager.php
BentiGorlich: nono if I throw that up then nothing has changed. The error can be catched here as this doesn't hinder the overall update process

melroy89: And actually catch the error higher in the chain..

BentiGorlich: I think this is the right place. The instance should be created, but the name and the version are just not essential...

PR #1102, opened 2024-09-09
Check if the openssl_sign went OK

melroy89:
In case `openssl_sign` fails the `$signature`  will be null. Resulting into errors like this:

```
{"message":"Warning: openssl_sign(): Supplied key param cannot be coerced into a private key","context":{"exception":{"class":"ErrorException","message":"Warning: openssl_sign(): Supplied key param cannot be coerced into a private key","code":0,"file":"/var/www/kbin.melroy.org/html/src/Service/ActivityPub/ApHttpClient.php:504"}},"level":400,"level_name":"ERROR","channel":"php","datetime":"2024-09-09T21:33:35.100575+02:00","extra":{}}
{"message":"Error thrown while handling message App\\Message\\ActivityPub\\Outbox\\DeliverMessage. Sending for retry #2 using 3600000 ms delay. Error: \"Handling \"App\\Message\\ActivityPub\\Outbox\\DeliverMessage\" failed: base64_encode(): Argument #1 ($string) must be of type string, null given\"","context":{"class":"App\\Message\\ActivityPub\\Outbox\\DeliverMessage","message_id":3420233,"retryCount":2,"delay":3600000,"error":"Handling \"App\\Message\\ActivityPub\\Outbox\\DeliverMessage\" failed: base64_encode(): Argument #1 ($string) must be of type string, null given","exception":{"class":"Symfony\\Component\\Messenger\\Exception\\HandlerFailedException","message":"Handling \"App\\Message\\ActivityPub\\Outbox\\DeliverMessage\" failed: base64_encode(): Argument #1 ($string) must be of type string, null given","code":0,"file":"/var/www/kbin.melroy.org/html/vendor/symfony/messenger/Middleware/HandleMessageMiddleware.php:124","previous":{"class":"TypeError","message":"base64_encode(): Argument #1 ($string) must be of type string, null given","code":0,"file":"/var/www/kbin.melroy.org/html/src/Service/ActivityPub/ApHttpClient.php:505"}}},"level":300,"level_name":"WARNING","channel":"messenger","datetime":"2024-09-09T21:33:35.100878+02:00","extra":{}}
```

We can better check on the return value of `openssl_sign` to validate if the signature actually succeeded or not.

If not, we now log a better error message. And we leave out the `Signature` header for now. And we throw an error as well.

---

Also we now free the memory of the `$key` using the appropriate `openssl_free_key` method. 

---

I also changed `InvalidApPostException` to `InvalidApGetException` only with the GET request, since well.. it's a get request..


melroy89:
The qrcode upgrade seems to be the only issue thus far. Not showing me the image and the logging gave me: 

```
{"message":"Uncaught PHP Exception Error: \"Class \"Endroid\\QrCode\\ErrorCorrectionLevel\\ErrorCorrectionLevelHigh\" not found\" at User2FAController.php line 133","context":{"exception":{"class":"Error","message":"Class \"Endroid\\QrCode\\ErrorCorrectionLevel\\ErrorCorrectionLevelHigh\" not found","code":0,"file":"/var/www/kbin.melroy.org/html/src/Controller/User/Profile/User2FAController.php:133"}},"level":500,"level_name":"CRITICAL","channel":"request","datetime":"2024-09-09T21:56:40.042112+02:00","extra":{}}
```

src/Service/ActivityPub/ApHttpClient.php
```diff
@@ -529,16 +529,26 @@ private function getHeaders(string $url, User|Magazine $actor, ?array $body = nu
...
+            $signature = base64_encode($signature);
+            $keyId = $actor instanceof User
+                ? $this->personFactory->getActivityPubId($actor).'#main-key'
+                : $this->groupFactory->getActivityPubId($actor).'#main-key';
+            $signatureHeader = 'keyId="'.$keyId.'",headers="'.$signedHeaders.'",algorithm="rsa-sha256",signature="'.$signature.'"';
+        } else {
+            $this->logger->error('Failed to sign headers for {url}: {headers}', [
+                'url' => $url,
+                'headers' => $headers,
+            ]);
```
BentiGorlich, line 546: Should you return here (and at 2nd comment) or throw an exception so this won't get processed any further?

src/Service/ActivityPub/ApHttpClient.php
```diff
@@ -554,11 +564,23 @@ private function getInstanceHeaders(string $url, ?array $body = null, string $me
...
+        openssl_free_key($key);
+        $signatureHeader = null;
+        if ($success_sign) {
+            $signature = base64_encode($signature);
+            $signatureHeader = 'keyId="'.$keyId.'",headers="'.$signedHeaders.'",algorithm="rsa-sha256",signature="'.$signature.'"';
+        } else {
+            $this->logger->error('Failed to sign headers for {url}: {headers}', [
+                'url' => $url,
+                'headers' => $headers,
+            ]);
```
BentiGorlich, line 578: 2nd

src/Service/ActivityPub/ApHttpClient.php
melroy89: Yea so what's the big question as stated in my pr description. What should we do? 🤔

It's ssl signing that important to fail or not? What are the consequences of both solutions?

BentiGorlich: This methods should never fail. Maybe add an additional check at the top of the method whether the actor is a local one. If the actor is local then there is literally no reason for not having a private key...

melroy89: OK. I will just throw an error.

src/Service/ActivityPub/ApHttpClient.php
```diff
@@ -318,7 +318,7 @@ private function getCollectionObjectImpl(string $apAddress): ?string
             // Accepted status code are 2xx or 410 (used Tombstone types)
             if (!str_starts_with((string) $statusCode, '2') && 410 !== $statusCode) {
                 // Do NOT include the response content in the error message, this will be often a full HTML page
-                throw new InvalidApPostException("Invalid status code while getting: $apAddress, status code: $statusCode");
+                throw new InvalidApGetException("Invalid status code while getting: $apAddress, status code: $statusCode");
```
BentiGorlich, line 321: I think you should keep the old exception. There are a lot of places that catch this exception and a few of them count on it. So for the ease of this PR I'd say keep the old one

BentiGorlich (changes requested):

src/Service/ActivityPub/ApHttpClient.php
melroy89:
```suggestion
                throw new InvalidApPostException("Invalid status code while getting: $apAddress, status code: $statusCode");
```

BentiGorlich (approved):

PR #1108, opened 2024-09-10
Add spoiler to markdown editor

melroy89:
Add spoiler insert button to markdown editor.

image

Demo: https://kbin.melroy.org/m/testing/t/452875/Spoiler-testing

### DIFF
--- a/assets/controllers/markdown_toolbar_controller.js
+++ b/assets/controllers/markdown_toolbar_controller.js
@@ -25,7 +25,7 @@ export default class extends Controller {
         }
 
         const spoiler = `
-::: spoiler spoiler
+::: spoiler spoiler-title
 ${spoilerBody}
 :::`;
 
--- a/assets/controllers/markdown_toolbar_controller.js
+++ b/assets/controllers/markdown_toolbar_controller.js
@@ -1,0 +1,39 @@
+// SPDX-FileCopyrightText: 2023-2024 /kbin & Mbin contributors
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+import { Controller } from '@hotwired/stimulus';
+
+/* stimulusFetch: 'lazy' */
+export default class extends Controller {
+    addSpoiler(event) {
+        event.preventDefault();
+
+        const input = document.getElementById(this.element.getAttribute('for'));
+        let spoilerBody = '_____';
+        let contentAfterCursor;
+
+        const start = input.selectionStart;
+        const end = input.selectionEnd;
+
+        const contentBeforeCursor = input.value.substring(0, start);
+        if (start === end) {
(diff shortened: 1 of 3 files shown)


assets/controllers/markdown_toolbar_controller.js
```diff
@@ -0,0 +1,39 @@
...
+        const contentBeforeCursor = input.value.substring(0, start);
+        if (start === end) {
+            contentAfterCursor = input.value.substring(start);
+        } else {
+            contentAfterCursor = input.value.substring(end);
+            spoilerBody = input.value.substring(start, end);
+        }
+
+        const spoiler = `
+::: spoiler spoiler
```
BentiGorlich, line 28:
```suggestion
::: spoiler spoiler title
```
I think this would encourage the user to replace this title. I was confused at first as to where the title comes from

assets/controllers/markdown_toolbar_controller.js
```diff
@@ -0,0 +1,39 @@
...
+
+import { Controller } from '@hotwired/stimulus';
+
+/* stimulusFetch: 'lazy' */
+export default class extends Controller {
+    addSpoiler(event) {
+        event.preventDefault();
+
+        const input = document.getElementById(this.element.getAttribute('for'));
+        let spoilerBody = '_____';
```
BentiGorlich, line 13:
```suggestion
        let spoilerBody = 'spoiler body';
```
So users know that they should replace this text

BentiGorlich:
Very cool addition. I always forget how to make a spoiler :D

assets/controllers/markdown_toolbar_controller.js
melroy89: this might still be confusing maybe, since now people might remove the whole line.. Including the first "spoiler" text,  I dunno. The current `spoiler spoiler`  seem to be a safe default.

assets/controllers/markdown_toolbar_controller.js
melroy89: Yea ok, seems to be redundant and definitely confusing. I agree.

melroy89:
> Very cool addition. I always forget how to make a spoiler :D

I may or may not looked at kbin.

assets/controllers/markdown_toolbar_controller.js
melroy89: Or we combine spoiler title together or something like `spoiler-title`

BentiGorlich: `spoiler-title` would good :+1:

melroy89:
New layout also still works luckily:

https://kbin.melroy.org/m/testing/t/453925/Second-spoiler-markdown-test

PR #1110, opened 2024-09-11
Fix the downvote error

BentiGorlich:
when we added the downvotes we overlooked a wrong condition leading to an error. We now just check the current mode and if downvotes are disabled and the user tries to downvote we just throw an exception


melroy89 (approved):
🥳

PR #1111, opened 2024-09-11
Clean up push controller js even more

melroy89:
I forgot to remove some additional console.log's. I also added a missing catch.

Also there seems to be a then, without doing anything with the data. Meaning this then might be not needed / can be removed. Can you confirm this?

### DIFF
--- a/assets/controllers/push_controller.js
+++ b/assets/controllers/push_controller.js
@@ -74,7 +74,6 @@ export default class extends Controller {
         this.askPermission()
             .then(() => window.navigator.serviceWorker.getRegistration())
             .then(registration => {
-                console.log("got service worker registration:", registration)
                 const subscribeOptions = {
                     userVisibleOnly: true,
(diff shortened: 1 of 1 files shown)


assets/controllers/push_controller.js
```diff
@@ -106,7 +103,7 @@ export default class extends Controller {
                 return response.json()
             })
             .then(data => {
-
+                // Nothing?
```
BentiGorlich, line 106: I think you can just leave this blank :D

BentiGorlich (commented):
apart from this one thing LGTM

assets/controllers/push_controller.js
melroy89: If you re not gonna use it, we can just remove this whole then.

melroy89: Removed the then. It can always be added, if we really want or need to.

BentiGorlich (approved):

PR #1116, opened 2024-09-12
Disable badges for now

melroy89:
- Hide the badges field (it is not implemented and not in scope)
- Comment out the fields in the PHP form builder as well
- Add missing `header` element on the headers that where missing them. So it's constant with the rest of the existing code.


BentiGorlich (changes requested):
You forgot the `EntryArticleType`:

grafik

Aside from that everything works fine :)

melroy89:
> You forgot the `EntryArticleType`:
> 
> grafik
> 
> Aside from that everything works fine :)

fixed.

BentiGorlich (approved):

PR #1117, opened 2024-09-12
Fix null error when an instance does not have a nodeinfo endpoint

BentiGorlich:
This was discovered when someone tried to pull in threads.net users


melroy89 (approved):
yup

PR #1118, opened 2024-09-12
Null error on reported post comment notifications

BentiGorlich:


melroy89 (approved):
That seems like to be a copy / paste kind of issue ;P

PR #1122, opened 2024-09-12
Fix 9th miscellaneous bug

BentiGorlich:


src/Service/ActivityPubManager.php
```diff
@@ -284,10 +284,14 @@ public function buildHandle(string $id): string
         $port = !\is_null(parse_url($id, PHP_URL_PORT))
             ? ':'.parse_url($id, PHP_URL_PORT)
             : '';
+        $apObj = $this->apHttpClient->getActorObject($id);
+        if (!isset($apObj['preferredUsername'])) {
+            throw new \InvalidArgumentException("$id does not supply a valid user object");
```
melroy89, line 289:
```suggestion
            throw new \InvalidArgumentException("Webfinger $id does not supply a valid user object");
```

melroy89: would this help clarify it even more..?

melroy89 (approved):

PR #1123, opened 2024-09-12
Fix 7th miscellaneous bug

BentiGorlich:


melroy89 (approved):
what does it mean if we can't retrieve the $entity here, could we still proceed further with the data we might have or not? Just brainstorming here.. Could we maybe parse the data better we get? 

Or is the data is really bad/wrong..

BentiGorlich:
the `entity` is the parent of the thing we need. If we do net get the entity, then we do not have the parent of the object we're trying to retrieve. E.g.: you want to create a comment of a comment on an entity. You cannot create the sub comment without having both the comment and the entry

PR #1124, opened 2024-09-12
Fix 2nd miscellaneous bug

BentiGorlich:
When creating a page we didn't look whether this object existed before, while we do it with note objects. So just copy over the code that checks whether a note already exists and we should be golden


melroy89 (approved):
yup make sense do first (get) check, before doing a create.

Issue #1119, opened 2024-09-12
Miscellaneous error/warning log items


melroy89:
**Describe the bug**

I see a long list of errors and warnings in the prod log.

We really need to clean-up some of those and try to fix the PHP warning or PHP or Twig errors we see...

I could make 20+ GitHub Issues out of this single ticket, but that won't help us. And will only cause more overhead. See the list of warnings/errors below under "List of issues".

**On which Mbin instance did you find the bug?**
kbin.melroy.org

**Which Mbin version was running on the instance?**
1.7.1

**To Reproduce**
1. Check-out the log: `tail -f var/log/prod-xx-x-x-x.log`
2. And follow it for a while...

**Expected behavior**

Minimal amount of errors and warnings in the production log. Only some error messages make sense if there is a real error and we still want to keep reporting it to admin of the instance.

Let's just see how much we can resolve in the first 1.7.2 milestone..

**List of issues**

Directly from the production log output:

- [x] the first - missing null check, see PR: https://github.com/MbinOrg/mbin/pull/1129

```json
{"message":"Uncaught PHP Exception Twig\\Error\\RuntimeError: \"Error rendering \"entry_comment\" component: Expected argument of type \"App\\Entity\\EntryComment\", \"null\" given at property path \"comment\".\" at view.html.twig line 49","context":{"exception":{"class":"Twig\\Error\\RuntimeError","message":"Error rendering \"entry_comment\" component: Expected argument of type \"App\\Entity\\EntryComment\", \"null\" given at property path \"comment\".","code":0,"file":"/var/www/kbin.melroy.org/html/templates/entry/comment/view.html.twig:49","previous":{"class":"Symfony\\Component\\PropertyAccess\\Exception\\InvalidTypeException","message":"Expected argument of type \"App\\Entity\\EntryComment\", \"null\" given at property path \"comment\".","code":0,"file":"/var/www/kbin.melroy.org/html/vendor/symfony/property-access/PropertyAccessor.php:203","previous":{"class":"TypeError","message":"Cannot assign null to property App\\Twig\\Components\\EntryCommentComponent::$comment of type App\\Entity\\EntryComment","code":0,"file":"/var/www/kbin.melroy.org/html/vendor/symfony/property-access/PropertyAccessor.php:520"}}}},"level":500,"level_name":"CRITICAL","channel":"request","datetime":"2024-09-12T12:18:16.360410+02:00","extra":{}}
```

---

- [x]  the second #1124

```json
[...]

PR #1128, opened 2024-09-13
Check on empty/null filePath image

melroy89:
Check if filePath is set, before continuing trying to resolve the image by LIIP.

Allowing to run: `./bin/console mbin:cache:build` again.

Fixes the the eleventh issue of: #1119


melroy89:
I can confirm this fixed the issue.

BentiGorlich (approved):

PR #1129, opened 2024-09-13
Check if EntryComment is defined and not null

melroy89:
[EntryComment can be null after all](https://github.com/MbinOrg/mbin/blob/4d930f6c8cf276bb2ec14ed5e579f4400af4222c/src/Controller/Entry/Comment/EntryCommentViewController.php#L35) during the __invoke. So `comment` can still be null. Let's update Twig to check for defined and not null, to avoid template errors..

---

100% repeatable now. Eg. https://kbin.melroy.org/m/games@sh.itjust.works/t/457088/The-entire-staff-of-beloved-game-publisher-Annapurna-Interactive-has/comment (so without any comment ID at the end provided in the URL)

Will give:

```json
{"message":"Uncaught PHP Exception Twig\\Error\\RuntimeError: \"Error rendering \"entry_comment\" component: Expected argument of type \"App\\Entity\\EntryComment\", \"null\" given at property path \"comment\".\" at view.html.twig line 49","context":{"exception":{"class":"Twig\\Error\\RuntimeError","message":"Error rendering \"entry_comment\" component: Expected argument of type \"App\\Entity\\EntryComment\", \"null\" given at property path \"comment\".","code":0,"file":"/var/www/kbin.melroy.org/html/templates/entry/comment/view.html.twig:49","previous": ... }
```

With error:

image

---

However, with this patch. You will no longer get this 500 error ⛱️ . And show an error instead:

image

### DIFF
--- a/templates/entry/comment/view.html.twig
+++ b/templates/entry/comment/view.html.twig
@@ -41,16 +41,20 @@
     </div>
     {% set autoAction = 'notifications:EntryCommentCreatedNotification@window->subject-list#addComment' %}
     {% set manualAction = 'notifications:EntryCommentCreatedNotification@window->subject-list#increaseCounter' %}
-    <div class="{{ html_classes('comments entry-comments comments-tree', {
-        'show-comment-avatar' : SHOW_COMMENT_USER_AVATARS is same as V_TRUE
-        }) }}"
-        data-controller="subject-list"
-        data-action="{{- DYNAMIC_LISTS is same as V_TRUE ? autoAction : manualAction -}}">
-        {{ component('entry_comment', {
-            comment: comment.root ?? comment,
-            showEntryTitle: false,
-            showMagazineName: false,
-            showNested: true,
-        }) }}
-    </div>
+
+    {% set entryComment = comment.root ?? comment %}
+    {% if entryComment is defined and entryComment is not null %}
+        <div class="{{ html_classes('comments entry-comments comments-tree', {
+            'show-comment-avatar' : SHOW_COMMENT_USER_AVATARS is same as V_TRUE
+            }) }}"
+            data-controller="subject-list"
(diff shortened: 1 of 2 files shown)


melroy89:
Also added an error with a back button in case of an empty/null entry comment...

melroy89:
Tested and works as expected now.

templates/entry/comment/view.html.twig
```diff
@@ -46,11 +46,23 @@
...
+                showMagazineName: false,
+                showNested: true,
+            }) }}
+        {% else %}
+        <div class="section alert alert__danger">
+            <p>
+                <a href="{{ entry_url(entry) }}" class="btn btn-link btn__danger">
+                    <i class="fa-solid fa-arrow-left-long" title="{{ 'back'|trans }}"></i> {{ 'back'|trans }}
+                </a>
+                &nbsp; {{ 'comment_not_found'|trans }}
```
BentiGorlich, line 63: What is the `&nbsp;` for?

melroy89: An additional space :)

BentiGorlich: I know what it does, but why is it needed? 😅

melroy89: For a bit more space after the back button...

BentiGorlich: I'd prefer it if you'd use margins for that

BentiGorlich: I mean you can just inline style it in such situations imo

melroy89: can we just merge it ;P We have more important stuff haha

melroy89: Using inline CSS

BentiGorlich (approved):

PR #1136, opened 2024-09-19
Disable CSRF checks on some most-used forms

melroy89:
We see a lot of issues with CSRF in your forms with Mbin... We disable CSRF checks for now on the forms that doesn't have actions that seems very harmful, even if it would be exploited by a CSRF attack..

Especially on the following **CSRF IDs**:

 - up_vote
 - down_vote
 - boost
 - subscribe
 - block
 - clear_notifications
 - read_notifications
 - follow

Since the impact on CSRF attacks on these kind of actions are minimal and slim. I will disable the validate check for now in the controller code. Until we fully have resolved this issue, since it affects a lot of people!

This is trying to resolve bullet 13th at: https://github.com/MbinOrg/mbin/issues/1119

### DIFF
--- a/src/Controller/BoostController.php
+++ b/src/Controller/BoostController.php
@@ -23,7 +23,7 @@ class BoostController extends AbstractController
     #[IsGranted('ROLE_USER')]
     public function __invoke(VotableInterface $subject, Request $request): Response
     {
-        // CSRF is causing a lot of issues, so we disable it for now.
+        // CSRF is causing a lot of issues, so we disable it for now. See PR: https://github.com/MbinOrg/mbin/pull/1136
         // $this->validateCsrf('boost', $request->getPayload()->get('token'));
 
         $this->manager->vote(VotableInterface::VOTE_UP, $subject, $this->getUserOrThrow());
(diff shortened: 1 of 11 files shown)


melroy89:
I think we can safely say this PR is a great success. 👍🏽

BentiGorlich (commented):
I think you should link this PR or the CSRF issue if we have on in the comments, so we can get the context of the problems. Additionally should the csrf token not also be removed on the forms? Or do you really just want to disable the verification of the tokens?

melroy89:
> I think you should link this PR or the CSRF issue if we have on in the comments, so we can get the context of the problems. 

I can add indeed this PR number.

> Additionally should the csrf token not also be removed on the forms? Or do you really just want to disable the verification of the tokens?

Officially we could remove all the CSRF code from the controller & twig templates if we really want to indeed. However, my aim was to "temporally" (for the foreseeable future) the CSRF checks on the less critical areas, solving the 500 errors and improve the user experience overall. Until we know how to better solve the issue... (if at all).

melroy89:
> I can add indeed this PR number.

Done!

BentiGorlich (approved):

PR #1138, opened 2024-09-19
Rename remaining article to thread

melroy89:
Rename all the remaining keywords from `article(s)` to `thread(s)`. Making the code more consistent.

Which also means I renamed the new valid link for creating a new thread: `new/thread` in the URL (same for the API).

This PR also finally solves the translation strings that then where still have the key `article`, `articles`, `edit_article` (which was always missing btw), `add_new_article`, etc. To the new keys: `type.thread`, `thread`, `threads`, `add_new_thread` and finally `edit_thread`.

To be backwards compatible with existing data in the database (specifically the `type` column in the `entry` table), we keep the `article` string in the `resolveType()` function, so t he old string "article" or "articles" are still valid. Same for some Twig templates, we extend with another `or` for backwards compatibly for now.

The next logical follow-up would be to rename `entry`  as well later.


templates/components/entry.html.twig
```diff
@@ -15,7 +15,7 @@
     {% elseif entry.cross %}
         {{ component('entry_cross', {entry: entry}) }}
     {% else %}
-        <article{{ attributes.defaults({
+        <thread{{ attributes.defaults({
```
BentiGorlich, line 18: I don't think that `thread` is a valid html5 element, `article` is: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/article

config/kbin_routes/magazine_api.yaml
```diff
@@ -1,7 +1,7 @@
-# Create an article entry in a magazine
-api_magazine_entry_create_article:
-  controller: App\Controller\Api\Entry\MagazineEntryCreateApi::article
-  path: /api/magazine/{magazine_id}/article
+# Create a thread entry in a magazine
+api_magazine_entry_create_athread:
+  controller: App\Controller\Api\Entry\MagazineEntryCreateApi::thread
+  path: /api/magazine/{magazine_id}/thread
```
BentiGorlich, line 4: changing an API path should never be done... You could mark the old path as deprecated and add a new API route at this new path and reference the new one. If you did this change suddenly on your server running latest main, an app (e.g. Interstellar) would not be able to create a thread anymore

config/kbin_routes/magazine_api.yaml
```diff
@@ -1,7 +1,7 @@
-# Create an article entry in a magazine
-api_magazine_entry_create_article:
-  controller: App\Controller\Api\Entry\MagazineEntryCreateApi::article
-  path: /api/magazine/{magazine_id}/article
+# Create a thread entry in a magazine
+api_magazine_entry_create_athread:
```
BentiGorlich, line 2: typo

BentiGorlich (changes requested):

templates/components/entry.html.twig
melroy89: ow haha, that could be right yes.

config/kbin_routes/magazine_api.yaml
melroy89: Correct. We might want to create a new API version number. Also a reason I targeted v1.8.0 version for Mbin.

Since Interstellar is the only client thus far, we could inform them about this backwards incompatible change for now.

BentiGorlich: I really don't want to introduce any unnecessary breaking changes. I understand the need to get rid of confusing nomenclature, but I think we should keep the old endpoint. It just doesn't hurt

melroy89: Sure ok, we will make it backwards compatible for now (and later remove it I suppose).

BentiGorlich:
Ok I will look at it closer tomorrow. Code wise it looks good now

melroy89:
> Ok I will look at it closer tomorrow. Code wise it looks good now

This pr is less urgent than the 1.7.x fixes. No worries here. It's not really a bug fix.

melroy89:
Rebased

BentiGorlich:
@melroy89 do you have an idea why there are so many commits linked to this PR? Doesn't really make sense to me especially since there are multiple commits of other PRs

melroy89:
> @melroy89 do you have an idea why there are so many commits linked to this PR? Doesn't really make sense to me especially since there are multiple commits of other PRs

yeah, I also found this weird. Maybe something went wrong with merging/rebasing?

melroy89:
I might need to redo this all, and revert several commits via reset hard or something. And try to rebase again. Maybe I started on the main branch initially? And later changed to the new `dev/new_features` branch?

PR #1140, opened 2024-09-19
Reduce error pollution even more on post request failures

melroy89:
This will fix the 16th [miscellaneous issue](https://github.com/MbinOrg/mbin/issues/1119), with too much verbose error logging.

I try to handle the post request basically the same way as the previously implemented `InvalidApPostException` exceptions like in the `getCollectionObjectImpl` function.

 
This meaning the content logging will be part of the catch -> `logRequestException` function instead. And the `InvalidApPostException` exception will only log the request body, but no longer the response content. Also we log the status code now as well for failed post requests. Plus some small refactoring.

Last but not least, I will only do a `json_encode` once.

### DIFF
--- a/src/Service/ActivityPub/ApHttpClient.php
+++ b/src/Service/ActivityPub/ApHttpClient.php
@@ -90,8 +90,8 @@ class ApHttpClient
     {
         $this->logger->debug("ApHttpClient:getActivityObject:url: $url");
 
-        $client = new CurlHttpClient();
         try {
+            $client = new CurlHttpClient();
             $r = $client->request('GET', $url, [
                 'max_duration' => self::MAX_DURATION,
                 'timeout' => self::TIMEOUT,
--- a/src/Service/ActivityPub/ApHttpClient.php
+++ b/src/Service/ActivityPub/ApHttpClient.php
@@ -329,7 +329,7 @@ class ApHttpClient
     }
 
     /**
-     * Helper function for logging get/post/.. requests to the error log with additional info.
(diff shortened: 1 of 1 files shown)


src/Service/ActivityPub/ApHttpClient.php
```diff
@@ -379,20 +379,28 @@ public function post(string $url, User|Magazine $actor, ?array $body = null): vo
...
 
-        if (!str_starts_with((string) $response->getStatusCode(), '2')) {
-            throw new InvalidApPostException("Post fail: $url, ".substr($response->getContent(false), 0, 1000).' '.json_encode($body));
+            $statusCode = $response->getStatusCode();
+            if (!str_starts_with((string) $statusCode, '2')) {
+                // Do NOT include the response content in the error message, this will be often a full HTML page
+                throw new InvalidApPostException("Post failed: $url, status code: $statusCode, request body: $jsonBody");
+            }
+        } catch (\Exception $e) {
+            $this->logRequestException($response, $url, 'ApHttpClient:post', $e);
```
melroy89, line 403: Apparently `logRequestException` is logging for "get failed..." .. I will either need to create a `logRequestException` or provide an additional parameter for the message "get failed" or when it's "post failed".

melroy89: or .. just remove the word "get"

melroy89: Solved.

melroy89:
I'm running this for 4 days without any issues at kbin.melroy.org

TheVillageGuy (approved):

PR #1152, opened 2024-10-02
Increase the interval for fetch_user_notifications

melroy89:
Users just leave the site open for hours and hours. And days after days.. Weeks after weeks.

With only a very small user base on kbin.melroy.org, I'm seeing 25694 requests in just 12 hours! Because it's getting trigger every 10 secs.

image

My proposal is to reduce the Ajax interval from 10 seconds to 30 seconds. After each new page load, this call will be executed anyways, so this `setTimeout` is only for checking for new notifications if people leave the site / browser tab open. 30 seconds is still plenty enough IMO.

I know the API request itself might not be very heavy, but these numbers are a bit absurd high for such a small instance.

### DIFF
--- a/assets/controllers/notifications_controller.js
+++ b/assets/controllers/notifications_controller.js
@@ -116,7 +116,7 @@ export default class extends Controller {
                 if (typeof data.messages === "number") {
                     this.setMessageCount(data.messages)
                 }
-                window.setTimeout(() => this.fetchAndSetNewNotificationAndMessageCount(), 10 * 1000)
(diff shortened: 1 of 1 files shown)


TheVillageGuy:
I see this as well, users leaving the site open for days

TheVillageGuy (approved):

BentiGorlich:
Maybe we can just make it a logarithmic delay with a max value of 5 minutes or so...
But I actually think that even the 10s are not a problem it is not a taxing query and just because you have 25k requests doesn't mean that it actually has a meaningful impact

PR #1166, opened 2024-10-06
Fix scheduler not running

BentiGorlich:
- fix the scheduler not working by making it stateful (only works with latest cache update)
- add a new message for clearing dead messages


melroy89:
Shouldn't we update the docs? So it's more explicit what is happening automatically?

BentiGorlich:
good point. I will do that 👍


Am 7. Oktober 2024 18:33:47 MESZ schrieb Melroy van den Berg ***@***.***>:

melroy89 (approved):

PR #1169, opened 2024-10-06
Fix wrong title showing while editing a thread

BentiGorlich:


melroy89:
yea I also notice that as well, the fix is also in: https://github.com/MbinOrg/mbin/pull/1138

melroy89 (approved):

PR #1173, opened 2024-10-07
Introduce the activity table

### LINKED ISSUE #772: Like activity ID is not dereferencable
**Describe the bug**

`Like` activities have IDs, but these IDs can not be dereferenced. Mbin server returns an empty JSON object.

https://moist.catsweat.com and others

1.5.3

**To Reproduce**

```
curl -H "Accept: application/activity+json" https://moist.catsweat.com/f/object/b808046e-41f2-43e6-ab98-56888a09cf0d
```

**Expected behavior**

Server should return the entire Like activity.

**Additional context**

When I receive an `Announce(X)` activity in my inbox, I want to verify that `X` activity is real and the group actor didn't make it up. If `Announce.actor` and `X.actor` belong to the same server, I can assume that activity is valid. However, when actors are on different servers, I need to verify the authenticity of `X` activity by fetching it from the server of origin. Otherwise the server of the group actor could impersonate actors on other servers.

You might not consider it a security risk, but I would appreciate if you make activity IDs dereferencable.

Currently I'm working only with announced `Like` activities, but other activities such as `Create` and `Delete` may also have this problem - I haven't verified that.

BentiGorlich:
- [x] Save activities in the database so that they can be called via URL
- [x] Add unit tests for the JSON used in outward federation
- [ ] Add functional tests for inbound federation
  - [x] tests with our own JSON
  - [ ] tests with lemmy's JSON

Closes #772


melroy89:
I think @BentiGorlich can better solve these conflicts, since he knows this code changes better. I will not touch this.

BentiGorlich:
I am still working on this, so do not think that this is not going anywhere... The tests are getting there, but a lot of work is still todo :)

BentiGorlich:
Left ToDos:
- add testing for undoing incoming dislikes
- add announcing of incoming reports to local magazines to all remote instances with moderators (although that might deserve it own issue and PR, what do you think @melroy89 ?)

melroy89:
> * add announcing of incoming reports to local magazines to all remote instances with moderators (although that might deserve it own issue and PR, what do you think @melroy89 ?)

That sounds indeed something for a separate PR. Especially if you want to limit the scope a bit of this PR in order to get it merged. Personally I'm always in favor of smaller PRs, merging more, but smaller amount of changes (if possible). Like small iterations. Which also as a benefit provide earlier feedback as well as leaner to adapt to new requests or changes in requirements. But I  digress :D

BentiGorlich:
Alright, then we'll do that in a different PR and I will make an issue about it

BentiGorlich:
@melroy89 one maybe open todo could be to make the new tests thread safe, but I was really trying and did the setup the way I though should work with paratest, but I really really struggled and gave up in the end. Problem is that the CI duration is now up to 20 minutes again (from the usual 13 minutes).
I don't know how much you know about parallelization, but maybe it would help if you take a look at it.
Btw. the changes to the settings repo for example were only done to fix a parallelization issue

BentiGorlich:
Yeah I know about that. The way I did the `setUp` methods everywhere should not be a problem I think...

BentiGorlich:
I have finally marked this as ready for review. I have it running on gehirneimer.de
It is of course a whole tone of code that is new. A lot of it are new snapshot tests, showing almost all of our federation jsons.

This issue is blocking basically all ActivityPub modifications, and it fixes some very important issues.

melroy89:
> * [x]  Save activities in the database so that they can be called via URL
> 
>     * [x]  Add unit tests for the JSON used in outward federation
> 
>     * [ ]  Add functional tests for inbound federation
>       
>       * [x]  tests with our own JSON
>       * [ ]  tests with lemmy's JSON
> 
> 
> Closes #772

Where the lemmy tests also done?

BentiGorlich:
Nope I didn't add any lemmy testing to the suite. But the mbin tests are a start and this PR is big enough and took long enough as it is...

melroy89:
No doubt indeed, this pr is big enough.

PR #1179, opened 2024-10-09
Use composer audit & npm audit

melroy89:
[Local PHP Security Checker repo is archived](https://github.com/fabpot/local-php-security-checker), and recommends to use `composer audit`:

```
COMPOSER_AUDIT_ABANDONED=ignore composer audit
```

So we don't depend anymore on this `local-php-security-checker` script, remove it from the Dockerfile for CICD and change both the README & workflow files.

Finally, I renamed the job from `security-check` to `audit-check`, to be consistent with the commands we are using (I will fix that in the branch settings when you approve it). Example of the audit check job: https://github.com/MbinOrg/mbin/actions/runs/11262147066/job/31317174935?pr=1179

More info: https://github.com/fabpot/local-php-security-checker

### DIFF
--- a/.github/workflows/action.yaml
+++ b/.github/workflows/action.yaml
@@ -95,15 +95,37 @@ jobs:
           SYMFONY_DEPRECATIONS_HELPER: disabled
         run: php bin/phpunit tests/Unit
 
-  security-check:
+  audit-check:
     runs-on: ubuntu-latest
     container:
       image: danger89/mbin-pipeline:1.2.0
     steps:
       - uses: actions/checkout@v4
 
-      - name: Run security checker
-        run: local-php-security-checker
+      - name: Get Composer Cache Directory
+        id: composer-cache
+        run: |
+          echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
+      - name: Calculate composer.lock hash
+        id: composer-lock-hash
+        run: |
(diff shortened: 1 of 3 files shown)


BentiGorlich:
should we add such a check for npm as well?

melroyvandenberg:
> should we add such a check for npm as well?

Yea, we might as well. Although NodeJS packages are not really running in production. Most packages are not fully used/deployment in production, like webpack for example.

I could include it as a step in the `build` job, another job will create again a lot of overhead for GitHub Actions.

BentiGorlich (commented):
Actually I think this should not run on commits, but on merged PRs. A PR should not fail because of insecure packages, should it?

melroy89:
> Actually I think this should not run on commits, but on merged PRs. A PR should not fail because of insecure packages, should it?

You also want to run it on (open) PRs, otherwise the pipeline fails after you merged it. Both `composer audit` and `npm audit` runs very quickly, so that isn't an issue either. 

We can however ask ourselves how important a failure is, in the branch settings we can make this job required to run & succeed. But we could also choice to not make it required, so you have feedback early without blocking the PR.

melroy89:
So for now the `audit-check` is not required to succeed. 

image


But we could make it mandatory, it's up to us.

BentiGorlich (approved):

melroy89:
> I think making it not required is a good idea. Might confuse new devs though, but 🤷

People can still fix it, if possible ;P

PR #1180, opened 2024-10-10
Exclude InvalidApPostException from rolling back the transaction

BentiGorlich:
Overwrite the default workWrapper method so the `transactional` method is not called. Instead implement a custom behaviour that does not roll back the db changes if the exception was a `InvalidApPostException`

### DIFF
--- a/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
+++ b/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
@@ -53,6 +53,7 @@ class DeliverHandler extends MbinMessageHandler
             $this->doWork($message);
             $conn->commit();
         } catch (InvalidApPostException $e) {
+            $conn->commit();
             // we don't roll back on an InvalidApPostException, so the failed delivery attempt gets written to the DB
             throw $e;
         } catch (\Exception $e) {
--- a/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
+++ b/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
@@ -42,6 +42,27 @@ class DeliverHandler extends MbinMessageHandler
         $this->workWrapper($message);
     }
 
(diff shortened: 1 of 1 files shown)


src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
```diff
@@ -42,6 +42,27 @@ public function __invoke(DeliverMessage $message): void
...
+    {
+        $conn = $this->entityManager->getConnection();
+        if (!$conn->isConnected()) {
+            $conn->connect();
+        }
+        $conn->beginTransaction();
+        try {
+            $this->doWork($message);
+            $conn->commit();
+        } catch (InvalidApPostException $e) {
```
melroyvandenberg, line 55: Also catch for this `InvalidApGetException` in advance now? Since I would like to introduce that at some point.


```suggestion
        } catch (InvalidApPostException|InvalidApGetException $e) {
```

melroyvandenberg: (and add: `use App\Exception\InvalidApGetException`  at the top)

BentiGorlich: Actually I think in this case it is not necessary. We only want to catch exception stemming from delivering the activity. That is actually only the `InvalidApPostException`

melroy89: If you say so :D

melroy89 (approved):

BentiGorlich:
why did my commit not dismiss your review :thinking:

melroy89:
> why did my commit not dismiss your review 🤔

Because of: 
image


But I trust you.

melroy89:
> Thanks for that :D But I think this box should be ticked. What do you think?

Yea maybe, if you want more work 😆

melroy89:
I checked the box.

PR #1181, opened 2024-10-10
Redesign federation page

BentiGorlich:
- 3 different sections
- TOC
- always show panels with an empty notice when there are no entries

grafik

### DIFF
--- a/templates/page/federation.html.twig
+++ b/templates/page/federation.html.twig
@@ -13,11 +13,25 @@
 {% endblock %}
 
 {% block body %}
-    <h1>{{ 'federation'|trans }}</h1>
-    <div id="content" class="section" style="overflow: auto;">
 
+    <h1>{{ 'federation'|trans }}</h1>
+    <div class="section">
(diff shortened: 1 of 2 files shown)


melroy89 (approved):
Thank you!

melroy89:
Hopefully an update like this `merge branch main into..` will not remove the approval?

BentiGorlich:
@melroy89 
> The base branch requires all commits to be signed

melroy89:
> @melroy89
> 
> > The base branch requires all commits to be signed

yea nice right..? I also try to enable that while I was there.. Not a good idea?

image

melroy89:
> > @melroy89
> > > The base branch requires all commits to be signed
> 
> yea nice right..? I also try to enable that while I was there.. Not a good idea?
> 
> image
> 
> So your first commit doesn't seems to be signed:
> 
> image

I guess I will disable this again.. but maybe a good thing to also enable soon...

BentiGorlich:
I just setup gpg signing on my end. Maybe in the future. I think this is a good idea in general. Don't know if it is necessary

PR #1204, opened 2024-11-04
Fix dead instance query by allowing lastSuccessfulDeliver is null

melroy89:
See title.

### DIFF
--- a/src/Entity/Instance.php
+++ b/src/Entity/Instance.php
@@ -100,6 +100,7 @@ class Instance
 
     public function isDead(): bool
     {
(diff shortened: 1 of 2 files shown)


BentiGorlich:
I think you need to edit this as well: https://github.com/MbinOrg/mbin/blob/c46b9c3fe2fa3367cffb819a5e5761e50834a35d/src/Entity/Instance.php#L101

melroy89:
> I think you need to edit this as well:
> 
> https://github.com/MbinOrg/mbin/blob/c46b9c3fe2fa3367cffb819a5e5761e50834a35d/src/Entity/Instance.php#L101

Uhm. I think so as well..

melroy89:
Fixed :)

BentiGorlich (approved):

PR #1206, opened 2024-11-06
increase pm.max_children

TheVillageGuy:
60 is too low, 70 runs a lot smoother


melroy89 (approved):

BentiGorlich:
I think the docs should state that this is a server specific config, depending on the size of your userbase and how active they are... It is like the postgresql tuning stuff just not one size fits all...

melroy89:
Yes agreed, but for now I also agreed with increasing the max_children.

PR #1211, opened 2024-11-08
Fix delivering to oneself

BentiGorlich:
When the `DeliverManager` tries to deliver to the local server prevent that and issue a warning

I am really not sure if that is the issue at hand for @Jerry-infosecexchange and @jwr1 but it is the only thing I can think of and it is theoretically possible...


melroy89:
Do you want an approval and merge? Or do you want them to test the branch first?

BentiGorlich:
I think we can just merge it as it is a good change either way... Even if it is a last minute failsafe...

PR #1217, opened 2024-11-12
Improve Nginx logging, by introducing a separate inbox log

melroy89:
- Split inbox requests and other access requests logs in Nginx, for both bare metal & Docker
   - Outbox is so less, so I didn't bother splitting that into a separate log file, it's less of an issue. But the **inbox** end-point is heavily requested.
- Extend troubleshooting page
- Combine the static assets section
- Improve Nginx configuration / improve markdown.
- Add `buffer=32k flush=5m` to the inbox.log, to reduce disk I/O!

**Rationale:** This makes debugging easier as well as reduce the strain on logging parsers and/or dashboards like Grafana or alike. Since the `inbox` boxes are receiving a lot of requests per second/minute or day. Filtering the inbox requests to a separate log will help a lot.

This PR allows you to manage the logs and processing of them (or exclude the inbox log for example) much easier when you're using external tools.

Inspired by: https://lemmy.cafe/u/Illecors

---

An alternative approach using variables in the log file was been tested as well, however this will hamper performance due to: `the file is opened and closed for each log write. ` See also: http://nginx.org/en/docs/http/ngx_http_log_module.html#access_log. So I'm not implementing a variable in the log file path.

### DIFF
--- a/docs/02-admin/01-installation/bare_metal.md
+++ b/docs/02-admin/01-installation/bare_metal.md
@@ -62,6 +62,8 @@ sudo php /tmp/composer-setup.php --install-dir=/usr/local/bin --filename=compose
 
 If you have a firewall installed (or you're behind a NAT), be sure to open port `443` for the web server. Mbin should run behind a reverse proxy like Nginx.
 
+For Nginx see: [Nginx configuration](../02-configuration/nginx.md).
+
 ## Install NodeJS (frontend tools)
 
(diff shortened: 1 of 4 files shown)


melroy89:
Tested on kbin.melroy.org Nginx server, works great! Much better.

TheVillageGuy (approved):

PR #1221, opened 2024-11-14
Use keepalive (requires a upstream block) + Open file cache

melroy89:
- Use `keepalive` in Nginx, to actually keep the connection open. keepalive is not the max. amount of connections, but the amount of idle connections.
- Also add `open_file_cache` 
- Increase `gzip_comp_level` to 5 for a better balance between performance/CPU usage.

For some reason Nginx think it's a good idea that this required an `upstream` block, so actually in order to leverage keepalive functionality in Nginx, you will need to define an upstream section...

Official keepalive documentation: http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive

And open file cache documentation here: http://nginx.org/en/docs/http/ngx_http_core_module.html#open_file_cache


docs/02-admin/01-installation/docker.md
```diff
@@ -243,17 +248,24 @@ server {
...
     access_log /var/log/nginx/mbin_inbox.log if=$inboxRequest buffer=32k flush=5m;
 
+    open_file_cache          max=1000 inactive=20s;
+    open_file_cache_valid    60s;
+    open_file_cache_min_uses 2;
+    open_file_cache_errors   on;
+
     location / {
+        proxy_http_version 1.1;
         proxy_set_header HOST $host;
```
BentiGorlich, line 258: Why http 1.1?

melroy89: Good question, so the default value is http 1.0, setting this to http v1.1 allows you to have persistent connections. This is very much recommended when using keep-alive (or even in order to have keep alive working). 

"Version 1.1 is recommended for use with [keepalive](http://nginx.org/en/docs/http/ngx_http_upstream_module.html#keepalive) connections"

http://nginx.org/en/docs/http/ngx_http_proxy_module.html#proxy_http_version 
https://www.rfc-editor.org/rfc/rfc2616#section-8.1.2

melroy89: Then your next question will be. Why `proxy_set_header Connection "";`? I will also answer that right away now. 

Like I said before, by default Ngix will use HTTP/1.0 for connections to upstream servers and adds the `Connection: close` header to the requests that it forwards to the servers. The result is that each connection gets closed when the request completes, despite the presence of the `keepalive` directive in the upstream block.

So this way we remove the close from the `Connection` header.

BentiGorlich: my question was more like "why not http 2" 😅

melroy89: ow. hahaa!

Keep alive is prohibited in HTTP/2 and HTTP/3. HTTP/2 has the notation of u multiple streams of data to be multiplexed over a single connection.. I might now go too far into all the details, more than I wanted. But yeah, the backends like in this case Mercure should also support these protocols if you want to enable HTTP/2 or HTTP/3. Which might not always be the best solution.

melroy89: Ps. there is also a stream module.. http://nginx.org/en/docs/stream/ngx_stream_core_module.html But that is overkill for what we want to do here.. Using keepalive for upstream services is just something I learned during fine-tuning tutorials of @antonputra (he is also presenting tutorials on YT).

BentiGorlich (approved):

PR #1232, opened 2024-11-19
[Feature] Admin approval

### LINKED ISSUE #722: Add option for new accounts to require admin approval

Right now instance owners still are lacking some measures to fight back against spam waves. One that may help is adding an option for account creation to require admin/global mod (maybe) approval before the account can be used

We might want to investigate how other platforms do this. One major consideration is, if enabled, should accounts be made (and the address for that account name burned)? Or should the accounts be held in a separate state until admin approval, at which point they are created and the fediverse can find them.

Assuming we don't want to create the accounts in a way that uses the address forever before approval:

- Potentially a new table like `users_awaiting_approval` with columns of the `username, hashed_password, email` etc that are required to create an account
- If enabled, registrations go to this new table instead of the user table, login says awaiting approval if possible (if not, maybe user does not exist is fine at least for first pass)
- New admin UI to list entries awaiting approval
- Ability to accept or reject those requests
  - On accept, actually make the `public.user` entry, email the user
  - On deny, email the user they have been denied?

If instead we don't care if the username is burned, another possible implementation is

- New column on `public.user` such as `requires_approval = true`
- Block login if the flag is `true`
- New admin UI to list users where `requires_approval` is `true`. New index on `where requires_approval is true` to make the query fast
- Same as above on accept/deny emailing user

There might be other options to curtail spam that are lower hanging fruit, if people can think of them. We could also just do something like an admin option to automatically ban users on registration, so admins could then unban as a form of "approval". The flow would be confusing for users, but it's all tech that exists today so would be faster to implement

BentiGorlich:
- admins can now decide to turn a mandatory approval on in the admin settings
- a new tab will appear labelled "Applications" where new users that are not yet approved will appear
- when the setting is enabled a new input will be displayed on the register page: "Application Text"
- This behavior will be apparent on registering with an added flash message informing the user about it
- Users will receive an email when they are accepted or rejected
- when a user is not approved they cannot log in and an appropriate message will be displayed
- the register site now redirects to the login page instead of the front page

## Tasks

- [x] API implemented
- [x] Updated Docs
- [x] Included Images
- [x] Implemented mails
- [x] Refine email texts (I am definitely not qualified to do that)
- [x] Refine Admin application page

# Pictures

## Register Page

Register page
After registration

## Admin application view

Admin application view

## Mails

Approved email

Rejected email

Closing #722

### DIFF
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -240,13 +240,24 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface, Visibil
     #[Column(type: 'string', nullable: false, options: ['default' => self::USER_TYPE_PERSON])]
     public string $type;
 
+    #[Column(type: 'string', nullable: true)]
+    public string $applicationText;
+
+    #[Column(type: 'boolean', nullable: false, options: ['default' => true])]
+    public bool $isApproved;
+
+    #[Column(type: 'boolean', nullable: false, options: ['default' => false])]
+    public bool $isRejected = false;
+
     public function __construct(
         string $email,
         string $username,
         string $password,
         string $type,
         ?string $apProfileId = null,
-        ?string $apId = null
+        ?string $apId = null,
+        bool $isApproved = true,
+        ?string $applicationText = null,
     ) {
         $this->email = $email;
         $this->password = $password;
@@ -280,6 +291,8 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface, Visibil
         $this->lastActive = new \DateTime();
         $this->createdAtTraitConstruct();
         $this->oAuth2UserConsents = new ArrayCollection();
+        $this->isApproved = $isApproved;
+        $this->applicationText = $applicationText;
     }
 
     public function getId(): int
--- a/src/Entity/User.php
+++ b/src/Entity/User.php
@@ -10,6 +10,7 @@ use App\Entity\Contracts\VisibilityInterface;
 use App\Entity\Traits\ActivityPubActorTrait;
 use App\Entity\Traits\CreatedAtTrait;
 use App\Entity\Traits\VisibilityTrait;
+use App\Enums\EApplicationStatus;
 use App\Repository\UserRepository;
 use App\Service\ActivityPub\ApHttpClient;
 use Doctrine\Common\Collections\ArrayCollection;
@@ -240,14 +241,11 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface, Visibil
     #[Column(type: 'string', nullable: false, options: ['default' => self::USER_TYPE_PERSON])]
     public string $type;
 
-    #[Column(type: 'string', nullable: true)]
+    #[Column(type: 'text', nullable: true)]
     public string $applicationText;
 
-    #[Column(type: 'boolean', nullable: false, options: ['default' => true])]
-    public bool $isApproved;
-
-    #[Column(type: 'boolean', nullable: false, options: ['default' => false])]
-    public bool $isRejected = false;
+    #[Column(type: 'enumApplicationStatus', nullable: false, options: ['default' => EApplicationStatus::Approved->value])]
+    private string $applicationStatus;
 
     public function __construct(
         string $email,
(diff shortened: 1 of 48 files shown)


BentiGorlich:
Do we have someone that is good in phrasing user facing stuff like the emails? @jwr1 or @TheVillageGuy maybe?

translations/messages.en.yaml
```diff
@@ -919,3 +921,15 @@ search_type_all: Threads + Microblogs
...
+new_users_need_approval: New users have to be approved by an admin before they can log in.
+applications: Applications
+application_text: Application text
+signup_requests_header: These users would like to join your server. They cannot log in until you've approved their application
+flash_application_info: Your admin needs to approve your account before you can log in.
+  You will receive an email when they processed your application.
+email_application_approved_title: Your application was approved
+email_application_approved_body: Your signup application was approved by the admins. You can log into the server at <a href="%link%">%siteName%</a>.
+email_application_rejected_title: Your application was rejected
+email_application_rejected_body: Your signup application was rejected by the admins
```
melroy89, line 933:
```suggestion
email_application_rejected_body: Thank you for your interest, but we regret to inform you that your registration request has been declined.
```

translations/messages.en.yaml
```diff
@@ -919,3 +921,15 @@ search_type_all: Threads + Microblogs
...
+email_application_pending: Your admin needs to approve your account before you can log in.
```
melroy89, line 934:
```suggestion
email_application_pending: Your account requires admin approval before you can log in.
```

translations/messages.en.yaml
```diff
@@ -919,3 +921,15 @@ search_type_all: Threads + Microblogs
...
+email_verification_pending: You have to verify your email address before you can log in.
```
melroy89, line 935:
```suggestion
email_verification_pending: Please verify your email address before you can log in.
```

translations/messages.en.yaml
melroy89: I would call it "registration request"?

src/Entity/User.php
```diff
@@ -240,13 +240,24 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface, Visibil
...
     public string $type;
 
+    #[Column(type: 'string', nullable: true)]
+    public string $applicationText;
+
+    #[Column(type: 'boolean', nullable: false, options: ['default' => true])]
+    public bool $isApproved;
+
+    #[Column(type: 'boolean', nullable: false, options: ['default' => false])]
+    public bool $isRejected = false;
```
melroy89, line 250: In theory you could potentially end up with `isApproved` and `isRejected` set to true, since there is no guard against this incorrect state.

Not sure how to prevent that.

melroy89:
The "One last step" in the message now contradict each other 😆 . Yes you still need to use the activation link, but if admin approval is required, this isn't the last step.. you than still need to wait.

BentiGorlich:
> The "One last step" in the message now contradict each other 😆 . Yes you still need to use the activation link, but if admin approval is required, this isn't the last step.. you than still need to wait.

Yeah I know :sweat_smile:, but I didn't know how to solve this. I didn't really want to mess up the existing translations... But if you think that it is necessary I will split the existing one up, so that both messages can be merged

src/Entity/User.php
melroy89: Ah yes of course, this needs to be an enum type. This will also allows you to introduce other states like "waiting" (waiting for approval) safely. 

This is the way.

BentiGorlich:
@melroy89 are you fine with the 2 flash messages upon signup? I think it is fine this way, but I can invest the work to fix it if you want to

melroy89:
> @melroy89 are you fine with the 2 flash messages upon signup? I think it is fine this way, but I can invest the work to fix it if you want to

It's fine for me.

src/Entity/User.php
melroy89: Seems to be fixed now, right? Looks much better with an enum 🥳

melroy89 (approved):

PR #1235, opened 2024-11-22
Throw UnrecoverableMessageHandlingException on 4xx

BentiGorlich:
When we try to deliver an activity to an inbox and receive a 4xx status code other than the 429 for rate limiting we now discard that message, so it will not get retried. This also doesn't count as a failed delivery to the instance anymore, because the instance is up, it just didn't accept the activity. TransportExceptions are now caught as well, since those are the ones we get when an instance is down

### DIFF
--- a/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
+++ b/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
@@ -26,6 +26,8 @@ use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
 #[AsMessageHandler]
 class DeliverHandler extends MbinMessageHandler
 {
+    public const HTTP_RESPONSE_CODE_RATE_LIMITED = 429;
+
     public function __construct(
         private readonly EntityManagerInterface $entityManager,
         private readonly ApHttpClient $client,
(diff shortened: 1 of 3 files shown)


src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
```diff
@@ -53,8 +59,21 @@ public function workWrapper(MessageInterface $message): void
             $this->doWork($message);
             $conn->commit();
         } catch (InvalidApPostException $e) {
+            if (400 <= $e->responseCode && 500 > $e->responseCode && 429 !== $e->responseCode) {
+                $conn->rollBack();
```
melroy89, line 63: Agreed, however, I would also like to enable debugging logging as an admin.

Allowing to see which servers (domain + response body) are giving me invalid HTTP requests back. And try to understand why.

src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
```diff
@@ -14,8 +16,12 @@
 use App\Service\ActivityPubManager;
 use App\Service\SettingsManager;
 use Doctrine\ORM\EntityManagerInterface;
+use Psr\Cache\InvalidArgumentException;
 use Psr\Log\LoggerInterface;
 use Symfony\Component\Messenger\Attribute\AsMessageHandler;
+use Symfony\Component\Messenger\Exception\RecoverableMessageHandlingException;
+use Symfony\Component\Messenger\Exception\UnrecoverableMessageHandlingException;
+use Symfony\Contracts\HttpClient\Exception\TransportExceptionInterface;
```
melroy89, line 24: Best practice would be to have a static constant variable defined for `429` error code, instead of a magic value.

src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
```diff
@@ -53,8 +59,21 @@ public function workWrapper(MessageInterface $message): void
             $this->doWork($message);
             $conn->commit();
         } catch (InvalidApPostException $e) {
+            if (400 <= $e->responseCode && 500 > $e->responseCode && 429 !== $e->responseCode) {
```
melroy89, line 62:
```suggestion
            if (400 <= $e->responseCode && 500 => $e->responseCode && 429 !== $e->responseCode) {
```

Also equal to 500 errors right?

BentiGorlich: nope 500 errors are server errors and therefore should subside on later tries.

melroy89: OK but 502, 503 & 504 (and more?) could potentiality also be some kind of server configuration issue (just like 500 error would give if something is wrong on the server itself).

melroy89: Ah yes, right! gotcha 🥟

melroy89 (approved):

PR #1236, opened 2024-11-22
Don't federate anything from the random magazine

### LINKED ISSUE #444: Stop federating /m/random
Unmoderated content from the random mag is federating to other servers.

The random mag as the local catch all for posts and entries which can not be put into another magazine should not accept followers and should not announce posts in any way.

Don't even return it for the webfinger

**Additional context**

BentiGorlich:
- don't return the random magazine on webfinger requests
- don't federate anything outwards from the random magazine

Closes #444

### DIFF
--- a/src/EventSubscriber/ActivityPub/GroupWebFingerProfileSubscriber.php
+++ b/src/EventSubscriber/ActivityPub/GroupWebFingerProfileSubscriber.php
@@ -55,6 +55,10 @@ class GroupWebFingerProfileSubscriber implements EventSubscriberInterface
 
     protected function getActor($name): ?Magazine
     {
+        if ('random' === $name) {
+            return null;
+        }
+
         return $this->magazineRepository->findOneByName($name);
     }
 }
--- a/src/EventSubscriber/ActivityPub/GroupWebFingerSubscriber.php
+++ b/src/EventSubscriber/ActivityPub/GroupWebFingerSubscriber.php
@@ -62,6 +62,10 @@ class GroupWebFingerSubscriber implements EventSubscriberInterface
 
     protected function getActor($name): ?Magazine
     {
+        if ('random' === $name) {
+            return null;
+        }
+
         return $this->magazineRepository->findOneByName($name);
     }
 }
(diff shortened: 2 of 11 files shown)


melroy89:
OK I think I agree maybe, but..

What if I got to a random thread like: https://kbin.melroy.org/m/random/t/64276/Lovecraft-s-Cthulhu-Mythos-Reading-Order#comments

And I place a comment.. Will that still just work? I think the users of my instance who comment there would like to get their comment federated out. At least I would.

BentiGorlich:
yeah you're absolutely right. The author should still get the comments from us even if the thread is jn the random magazine... Gotta adjust the PR, pribably a lot...

melroy89:
Let me try to run this on my instance. Just to check.

BentiGorlich:
Sounds like a good idea

melroy89:
> Sounds like a good idea

Seems to work? https://wehavecookies.social/@BentiGorlich/113538159661153615

melroy89:
Can you reply one more time? I'm now for sure running the branch. Before it could have been still running on main.

BentiGorlich:
The important part is you replying to me. Inbound federation was not changed :)

melroy89 (approved):

melroy89:
> The important part is you replying to me. Inbound federation was not changed :)

I tried to reply again: https://kbin.melroy.org/m/random/p/186322/at-melroy-at-kbin-melroy-org-Here-is-a-test-toot-from-mastodon-for-you

melroy89:
What about comments under threads??

BentiGorlich:
Threads use the same `CreateHandler` so should also work fine

melroy89:
Yup that also works: https://kbin.melroy.org/m/random/t/295162/Crane-on-a-plant and https://lemmy.world/post/16073167

melroy89:
Like also works: https://wehavecookies.social/@BentiGorlich/113538353072337133/favourites

BentiGorlich:
And the likes are also working :)

PR #1241, opened 2024-11-23
Improved action workflow

melroy89:
- Let's try to use the built-in `hashFiles` function, and clean-up the yaml file

### DIFF
--- a/.github/workflows/action.yaml
+++ b/.github/workflows/action.yaml
@@ -37,7 +37,6 @@ jobs:
           restore-keys: ${{ runner.os }}-npm-
 
       - uses: actions/cache@v4
-        id: vendor-cache
         with:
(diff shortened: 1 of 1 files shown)


melroy89:
Interesting right? We do have a cache (cache hit).. And it does restore the cache..

But still unable to find the vendor binary from the directory? huh? Sure we can always trigger composer install, but shouldn't the cache restore not just contain these files.

image

BentiGorlich (approved):
I am just gonna trust you on this, as I don't know much about the github actions :)

melroy89:
> I am just gonna trust you on this, as I don't know much about the github actions :)

So instead of calculating the md5sum yourself. We use the built-in hashFiles function to hash the file. Which is used as a unique key index for caching. If the lock files didn't change, the same hash will be lookup. And therefore we have match (cache hit).

PR #1242, opened 2024-11-24
[Feature] Admin Signup Notifications

### LINKED ISSUE #838: Admin Notification for signups
I want to be notified when a user signs up, so I can check whether that user is a spam account or a real user

A new notification only going to admins, created when a user signs up, after verifying their mail

none

**Additional context**
mastodon has that and it is super useful

BentiGorlich:
When a user is created send a notification to all admins

Notification:
grafik

Push Notification:
grafik

closes #838

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -7,6 +7,7 @@ doctrine:
         mapping_types:
             user_type: string
             citext: citext
+            enumApplicationStatus: string
 
         # IMPORTANT: You MUST configure your server version,
         # either here or in the DATABASE_URL env var (see .env file)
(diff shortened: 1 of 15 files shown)


melroy89:
Could we please disable this feature at settings as well; `admin/settings`?

BentiGorlich:
Never saw that message, sure I'll add it :)

melroy89:
I see, you added it now to the user profile. aka `settings/general`?

melroy89 (approved):

melroy89:
> yeah correct. Each admin account can enable/disable as they wish. It is on by default though. The option is not displayed to non admins

Yup! Perfect. I was just thinking like this is the first option for admins under the profile settings? I was a bit surprised to be honest 😆 . I was expecting to have more options for admins. But apparently not (yet).

PR #1245, opened 2024-11-24
Fix min length issues for magazine description and rules

BentiGorlich:
- lower the min length for the magazine description and rules from 3 to 0

Fixes the mentioned issue in #1001


jwr1 (approved):
Just tried it on kbin.earth, looks like it works!

PR #1254, opened 2024-11-26
Check for banned instances + improved logging

melroy89:
- Check for banned instances on missing places
- Improve logging
- Extend comment & code docs (please, do this always if you finally understand the method, this will help other devs)

Fixes :  18th issue of #1175

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -140,7 +140,7 @@ class ActivityPubManager
         $this->logger->debug('ActivityPubManager:findActorOrCreate: searching for actor at "{handle}"', ['handle' => $actorUrlOrHandle]);
         if (str_contains($actorUrlOrHandle, $this->settingsManager->get('KBIN_DOMAIN').'/m/')) {
             $magazine = str_replace('https://'.$this->settingsManager->get('KBIN_DOMAIN').'/m/', '', $actorUrlOrHandle);
-            $this->logger->debug('found magazine "{magName}"', ['magName' => $magazine]);
+            $this->logger->debug('Found magazine: "{magName}"', ['magName' => $magazine]);
 
             return $this->magazineRepository->findOneByName($magazine);
         }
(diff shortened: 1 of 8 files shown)


melroy89:
OK it seems to work as expected now on my instance (kbin.melroy.org).

src/Service/ActivityPubManager.php
```diff
@@ -741,16 +743,20 @@ private function handleMagazineFeaturedCollection(string $actorUrl, Magazine $ma
...
                             } else {
                                 $object = $item;
                                 if ($isString) {
-                                    $this->logger->debug('getting {url} because we dont have it', ['url' => $apId]);
-                                    $object = $this->apHttpClient->getActivityObject($apId);
+                                    if (!$this->settingsManager->isBannedInstance($apId)) {
+                                        $this->logger->debug('Getting {url} because we dont have it', ['url' => $apId]);
+                                        $object = $this->apHttpClient->getActivityObject($apId);
+                                        $this->logger->debug('Dispatching create message for entry: {e}', ['e' => json_encode($object)]);
+                                        $this->bus->dispatch(new CreateMessage($object, true));
```
BentiGorlich, line 755: This has to be outside of the `if ($isString)` 

It can and should be behind its own `if (!$this->settingsManager->isBannedInstance($apId))`

BentiGorlich (changes requested):
The one thing I commented on in the code definitely needs to be changed. Otherwise the changes look good to me, though I think it could be useful to prepend the origin class to the log statements like 
> The instance is banned, url: {url}

Maybe something like:
> [ActivityPubManager::getEntityObject] The instance is banned, url: {url}

src/Service/ActivityPubManager.php
melroy89: You mean this? https://github.com/MbinOrg/mbin/pull/1254/commits/042ab8b433f1abc170ae33d52fb1d29ead3db945

BentiGorlich: exactly :+1:

melroy89: In that case, can't we just wrap it all behind a single `isBannedInstance` check?

BentiGorlich: would work as well

BentiGorlich (approved):

PR #1272, opened 2024-12-08
[Feature] Custom notification settings

BentiGorlich:
You can now set custom notification settings for users, magazine, entries and posts: `Loud`, `Default` or `Muted`.

The heart of the logic is in the method `NotificationSettingsRepository::findNotificationSubscribersByTarget` which replaced some logic in the notification managers. 

The logic works in tiers:
1. user settings
2. entry/post settings
3. magazine settings
4. default settings

So if you muted the magazine and thread, but the author is set to "Loud" then you will still get notifications for threads created by them.
If you muted the user, but set the magazine to "Loud" then you will get notifications for threads that are not created by that muted user.

---

| UI | |
| --- | --- |
| Magazine Box | Magazine Box |
| Thread Info | Thread Info |
| Entry | Entry |
| Post | Post |
| User Popover | User Popover |


melroy89:
Depending on the order of merging. Do you want the branch to be merged to main. Or in the dev branch. I leave the merge order up to you.

melroy89 (approved):

src/Repository/NotificationSettingsRepository.php
```diff
@@ -0,0 +1,223 @@
...
+                $dontNeedToBeAuthor = false;
+            }
+        }
+
+        $isMagazineLevelString = $isMagazineLevel ? 'true' : 'false';
+        $isNotMagazineLevelString = !$isMagazineLevel ? 'true' : 'false';
+        $dontNeedSubscriptionString = $dontNeedSubscription ? 'true' : 'false';
+        $dontNeedToBeAuthorString = $dontNeedToBeAuthor ? 'true' : 'false';
+
+        $sql = "SELECT u.id FROM \"user\" u
```
melroy89, line 144: This query looks very heavy. Did you perform benchmarks on this new query? We already have too many long duration queries, so hopefully this query is not taking long to execute?

And I'm talking about a server with many users and magazines of course (a real world scenario).

BentiGorlich: Nope I didn't test or benchmark this query, but imo it only looks heavy, but is actually not for a database server. It has just a big where statement which means it will not go through that many table rows.
I would love to benchmark it, but I have no idea how I would do that

melroy89: I would love to have some performance tracing lines in Mbin, but I digress..

Regarding PostgreSQL queries, try to get real life performance duration by setting `log_min_duration_statement` in postgresql. And check the postgresql logs.  Like in this issue: https://github.com/MbinOrg/mbin/issues/1230#issuecomment-2486540382

BentiGorlich: Ok, I activated it with a 1 second `log_min_duration_statement`. But I doubt that my server is a good benchmark

melroy89: maybe set it to 500ms instead ;)? Which is also very long (too long) for a query.

BentiGorlich: I had it at the setting for 5 days and I have gotten notifications in the mean time: when I run `cat /var/log/postgresql/postgresql-16-main.lo1g | grep LEFT\ JOIN\ notification_settings` I get an empty response. So I'd say the performance is safe :)

melroy89 (approved):

PR #1277, opened 2024-12-09
Include last receive in the formula for dead instances

BentiGorlich:
An instance now counts as dead only if last successful deliver AND last successful receive are older than a week and there were 10 failed delivers in a row

This is of course open for discussion, but imo we have to include the receive in the formula. I looked at my instances and at fedia and there were just far too many instances that don't belong there


melroy89 (approved):

melroy89:
I assume `lastSuccessfulReceive` has an index, so the DB query performance is not impacted that much.

BentiGorlich:
I actually think that there is no index on the instance table other than the `domain` column

PR #1256, opened 2024-11-28
Rename routes, serialization, utils controller and mixins to mbin

melroy89:
Rename some stuff to Mbin, without any impact for users or admins.

- `config/kbin_routes -> config/mbin_routes`
- `config/kbin_serialization -> config/mbin_serialization`
- `utils/kbin.js -> utils/mbin.js`
- `styles/mixins/kbin.scss -> styles/mixins/mbin.scss` + `mbin-btn-link` mixin
- `controllers/kbin_controller.js -> controllers/mbin_controller.js`
- Fixes in various code places to make the rename work again of course (like `config/services.yaml`, `src/Kernel.php`, JS files and scss files)

PR #1278, opened 2024-12-09
Fix data action kbin -> mbin rename

melroy89:
- I renamed `kbin_controller.js` to `mbin_controller.js`, however then you apparently **ALSO need to update the `data-controller` as well with this name. So Stimulus would load the controller.**
- Then I also did the rename on the `data-action` to `mbin#` (which is required after you renamed the data-controller).
- In the swagger template (by NelmioApiDocBundle), we don't even use this navigation menu. This should be removed, because this is dead-code there.
   -  While I was at here, I also improved the CSS in the Swagger template, so the logo actually is better displayed now within the SwaggerUI template.
- Also update `kbin_lang` to `mbin_lang` while I found it anyways how it works now.
- ALso rename `kbin-container` to `mbin-container`, while I'm here now.

Regression of: https://github.com/MbinOrg/mbin/pull/1256

### DIFF
--- a/assets/controllers/mbin_controller.js
+++ b/assets/controllers/mbin_controller.js
@@ -60,6 +60,6 @@ export default class extends ApplicationController {
     }
 
     changeLang(event) {
-        window.location.href = '/settings/theme/kbin_lang/' + event.target.value;
+        window.location.href = '/settings/theme/mbin_lang/' + event.target.value;
     }
 }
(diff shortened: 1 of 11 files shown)


TheVillageGuy (approved):

melroy89:
image

Nice

TheVillageGuy (approved):

PR #1280, opened 2024-12-10
Remove default local domain for entries

BentiGorlich:
When an entry is created it will be assigned a domain depending on its url. If it does not have one it previously fell back to the local domain. This commit removes this fallback as it can introduce long-running queries, when calculating the new number of entries in the domain

Also added a command to remove all remote entries from this local domain (not a migration as we don't know which id the local domain has).

What this looked like on Gehirneimer: there were >50,000 entries assigned to the domain `gehirneimer.de`. After removing the remote one it is down to 67. The problem with having this fallback is that you have to recount the number of entries in the domain. That could have been solved by just counting +1 instead of running a query (maybe in the future). The query for counting these entries took ~1.7 seconds, which is not terribly long, but it is avoidable and the domain should actually match the entry. No idea why the fallback was introduced


melroy89 (commented):
Should we update the docs?

BentiGorlich:
will do

melroy89 (dismissed):

melroy89 (dismissed):

docs/02-admin/04-running-mbin/05-cli.md
```diff
@@ -346,3 +346,13 @@ Usage:
...
 ```
+
+### Update-Local-Domain
+
+This command will remove all remote posts from belonging to the local domain. This command is only relevant for instances
+created before v1.7.4 as the local domain was the fallback if no domain could be extracted from a post.
+
+Usage:
+```bash
+php bin/console mbin:update:local-domain [--]
```
melroy89, line 357: What is --?

BentiGorlich: that's just something symfony creates automatically :D should I remove it?

melroy89: Uh. Yes please, maybe the command has an incorrect parameter config? Anyhow, I don't think it should have this optional --. Or at least remove it from the docs indeed, it's very confusing to me.

melroy89 (approved):

PR #1281, opened 2024-12-10
Remove using the AP id as the URL

BentiGorlich:
Entries exposed their AP id as their URL if no URL was set. This commit removes that

As far as I can tell this is only relevant for `Article` AP types, so only the entries. Mastodon posts have their id as the url, so I'd keep it for `Note` AP types (posts and comments)

On Articles it had the effect that mbin instances had domains for each other mbin instance, but also the mbin logo on each mbin article when it didn't have an image attached or pointed to a link. That behaviour has annoyed me forever, so here it gets removed

### DIFF
--- a/src/Factory/ActivityPub/EntryPageFactory.php
+++ b/src/Factory/ActivityPub/EntryPageFactory.php
@@ -131,6 +131,6 @@ class EntryPageFactory
             return $this->imageManager->getUrl($entry->image);
         }
 
-        return $entry->url ?? $this->getActivityPubId($entry);
+        return $entry->url;
     }
 }


melroy89:
I might not fully understand your story in your description to be honest.. But I believe you

melroy89:
Did you ran this for let's say 12 or 24 hours on a live instance? No issues?

BentiGorlich:
Ok I try to explain it better: The `url` field is the thing we show behind the title: 
grafik

In this case `lemmy.world` with the link to the image attached to the post.

We populate this field with
1. The url of the link you posted (if you clicked "add new link")
2. The url of the image that is attached to the post (if you clicked "add new photo")
3. The url of the post itself (if you clicked "add new thread")

The last one does not make sense in my opinion. Lemmy has no `url` field if the post is not linking to a url or an image, we shouldn't either. 

The consequence of having this 3. point (that gets removed with this PR) is: that every post originating from an mbin instance has a url attached to it. From attached urls we generate thumbnails, which is why even a simple post saying "Hi" in the title will have the mbin logo as an image attached to it. If we implemented an embedded view in the future this embedded view would be attached to the post, which just shouldn't happen.

The only upside of this behaviour is, that we can filter for posts by fedia.io for example if we go to `/d/fedia.io`, though this will not be a complete list as posts pointing to imgur for example will be excluded here. We should introduce something to actually filter for instances, not for the domains of the url posts are pointing to.

As for the testing part, no I have not, but I do not expect any problems because lemmy is handling it this way and the url field just isn't a requirement. You can see here: https://browser.pub/https://lemmy.world/post/8715949 that this post doesn't have one

melroy89:
> Ok I try to explain it better: The `url` field is the thing we show behind the title: grafik
> 
> In this case `lemmy.world` with the link to the image attached to the post.
> 
> We populate this field with
> 
>     1. The url of the link you posted (if you clicked "add new link")
> 
>     2. The url of the image that is attached to the post (if you clicked "add new photo")
> 
>     3. The url of the post itself (if you clicked "add new thread")
> 
> 
> The last one does not make sense in my opinion. Lemmy has no `url` field if the post is not linking to a url or an image, we shouldn't either.
> 
> The consequence of having this 3. point (that gets removed with this PR) is: that every post originating from an mbin instance has a url attached to it. From attached urls we generate thumbnails, which is why even a simple post saying "Hi" in the title will have the mbin logo as an image attached to it. If we implemented an embedded view in the future this embedded view would be attached to the post, which just shouldn't happen.

OK this helps a lot! Thanks, now I get it. And yes I also believe this is actually wrong and annoying.

> The only upside of this behaviour is, that we can filter for posts by fedia.io for example if we go to `/d/fedia.io`, though this will not be a complete list as posts pointing to imgur for example will be excluded here. We should introduce something to actually filter for instances, not for the domains of the url posts are pointing to.

Well we do have `ap_id`.. is that something we could leverage?

> As for the testing part, no I have not, but I do not expect any problems because lemmy is handling it this way and the url field just isn't a requirement. You can see here: https://browser.pub/https://lemmy.world/post/8715949 that this post doesn't have one

I mean it also doesn't break searches (`/search`) or other existing DB queries we already have in Mbin now that may or may misuse this url column?

melroy89 (approved):

BentiGorlich:
We can just join in the author and then we have a `ap_domain` column.

> I mean it also doesn't break searches (/search) or other existing DB queries we already have in Mbin now that may misuse this url column?

Like I said it is completely normal for lemmy posts to not have this field so our code already handles it :)

PR #1297, opened 2024-12-14
Update front.html.twig Add border to non federated notifications for admins and moderators

TheVillageGuy:
Add border to non federated notifications for admins and moderators

This allows admins/moderators (when they receive notifications in the future) to quickly see if there has been any activity by local users on their instance

image


melroy89:
I'm not so sure about this change. Not that I'm not in favor to make it clear whether it's federated or non-federated notification. It's more about how to solve this issue.

I believe adding a border would only add confusing. Only you now know what this means, but its not clear at all what a border means around a notification at all. 

If you would try to distinguish between local or federated messages. I believe we should solve it differently. Maybe with an icon + alt text or something, not a "random" border.

TheVillageGuy:
> I'm not so sure about this change. Not that I'm not in favor to make it clear whether it's federated or non-federated notification. It's more about how to solve this issue.
> 
> I believe adding a border would only add confusing. Only you now know what this means, but its not clear at all what a border means around a notification at all.
> 
> If you would try to distinguish between local or federated messages. I believe we should solve it differently. Maybe with an icon + alt text or something, not a "random" border.

I can add an alt text, that's a great idea. I've been using this for a while now and it works really well for me. The border is subtle but still catches my attention. I think it works better than an icon, I know from experience icons can be missed when skimming lists. These borders are hard to miss

TheVillageGuy:
Tooltip added:

image

TheVillageGuy:
image

Added tooltip with username. Don't remember who but someone said it would be nice to be able to see which instance something was posted from. 2 birds with one stone. I don't feel this should be iconified

melroy89 (approved):

templates/notifications/front.html.twig
```diff
@@ -42,11 +42,24 @@
                 </button>
             </div>
         </div>
-
     </div>
 
     {% for notification in notifications %}
-        <div class="{{ html_classes('section section--small notification', {'opacity-50': notification.status is not same as 'new' }) }}">
+        <div class="{{ html_classes('section section--small notification', {'opacity-50': notification.status is not same as 'new' }) }}"
+            {% if not notification.entry.magazine.apId and not notification.post.magazine.apId and (app.user.admin or app.user.moderator) %}
```
melroy89, line 49: did you also tested on if the variable is not-set / null or empty?? And if these checks are not causing any 503 errors?

TheVillageGuy: Yes I tested the lot and it's live on my instance, the variables are always set, which surprised me. Check me in mbin development as there is one weird situation with comments

melroy89: > Yes I tested the lot and it's live on my instance, the variables are always set, which surprised me. 

Well, because you are an admin? Try to login with a regular account?

TheVillageGuy: > > Yes I tested the lot and it's live on my instance, the variables are always set, which surprised me.
> 
> Well, because you are an admin? Try to login with a regular account?

I did, and if the user is not an admin they never reach the code. I'm stil looking at moderators, it is not working as I had hoped.

melroy89: So it doesn't work as expected?

TheVillageGuy:
Yeah it does but I've redone it with a different approach with less chance
of failure as it's impossible to properly check if the vars are set in
twig. I'll update the pr soon and then pass it by you again

templates/notifications/front.html.twig
TheVillageGuy: It does, to clarify;

> I did, and if the user is not an admin they never reach the code

is intended behavior. 

I was experimenting with the code though and found out that under the 'right' circumstances var checks can raise exceptions in Twig while in fact trying to prevent them. I don't like that, So, despite it working as is I want it to, I've rewritten it and am currently testing a different strategy

TheVillageGuy:
> Don't work yet as expected

It does, see my comment above

TheVillageGuy:
My new approach comes with the advantage of having more control over what the notifications look like, I've resisted the urge to go wild but have added a little envelope to notifications when they are messages

image

let me know what you think. It's just an ✉ emoticon so doesn't need additional image handling

melroy89 (approved):

BentiGorlich:
@TheVillageGuy I bet the failing tests are caused by your changes regarding the css classes. The tests will have css selectors on them that just don't match anymore

TheVillageGuy:
> @TheVillageGuy I bet the failing tests are caused by your changes regarding the css classes. The tests will have css selectors on them that just don't match anymore

The changes are the same as before though, when they did pass the tests. Just achieved in a different way from the code. Haven't had time to look at the tests yet

melroy89:
Sorry, you didn't made 1.7.4 release.

PR #1300, opened 2024-12-17
Get the integration tests up and running again

BentiGorlich:
- The messenger transactional wrap is not working with the tests, as it conflicts with the DoctrineTestBundle
- Fix some type problems (DateTime vs DateTimeImmutable, nulls and initialization errors)
- `SearchManager::findActivityPubActorsByUsername` now converts `!user@domain` into `@user@domain` so searching for a magazine by `!mag@domain` works from the API as well
- The `StatsContentRepository` had a sign error
- add a new docker compose for the test dependencies, as well as oauth keys for the tests
- remove the purge user test, as this is no longer an option
- change the .env.test to fit the new dependencies


.github/workflows/action.yaml
```diff
@@ -81,6 +81,61 @@ jobs:
...
+          POSTGRES_PASSWORD: ChangeThisPostgresPass
+        # Set health checks to wait until postgres has started
+        options: >-
+          --health-cmd pg_isready
+          --health-interval 10s
+          --health-timeout 5s
+          --health-retries 5
+      redis:
+        # Docker Hub image
+        image: redis
```
melroy89, line 128:
```suggestion
        image: valkey/valkey
```

If you believe in open source licenses :)

BentiGorlich:
Ok, so it is now working, but the tests take forever: 23 minutes. There is a library called [paratest](https://github.com/paratestphp/paratest) that utilizes parallel testing, however the transaction isolation is not working at the moment (I have no idea why) so we have random assert failures with it (just tested it)
@melroy89 if you have any idea what could cause the [doctrine test bundle](https://github.com/dmaicher/doctrine-test-bundle) to not work, please let me know as this already caused some headache for me :sweat_smile:

BentiGorlich:
Ok, so just checked what actually happens. It does not commit anything to the DB, however cahing is screwing us a little bit and the static settings were a problem. I am working on it :D

melroy89:
If you need more php extensions installed. Let me know right.

melroy89:
That is strange you need this kernel object while not using it?

melroy89 (approved):

BentiGorlich:
it is being passed to the parent which needs it to know in which environment it is

BentiGorlich:
the unit test job is required, but I renamed/removed it 😅

BentiGorlich:
Ok so your comment reads relatively offensive 😅
My intention was to not have to have two redundant build steps only differing in one command. Additionally I wanted to make the build shorter as the unit tests will have to have the postgres and valkey services as well since the db bootstrap is always executed now. I can change it back if ypu really hate it. But since it is in two steps of The same check/run its easy to spot either way. Additionally for me it is only relevant that tests failed, not which kind of test failed

melroy89:
> Ok so your comment reads relatively offensive 😅 My intention was to not have to have two redundant build steps only differing in one command. Additionally I wanted to make the build shorter as the unit tests will have to have the postgres and valkey services as well since the db bootstrap is always executed now. I can change it back if ypu really hate it. But since it is in two steps of The same check/run its easy to spot either way. Additionally for me it is only relevant that tests failed, not which kind of test failed

Ok Continue. I actually blame github.

melroy89 (approved):

PR #1304, opened 2024-12-18
Improve PostgreSQL config even further

melroy89:
See title. Improve PostgreSQL disk I/O significantly.  Especially on a service like Mbin, with a lot of writes.

Proof of reduce I/O and improve disk write times:

image

image

### DIFF
--- a/docs/02-admin/02-configuration/04-postgresql.md
+++ b/docs/02-admin/02-configuration/04-postgresql.md
@@ -60,7 +60,7 @@ commit_delay = 300
 checkpoint_timeout = 40min
 checkpoint_completion_target = 0.9
 # Write ahead log sizes (so the WAL file can contain around 1 hour of data)
-max_wal_size = 30GB
+max_wal_size = 10GB
 min_wal_size = 2GB
 
 # Query tuning
--- a/docs/02-admin/02-configuration/04-postgresql.md
+++ b/docs/02-admin/02-configuration/04-postgresql.md
@@ -16,7 +16,7 @@ However, the following settings are a good starting point when your serve is aro
 
 ```ini
 # Increase max connections
-max_connections = 100
+max_connections = 200
 
 # Increase shared buffers
 shared_buffers = 8GB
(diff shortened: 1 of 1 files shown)


docs/02-admin/02-configuration/04-postgresql.md
```diff
@@ -48,14 +48,20 @@ max_parallel_maintenance_workers = 4
...
 
 # Group write commits to combine multiple transactions by a single flush (this is a time delay in μs)
 commit_delay = 300
 
+# Increase the checkpoint timeout (time between two checkpoints) to reduce the disk I/O
+# This will significantly reduce the disk I/O and speed-up the write times to disk. The only downside is time needed for crash recovery.
+checkpoint_timeout = 40min
 checkpoint_completion_target = 0.9
+# Write ahead log sizes (so the WAL file can contain around 1 hour of data)
+max_wal_size = 30GB
```
BentiGorlich, line 63: I think 40 min and 30 GB is abit much for the default...

melroy89: Nope. 40 min isn't for sure, this is very much recommended for every write database queries, we have with mbin. And 30gb maybe, but actually not for big instances.

See also https://www.crunchydata.com/blog/tuning-your-postgres-database-for-high-write-loads


The wall size was just too small even on my instance together with the checkpoint interval. My server was therefore constantly busy writing checkpoints. This is bad. I increased the max Wal size to be big enough, that the checkpoint will be triggered by the timeout.

melroy89: I believe 30GB is fine. It doesn't matter too much, the interval will trigger the checkpoint.

Even my small instance benefit of these values a lot as you can see in the grafana dashboard.

BentiGorlich: Lets put it this wqy: people hosting from VPSes in the cloud will not have 30GB to spare. I dont on Gehirneimer for example.
Yes your instance might be smaller, but you have a gargantous server behind it compared to most servers on the fediverse 😁
If you say that it cannot cause problems, then I'll approve, but J am not sure 😅

melroy89: The wal file is stored in `$PGDATA/pg_wal`. I can decrease the wal size. But.. my database alone (which is stored in the $PGDATA is 48GB already..

melroy89: I decrease it  to 10GB, is that OK?

BentiGorlich: Yeah that is fine. My DB is at about 22 GB I believe

BentiGorlich (approved):

PR #1333, opened 2024-12-30
Devcontainer improvements

garrettw:
added missing env file, fixed vhost config


melroy89:
Is it possible to create maybe some sort of symbolic link for this dot env file? To avoid duplicate info? And more importantly to avoid forgetting to update also this dot env file in the future.

garrettw:
The purpose of this file is not to get it into a correct path to be used — quite the opposite, in fact. This file is meant to be used exactly as is with the devcontainer so that it works out of the box. What happens when the container is created is that this file gets copied to `/.env` where the app can use it.

melroy89:
Would have been nice to create a symlink to: https://github.com/MbinOrg/mbin/blob/main/.env.example_docker

But I think the settings are too different ;(

melroy89 (approved):

Issue #1340, opened 2024-12-30
Fix PHPStan errors


garrettw:
Just making this issue to track my work on fixing PHPStan errors. There's over 2000 of them so I won't be doing it all in one PR.

PR #1341, opened 2024-12-30
PHPStan (+PHPUnit) corrections, pt 1

### LINKED ISSUE #1338: Fix unit tests that trigger deprecation warnings
5-6 tests are triggering deprecation warnings, so I want to fix those.

For my reference: https://stackoverflow.com/questions/76152966/how-to-make-phpunit-explain-deprecations-and-warnings

garrettw:
Addresses #1340 

Decided to include PHPUnit deprecation fixes while I'm at it, since it was quick and easy.
This one should be pretty straightforward; there aren't many changes to actual code that could be impactful.

Fixes #1338


melroy89:
Since we don't see anything in the workflow regarding these warnings. Would it be a good idea to introduce a PHPStan workflow/step as well?

garrettw:
Yes, but I don’t think we’re ready for it yet. Even if it were marked as not required, it would make every test run fail with a red X until all ~2200 issues are fixed, which I don’t think is helpful. I would support adding it after I finish going through and fixing everything, though.

melroy89 (approved):

PR #1345, opened 2024-12-31
Enable multi-core parallel PHP CS Fixer

melroy89:
Enable the parallel config, which auto-detect the cores on the system and run PHP CS Fixer in parallel. See also:

https://cs.symfony.com/doc/usage.html

Tested and works perfectly fine! Especially on my new threadripper system :)


garrettw:
Looks like the job went from 37s to 17s with this change. 👍

garrettw (approved):

PR #1348, opened 2025-01-03
PHPStan corrections, pt 2

garrettw:


.github/workflows/action.yaml
```diff
@@ -93,6 +93,11 @@ jobs:
       - name: Build frontend (production)
         run: npm run build
 
+      - name: Dump event payload
+        env:
+          EVENT_PAYLOAD: ${{ tojson(github) }}
+        run: echo "EVENT_PAYLOAD is $EVENT_PAYLOAD"
```
melroy89, line 99: Is this necessary? It seems like some sort of debug info?

melroy89: nevermind, the PR is still draft

PR #1355, opened 2025-01-07
Fix signup notifications for deleted accounts

BentiGorlich:
Admins shouldn't get notifications when an account is created as a deleted account. This happens when we delete users, as a new user is created with deleted set to true to block the username


melroy89 (approved):

BentiGorlich:
There seems to be a problem with the actions... They were not triggered

melroy89:
> There seems to be a problem with the actions... They were not triggered

Maybe because you changed branch target branch names.. GitHub is buggy ;P

PR #1376, opened 2025-01-17
Move from dotenv to ini codeblocks

BentiGorlich:
- prismjs which we use to highlight our codeblocks does not support `dotenv` as a language, so move to ini (which was already used in a lot of places)


melroy89 (approved):
Agreed.

PR #1390, opened 2025-01-25
Make touch click open post

BentiGorlich:
When on a touch device the initial though is that clicking on the text of a post the post should be loaded in the single view. This implements exactly that

### DIFF
--- a/assets/controllers/subject_controller.js
+++ b/assets/controllers/subject_controller.js
@@ -36,6 +36,19 @@ export default class extends Controller {
         }
 
         this.checkHeight();
+
+        // if in a list and the click is made via touch, open the post
+        if (!this.element.classList.contains('isSingle')) {
+            this.element.addEventListener('click', (e) => {
+                if (e.defaultPrevented) {
+                    return
+                }
+                if ("touch" === e.pointerType) {
+                    let link = this.element.querySelector("header a:not(.user-inline)")
+                    document.location.href = link.getAttribute('href')
+                }
+            })
+        }
     }
 
     async getForm(event) {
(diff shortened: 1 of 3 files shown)


assets/controllers/subject_controller.js
```diff
@@ -36,6 +36,19 @@ export default class extends Controller {
...
+
+        // if in a list and the click is made via touch, open the post
+        if (!this.element.classList.contains('isSingle')) {
+            this.element.addEventListener('click', (e) => {
+                if (e.defaultPrevented) {
+                    return
+                }
+                if ("touch" === e.pointerType) {
+                    let link = this.element.querySelector("header a:not(.user-inline)")
+                    document.location.href = link.getAttribute('href')
```
melroy89, line 48: You might want to check that `link` is not null or undefined.. Yes I know, in the good case scenario this might be true..

BentiGorlich: done :)

assets/controllers/subject_controller.js
```diff
@@ -36,6 +36,24 @@ export default class extends Controller {
...
         this.checkHeight();
+
+        // if in a list and the click is made via touch, open the post
+        if (!this.element.classList.contains('isSingle')) {
+            this.element.addEventListener('click', (e) => {
+                if (e.defaultPrevented) {
+                    return
+                }
+                if ("touch" === e.pointerType) {
+                    let link = this.element.querySelector("header a:not(.user-inline)")
```
melroy89, line 47: both `let` here and the `let` below can be `const` :)

BentiGorlich: done :)

melroy89 (approved):

PR #1394, opened 2025-01-25
Introducing the mbin:images:delete command

melroy89:
- As the title says; introducing a new command for removing old images from your instance.
- By default will try to delete old federated images from threads, thread comments, posts and post comments.
- By default the `monthsAgo` arg is set to 12 months (1 year)
- By default the `type` arg is set to "all", meaning delete all images from threads, threads comments, posts and post comments, with the exceptions of users. Since I believe users can still be activate after 12 months, right ;)..
- Extended the docs

---

In my case, each time running `./bin/console mbin:images:delete`, will delete around 1GB of data from my media directory. Just for info.

### DIFF
--- a/src/Command/RemoveOldImagesCommand.php
+++ b/src/Command/RemoveOldImagesCommand.php
@@ -91,6 +91,8 @@ class RemoveOldImagesCommand extends Command
 
     /**
      * Call all delete methods below, _except_ for the delete users images.
+     * Since users on the instance can be several years old and not getting fetched,
+     * however we shouldn't remove their avatar/cover images just like that.
      */
     private function deleteAllImages($output): void
     {
@@ -251,7 +253,7 @@ class RemoveOldImagesCommand extends Command
     }
 
     /**
-     * Delete user avatar and user cover images.
+     * Delete user avatar and user cover images. Check ap_fetched_at column for the age.
      * Limit by batch size.
      */
     private function deleteUsersImages(OutputInterface $output)
--- a/src/Command/RemoveOldImagesCommand.php
+++ b/src/Command/RemoveOldImagesCommand.php
@@ -67,17 +67,17 @@ class RemoveOldImagesCommand extends Command
         $this->batchSize = (int) $input->getOption('batchSize');
 
         if ('all' === $type) {
-            $this->deleteAllImages($output); // Except for user avatars and covers
+            $nrDeletedImages = $this->deleteAllImages($output); // Except for user avatars and covers
         } elseif ('threads' === $type) {
-            $this->deleteThreadsImages($output);
+            $nrDeletedImages = $this->deleteThreadsImages($output);
         } elseif ('thread_comments' === $type) {
-            $this->deleteThreadCommentsImages($output);
+            $nrDeletedImages = $this->deleteThreadCommentsImages($output);
         } elseif ('posts' === $type) {
-            $this->deletePostsImages($output);
+            $nrDeletedImages = $this->deletePostsImages($output);
         } elseif ('post_comments' === $type) {
-            $this->deletePostCommentsImages($output);
+            $nrDeletedImages = $this->deletePostCommentsImages($output);
         } elseif ('users' === $type) {
-            $this->deleteUsersImages($output);
+            $nrDeletedImages = $this->deleteUsersImages($output);
         } else {
             $io->error('Invalid type of images to delete. Try \'all\', \'threads\', \'thread_comments\', \'posts\', \'post_comments\' or \'users\'.');
 
@@ -86,6 +86,8 @@ class RemoveOldImagesCommand extends Command
 
         $this->entityManager->clear();
 
+        $output->writeln(\sprintf('Total images deleted during this run: %d', $nrDeletedImages));
+
         return Command::SUCCESS;
     }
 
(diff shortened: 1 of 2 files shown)


src/Command/RemoveOldImagesCommand.php
```diff
@@ -0,0 +1,291 @@
...
+            ->from(Entry::class, 'e')
+            ->where(
+                $queryBuilder->expr()->andX(
+                    $queryBuilder->expr()->lt('e.createdAt', ':timeAgo'),
+                    $queryBuilder->expr()->neq('i.id', 1),
+                    $queryBuilder->expr()->isNotNull('e.apId'),
+                    $this->noActivity ? $queryBuilder->expr()->eq('e.upVotes', 0) : null,
+                    $this->noActivity ? $queryBuilder->expr()->eq('e.commentCount', 0) : null,
+                    $this->noActivity ? $queryBuilder->expr()->eq('e.favouriteCount', 0) : null,
+                    $this->noActivity ? $queryBuilder->expr()->isNotNull('e.image') : null
```
BentiGorlich, line 126: I think this should just be in the query no matter the value of `$this->noActivity` right? Or you could just do a inner join instead of a left join and remove this condition

src/Command/RemoveOldImagesCommand.php
```diff
@@ -0,0 +1,291 @@
...
+            ->orderBy('e.id', 'ASC')
+            ->setParameter('timeAgo', $timeAgo)
+            ->setMaxResults($this->batchSize)
+            ->getQuery();
+
+        $entries = $query->getResult();
+
+        foreach ($entries as $entry) {
+            $output->writeln(\sprintf('Deleting image from thread ID: %d, with ApId: %s', $entry->getId(), $entry->getApId()));
+            $this->entryManager->detachImage($entry);
```
BentiGorlich, line 139: if you call `detachImage` with an image that is null this will throw an error. Same for all the other `...Manager->detachImage(` calls. Especially for the user one where you have to have the or in the query

BentiGorlich (changes requested):
Some remarks. Additionally I would like to see a counter at the end how many images were deleted, but that is just icing on the cake :)

src/Command/RemoveOldImagesCommand.php
melroy89: Fixed. I indeed went for left join. And remove this condition.

src/Command/RemoveOldImagesCommand.php
melroy89: This will never happen anymore, we use inner join now. It already worked before, I removed several GBs of data without issues. But with `inner join` now, I don't even think that is possible at the moment.

melroy89: However, I did found an issue with deleting users images:

```
./bin/console mbin:images:delete --batchSize=2 users
Deleting image from username: @norb@lemmy.norbz.org
Deleting image from username: @zach@bigfoot.ninja
{"message":"Error thrown while running command \"mbin:images:delete --batchSize=2 users\". Message: \"Unsupported operand types: array * int\"","context":{"exception":{"class":"TypeError","message":"Unsupported operand types: array * int","code":0,"file":"/var/www/kbin.melroy.org/html/src/Command/RemoveOldImagesCommand.php:317"},"command":"mbin:images:delete --batchSize=2 users","message":"Unsupported operand types: array * int"},"level":500,"level_name":"CRITICAL","channel":"console","datetime":"2025-01-29T00:45:19.821603+01:00","extra":{}}
00:45:19 CRITICAL  [console] Error thrown while running command "mbin:images:delete --batchSize=2 users". Message: "Unsupported operand types: array * int" ["exception" => TypeError { …},"command" => "mbin:images:delete --batchSize=2 users","message" => "Unsupported operand types: array * int"]
{"message":"Uncaught Error: Unsupported operand types: array * int","context":{"exception":{"class":"TypeError","message":"Unsupported operand types: array * int","code":0,"file":"/var/www/kbin.melroy.org/html/src/Command/RemoveOldImagesCommand.php:317"}},"level":500,"level_name":"CRITICAL","channel":"php","datetime":"2025-01-29T00:45:19.834248+01:00","extra":{}}

In RemoveOldImagesCommand.php line 317:
                                          
  Unsupported operand types: array * int  
                                          

mbin:images:delete [--noActivity [NOACTIVITY]] [--batchSize [BATCHSIZE]] [--] [<type> [<monthsAgo>]]

```

src/Command/RemoveOldImagesCommand.php
```diff
@@ -0,0 +1,319 @@
...
+            ->setParameter('timeAgo', $timeAgo)
+            ->setMaxResults($this->batchSize)
+            ->getQuery();
+
+        $users = $query->getResult();
+
+        foreach ($users as $user) {
+            $output->writeln(\sprintf('Deleting image from username: %s', $user->getUsername()));
+            $this->userManager->detachCover($user);
+            $this->userManager->detachAvatar($user);
```
BentiGorlich, line 313: I think it would be a good practice to check for the the variables, even if it is already checked down the line
```suggestion
            if (null !== $user->cover) {
                $this->userManager->detachCover($user);
            }
            if (null !== $user->avatar) {
                $this->userManager->detachAvatar($user);
            }
```

BentiGorlich (changes requested):
Just small annotations. Otherwise it looks fine :)

BentiGorlich (approved):

PR #1405, opened 2025-01-31
Fix nodeinfo not available when federation is disabled

### LINKED ISSUE #1292: Nodeinfo route is inaccesible when federation is disabled
**Describe the bug**

Copied from matrix:

> I just had someone ask me why Interstellar wouldn't work with their Mbin instance that has federation disabled, and it turns out Mbin completely hides the nodeinfo route whenever federation is disabled (which I was not aware of). This route is what Interstellar uses in order to check and verify that a server is running a certain software (Mbin, Lemmy, etc.), and not just Interstellar, even several Lemmy apps (including the official Lemmy app), the [joinmbin.org](http://joinmbin.org/) website, fediverse crawlers, etc. check this endpoint. I understand that the nodeinfo route is meant for federation purposes, but at this point, it's used by so many other things, I think it should still be available even if server federation is disabled.

Upon which it was agreed that nodeinfo should still be available even if federation is disabled.

**To Reproduce**

Disable federation in the admin settings, and you will notice the nodeinfo route no longer works.

**Expected behavior**

The nodeinfo route should be available even if federation is disabled.

BentiGorlich:
Closes #1292

### DIFF
--- a/src/EventListener/FederationStatusListener.php
+++ b/src/EventListener/FederationStatusListener.php
@@ -22,7 +22,7 @@ class FederationStatusListener
 
         $route = $event->getRequest()->attributes->get('_route');
 
-        if (str_starts_with($route, 'ap_')) {
+        if (str_starts_with($route, 'ap_') && 'ap_node_info' !== $route && 'ap_node_info_v2' !== $route) {
             throw new NotFoundHttpException();
         }
     }


jwr1:
Do you think it would make sense to have the `/.well-known/nodeinfo` route available as well?

jwr1:
I see. I didn't realize the `ap_node_info` route corresponded to the well-known.

melroy89 (approved):
LGTM

PR #1406, opened 2025-01-31
Always put thumb images in a lightbox

### LINKED ISSUE #1127: Inline image view
id like the ability to tap an image inline in the feed to open it in full without having to open the post first

BentiGorlich:
- this commit puts all thumb images in a lightbox (so you can now click the thumb image in a list to open a lightbox of the image)
- additionally we now set the `data-gallery` so only the images belonging to one post are in a gallery together (which will currently always be one)

Closes #1127


melroy89 (dismissed):
I trust you here, hopefully everything still works.

BentiGorlich:
One thing I noticed is that opening the lightbox sometimes jumps to the bottom of the page, but only on mobile, not even with the mobile emulator. I gotta fix that first, cause it is really annoying when this is actively used...

TheVillageGuy:
Clicking thumbnails will still open images, right?

BentiGorlich:
> Clicking thumbnails will still open images, right?

Yes of course. This PR will let you click the thumb images in lists to get the lightbox containing the image. You can test it on gehirneimer.de I have it running there

BentiGorlich:
The scroll issue seems to be cause by the library and not our code. I created an issue for it: biati-digital/glightbox/issues/553

melroy89:
> The scroll issue seems to be cause by the library and not our code. I created an issue for it: [biati-digital/glightbox/issues/553](https://github.com/biati-digital/glightbox/issues/553)

Can this be replicated with their demo site? https://biati-digital.github.io/glightbox/#examples if not, maybe it's not their library.

TheVillageGuy:
> The scroll issue seems to be cause by the library and not our code. I created an issue for it: [biati-digital/glightbox/issues/553](https://github.com/biati-digital/glightbox/issues/553)

I have lightbox disabled on my instance because it behaves poorly with large images in general and doesn't work properly on iphones. Scrolling/zooming in iirc was problematic

BentiGorlich:
> Can this be replicated with their demo site? https://biati-digital.github.io/glightbox/#examples if not, maybe it's not their library.

No it cannot be replicated on their demo site, but that doesn't mean that it is not a library error. The super frustrating part is that is only on gehirneimer.de and I cannot even replicate it locally...

BentiGorlich:
Interestingly enough it seems like clearing the website data and cookies in my browser solved it. I had the error in FF and tried in Fennec without the problem, went ahead and delete the website data and cookies and not I don't have this problem in FF anymore... No idea where it came from, but if you cannot reproduce this problem on your phone @melroy89 I think we can safely merge this PR

BentiGorlich:
@TheVillageGuy do you have an iPhone and can test gehirneimer.de with the lightbox? I have no zoom problems on android, but I do not own or know anyone with an iPhone to test it...

BentiGorlich:
That would be great :)

melroy89 (approved):
I see no issues with the lightbox feature under iPhone 6s under Safari web browser. The images are opening in lightbox, and I'm able to close the box again. Everything seems to work fine.

melroy89:
@BentiGorlich I improvement would be that if you close outside of the image when in lightbox, the box should also close..? So you don't necessarily need to use the 'X' icon.

BentiGorlich:
The question with using an iPhone was whether zooming and moving the image is also working fine.

As for the click outside: that is actually a bug. I could implement the fix someone mentioned in the issue I created, but on desktop clicking outside the lightbox closes it

melroy89:
> The question with using an iPhone was whether zooming and moving the image is also working fine.

Owh, my bad. Let me test that as well now. I will also try to load maybe some large images like these posts with maps from rim world.

melroy89:
I can confirm that zooming as well as panning works as expected, no issues on this iPhone 6s in Safari.

TheVillageGuy:
> @TheVillageGuy do you have an iPhone
No, sorry

TheVillageGuy:
> I can confirm that zooming as well as panning works as expected, no issues on this iPhone 6s in Safari.

I don't know anything about iPhones but I'd recommend testing on one models as well

melroy89:
> > I can confirm that zooming as well as panning works as expected, no issues on this iPhone 6s in Safari.
> 
> I don't know anything about iPhones but I'd recommend testing on older models as well

I AM testing it on a very old iPhone. Like I said an iPhone 6s from 2015. We currently are already at iPhone 16.

TheVillageGuy:
> > > I can confirm that zooming as well as panning works as expected, no issues on this iPhone 6s in Safari.
> > 
> > 
> > I don't know anything about iPhones but I'd recommend testing on older models as well
> 
> I AM testing it on a very old iPhone. Like I said an iPhone 6s from 2015. We currently are already at iPhone 16.

Oh excellent, in that case manager the lightbox developers have done something with my feedback at the time

PR #1409, opened 2025-01-31
Fix NRE within the `ExternalLinkRenderer`

BentiGorlich:


melroy89 (approved):
Yup fixes the problem I was seeing on my instance. thx for the quick fix.

PR #1410, opened 2025-01-31
Add endpoint to create bookmark list

BentiGorlich:
- add an endpoint to create a bookmark list
- change the http method to set a list as default to PUT


jwr1 (changes requested):
Sorry to add more work, but could the bookmark-list create route be `POST` and the edit route be `PUT`.

And I don't know if you'd want to go ahead and change the four bookmark adding/removing routes to `PUT` in this PR also.

BentiGorlich:
Sure no problem, I'll change that

jwr1 (approved):
Thanks, LGTM!

PR #1415, opened 2025-02-03
Fix notification api route methods

jwr1:
I found two more routes that need their methods changed:

- The notification settings route should use `PUT` to indicate that it updates a setting.
- The notification test route should use `POST` to indicate that something happens on the backend (sends a test notification). This route was already released, but I do know both Interstellar and joinmbin.org don't use this route, and I did a quick check through a GitHub search and didn't find anything else, so I think we'll be fine updating this route. I'd say out of all the routes that exist, this is probably one of the least important ones anyway.

---

Just as an aside, for those who don't know, HTTP requests with the `GET` method are not supposed to have side effects. They should exclusively be a read-only operation that returns data from the server (nothing else should happen). For data mutating routes, `POST` is best for a create operation, `PUT` for update, and `DELETE` for deleting a resource.

### DIFF
--- a/config/mbin_routes/notification_api.yaml
+++ b/config/mbin_routes/notification_api.yaml
@@ -62,5 +62,5 @@ api_notification_push_unregister:
 api_notification_push_test:
     controller: App\Controller\Api\Notification\NotificationPushApi::testSubscription
     path: /api/notification/push/test
-    methods: [ GET ]
+    methods: [ POST ]
     format: json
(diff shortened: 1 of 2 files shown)


melroy89:
I agree with this RESTful approach. I think it's indeed a common practice. In both cases.

jwr1:
Both of the routes I changed should only be used in the external facing API, not the ajax routes that the Mbin web client uses. For example, I just tried the "Test Push Notifications" button, and the browser made a request to `/ajax/test_push` not `/api/notification/push/test`. We can let @BentiGorlich verify though, just to make sure.

BentiGorlich:
if it is in the `/api` route than it is not used by the web frontend :)

BentiGorlich (approved):

PR #1417, opened 2025-02-03
Update npm packages

melroy89:
Small **minor** release updates in NPM packages. So basically `npm update`.

Only updated the `simple-icons-font` package to a new major release. So far running at kbin.melroy.org, without issues.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -29,7 +29,7 @@
                 "regenerator-runtime": "^0.14.0",
                 "sass": "^1.69.5",
                 "sass-loader": "^16.0.0",
-                "simple-icons-font": "^13.21.0",
+                "simple-icons-font": "^14.5.0",
                 "stimulus-textarea-autogrow": "^4.1.0",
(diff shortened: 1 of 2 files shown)


jwr1 (approved):
Are we really using a whole font for two icons? All I can see are these two:
https://github.com/MbinOrg/mbin/blob/d7ed2a837735b864467590d9229d3b33b48d8012/templates/components/login_socials.html.twig#L29
https://github.com/MbinOrg/mbin/blob/d7ed2a837735b864467590d9229d3b33b48d8012/templates/components/login_socials.html.twig#L37

Either way, upgrade looks good :+1:.

melroy89:
> Are we really using a whole font for two icons? All I can see are these two:

haha yes apparently. Maybe we could use other fonts we already have like font awesome (`fa`)..? Maybe a good follow-up. Will create a ticket for that.

melroy89:
https://github.com/MbinOrg/mbin/issues/1421

PR #1420, opened 2025-02-03
Add missing cascade deletes to NotificationSettings

BentiGorlich:
- add the missing cascade deletes to the foreign keys of NotificationSettings


BentiGorlich:
I have no idea why the tests are failing... I doubt that it has anything to do with the changes...

melroy89 (approved):
Now the PHP file is added as well. Great! thx

melroy89:
However.. integration test is still failing consistently. you might want to check that out still. Integration test does not fail on other PRs like: https://github.com/MbinOrg/mbin/pull/1426

melroy89:
_I triggered the integration test manually again_

BentiGorlich:
Well I think the problem is this line:
>  1) Exception in third-party event subscriber: There is already an active transaction

Which I also have in the AP PR and I have no idea where that should come from... I cannot reporoduce it locally...
This command runs all tests like in the CI and it just works:
`SYMFONY_DEPRECATIONS_HELPER=disabled ./vendor/bin/paratest tests/Unit && php vendor/bin/phpunit tests/Functional --group NonThreadSafe && php vendor/bin/paratest tests/Functional --exclude-group NonThreadSafe`

BentiGorlich:
Normally each test is executed in its own transaction which leads to an isolation, but the error I sent says that that isn't working. I just have no clue as to why

melroy89:
is it conflicting with other runs like from the main branch maybe? Just thinking out loud here.

BentiGorlich:
what the actual fuck

melroy89:
Maybe we need to add try/catch with a `rollBack()` method. This isn't an error in LoginControllerTest. But maybe in setUp or where does it get triggered? 

The HTML response is also not helpful. Maybe we should add the logs as an artifact to GitHub? So you can see the logs. This can be achieved via https://github.com/actions/upload-artifact. And needs to be used via `if: always()` so also in case the job fails, it will then upload the log file.

BentiGorlich:
All of the duplicate errors are caused by the transaction wrapping not working due to the error message I linked

> Exception in third-party event subscriber: There is already an active transaction

I just have no idea why there would be an active transaction in there...

It seems like it is just the non thread safe integration tests, but still no idea why as I cannot reproduce it locally

melroy89:
What about creating a new branch? And try again. And see when the integration test fails..? Unless you already know the root cause.

BentiGorlich:
I did :( #1429 #1428

BentiGorlich:
It is definitely the migration, I just don't know why...

BentiGorlich:
I'm 100% certain that the cause of all the errors is that the DB isolation that the doctrine test bundle is supposed to do is not working, because there already is a transaction in every of the test cases... I have no idea where that would be coming from and I tried to put every `beginTransaction` that we are calling behind an if so that it is not executed in the test environment which also didn't help... I added a commit and a close connection to the bootstrap which also did nothing... I am clueless where this other transaction would be coming from...

melroy89:
...Rebasing trying the new docker image now..

melroy89 (approved):

PR #1425, opened 2025-02-04
Fix the bookmark API and add tests for it

BentiGorlich:
- Add the `bookmarks` property to entry, entry_comment, post and post_comment responses
- Fix problem with scopes
- `makeDefault` API endpoint now returns the list
- The response from editing a list now contains the correct values
- Add tests for bookmarking entry, entry_comment, post and post_comment to the default list and a specific list
- Add tests for creating, editing and deleting lists as well as making it default


melroy89:
Deployed on kbin.melroy.org atm

jwr1 (changes requested):
Could it be made so the default when you're signed out is `null`, to indicate that you couldn't have any bookmarks anyway? But when you're signed in, it keeps the empty array (assuming it's not in a list).

jwr1:
For some reason, the new `bookmarks` property is showing up in the API docs for the EntryResponse, but not for posts, entry-comments, or post-comments. When actually testing the API, it does look like the property is there though.

melroy89:
> > > > `makeDefault` API endpoint now returns the list
> > > 
> > > 
> > > Maybe the docs just didn't get updated correctly, because the API docs still says the makeDefault route doesn't return anything.
> > 
> > 
> > Be sure to check https://kbin.melroy.org/api/docs. Other documentation pages will only be update later if it's merged.
> 
> Yup, that's what I was using.

OK great. That is good.

This is what it shows:

image

The 200 OK response only says: "Sets the provided list as the default". There is no type definition added, so no response schema. And no example.

So you only see this description text. If that is not sufficient and you want more details. Then this file needs to be extended as well: https://github.com/MbinOrg/mbin/blob/649dcdfcd8be7aa92524cf391c9a9c191c5f983d/src/Controller/Api/Bookmark/BookmarkListApiController.php#L222

melroy89:
kbin.server.melroy.org is rebased. `makeDefault` does now have a `content` and swagger will show:

image

If this is an array instead of a single object? Then the PHP code still needs to be adapted.

jwr1:
Two more things:

- The api docs looks mostly correct now, but I did see the EntryResponseDto still has the bookmarks property marked as required, even though it is optional.
- Whenever I try to use the `/api/bookmark-lists` route, all I'm getting is a 403 Forbbiden error.

BentiGorlich:
Did you use the correct scope? Because there are new ones

jwr1:
Do you mean the `user:bookmark_list` scope, because I thought the `user` scope covered anything that was prefixed with `user:`?

BentiGorlich:
Yeah, I think it doesn't... Getting to know these scopes and I think it just doesn't belong in the user scope and I think this inheritance is not happening automagically

jwr1 (dismissed):
Adding the `user:bookmark_list` scope did fix it.

 I'm still testing a few more things, but so far I'd say everything is looking good now.

melroy89:
`BookmarkListDto2` schema is missing on `PUT /api/bookmark-lists/{list_name}`:

image

melroy89:
I will move away from this branch, just saying @jwr1 .

BentiGorlich:
I will fix the scope thing and the missing response type :)

jwr1 (approved):

PR #1426, opened 2025-02-04
Fix doctrine weirdness

BentiGorlich:
- remove an unnecessary mapping type
- remove a column comment that is confusing doctrine

If we only removed the mapping type the `php bin/console doctrine:migrations:diff` command would throw an error because of the comment on the column


melroy89 (approved):
That seems to be a copy/paste issue from the past

PR #1431, opened 2025-02-06
Update messages.en.yaml

TheVillageGuy:
Shorten subscribe/unsubscribe to prevent buttons forming two rows

### DIFF
--- a/translations/messages.en.yaml
+++ b/translations/messages.en.yaml
@@ -68,8 +68,8 @@ subscribe_for_updates: Subscribe to start receiving updates.
 federated_user_info: This profile is from a federated server and may be incomplete.
 go_to_original_instance: View on remote instance
 empty: Empty
-subscribe: Subscribe
-unsubscribe: Unsubscribe
+subscribe: Sub
+unsubscribe: Unsub
 follow: Follow
 unfollow: Unfollow
 reply: Reply


melroy89:
Not sure why the integration tests are failing. Anyhow, you could also change it here: https://hosted.weblate.org/engage/mbin/ and sync it back. But both works.

melroy89:
Also we might still want to show the full line "Subscribe" and "Unsubscribe" to the users when they mouse-over the button, using the HTML `title` tag. Hence it might be better to create a new translation string for the abbreviation.

BentiGorlich:
@TheVillageGuy some tests do use text to recognize some buttons so when you change it it will not find those buttons anymore and then you'll get errors like "current node list is empty".
When you go to the failed pipeline you will get the test errors with the files they originate from. Then you can fix them

TheVillageGuy:
> Also we might still want to show the full line "Subscribe" and "Unsubscribe" to the users when they mouse-over the button, using the HTML `title` tag. Hence it might be better to create a new translation strings for the abbreviations.

Good point, I'll add them soon. 

@BentiGorlich I'm going to have to figure out how those tests work then, that's going to take some time

blued-gear:
Updated and should be ready.

But honestly I don't know if I like the look of a button with just 'Sub' on non-mobile sized screens...

PR #1432, opened 2025-02-06
Add border to non federated notifications for admins and moderators

TheVillageGuy:
Added new_signup block


melroy89:
What are those random `><div>` doing in the _blocks.html.twig??

templates/notifications/_blocks.html.twig
```diff
@@ -1,106 +1,152 @@
...
 {% endblock entry_comment_edited_notification %}
 
 {% block entry_comment_reply_notification %}
+    ><div>
     {{ component('user_inline', {user: notification.comment.user}) }} {{ 'replied_to_your_comment'|trans|lower }} - <a
         href="{{ entry_comment_view_url(notification.comment) }}#{{ get_url_fragment(notification.comment) }}">{{ notification.comment.shortTitle }}</a>
 {% endblock entry_comment_reply_notification %}
 
 {% block entry_comment_deleted_notification %}
+    ><div>
```
BentiGorlich, line 54: I honestly thing that this is just bad practice... It might do the job but it is not easy to maintain or to read...

BentiGorlich (changes requested):

templates/notifications/_blocks.html.twig
TheVillageGuy: I agree, though I don't see another way of achieving it without reorganizing the entire notifications code

BentiGorlich: You can set the class in an if in the `front.html.twig` inside of the notifications folder. As far as I can tell you just want to do it for specific notification types right? You have the `notification.type` variable there to check that

TheVillageGuy:
> What are those random `><div>` doing in the _blocks.html.twig??

I apologize for missing your comment, they are not random, they are there to carry the notification div into the blocks twig. Due to the way the blocks are called this was the only way to allow for specific logic/css to be applied to them without having to rewrite a lot of code, just to achieve a visual indicator

templates/notifications/_blocks.html.twig
TheVillageGuy: So creating a new class for every kind of notification, I guess that would work. The benefit of having the logic in the blocks is that in the future we can do more than just a coloured border, line adding a symbol for different situations, I'm not sure what idea I had on that about though, I think indicating when something is done by a local user, not just locally.

PR #1435, opened 2025-02-06
Add missing colons to GitHub Issues template for bug reports

MHLoppy:
The GitHub Issues template for bug reports is inconsistent with its usage of colons. This looks like it wasn't intentional (?) so I added colons to the points missing it.


melroy89 (approved):
Thank you!

Issue #1408, opened 2025-01-31
New registration notification is not marked as read


melroy89:
**Describe the bug**

When a new user is register, a new notification will now pop-up for admins to approve or reject.

**On which Mbin instance did you find the bug?**
kbin.melroy.org

**Which Mbin version was running on the instance?**
main

**To Reproduce**
1. Go to Notifications
2. See the registration notification

**Expected behavior**

A link to the approval registration page. (incl the correct page number), which is required when you the "New users have to be approved by an admin before they can log in." feature is checked.

The username link will never work when admin approval is required. Only show a link to the user if the admin approval feature is disabled.

**Screenshots**

![Image](https://github.com/user-attachments/assets/c8714236-8779-465e-93e4-f737eb03e330)

BentiGorlich:
Again this is only true when approvals are enabled in the admin settings. When approvals are not needed the link is working fine :D

melroy89:
I edited my desc.

melroy89:
Fixed in https://github.com/MbinOrg/mbin/pull/1411? Right?

melroy89:
Well except for "Only show a link to the user if the admin approval feature is disabled."

BentiGorlich:
Hmm... Do you think we should remove the user_inline component? I mean maybe we could change the page it links to to the same signup request?

melroy89:
Yes you can either disable the link I guess. Or also point to the approval page.

melroy89:
@BentiGorlich Can we also "mark as read" the notification when clicking on the 'Open signup request' link?

melroy89:
Notification doesn't get marked as read after clicking on the Open signup request link. Hence I reopen the ticket.

melroy89:
I suspect an issue here somewhere: https://github.com/MbinOrg/mbin/blob/ff07ef7e0db00d709c511c1e9f21c148e04de17e/src/Controller/User/UserFrontController.php#L58

Why is this call in the UserFrontController? And more importantly why the `if ($loggedInUser = $this->getUser())` check? As the admin I'm never the same person who is logged in.

Furthermore, I would like to at least add `markUserSignupNotificationsAsRead` to the src/Controller/Admin/AdminSignupRequestsController.php file.

BentiGorlich:
> Can we also "mark as read" the notification when clicking on the 'Open signup request' link?

I actually thought that I did that...

melroy89:
[...]

PR #1437, opened 2025-02-06
Mark signup notification as read, after clicking on the link

melroy89:
- Mark the new signup notificaction "as read" after the admin clicked on the "Open signup request" link (just like we have with other notifications like the report notification, which is using the `markReportNotificationsAsRead` in the AdminReportController.php for example)

Fixes: https://github.com/MbinOrg/mbin/issues/1408


src/Controller/Admin/AdminSignupRequestsController.php
```diff
@@ -31,6 +33,7 @@ public function requests(#[MapQueryParameter] ?int $page = 1, #[MapQueryParamete
             if ($signupRequest = $this->repository->findSignupRequest($username)) {
                 $requests[] = $signupRequest;
             }
+            $this->notificationRepository->markUserSignupNotificationsAsRead($this->getUserOrThrow(), $username);
```
BentiGorlich, line 36: I would put that in the `if`  just above this code

melroy89: I rejected the the spam user, so I get:

image

melroy89: But still the notification stays unread now...
image

melroy89: So sorry, I will keep it after the `if` check.. now it works! So it get also marked as read in case you already rejected the request via another way and `findSignupRequest` is empty.

BentiGorlich (approved):

PR #1440, opened 2025-02-07
Adjust our pipeline docker file to use the prebuild ubuntu packages

BentiGorlich:
- Introduce a newer Dockerfile based on Ubuntu LTS image
- Add a workflow file for building & deploying the docker image to ghcr inside GitHub Actions (and also triggers on changes in the ci/Dockerfile when merged to main)

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -1,21 +1,8 @@
-FROM php:8.3
+# Using the latest LTS release
+FROM ubuntu:latest
 
-# Add composer
-COPY --from=composer:2.7.8 /usr/bin/composer /usr/bin/composer
-
-# Add NodeJS
-RUN apt-get update && apt-get install -y nodejs npm
-
-# Add additional APT packages
-RUN apt-get update && apt install -y unzip postgresql
-
-# Add additional PHP packages
-RUN curl -sSLf \
-  -o /usr/local/bin/install-php-extensions \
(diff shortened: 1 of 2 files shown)


ci/Dockerfile
```diff
@@ -1,21 +1,7 @@
-FROM php:8.3
+FROM ubuntu:latest as base
```
melroy89, line 1: Maybe add a small remark that `latest` is actually LTS.

```suggestion
# Using the latest tag actually means LTS
FROM ubuntu:latest as base
```

according to https://hub.docker.com/_/ubuntu

melroy89, line 1: Also you do not need `as base`...  This is only needed when you are doing complex multi stage Docker builds: https://docs.docker.com/build/building/multi-stage/#name-your-build-stages

Which you aren't doing here.

melroy89:
I also added a workflow file for you, which will be triggered in the future automatically when the Dockerfile is changed on the main branch. Or in case you trigger it manually.

melroy89:
I also set `DEBIAN_FRONTEND=noninteractive`, perform an apt upgrade first. And write the ini files to the correct directory.

melroy89 (dismissed):

melroy89:
Once this is merged and the first image is published, we can create a follow-up PR that will use this new Docker image from ghrc and see if it still works!

melroy89 (approved):

PR #1451, opened 2025-02-12
Push notification: add notification id

BentiGorlich:
- received push notification objects will have the notification id from now on (if it is not a test notification)


jwr1 (approved):
Looks good, thanks!

PR #1454, opened 2025-02-14
Render lemmy community links as preview

BentiGorlich:
A lemmy community link often looks like: `[!community@domain.tld](https://domain.tld/c/community)` which is just rendered as a link at the moment, render such links as the new inline preview

### DIFF
--- a/src/Markdown/CommonMark/CommunityLinkParser.php
+++ b/src/Markdown/CommonMark/CommunityLinkParser.php
@@ -97,7 +97,7 @@ class CommunityLinkParser implements InlineParserInterface
         if ('[' === $prev && ']' === $next && '(' === $nextNext) {
             $closing = null;
             $link = '';
-            for ($i = 2; $char = $cursor->peek($i); ++$i) {
+            for ($i = 2; null !== ($char = $cursor->peek($i)); ++$i) {
                 if (')' === $char) {
                     $closing = $i;
                     break;
--- a/src/Markdown/CommonMark/CommunityLinkParser.php
+++ b/src/Markdown/CommonMark/CommunityLinkParser.php
@@ -43,6 +43,8 @@ class CommunityLinkParser implements InlineParserInterface
         $isRemote = $this->isRemoteCommunity($domain);
         $magazine = $this->magazineRepository->findOneByName($isRemote ? $fullHandle : $handle);
 
+        $this->removeSurroundingLink($ctx, $handle, $domain);
+
         if ($magazine) {
             $ctx->getContainer()->appendChild(
                 new CommunityLink(
@@ -61,7 +63,7 @@ class CommunityLinkParser implements InlineParserInterface
             $ctx->getContainer()->appendChild(
                 new ActorSearchLink(
                     $this->urlGenerator->generate('search', ['search[q]' => $fullHandle], UrlGeneratorInterface::ABSOLUTE_URL),
-                    '!'.$handle,
+                    '!'.$fullHandle,
                     '!'.$fullHandle,
                 )
             );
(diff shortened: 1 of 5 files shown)


melroy89 (approved):

BentiGorlich:
I think I want to make one change to this. As learned from previous inline preview changes: the link should only be rendered as inline preview if the linktext is the same as the magazine name

melroy89:
Maybe related, but I don't fully get this message from Arthur: https://lemmy.ml/comment/16776661

src/Markdown/CommonMark/ExternalLinkRenderer.php
```diff
@@ -136,7 +141,47 @@ public function render(Node $node, ChildNodeRendererInterface $childRenderer): H
...
+                    // technology@lemmy.world
+                    $name = $magazine->name;
+                    if (!str_starts_with($name, '@') && !str_starts_with($name, '!')) {
+                        // @technology@lemmy.world
+                        $name = "@$name";
+                    }
+                    // technology
+                    $name2 = $this->mentionManager->getUsername($name);
+                    // !technology
+                    $name3 = '!'.$this->mentionManager->getUsername($name);
```
melroy89, line 171: You're calling `his->mentionManager->getUsername` twice, which sounds inefficient.

melroy89: Same for above on line 150.

BentiGorlich:
Ok so I moved the logic into the parsers now, where I should have put them initially. I also wrote a few tests :)

src/Markdown/CommonMark/CommunityLinkParser.php
```diff
@@ -79,4 +81,40 @@ private function isRemoteCommunity(?string $domain): bool
...
+    /**
+     * Removes a surrounding link from the parsing container if the link contains $handle and $domain.
+     *
+     * @param string $handle the user handle in [!@]handle@domain
+     * @param string $domain the domain in [!@]handle@domain
+     */
+    public static function removeSurroundingLink(InlineParserContext $ctx, string $handle, string $domain): void
+    {
+        $cursor = $ctx->getCursor();
+        $prev = $cursor->peek(-1 - $ctx->getFullMatchLength());
```
melroy89, line 94: What happens when `$prev`, `$next`  and/or `$nextNext` are null? I quote from the docs:

 "Returns the next character (or null, if none) without advancing forwards" 

https://github.com/thephpleague/commonmark/blob/64cba7fd2dcc04e33b1ae25dfd0b17a1f2309c52/src/Parser/Cursor.php#L187

We could skip all the code if one of these values are null right?

tests/Unit/Utils/MarkdownTest.php
```diff
@@ -0,0 +1,92 @@
...
+
+declare(strict_types=1);
+
+namespace App\Tests\Unit\Utils;
+
+use App\Markdown\MarkdownConverter;
+use App\Markdown\RenderTarget;
+use App\Tests\WebTestCase;
+
+class MarkdownTest extends WebTestCase
```
melroy89, line 11: Nice 👍🏽

src/Markdown/CommonMark/CommunityLinkParser.php
```diff
@@ -79,4 +81,40 @@ private function isRemoteCommunity(?string $domain): bool
...
+            $link = '';
+            for ($i = 2; null !== ($char = $cursor->peek($i)); ++$i) {
+                if (')' === $char) {
+                    $closing = $i;
+                    break;
+                }
+                $link .= $char;
+            }
+            if (null !== $closing && str_contains($link, $handle) && str_contains($link, $domain)) {
+                // this is probably a lemmy community link a lá [!magazine@domain.tld](https://domain.tld/c/magazine]
```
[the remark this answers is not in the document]
BentiGorlich: No, we do not have to check for `)` because closing is either `null` or `)` just because of the code before

src/Markdown/CommonMark/CommunityLinkParser.php
BentiGorlich: since all three values are checked in the if just beneath it this is already prevented

melroy89 (approved):

PR #1460, opened 2025-02-17
Some opcache and composer optimizations

BentiGorlich:
- add opcache preload to docs and docker config
- add disable opcache timestamp validation in docs and docker config
- enable composers authoritative classmap

### DIFF
--- a/docs/02-admin/01-installation/01-bare_metal.md
+++ b/docs/02-admin/01-installation/01-bare_metal.md
@@ -265,11 +265,12 @@ realpath_cache_size = 4096K
 realpath_cache_ttl = 600
 ```
 
-Optionally also enable OPCache for improved performances with PHP (recommended for both fpm and cli ini files) (replace `/var/www/mbin` with `/var/www/kbin` if needed):
+Optionally also enable OPCache for improved performances with PHP (recommended for both fpm and cli ini files):
 
 ```ini
 opcache.enable = 1
 opcache.enable_cli = 1
+;replace `/var/www/mbin` with `/var/www/kbin` if needed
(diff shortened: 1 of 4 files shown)


docs/02-admin/01-installation/01-bare_metal.md
```diff
@@ -265,16 +265,19 @@ realpath_cache_size = 4096K
 realpath_cache_ttl = 600
 ```
 
-Optionally also enable OPCache for improved performances with PHP (recommended for both fpm and cli ini files):
```
melroy89, line 268: maybe not use parentheses `()` after existing parentheses ;). A new line or note should make more sense. Or even as a comment in the ini file `;`

melroy89 (approved):

melroy89 (changes requested):
Owh. We should also update the docs for `opcache.preload=/var/www/mbin/config/preload.php`

BentiGorlich:
I thought that that is just the preload and there can always only be one preload... But not that it would cause issues with other websites...

melroy89:
> I thought that that is just the preload and there can always only be one preload... But not that it would cause issues with other websites...

Here is your proof:

```
Fatal error: Cannot redeclare Aws\constantly() (previously declared in /var/www/kbin.melroy.org/html/vendor/aws/aws-sdk-php/src/functions.php:20) in /var/www/nextcloud/html/3rdparty/aws/aws-sdk-php/src/functions.php on line 20
```

BentiGorlich:
oooooohhhhhhhh

melroy89 (approved):

melroy89:
See follow-up: https://github.com/MbinOrg/mbin/pull/1460

PR #1461, opened 2025-02-17
Remove the `opcache.preload` setting for the docker build

BentiGorlich:


melroy89 (approved):
It's weird this won't work in docker, owh well.

PR #1464, opened 2025-02-18
Translations update from Hosted Weblate

weblate:
Translations update from [Hosted Weblate](https://hosted.weblate.org) for [Mbin/Mbin](https://hosted.weblate.org/projects/mbin/mbin/).

Current translation status:

Weblate translation status

PR #1465, opened 2025-02-18
Update messages.en.yaml

TheVillageGuy:
List thumbs open images vs topics


melroy89:
See PR: https://github.com/MbinOrg/mbin/pull/1464

PR #1468, opened 2025-02-19
Add all needed instructions for php install on debian 12

TheVillageGuy:


melroy89 (approved):
Exactly, as discussed via chat. This is the correct way of doing it today.

PR #1472, opened 2025-02-19
Improve modlog (use showAvatars on all user_inline + possibility to hide new icon from modlog)

melroy89:
Goal: Have a very clean modlog if you want to.

- Currently the modlog was still showing user avatars despite the `MBIN_MODERATION_LOG_SHOW_USER_AVATARS` is set to false. This PR fixes that.
- Give the inline user and inline magazine the ability to hide the new icon (enabled by default across mbin)
- Add only an extra option for the modlog to disable the "new icon" to have a **clean modlog** (new icon is enabled by default in modlog)

This will now finally give a clean modlog (except for the new icons on users & magazines):

image

If you wish to have icons/avatars again, use the options under "Moderation log". If you wish to have no new icon, you can disable this by the new option to hide the new icon.

The hide "new icon" mod log option (by default its showing new icons), will look like this in the modlog:

image


BentiGorlich:
Why hide the new user icon? This is actually important information especially in the modlog imo

melroy89:
> Why hide the new user icon? This is actually important information especially in the modlog imo

It's an icon that is actually distracting IMO, if you want icons you can show the icons/avatar option in the settings. I only implemented this for inline user and inline magazine btw.

melroy89:
We could also make an additional setting option for the new icon?

BentiGorlich:
We do not hide it anywhere depending on any kind of setting, why introduce it in the modlog? And if we were to introduce another setting will we have a battle over the default again? 

We have way too many settings already, I am against introducing new settings. We might consolidate settings with just a general "display icons" setting which switches every avatar and icon on or off. And if you want we can include that icon und that new setting as well, but imo it should be a separate PR, no matter which option you/we may choose

melroy89:
> And if we were to introduce another setting will we have a battle over the default again?

Yes I'm also getting sick of this.

melroy89:
I will make it default to show new icons.

melroy89:
I know it's _again a sensitive topic_ (what is not anymore nowadays?), but the new icon was created by you in this PR:  https://github.com/MbinOrg/mbin/pull/969. 

There was no ability to control this new feature within that PR. This PR fixes that as well and makes it by default `true` everywhere (so there is no impact for you and others).

The new option is only introduced for moderation log, in all other cases it's by default set to `true` like I said (just like `showAvatar: true` is also set to true in most cases). Even the default value for the new icon in the moderation log is set to true for you. 

@BentiGorlich  Are you using the modlog a lot?

melroy89:
> We need to create a workflow for changing the UI... UI is a very subjective thing and we should not emd up with every change behind a setting. Basically we need to find something at least most of us are fine with regarding the UI.
> 
> An example: I add this icon. Melroy saying that is very distracting in the modlog and then finding the conpromise to not make the icon green.
> 
> Again just an example, but something like that

To by fully transparent here, this the background story: It started small (as you can see at the branch name: `use_showAvatar_on_all_userinline_modlog`) just to cleaning-up the modlog.

Since I notice the avatar icon should be disabled by default in the mod log option, yet it was still showing in the modlog... 

This is where I started, then I fixed this with the user avatars issue.. And the next thing that popped-up was the still showing icons (the "new" icon). One thing led to another. And now we ended up again with UI changes indeed.

src/Twig/Components/MagazineInlineComponent.php
```diff
@@ -15,4 +15,5 @@ final class MagazineInlineComponent
     public bool $fullName = false;
     public bool $stretchedLink = false;
     public bool $showAvatar = false;
```
melroy89, line 17: TIL: These variables are defined in the Twig component as well with default values. I believe this `showAvatar` should be set to `true` as well, since that is the default option we also have in the GUI. And then its consistent with the `showAvatar` value as well in `UserInlineComponent.php` (which was already true).

After that, I believe we can remove a lot of unnecessary setting booleans to true, and see if that this works. Since then Twig will just use the default boolean.

BentiGorlich (approved):
We sure have a lot of uses of the magazine and user inline components huh? :D

melroy89:
> We sure have a lot of uses of the magazine and user inline components huh? :D

Yes we have indeed. Maybe a follow-up PR later can clean-up the Twig mess a bit, if we for example would change the default: `public bool $showAvatar = false;`  to `true`, so all the cases where we now need to set booleans, shouldn't be needed and only when you want to deviate from the default values should need to change.

PR #1487, opened 2025-03-03
Extend public access roles for three API end-points

melroy89:
When the server is in "Force users to login before they can access any content" mode, still allow the following 3 API end-points to be accessible for requests. Which is also used for: https://joinmbin.org/servers

- `/api/info`
- `/api/instance`
- `/api/defederate`

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -123,6 +123,7 @@ security:
         - { path: ^/authorize, roles: PUBLIC_ACCESS }
         - { path: ^/token, roles: PUBLIC_ACCESS }
         - { path: ^/api/doc, roles: PUBLIC_ACCESS }
+        - { path: ^/api/client, roles: PUBLIC_ACCESS }
         - { path: ^/api/info, roles: PUBLIC_ACCESS }
         - { path: ^/api/instance, roles: PUBLIC_ACCESS }
         - { path: ^/api/defederate, roles: PUBLIC_ACCESS }
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -123,7 +123,9 @@ security:
         - { path: ^/authorize, roles: PUBLIC_ACCESS }
         - { path: ^/token, roles: PUBLIC_ACCESS }
         - { path: ^/api/doc, roles: PUBLIC_ACCESS }
-        - { path: ^/api/client, roles: PUBLIC_ACCESS }
+        - { path: ^/api/info, roles: PUBLIC_ACCESS }
+        - { path: ^/api/instance, roles: PUBLIC_ACCESS }
+        - { path: ^/api/defederate, roles: PUBLIC_ACCESS }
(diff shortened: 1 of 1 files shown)


config/packages/security.yaml
```diff
@@ -124,6 +124,9 @@ security:
         - { path: ^/token, roles: PUBLIC_ACCESS }
         - { path: ^/api/doc, roles: PUBLIC_ACCESS }
         - { path: ^/api/client, roles: PUBLIC_ACCESS }
+        - { path: ^/api/info, roles: PUBLIC_ACCESS }
+        - { path: ^/api/instance, roles: PUBLIC_ACCESS }
+        - { path: ^/api/defederate, roles: PUBLIC_ACCESS }
```
BentiGorlich, line 129: I think we cannot do this one (as it is used for admin purposes as well): 
grafik

Maybe if we restricted the method (don't know if that is possible here), but I think I would just exclude it

BentiGorlich (changes requested):

config/packages/security.yaml
melroy89: Uhm I only want to allow `GET` calls indeed. These PUT calls are behind authorization anyways.

melroy89: The PUT call still seems to trigger a 401 response (which is what we want):

image

melroy89: Now let me try with a normal account logged in.

BentiGorlich: Any you gotta change it to:
```suggestion
        - { path: ^/api/defederated, roles: PUBLIC_ACCESS }
```

melroy89: OK, I now created

1. A new oauth client with all the scopes.
2. Authorize it against a normal user account
3. Retrieved a bearer access token
4. Executed the PUT `api/defederated` call with the bearer token
5. I got a 403 (forbidden)

image

BentiGorlich (approved):

BentiGorlich:
maybe also add the federated and dead routes?

melroy89:
Your review is now auto dismissed 😢

BentiGorlich (approved):

jwr1:
So, then should we make the corresponding web pages public as well? Like the about, privacy policy, faq, terms, federation page, etc.

melroy89:
> So, then should we make the corresponding web pages public as well? Like the about, privacy policy, faq, terms, federation page, etc.

Some of these makes sense yes. Especially about, privacy policy a d terms.

jwr1:
Well, all the ones I listed are now public through the API, right? So, the corresponding web pages should be made public as well.

melroy89:
> Well, all the ones I listed are now public through the API, right? So, the corresponding web pages should be made public as well.

Yes, only as long as there is a Twig caching in place for these pages. So the impact on the server-side should be minimal.

PR #1492, opened 2025-03-04
Fix missing `Update` activity add cover and avatar removal

BentiGorlich:
- add option to remove the user avatar or cover in the profile settings. This uses the same style as the media removal in the entry edit
- fix missing outwards federation when deleting the image through the API
- fix inwards federation when an image was removed, this was previously ignored
- add error message when the user submits an image that is too large

Delete user iamges:
Bildschirmfoto vom 2025-03-04 16-42-30

Error message:
Bildschirmfoto vom 2025-03-04 16-42-55


BentiGorlich:
This PR is a bit of a mash up, sorry :sweat_smile:

melroy89 (approved):

PR #1499, opened 2025-03-14
Extend mention regex with period + no char limit of 30

melroy89:
Seems to work correctly now, tested it with: https://regex101.com/

- Removing the 30 char limit, just use `+` meaning one or many matches
- Add the `.` (period) symbol as allowed chars

There are several fediverse platforms (software) that support and allow dots in the username. This will take care of this, and render the inline correctly.


melroy89:
This code is wrong.

melroy89:
New PR: https://github.com/MbinOrg/mbin/pull/1533

PR #1501, opened 2025-03-15
Add mbin logo bw (transparent)

melroy89:
I will help you out a bit, here is a SVG black / transparent logo. Allowing any color to be set in the code to override the black, and any background color can also be set to change the background.

Currently this SVG is just stored under `images/sources`.

I will use this SVG also for submitting a PR towards https://fediverse.observer/


TheVillageGuy:
For it to work with my automatic colour by username pr it has to be non black/white. Preferably a bright colour like FF0000

melroy89:
Ignore the 'bw' in the title. Like I said, you can use the magic of CSS with `color: ` to set the logo color (any color you want). And use the `background-color:` to set the background color (any color you want).

PR #1505, opened 2025-03-17
Add .vs directory to gitignore and dockerignore

melroy89:
See title.


TheVillageGuy (approved):
This is an example of sublime coding! Screw VS

PR #1514, opened 2025-03-23
Remove DMs and ban sender users command

melroy89:
Reducing spam messages (like the infamous fediverse chick) by introducing a command. Meaning admins no longer need to use manual DB queries. Its just simpler.

This new command (`./bin/console mbin:messages:remove_and_ban`), allows you to easily search on direct messages. And remove all the matched messages and ban the senders user from the instance.

Hint: If you want to search on strings with spaces its just a shell script basically, so use quotes around the body query in those cases and it will work fine:

```sh
./bin/console mbin:messages:remove_and_ban "Fediverse Chick"
```

----

1. Search query is executed. And displayed in a very nice ASCII table:
image
2. User input requested to continue: 
image
3. If y(es), then we continue with banning the users and removing the direct messages.

### DIFF
--- a/src/Command/RemoveDMAndBanCommand.php
+++ b/src/Command/RemoveDMAndBanCommand.php
@@ -1,0 +1,134 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Command;
+
+use App\Repository\UserRepository;
+use App\Service\UserManager;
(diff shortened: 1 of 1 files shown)


TheVillageGuy (approved):

melroy89:
Documentation follow up PR: https://github.com/MbinOrg/mbin/pull/1515

PR #1526, opened 2025-03-27
Inline all content previews

### LINKED ISSUE #1493: Link previews fail to render if the thread title is included in the URL
**Describe the bug**
Image
*What I wrote*

Image
*What renders*

Image
*HTML contents*

Image
*It gets federated correctly (screenshot taken from the Lemmy instance of the user I replied to)*

As you can see above, it seems Mbin tries to replace my link with something else but fails and only inserts a div despite removing the actual link, completely swallowing the actual link.

kbin.earth

1.8.0

**To Reproduce**
Create a comment containing this link: https://kbin.earth/m/fediversenews@piefed.social/t/976189/PieFed-adds-multi-reddit-multi-community-feature/favourites

**Expected behavior**
Either my link stays intact or gets replaced with the new inline preview feature, either way I expect it to display in some manner.

 - OS: Windows 11
 - Browser: Microsoft Edge
 - Browser Version:  132.0.2957.140

**Additional context**
[Link to the comment](https://kbin.earth/m/technology@lemmy.world/t/1006524/-/comment/5401565)
There's no errors in the console.

BentiGorlich:
- make them inline, so when you have `look here (some content link)` it does not get converted to: ``` look here ( some content preview ) ```
- fix links getting completely removed. The cause was the link-matching in the `ApActivityRepository` which sometimes returned an object with `0` as the id which of course does not yield results from the DB. Since that was just not expected we showed an empty div for these links
grafik

Closes #1493

### DIFF
--- a/src/Repository/ApActivityRepository.php
+++ b/src/Repository/ApActivityRepository.php
@@ -95,6 +95,10 @@ class ApActivityRepository extends ServiceEntityRepository
                         'type' => Post::class,
                     ];
                 } else {
+                    if (0 === $id) {
+                        return null;
+                    }
+
                     return [
                         'id' => $id,
                         'type' => PostComment::class,
@@ -108,7 +112,17 @@ class ApActivityRepository extends ServiceEntityRepository
                         'id' => $id,
                         'type' => Entry::class,
                     ];
+                } elseif (5 === \count($exploded)) {
+                    // entry url with slug (non-ap route)
+                    return [
+                        'id' => \intval($exploded[4]),
+                        'type' => Entry::class,
+                    ];
                 } else {
+                    if (0 === $id) {
+                        return null;
+                    }
+
                     return [
                         'id' => $id,
                         'type' => EntryComment::class,
--- a/src/Repository/ApActivityRepository.php
+++ b/src/Repository/ApActivityRepository.php
@@ -95,6 +95,7 @@ class ApActivityRepository extends ServiceEntityRepository
                         'type' => Post::class,
                     ];
                 } else {
+                    // since the id is just the intval of the last part in the url it will be 0 if that was not a number
                     if (0 === $id) {
                         return null;
                     }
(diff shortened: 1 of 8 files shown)


melroy89:
See it in action as well here: https://kbin.melroy.org/m/testing/t/728880/Test-markdown

It does seem to work now indeed! Thanks. It only doesn't work when the span is too large like a big title or something. Then the span will just be too large and cause the text to wrap anyways (or even the inside text in the span will wrap).

melroy89 (approved):

tests/Unit/Utils/MarkdownTest.php
```diff
@@ -89,4 +92,49 @@ public function testExternalMentionLink(): void
         // assert that this community does not exist, and we get a search link for it that does not link to the external instance
         self::assertStringContainsString('https://kbin.test2', $markdown);
     }
+
+    public function testExternalMagazineLocalPostLink(): void
```
melroy89, line 96: 👍🏽  for an additional test.

src/Repository/ApActivityRepository.php
```diff
@@ -95,6 +95,10 @@ public function findLocalByApId(string $apId): ?array
                         'type' => Post::class,
                     ];
                 } else {
+                    if (0 === $id) {
```
melroy89, line 98: Maybe also explain this code, just for future references so we still understand why this was added.

```suggestion
                    // Ignore processing $id of zero, because ...
                    if (0 === $id) {
```

src/Repository/ApActivityRepository.php
```diff
@@ -108,7 +112,17 @@ public function findLocalByApId(string $apId): ?array
...
                         'type' => Entry::class,
                     ];
+                } elseif (5 === \count($exploded)) {
+                    // entry url with slug (non-ap route)
+                    return [
+                        'id' => \intval($exploded[4]),
+                        'type' => Entry::class,
+                    ];
                 } else {
+                    if (0 === $id) {
```
melroy89, line 122: Same here :)

BentiGorlich: done :)

melroy89 (approved):

PR #1533, opened 2025-03-30
Fix period in remote usernames

melroy89:
- Fix period in remote usernames (`\w` for word, will not cut it), since some Fediverse software allows `@user.name`, for example, as a valid username)
- Removing the arbitrary 30 char limit as well
- Add code documentation

I tested it regex again with https://regex101.com/


jwr1 (approved):
Thank you! This looks good.

PR #1538, opened 2025-03-31
Fix SendApplicationAnswerMailHandler not supplying enough parameters

BentiGorlich:
SendApplicationAnswerMailHandler was not supplying the kernel interface parameter to the base class, fix that


BentiGorlich:
I looked through all the message handlers and could not find another one missing the second parameter

melroy89 (approved):

melroy89:
I think this was not the first similar error we discovered. So I'm saying this is the second time we found this bug.

Q: Is it possible to find any remaining issues in the code base? Maybe some intelligent PHP analyzer tool or something?

BentiGorlich:
I searched for it with a reference search in PHPStorm, I don't think that we can search more in depth than that. I think we had this problem before, but with a different message handler. I think the problem was that the message handler adaption was being done at the same time as this handler was introduced and then it was just forgotten...

PR #1542, opened 2025-04-04
Rework docker setup

jwr1:
## Changes to Docker setup

- Uses the https://github.com/dunglas/symfony-docker template, and in turn, https://github.com/dunglas/frankenphp.
- An external reverse proxy is not required, as the Caddy server built into FrankenPHP enables automatic HTTPS.
- The `.env` and `compose.override.yaml` files are now stored in the root of the Mbin directory, with all other server specific data stored in `storage/` (including OAuth keys).
- RabbitMQ now correctly persists data.
- All containers are now set up to work with non-root users.
- There's a new bash script that automates the setup of the docker environment.
- The default user for all services has changed from `kbin` to `mbin`.
- The default Postgres version has been bumped to `17`.
- Switch to using Valkey instead of Redis.
- Include a Valkey config with recommended default.
- Automatically enable the S3 FS adapter when an `S3_KEY` is provided.
- Enable permanent image redirect responses (`301` status code) by default.
- Includes a development server, making contributing much easier.

## Quick Start Guide

Use these steps to install Mbin on a new server. It is assumed that docker is already installed and set up.

1. Clone Mbin and change directories: `git clone https://github.com/MbinOrg/mbin.git && cd mbin` (make sure to switch to this docker branch).
2. Run the auto setup script: `./docker/setup.sh prod DOMAINHERE` (use `localhost` for the domain if you want to test locally).
3. Optionally configure SMTP, hCaptcha, S3, and/or SSO in `.env`.
4. Use `docker build -t mbin -f docker/Dockerfile .` to build the Docker image.
5. Run `docker compose up -d` to start the Docker containers.
6. Assuming your DNS and firewall (ports `80` and `443` allowed) are correctly configured, then you should be able to access your new Mbin instance from your domain (or [https://localhost_](https://localhost/) if you used `localhost`)!
7. Don't forget to follow the [Mbin first setup](https://docs.joinmbin.org/admin/running-mbin/first_setup) instructions!

## Migration Guide

Use these steps to migrate from the old Docker setup to this one. The big thing to keep in mind is that all your server data is now in one of three locations: `.env`, `compose.override.yaml`, and `storage/`. So, all server data will be going under here from now on.

Additionally, this new setup **does not require a reverse proxy** (and actually recommends not using one), as the PHP server it's using comes with **built-in HTTPS support** via Let's Encrypt and Caddy. You will need to ensure ports `80` and `443` are directly exposed and available on your server in order for this migration guide to work.

1. Keep your old Mbin directory for now; just create a new one by cloning Mbin again (`git clone https://github.com/MbinOrg/mbin.git`)! I'll reference `old` as your old Mbin directory and `new` as your new one.
2. Run the auto setup script in `new` (with this docker branch of course): `./docker/setup.sh prod DOMAINHERE`.
3. Run `docker compose down` in `old/docker` if you haven't already.
4. Copy your Postgres database from `old/docker/storage/postgres` to `new/storage/postgres`.
5. Copy your media files from `old/docker/storage/media` to `new/storage/media`.
6. Copy your OAuth keys from `old/config/oauth2` to `new/storage/oauth` (both `private.pem` and `public.pem`).
8. Open `old/docker/.env` and `new/.env` side by side so you can copy values from the old `.env` file to the new one as needed.
9. You'll want to copy over most `.env` values, including `KBIN_*`, `MBIN_*`, `MAX_IMAGE_BYTES`, `HCAPTCHA_*`, `S3_*`, the SSO fields, `EXIF_*`, `APP_SECRET`, `POSTGRES_*`, `MAILER_DSN`, `MERCURE_JWT_SECRET`, `OAUTH_PASSPHRASE`, and `OAUTH_ENCRYPTION_KEY`. Everything else, you can leave alone; feel free to ask if you have any questions about this.
10. Now that non-root containers are supported, you'll have to update the permissions of the files in `new/storage/` to match `MBIN_USER` in `.env`. Run `chown -R 1000:1000 new/storage/` if your `MBIN_USER` is `1000:1000`. Running the new setup under root has not been tested yet, but you will likely not need to change any file permissions if you were running as root previously.
11. In `new/compose.override.yaml`, uncomment the section that enables you to build the Docker image locally.
12. In `new`, run `docker compose up -d` to build and start the Docker containers.
13. If all goes well, then you should be able to access your Mbin instance from your domain!

## To Do

- [x] Document using behind another reverse proxy
- [x] Work on the dev environment
- [x] Update GH actions
- [x] Fix volume permissions
- [x] Look into whether rabbitmq volumes are actually needed
- [x] Change docs guide to edit `compose.override.yaml` instead of `compose.prod.yaml` to use the prebuilt image.


compose.yaml
```diff
@@ -55,7 +55,7 @@ services:
         user: ${MBIN_USER}
 
     redis:
-        image: redis:alpine
+        image: redis:bookworm
```
melroy89, line 58: I would love to see valkey here instead of Redis: https://github.com/valkey-io/valkey

Valkey is truly open-source, and is a fork of Redis before they changed their license model. Also Valkey is more performant today.

.env.example
```diff
@@ -120,13 +120,13 @@ EXIF_EXIFTOOL_TIMEOUT=10
...
 
 ###> doctrine/doctrine-bundle ###
 # Format described at https://www.doctrine-project.org/projects/doctrine-dbal/en/latest/reference/configuration.html#connecting-using-a-url
-POSTGRES_DB=kbin
-POSTGRES_USER=kbin
+POSTGRES_DB=mbin
+POSTGRES_USER=mbin
 POSTGRES_PASSWORD=!ChangeThisPostgresPass!
 # IMPORTANT: You MUST configure your PostgreSQL server version!
 POSTGRES_VERSION=16
```
melroy89, line 132: Latest version of postgres is v17. Maybe its a good idea to new people to just start using the latest stable PostgreSQL version. Which is again `17`.

jwr1: Actually, I did update it to `17` in the `.env.example_docker` file, but good catch, I will update it here as well for bare metal.

compose.yaml
jwr1: Sure, I wouldn't mind switching it out. Are there any changes I need to be aware of for the migration? (like configuration changes)

melroy89: It's a fully drop in replacement. So no changes are needed. You could however add additional io threads maybe to speedup the current configuration in docker.

We have a very nice website 😬: https://docs.joinmbin.org/admin/configuration/redis

jwr1: Does this look good: https://github.com/MbinOrg/mbin/pull/1542/commits/8506a510b9ab6e19ce5537543a1db62a85aa2638?

melroy89: Yes it does look good. Assuming it also just still works.

jwr1: Yeah, I did test it briefly :)

docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,57 @@
...
+# Use 301 response for image redirects to reduce server load
+sed -i 's|redirect_response_code: 302|redirect_response_code: 301|' config/packages/liip_imagine.yaml
+
+# Use S3 file system adapter when S3_KEY env var is not empty
+if [ -n "$S3_KEY" ]; then
+	sed -i 's|adapter: default_adapter|adapter: kbin.s3_adapter|' config/packages/oneup_flysystem.yaml
+fi
+
+# Needed to apply the above config changes
+composer run-script --no-dev post-install-cmd
```
jwr1, line 16: @melroy89, I just wanted to confirm, after editing either of the config files above (liip_imagine.yaml or oneup_flysystem.yaml), is this command required in order to apply the changes? I was not able to get the S3 adapter, for instance, applied without running this post-install command.

melroy89: This could make sense yes. However, you need to know what post-install-cmd is actually doing. If you look in composer.json, you will notice:

```
    "scripts": {
        "auto-scripts": {
            "cache:clear": "symfony-cmd",
            "assets:install %PUBLIC_DIR%": "symfony-cmd"
        },
        "post-install-cmd": [
            "@auto-scripts"
        ],
        "post-update-cmd": [
            "@auto-scripts"
        ],
        "codestyle:fix": [
            "tools/vendor/bin/php-cs-fixer fix"
        ]
    }
```

This means that post-install-cmd will just trigger the `auto-scripts` section. This is the default Symfony setup scripts. If will do two thinks as you can see, it triggers two commands:

1. Symfony cache clear command
2. Symfony install assets

Meaning in this particular case, I believe a simple symfony clear cache could be sufficient after changing config. So I think a Symfony cache clean might always be a good idea to keep in the docker-entry, since you never know when a user changed one or more config files before start-up?

This cache clean command is actually just the same as running: `bin/console cache:clear`. Just saying.

You can also take a look at https://github.com/MbinOrg/mbin/blob/main/bin/post-upgrade script, which is also doing something similar, with additional `APP_ENV` vars...

.env.example_docker
```diff
@@ -67,7 +68,7 @@ S3_REGION=
 S3_ENDPOINT=
 S3_VERSION=
 
-# Only let admins generate oauth clients
+# Only let admins generated oauth clients
```
melroy89, line 71:
```suggestion
# Only let admins generate oauth clients
```

English might not be my first language, but I think it was correct before right? "Only let admins generate oauth clients"?

jwr1: You're right! Not sure how that happened, lol :)

composer.json
```diff
@@ -59,6 +59,7 @@
         "phpstan/phpdoc-parser": "^2.0.0",
         "predis/predis": "^2.2.2",
         "privacyportal/oauth2-privacyportal": "^0.1.1",
+        "runtime/frankenphp-symfony": "*",
```
melroy89, line 62: Normally, for reproducible its a good practice to pin the package to at least a major version like `^0.2.0`

docker/setup.sh
```diff
@@ -0,0 +1,113 @@
...
+EOF
+else
+   cat > compose.override.yaml << EOF
+# Customizations to the docker compose should be added here.
+EOF
+fi
+
+echo Setting up storage directories...
+mkdir -p storage/{caddy_config,caddy_data,media,messenger_logs,oauth,php_logs,postgres,rabbitmq_data,rabbitmq_logs}
+echo Configuring OAuth2 keys...
```
melroy89, line 107:
```suggestion
echo "Configuring OAuth2 keys..."
```

docker/setup.sh
```diff
@@ -0,0 +1,113 @@
...
+openssl genrsa -des3 -out ./storage/oauth/private.pem -passout "pass:$OAUTH_PASS" 4096
+openssl rsa -in ./storage/oauth/private.pem --outform PEM -pubout -out ./storage/oauth/public.pem -passin "pass:$OAUTH_PASS"
+
+echo
+echo Mbin environment complete!
```
melroy89, line 112: Same here etc.

docker/docker-entrypoint.sh
```diff
@@ -0,0 +1,61 @@
...
+
+		if [ $ATTEMPTS_LEFT_TO_REACH_DATABASE -eq 0 ]; then
+			echo 'The database is not up or not reachable:'
+			echo "$DATABASE_ERROR"
+			exit 1
+		else
+			echo 'The database is now ready and reachable'
+		fi
+
+		if [ "$( find ./migrations -iname '*.php' -print -quit )" ]; then
```
melroy89, line 50: I think this `if` always returns true in our project right?

docker/setup.sh
```diff
@@ -0,0 +1,113 @@
...
+
+echo Creating compose.override.yaml file... Any additional customizations to the compose setup should be added here.
+if [[ $mode == "dev" ]]; then
+   cat > compose.override.yaml << EOF
+include:
+  - compose.dev.yaml
+EOF
+else
+   cat > compose.override.yaml << EOF
+# Customizations to the docker compose should be added here.
```
melroy89, line 101: Maybe we should help users better there with more templating (aka examples most used settings)? Can still be with comments.  Or do you expect that this file is kept empty on production in 99% of the cases?

docs/03-contributing/01-getting_started.md
```diff
@@ -159,10 +180,7 @@ If you have messenger jobs configured, be sure to stop them:
 - Docker: `docker compose stop messenger`
 - Bare Metal: `supervisorctl stop messenger:*`
 
-If you are using the Docker setup and want to load the fixture.
-
-1. Go to the `docker` directory.
-2. Execute: `docker compose exec php bin/console doctrine:fixtures:load --append --no-debug`
+If you are using the Docker setup and want to load the fixture, execute: `docker compose exec php bin/console doctrine:fixtures:load --append --no-debug`
```
melroy89, line 183:
```suggestion
If you are using the Docker setup and want to load the fixture, execute: 

````
docker compose exec php bin/console doctrine:fixtures:load --append --no-debug
```
```

Lets move it to a next line with a bigger code block.

File: docs/03-contributing/01-getting_started.md [hunk 11]
```diff
@@ -4,6 +4,27 @@ There are several ways to get started. Like using the Docker setup or use the de
 
 The code is mainly written in PHP using the Symfony framework with Twig templating and a bit of JavaScript & CSS and of course HTML.
 
+## Docker as a dev server
+
+To save yourself much time setting up a development server, you can use our Docker setup instead of a manual configuration:
+
+1. Make sure you are currently in your Mbin directory.
```
melroy89, line 11:
```suggestion
1. Make sure you are currently in the root of your Mbin directory.
```

docs/03-contributing/01-getting_started.md
```diff
@@ -4,6 +4,27 @@ There are several ways to get started. Like using the Docker setup or use the de
...
+5. From here, you should be able to access your server at [https://localhost/](https://localhost/). Any edits to the source files will automatically rebuild your server.
+6. Optionally, follow the [Mbin first setup](../02-admin/04-running-mbin/01-first_setup.md) instructions.
+
+> [!NOTE]
+> If you'd prefer to manually configure your Docker environment (instead of using the setup script) then follow the manual environment setup steps in the [Docker install guide](../02-admin/01-installation/02-docker.md), but while you're creating `compose.override.yaml`, use the following:
+>
+> ```yaml
+> include:
+>   - compose.dev.yaml
+> ```
```
melroy89, line 24: We can add maybe another `> [!INFO]` section here about how to stop it again using docker compose down.

Also you might want to explain to use `composer up -d` with the `-d` option. To execute in the background (daemon). Of course this all depends a bit on what kind of knowledge level we expect the admins to have of Docker.

docker/setup.sh
jwr1: At the moment, the docs already have you editing this file you are using a reverse proxy, or if you are using the pre-built image. I agree with you, some sort of commented template could be good.

docker/docker-entrypoint.sh
jwr1: Yup, I didn't even think about it. That was just copied from https://github.com/dunglas/symfony-docker.

docker/setup.sh
```diff
@@ -0,0 +1,113 @@
...
+    *)
+        echo "Invalid mode provided: $1"
+        echo "Must be either prod (recommended for most cases) or dev."
+        exit 1
+        ;;
+esac
+
+domain=$2
+if [[ -z $domain ]]; then
+  echo "DOMAIN must be provided. Use "localhost" if you are just testing locally."
```
melroy89, line 37:
```suggestion
  echo "DOMAIN must be provided. Use \"localhost\" if you are just testing locally."
```

And now we do need to backlash these quotes :P  sorry about that

docs/03-contributing/01-getting_started.md
jwr1: I didn't mention the `-d` option because I thought it would be unnecessary in a development environment. You want to be able to see the logs immediately as they come through, and be able to stop the server as needed with `Ctrl+C`.

I could add a note about stopping the server with `Ctrl-C` though. I'm not sure if that's completely necessary though, as the guide doesn't mention stopping the `symfony server:start` command either (as `Ctrl-C` is assumed).

melroy89: > I could add a note about stopping the server with `Ctrl-C` though. I'm not sure if that's completely necessary though, as the guide doesn't mention stopping the `symfony server:start` command either (as `Ctrl-C` is assumed).

That doesn't mean we can't improve the guide :D :D

jwr1: Done: https://github.com/MbinOrg/mbin/pull/1542/commits/b9fdfa4e41cd4b3883dcb64cc1e7eb2910d8bd82!

melroy89:
I didn't yet test it myself, but just a general note. Did you test what happens when there are changes in the config directory locally on disk. So let's say you want to change a yaml configuration file in `config/packages`. Does a docker down & up will apply the new configuration correctly?

Same question for other files that admins might change over time that needs to be applied after a simple restart of the container.

melroy89:
I actually fully agree with you. It should be as simple as ideally changing one file like `.env`.

The only two things I can think of (for now) that might make a lot of sense to support via `.env` in Docker is:

- Changing the log level.  Normally located in `config/packages/monolog.yaml`. Specially the `action_level` under `main` under `@prod`:

```yaml
when@prod:
    monolog:
        handlers:
            main:
                type: fingers_crossed
                action_level: warning <------- Allowing error to warning or info, .....
```

- Allow admins to update / extend the robots.txt fle? Normally located at: `public/robots.txt`

What do you think?

jwr1:
Yeah, having an option for log level would be great; I'll add that for sure.

I'm less sure about robots.txt. Have you found yourself needing to edit it manually? For me personally, if I really want to block a user agent, I don't even trust the robots.txt file to be used correctly, so I just block requests at the network level.

melroy89:
> I'm less sure about robots.txt. Have you found yourself needing to edit it manually? For me personally, if I really want to block a user agent, I don't even trust the robots.txt file to be used correctly, so I just block requests at the network level.

Understandably lately as well with all those AI scrapers haha. Its a sad story, I shouldn't laugh.

jwr1:
> Yeah, having an option for log level would be great; I'll add that for sure.

And, done!

melroy89:
> > Yeah, having an option for log level would be great; I'll add that for sure.
> 
> And, done!

Nice!

docs/03-contributing/01-getting_started.md
```diff
@@ -4,6 +4,30 @@ There are several ways to get started. Like using the Docker setup or use the de
...
+## Docker as a dev server
+
+To save yourself much time setting up a development server, you can use our Docker setup instead of a manual configuration:
+
+1. Make sure you are currently in the root of your Mbin directory.
+2. Run the auto setup script with `./docker/setup.sh dev localhost` to configure `.env`, `compose.override.yaml`, and `storage/`.
+3. Use `docker build -t mbin -f docker/Dockerfile --target dev .` to build the dev Docker image.
+4. Run `docker compose up` to start the Docker containers.
+5. From here, you should be able to access your server at [https://localhost/](https://localhost/). Any edits to the source files will automatically rebuild your server.
+6. Optionally, follow the [Mbin first setup](../02-admin/04-running-mbin/01-first_setup.md) instructions.
```
melroy89, line 16: OK. I followed the guide from a scratch. Step by step.

I get the following errors two errors:

```sh
php-1        | 2025/04/11 13:32:17.425  ERROR   pki.ca.local    failed to install root certificate      {"error": "failed to execute tee: exit status 1", "certificate_file": "storage:pki/authorities/local/root.crt"}
```

And later, which is constantly spamming (repeating):

```sh
php-1        | 2025/04/11 13:32:20.977  INFO    frankenphp      <br />
php-1        | <b>Fatal error</b>:  Uncaught Error: Class &quot;Runtime\FrankenPhpSymfony\Runtime&quot; not found in /app/vendor/autoload_runtime.php:16
php-1        | Stack trace:
php-1        | #0 /app/public/index.php(7): require_once()
php-1        | #1 {main}
php-1        |   thrown in <b>/app/vendor/autoload_runtime.php</b> on line <b>16</b><br />
php-1        | 
php-1        | 2025/04/11 13:32:20.977  ERROR   frankenphp      worker script has not reached frankenphp_handle_request {"worker": "/app/public/index.php"}
php-1        | 2025/04/11 13:32:20.977  WARN    frankenphp      many consecutive worker failures        {"worker": "/app/public/index.php", "failures": 6}
php-1        | 2025/04/11 13:32:20.977  ERROR   frankenphp      worker script has not reached frankenphp_handle_request {"worker": "/app/public/index.php"}
php-1        | 2025/04/11 13:32:20.977  WARN    frankenphp      many consecutive worker failures        {"worker": "/app/public/index.php", "failures": 6}
php-1        | 2025/04/11 13:32:20.978  ERROR   frankenphp      PHP Fatal error:  Uncaught Error: Class "Runtime\FrankenPhpSymfony\Runtime" not found in /app/vendor/autoload_runtime.php:16
php-1        | Stack trac
```

melroy89: I was hoping during the docker build it will install the packages right in DOcker?

Anyhow, this might be related.. I try to run `composer install` on my computer (outside of Docker), giving the following error:

```sh
$ composer install

Installing dependencies from lock file (including require-dev)
Verifying lock file contents can be installed on current platform.
Warning: The lock file is not up to date with the latest changes in composer.json. You may be getting outdated dependencies. It is recommended that you run `composer update` or `composer update <package name>`.
Your lock file does not contain a compatible set of packages. Please run composer update.
```

melroy89: Wait.. is the Docker setup using my local PHP composer packages!?? That is weird...

jwr1: Thank you. I have fixed the issue and updated the docs as well. You should be able to try again successfully (which I verified as well). The issue was that on first run, dependencies need to be installed, and both the `php` and `messenger` containers were trying to do so at the same time to the same folder. I have disabled both the `messenger` and `rabbitmq` containers by default (which probably wasn't needed anyway unless you are developing/testing federation), and this also has fixed the dependency issue.

And yes, the whole Mbin directory is mounted into the Docker containers in the dev environment (which I know is completely the opposite of how it's done in production). This of course, means that dependencies are saved to your local directory (not just stored and installed in the Docker image).

jwr1: > ```shell
> php-1        | 2025/04/11 13:32:17.425  ERROR   pki.ca.local    failed to install root certificate      {"error": "failed to execute tee: exit status 1", "certificate_file": "storage:pki/authorities/local/root.crt"}
> ```

You can ignore that. Here's the technical explanation:

Since you aren't using a real domain and instead using `localhost`, then Caddy needs to generate a self-signed HTTPS certificate. This causes browsers to show that warning message about the page using a self-signed certificate, where you have to confirm you really do want to view the page. Caddy is nice enough to attempt to install a root certificate on your device, which would allow you to visit a page in the browser without it displaying a warning; it is unsuccessful though, because it needs to install it as `root` (and we have Caddy/Frankenphp running as non-root). It wouldn't be beneficial to us anyway though, because Caddy would be installing the root certificate in the docker container, not your actual computer's operating system, which is likely where your browser actually is. So again, that error can be safely ignored.

melroy89: I see, makes all sense. Thanks for fixing.

melroy89: Small remark.. I also have a web server running locally (eg. Apache2 or Nginx) for other developments.. Would it be a good idea to start the dev server on another port? So something else like port `8008`?

docs/02-admin/01-installation/02-docker.md
```diff
@@ -48,83 +45,57 @@ git clone https://github.com/MbinOrg/mbin.git
...
-#### Build our own Docker image
-
-If you want to build our own image, run (_no_ need to update the `compose.yml` file):
-
-```bash
-docker build --no-cache -t mbin -f Dockerfile  ..
-```
-
-#### Use Mbin pre-build image
+#### Automatic setup
```
melroy89, line 55:
```suggestion
#### Automatic setup script
```

docs/02-admin/01-installation/02-docker.md
```diff
@@ -48,83 +45,57 @@ git clone https://github.com/MbinOrg/mbin.git
...
 
-_OR_ use our pre-build images from [ghcr.io](https://ghcr.io). In this case you need to update the `compose.yml` file:
+Run the setup script and pass in a mode (either `prod` or `dev`) and your domain (which can be `localhost` if you plan to just test locally):
 
 ```bash
-nano compose.yml
+./docker/setup.sh prod mbin.domain.tld
 ```
```
melroy89, line 61: So let's say I just followed this guide. And I went for the setup script.

 It would be nice to point the user to the next step after existing this command.. Since what now? I see a lot of sections below about manual configuration. 
 
Explain to the user what are the follow up steps and/or point to the next step in the guide / link to the next heading..?

melroy89:
Add:

- `compose.override.yaml`

and:

- `storage/`

to .gitignore?

docs/02-admin/01-installation/02-docker.md
jwr1: Does this look good: https://github.com/MbinOrg/mbin/pull/1542/commits/8ce74401d5319e7a76684d5d098045376e92a414?

docs/03-contributing/01-getting_started.md
jwr1: > 1. I also have a web server running locally (eg. Apache2 or Nginx) for other developments.. Would it be a good idea to start the dev server on another port? So something else like port `8008`?

I think it would be best to leave it as is. Changing the default port would just make the dev setup diverge more from prod and complicate it more. How about I add a little note in the Contributing (dev) section about changing the port numbers away from default?

> 2. What about also reporting a small _info_ message after: `php-1       | 2025/04/12 00:15:14.040   INFO    watcher watching config file for changes        {"config_file": "/etc/caddy/Caddyfile"}`
> 
> 
> That the dev server is now ready on localhost on port xy on the terminal output?

I'm not sure that this would actually be feasible. For one, there's no way to know once Caddy has completed its setup and then to trigger that message to display. Secondly, there are so many different things we'd have to take into account in order to display a correct message: is it HTTP or HTTPS, is it `localhost` or some other domain, has the port number been switched, etc?

> 3. Does the production docker images will just fine in the production set? So people do not need to build it locally after all. They should be able to just use your officially published docker images (which are also version tagged) her: [MbinOrg/mbin/pkgs/container/mbin](https://github.com/MbinOrg/mbin/pkgs/container/mbin). Instead of needing to clone the git repo & build the docker file from scratch, like now:
> 
> 
> ```
>         build:
>             dockerfile: docker/Dockerfile
>             context: .
>             target: prod
> ```
> 
> You might need to also update the `.github/workflows/action.yaml` file all the way to the bottom. is where we build the docker image.

The production image should work, and the GH action build workflow should not need to be updated in order to work (the `Dockerfile` is still in the same location). Obviously, I have not been able to test that though.

melroy89:
> > Add:
> > 
> > * `compose.override.yaml`
> > 
> > and:
> > 
> > * `storage/`
> > 
> > to .gitignore?
> 
> Actually already done: https://github.com/MbinOrg/mbin/pull/1542/files#diff-bc37d034bad564583790a46f19d807abfe519c5671395fd494d8cce506c42947

ow yes. maybe because I switched branches in between testing. My bad!

docs/02-admin/01-installation/02-docker.md
melroy89: almost ;)

melroy89: ow nevermind its good now. I thought `> [!WARNING]` was on the same line. That would cause issues, but it seems to work fine now.

docs/03-contributing/01-getting_started.md
jwr1: > How about I add a little note in the Contributing (dev) section about changing the port numbers away from default?

I have added said note: https://github.com/MbinOrg/mbin/pull/1542/commits/63fa6563818432f86a31d354d5d9dd08d4bdb4cd.

docs/02-admin/04-running-mbin/03-upgrades.md
```diff
@@ -17,7 +17,7 @@ And when needed also execute: `sudo redis-cli FLUSHDB` to get rid of Redis/KeyDB
 1. (Re)build a new Mbin docker image (without using cached layers):
 
 ```bash
-docker build --no-cache -t mbin -f docker/Dockerfile .
+docker compose build --no-cache
```
melroy89, line 20: Nice! I didn't knew this compose command.

melroy89 (approved):

melroy89:
> I think it's really ready this time!
> 
> Of course, if I find anything else, I'll just make another PR.
> 
> Once this is merged, I'm going to put it on kbin.earth. That will be the real test!

Yeah! I'm really happy with this! Thank you so much!

melroy89 (approved):

PR #1563, opened 2025-04-16
Sort and filter comment replies the same as root comments

### LINKED ISSUE #1397: Blocked users' comments still visible in the Interstellar app
**Describe the bug**
Blocked users' comments are still visible in the Interstellar app [please see additional context]

kbin.earth

1.7.4

**To Reproduce**
1. Block some obnoxious reply guy
2. In Interstellar, navigate to communities where you encountered them in the past
3. Find them in the comment section

I did a side-by-side comparison of the same thread in the web UI and in Interstellar: in the browser blocked users' comments were hidden, in the app they were shown.

**Expected behavior**
I expect blocked users to not be shown at top or comment level, in the web UI or apps.

**Smartphone:**
 - Device: Fairphone 4
 - OS: LineageOS for MicroG 22.1
 - Browser Fennec
 - Browser Version: 134.0.0

**Additional context**
I raised this issue [in the Interstellar repo](https://github.com/jwr1/interstellar/issues/92), and the developer asked me to report it here as well. I hope you guys can hash it out?

BentiGorlich:
- pass the criteria through to match child comments by it
- included criteria:
  - ordering
  - banned hashtag checking
  - preferred languages
  - federation filter (not used with comments, but still included)
  - time filter (also not used for comments atm)
- filter out comments from banned authors

Closes #1397

### DIFF
--- a/src/Controller/Api/Entry/Comments/EntryCommentsRetrieveApi.php
+++ b/src/Controller/Api/Entry/Comments/EntryCommentsRetrieveApi.php
@@ -153,6 +153,9 @@ class EntryCommentsRetrieveApi extends EntriesBaseApi
 
         $comments = $commentsRepository->findByCriteria($criteria);
 
+        $commentsRepository->hydrate(...$comments);
+        $commentsRepository->hydrateChildren(...$comments);
+
         $dtos = [];
         foreach ($comments->getCurrentPageResults() as $value) {
             try {
--- a/src/Controller/Api/Entry/Comments/EntryCommentsRetrieveApi.php
+++ b/src/Controller/Api/Entry/Comments/EntryCommentsRetrieveApi.php
@@ -143,7 +143,8 @@ class EntryCommentsRetrieveApi extends EntriesBaseApi
 
         $request = $this->request->getCurrentRequest();
         $criteria = new EntryCommentPageView($this->getPageNb($request), $security);
-        $criteria->showSortOption($criteria->resolveSort($request->get('sortBy', Criteria::SORT_HOT)));
+        $sort = $criteria->resolveSort($request->get('sortBy', Criteria::SORT_HOT));
+        $criteria->showSortOption($sort);
         $criteria->entry = $entry;
         $criteria->perPage = self::constrainPerPage($request->get('perPage', EntryCommentRepository::PER_PAGE));
         $criteria->setTime($criteria->resolveTime($request->get('time', Criteria::TIME_ALL)));
@@ -152,15 +153,12 @@ class EntryCommentsRetrieveApi extends EntriesBaseApi
 
         $comments = $commentsRepository->findByCriteria($criteria);
 
-        $commentsRepository->hydrate(...$comments);
-        $commentsRepository->hydrateChildren(...$comments);
-
         $dtos = [];
         foreach ($comments->getCurrentPageResults() as $value) {
             try {
                 \assert($value instanceof EntryComment);
                 $this->handlePrivateContent($value);
-                array_push($dtos, $this->serializeCommentTree($value));
+                $dtos[] = $this->serializeCommentTree($value, $criteria);
             } catch (\Exception $e) {
                 continue;
             }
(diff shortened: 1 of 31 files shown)


src/Controller/Api/Entry/Comments/EntryCommentsRetrieveApi.php
```diff
@@ -152,15 +153,12 @@ public function __invoke(
 
         $comments = $commentsRepository->findByCriteria($criteria);
 
-        $commentsRepository->hydrate(...$comments);
```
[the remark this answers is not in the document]
BentiGorlich: I am not 100% sure, but I think it was supposed to fetch everything needed at once instead of fetching the collections one by one... I am not sure if it actually worked in any way shape or form, but that is my guess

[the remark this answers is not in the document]
melroy89: I also didn't fully get it, despite the name "hydrate" which is of course a hint. Meaning its pre-loading or preparing something. Just like Angular Hydrate.

In this case it seems that its used in Doctrine to partial load the records in one big good rather than one query at a time, see also: https://github.com/doctrine/orm/issues/8471#issuecomment-2400932188

And also: https://www.doctrine-project.org/projects/doctrine-orm/en/3.3/reference/partial-objects.html

This is normally done to improve the speed & performance, so  Doctrine’s identity map is populated with all needed associations so later access doesn’t trigger additional SQL.

Whether its actually working or not, I do not know, but maybe we should keep it, unless you really know for sure if it has no benefit now.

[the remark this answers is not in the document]
BentiGorlich: I can check that when I am at home. With a larger comment tree I should be able to see the queries decrease if it works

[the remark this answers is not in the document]
BentiGorlich: I will bring it back. It reduces the DB queries from 278 to 247 on an entry with 40 comments

BentiGorlich:
I have it running on gehirneimer.de

src/Controller/Api/Entry/Comments/EntryCommentsRetrieveApi.php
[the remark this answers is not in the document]
melroy89: Thank you for your confirmation

src/Repository/EntryCommentRepository.php
```diff
@@ -314,37 +313,4 @@ public function hydrate(EntryComment ...$comments): void
             ->getQuery()
             ->execute();
     }
-
-    public function hydrateParents(EntryComment ...$comments): void
```
melroy89, line 318: Could you also bring back this hydration code?

src/Repository/PostCommentRepository.php
```diff
@@ -221,27 +220,4 @@ public function hydrate(PostComment ...$comment): void
                 ->getResult();
         }
     }
-
-    public function findToDelete(User $user, int $limit): array
```
melroy89, line 225: Could you also bring back this hydration code?

BentiGorlich: This is not hydration code

src/Repository/EntryCommentRepository.php
BentiGorlich: This was just not used, so I think it is not necessary...

melroy89: Ow. Maybe we could now leverage hydration to reduce the db calls. Now you know it actually works?

Could we call this method?

BentiGorlich: I cannot think of a place where this might be useful... We basically always already have the parent of a comment and we can hydrate that with the method in the EntryRepository

melroy89 (approved):

melroy89:
I'm getting this weird error with Docker now. When I try to use `./docker/setup.sh dev localhost`

I did change the docker image to the main tag: `ghcr.io/mbinorg/mbin:main` in my compose.yaml.

But when running:

```sh
docker compose up -d
```

I see:

```
In FileLoader.php line 356:
                                                                               
  Expected to find class "App\Utils\ArrayUtils" in file "/app/src/Utils/Array  
  Utils.php" while importing services from resource "../src/", but it was not  
   found! Check the namespace prefix used with the resource.                   
                                                                               


In FileLoader.php line 177:
                                                                               
  Expected to find class "App\Utils\ArrayUtils" in file "/app/src/Utils/Array  
  Utils.php" while importing services from resource "../src/", but it was not  
   found! Check the namespace prefix used with the resource in /app/config/se  
  rvices.yaml (which is being imported from "/app/src/Kernel.php").            
                                                                               

In FileLoader.php line 356:
                                                                               
  Expected to find class "App\Utils\ArrayUtils" in file "/app/src/Utils/Array  
  Utils.php" while importing services from resource "../src/", but it was not  
   found! Check the namespace prefix used with the resource.                   
```

Any clue?

BentiGorlich:
As far as I know it is the autoloader generated by composer that fucked up... So rerun `composer install` and restarting php to clear out opcache should fix it

melroy89:
> As far as I know it is the autoloader generated by composer that fucked up... So rerun `composer install` and restarting php to clear out opcache should fix it

This is what the script should do by default in docker. Right @jwr1? It seems we missed something with the docker setup.

BentiGorlich:
I thought that that fixed it for me locally when I was developing (also with docker)... Not sure how I fixed it if it was not with these commands...

PR #1565, opened 2025-04-17
Pin docker images to version tags

melroy89:
- Extends the documentation and explain about using Docker image tags, to pin the docker image to a specific release. To avoid sudden upgrades or causing unplanned down-time due to that. It's a best practice to use pinned versions and let the server admin decide when to update/upgrade.
- Extend `setup.sh` override template with docker tag version example. Plus a hint at the top about combining settings, so people won't define the service multiple times in the yaml file.


jwr1 (approved):
LGTM!

PR #1573, opened 2025-04-22
Check whether the actor is a User

BentiGorlich:
- when creating a post of any kind the actor needs to be a user in the mbin context, so check for that. If the actor is not a user we get a bunch of type errors which severely slow the messengers down

I had this problem with a pleroma post from a group actor. My message rate plummeted: 
grafik

Which lead to me banning the origin instance. I could not check whether this fix worked, but it should. Either way, since we can get a Magazine from `findActorOrCreate` it should be done anyway.

### DIFF
--- a/src/Service/ActivityPub/Note.php
+++ b/src/Service/ActivityPub/Note.php
@@ -131,7 +131,7 @@ class Note
         }
 
         $actor = $this->activityPubManager->findActorOrCreate($object['attributedTo']);
-        if (!empty($actor)) {
+        if ($actor instanceof User) {
             if ($actor->isBanned) {
                 throw new UserBannedException();
             }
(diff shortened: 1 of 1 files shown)


BentiGorlich:
It seems like this does not speed up the processing, but at least the error message is clearer and the messages just get removed from the transport now. I think I just have a ridiculous amount of them in my queue. I just don't know where they are coming from

melroy89 (approved):
Great catch!

melroy89:
> It seems like this does not speed up the processing, but at least the error message is clearer and the messages just get removed from the transport now.

Less messengers in transport/queue means less CPU and memory usage of RabbitMQ. Which in general could be beneficial for performance. Especially on very large instances.

PR #1577, opened 2025-04-30
Url Ids should not be extracted as a hashtag

### LINKED ISSUE #1548: Undesired tagging of a post when a link has the # character
**Describe the bug**
Undesired tagging of a post when a link in the post body includes the # character.

kbin.melroy.org

1.8.1

**To Reproduce**
1. Create a new thread.
2. Put a link in the body that includes the `#` character. For example, put in `https://test.notrealwebsite284/#hello`.
3. Click the top left hamburger menu icon and scroll to the Thread section, and see that you now have a tag that matches whatever followed the `#` character in the link. If you followed my example, your post is now tagged #hello.
4. Edit the post, clicking the X next to the tag to attempt to remove the tag. Notice that it stays anyways, because of the content in the link following the `#` character.

**Expected behavior**
I expect the `#` character to be ignored for tagging purposes when it is in a link. It can cause trouble otherwise. Imagine linking https://news.site375/the-summer-blockbuster-movie-broke-sales-records/#heading. The way things are now, you will end up forced to have the tag #heading on your post. 

 - Device: iPhone 13 mini
 - OS: iOS 17.1.2
 - Browser: Safari
 - Browser Version: 17.1

BentiGorlich:
- our `TagExtractor` extracted url ids as hashtags before. Add a testcase for it and fix that

Closes #1548


src/Utils/UrlUtils.php
```diff
@@ -39,4 +39,17 @@ public static function getCacheKeyForMarkdownMagazineMention(string $url): strin
 
         return "markdown_magazine_mention_$key";
     }
+
+    public static function extractUrlsFromString(string $text): array
+    {
+        $words = preg_split('/[ \n\[\]()]/', $text);
```
melroy89, line 45: Would you mind to move the regex to our RegPatterns Utils file? https://github.com/MbinOrg/mbin/blob/main/src/Utils/RegPatterns.php

melroy89: Nevermind.. I do not think this is really a regex. right?

BentiGorlich: will di

melroy89 (approved):

PR #1612, opened 2025-05-24
Use baseData instead

melroy89:
Try to use `$baseData` instead of `$data` to pass along to the `renderView` of the `*_list.html.twig` template.


melroy89:
While this does solve the issue with `criteria`, etc. So that is no longer causing issues..


We trigger another 500 error now..

image

melroy89:
Doesn't seem to solve it tho 😢

BentiGorlich:
I fixed the problem in #1613

PR #1613, opened 2025-05-26
Fix the post comment expanding not working

BentiGorlich:
- the criteria necessary for sorting the comments was not passed through in all cases, add the missing ones


melroy89 (approved):
Thanks!

PR #1646, opened 2025-06-19
Fix getting bookmarks from a bookmark list.

### LINKED ISSUE #1644: Bookmarks API only returns the default list.
**Describe the bug**
When trying to fetch bookmarked items from a bookmark list it always returns the default list.

I skimmed the mbin code and it looks like the list parameter is ignored and list_id is used instead. Problem is the list_id doesn't seem to be returned by any of the other endpoints so all the client knows is the list name.

kbin.earth

1.8.2

**To Reproduce**
1. Make request to `/api/bookmark-lists/show` for a bookmark list other than the default.
2. Get default list in response.

**Expected behavior**
Expect the requested bookmark list to be returned.

olorin99:
- Falls back to finding bookmark list by name if couldn't find by id.
- Adds id to BookmarkListDto.

Fixes #1644

### DIFF
--- a/src/Controller/Api/Bookmark/BookmarkListApiController.php
+++ b/src/Controller/Api/Bookmark/BookmarkListApiController.php
@@ -137,7 +137,6 @@ class BookmarkListApiController extends BaseApi
     #[IsGranted('ROLE_OAUTH2_BOOKMARK_LIST:READ')]
     public function front(
         #[MapQueryParameter] ?string $list,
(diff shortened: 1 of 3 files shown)


BentiGorlich:
From my perspective we should not include the id of the list in the dto as the name is already unique per user. 
So I think the front method should not take the id of a list but just the name and resolve it by the user name, like you do in this PR.

PR #1636, opened 2025-06-10
Take federation status into account when creating content

BentiGorlich:
- display an info alert when a user visits a magazine from a banned instance, that the magazine will not receive any updates, as the instance is banned
- when creating threads, microblogs or comments, throw an exception that the instance is banned and show an error about it
- remove the subscribe/unsubscribe option for magazines from defederated instances

User overview:
grafik

Magazine view:
Bildschirmfoto vom 2025-06-10 22-13-06

Creating thread error:
Bildschirmfoto vom 2025-06-10 22-15-41

Closes #1595

PR #1666, opened 2025-07-05
Fix reading the wrong `magazine` property

BentiGorlich:
`$dto->magazine` seems to be null relatively often, so switch to the already used `$dto->entry->magazine` 

Fixes this error:
>ERROR: php: Warning: Attempt to read property "apId" on null

Introduced in #1636


melroy89 (approved):
Make sense

PR #1670, opened 2025-07-06
Improve the generic search and add users and magazines to it

BentiGorlich:
- Order results by rank instead of creation date
- Add new creation date filter because we don't order by it anymore
- do not show the parent of the comments found via search anymore, that info is already contained in them and displaying the parents is distracting. Also adding magazine and user support to that code block would have been hell of a lot of work
- Add ts columns to user and magazine entities
- Add user and magazine to the generic search results
- Add new component for user inline boxes, displayed in the search results
- improve CSS of user_box
  - move negative offset to the image part (the thing that is actually offset), and subtract it from the height -> weird offset is only there
  - about section now has to get a proper margin instead of offsetting it negatively to the top

Images:
Some user search results:
grafik

Some combined results:
grafik

Magazine result:
grafik

### DIFF
--- a/src/Repository/SearchRepository.php
+++ b/src/Repository/SearchRepository.php
@@ -103,7 +103,7 @@ class SearchRepository
         $sqlEntry = "SELECT e.id, e.created_at, e.visibility, 2 * ts_rank_cd(e.title_ts, plainto_tsquery(:query)) + ts_rank_cd(e.body_ts, plainto_tsquery(:query)) as rank, 'entry' AS type FROM entry e
             INNER JOIN public.user u ON u.id = user_id
             INNER JOIN magazine m ON e.magazine_id = m.id
-            WHERE (e.body_ts @@ plainto_tsquery( :query ) = true OR e.title_ts @@ plainto_tsquery( :query ) = true)
+            WHERE (e.body_ts @@ plainto_tsquery( :query ) = true OR e.title_ts @@ plainto_tsquery( :query ) = true OR e.title LIKE :likeQuery OR e.body LIKE :likeQuery)
                 AND e.visibility = :visibility
                 AND u.is_deleted = false
                 AND NOT EXISTS (SELECT id FROM user_block ub WHERE ub.blocked_id = u.id AND ub.blocker_id = :queryingUser)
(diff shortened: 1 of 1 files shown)


BentiGorlich:
I did not change the magazine box, but maybe I should have. But we can do that in a different PR I think

melroy89:
While you at it. I have a small request regarding the generic search.

This search is also used to lookup magazine and retrieve data like comments or posts from remote servers.

However, the current placeholder text is not reflecting that.

Screenshot_20250707-172415 (1)


How could we make this more clear for the users?

BentiGorlich:
Maybe `search term, url or handle`?

melroy89:
> Maybe `search term, url or handle`?

Sounds good!

BentiGorlich:
What do you think about this: 
grafik

melroy89:
> What do you think about this: grafik

It's good!

The only thing is the date format. In the Netherlands we use never dots but dashes. And I think in English dashes are also the default format for dates. So `dd-mm-yyyy`.

You could argue about the order. America uses mm-dd-yyyy, which is really annoying. Or we use yyyy-mm-dd. But that aside, it's never separated by dots.

BentiGorlich:
This is just a date input field thing, it will look different on your browser

melroy89:
> This is just a date input field thing, it will look different on your browser

Ah OK! Nevermind in that case.

melroy89 (dismissed):

BentiGorlich:
But you should probably check it out on your browser :)
And I did not push those changes, yet. Will do that shortly :)

BentiGorlich:
I put the changes on gehirneimer.de but you'd need an account to check it out...

melroy89:
I also upgraded to this branch. And running this branch currently at https://kbin.melroy.org. 

Eg. searching on just the letter `a` will give no results for some reason? https://kbin.melroy.org/search?search%5Bq%5D=a&search%5Bmagazine%5D=&search%5Buser%5D=&search%5Btype%5D=0&search%5Bsince%5D=

melroy89:
I removed / deleted the magazine `tech@kbin.social` on my instance. Yet it seems to still be part of the list?

image
 
 
 Most likely we are missing some SQL `where` filter in the query.

BentiGorlich:
Well that is not something I added here. That has to be resolved separately. That is just the normal magazine search box

BentiGorlich:
The psql search thing only uses "whole" words I guess... I am not certain, but maybe we should add some results just based on a simple string search in titles

BentiGorlich:
> I removed / deleted the magazine `tech@kbin.social` on my instance. Yet it seems to still be part of the list?

How did you delete the magazine? I want to fix it, just in a different PR. It already checks for the visibility... See https://github.com/MbinOrg/mbin/blob/main/src/Form/Type/MagazineAutocompleteType.php#L46-L50

melroy89 (approved):

PR #1696, opened 2025-08-10
Implement a combined front page

### LINKED ISSUE #484: Combined View of Threads and Microblogs
Blocked by #453

BentiGorlich:
- new `ContentRepository` that can find entries and threads by criteria
- improved `ContentPopulationTransformer` by hydrating content -> fewer SQL requests
- add API for fetching entries and posts at the same time
- add new `getTagsOfContent` and make the class specific methods private in the `TagLinkRepository`
- remove the front templates for entry and post as they are now taken care of by the `content/front` template

Closes #484


BentiGorlich:
Sorry for such a big PR again @melroy89 , however 49 / 80 files are just renamed methods. It looks a bit bigger than it is, but it is still a lot :sweat_smile:

melroy89:
I think I will test this pr (together with the other related pr) later on my own server as well if you don't mind.

BentiGorlich:
of course, but you should try the other branch with the performance improvements that is built on top of this in #1697

src/Repository/ContentRepository.php
```diff
@@ -0,0 +1,242 @@
...
+
+        $contentClauseEntry = '';
+        $contentClausePost = '';
+        if ('all' !== $criteria->content) {
+            if ('threads' === $criteria->content) {
+                $contentClausePost = 'false';
+            } elseif ('microblog' === $criteria->content) {
+                $contentClauseEntry = 'false';
+            } else {
+                throw new \LogicException("wtf is $criteria->content");
```
melroy89: Better logging text?

src/Utils/SqlHelpers.php
```diff
@@ -25,16 +25,51 @@ public static function makeWhereString(array $whereClauses): string
...
+        $newSql = $sql;
+        foreach ($parameters as $name => $value) {
+            if (\is_array($value)) {
+                $size = \sizeof($value);
+                $newParameterNames = [];
+                for ($i = 0; $i < $size; ++$i) {
+                    $newParameters["$name$i"] = $value[$i];
+                    $newParameterNames[] = ":$name$i";
+                }
+                $newSql = str_replace(":$name", join(',', $newParameterNames), $sql);
```
melroy89, line 61: Look out for SQL injections here.
Try to use prepared statements (with Parameterized Queries). Or use the `createQueryBuilder()` for example. 

Either way, be really really sure you are not using any user input directly into SQL strings.

templates/content/front.html.twig
```diff
@@ -49,15 +63,24 @@
...
+                {% include 'post/_form_post.html.twig' %}
+            </section>
+            {% include 'post/_options.html.twig' %}
+        {% else %}
+            {% include 'entry/_options.html.twig' %}
+        {% endif %}
+    {% endif %}
     {% include 'layout/_flash.html.twig' %}
     {% if magazine is defined and magazine %}
         {% include 'magazine/_restricted_info.html.twig' %}
```
melroy89, line 78: Since you changed the `{%` syntax for executing statements to `{{` used for outputting the result of an expression on line 83.. Maybe we should do the same here as well? And on other lines.. like line 71, line 73.. etc?

tests/Functional/Controller/Post/PostFrontControllerTest.php
```diff
@@ -216,6 +214,9 @@ private function prepareEntries(): KernelBrowser
             $this->getUserByUsername('JaneDoe')
         );
 
+        // sleep so the creation time is actually 1 second apart for the sort to reliably be the same
+        sleep(1);
```
melroy89, line 218: Introducing sleep is never a good idea, also not in tests. Depending on the CPU usage, performance, resources etc. This could change..

If you want to make it more robust, look into a polling technique or maybe a notify system. But `sleep()` is actually always guaranteed to be a problem sooner or later.

src/Repository/TagLinkRepository.php
```diff
@@ -32,7 +32,26 @@ public function __construct(
...
+            return $this->getTagsOfEntry($content);
+        } elseif ($content instanceof EntryComment) {
+            return $this->getTagsOfEntryComment($content);
+        } elseif ($content instanceof Post) {
+            return $this->getTagsOfPost($content);
+        } elseif ($content instanceof PostComment) {
+            return $this->getTagsOfPostComment($content);
+        } else {
+            // this is unreachable because of the strict types
+            throw new \LogicException();
```
melroy89: Maybe still add some text here.. even though it should never be triggered.

src/Factory/ActivityPub/ActivityFactory.php
```diff
@@ -27,10 +27,10 @@ public function __construct(
...
-            $activity instanceof Post => $this->postNoteFactory->create($activity, $this->tagLinkRepository->getTagsOfPost($activity), $context),
-            $activity instanceof PostComment => $this->postCommentNoteFactory->create($activity, $this->tagLinkRepository->getTagsOfPostComment($activity), $context),
+            $activity instanceof Entry => $this->pageFactory->create($activity, $this->tagLinkRepository->getTagsOfContent($activity), $context),
+            $activity instanceof EntryComment => $this->entryNoteFactory->create($activity, $this->tagLinkRepository->getTagsOfContent($activity), $context),
+            $activity instanceof Post => $this->postNoteFactory->create($activity, $this->tagLinkRepository->getTagsOfContent($activity), $context),
+            $activity instanceof PostComment => $this->postCommentNoteFactory->create($activity, $this->tagLinkRepository->getTagsOfContent($activity), $context),
             $activity instanceof Message => $this->messageFactory->build($activity, $context),
             default => throw new \LogicException(),
```
melroy89: Also add some better log exception message here, just to be sure.

tests/Functional/Controller/Post/PostFrontControllerTest.php
BentiGorlich: I'd agree in general, but here the creation time just has to be one second more as in the previous post. Otherwise the sorting is random as the creation time is the same

templates/content/front.html.twig
BentiGorlich: Actually we cannot do that. They actually differ in what they are doing. The `{{ }}` one shares the variables where as the `{% %}` one does not (if I am remembering correctly. I found out while working on this and I really do hate it... It should be `{{ }}` everywhere, but we cannot guarantee that it will behave the same as before.

src/Utils/SqlHelpers.php
BentiGorlich: This method only changes something like this:
```sql
SELECT * FROM entry WHERE id IN (:ids)
```

To something like this (with `ids` being set to `[1, 2, 3]`):
```sql
SELECT * FROM entry WHERE id IN (:ids0, :ids1, :ids2)
```

And the actual values are still using sql parameters to be sanitized. So no user input is getting added to the sql query in any way.

melroy89 (approved):

tests/Functional/Controller/Post/PostFrontControllerTest.php
melroy89: Ah I got it. This might be robust enough.

BentiGorlich:
I will merge this once #1697 is also approved, just so main is not unnecessarily slow

tests/Functional/Controller/Entry/EntryCreateControllerTest.php
```diff
@@ -44,7 +44,7 @@ public function testUserCanCreateEntryLinkFromMagazinePage(): void
             )
         );
 
-        $this->assertResponseRedirects('/m/acme/threads/newest');
+        $this->assertResponseRedirects('/m/acme/all/newest');
```
melroy89, line 47: Why is this changed from `threads` to `all`?

BentiGorlich:
Because the default was set to the new combined vue which is identifued by `all`
 

Am 27. September 2025 21:35:54 MESZ schrieb Melroy van den Berg ***@***.***>:

tests/Functional/Controller/Entry/EntryCreateControllerTest.php
melroy89: owh sorry I was also confused about `acme` lol.. Its just a random magazine name in the tests

PR #1697, opened 2025-08-11
Performance Improvements for the new combined front page

BentiGorlich:
- add caching to the `ContentRepository` (it still works without the cache being populated) for
  - user follows
  - user magazine subscriptions
  - user domain subscriptions
  - user blocks
  - user magazine blocks
  - user domain blocks
  - user moderated magazines
- add event subscribers to each event changing one of those collections to clear the cache
- add caching to the count query of the `NativeQueryAdapter` (for pagination)
- hard set the count to 1000 pages if the result set is expected to be large

I separated it from the other PR (#1696) so it is less convoluted to look at :)

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,7 @@
 {
     "name": "PHP",
     // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-    "image": "mcr.microsoft.com/devcontainers/php:1-8.2-bullseye",
+    "image": "mcr.microsoft.com/devcontainers/php:1-8.3-bullseye",
 
     // Features to add to the dev container. More info: https://containers.dev/features.
     "features": {
@@ -71,4 +71,11 @@
 
     // Uncomment to connect as root instead. More info: https://aka.ms/dev-containers-non-root.
     // "remoteUser": "root"
+
+    // Uncomment if you are using Podman
+    //runArgs: [
+    //    "--userns=keep-id",
+    //    "--security-opt=label=disable"
+    //],
+    //"updateRemoteUserUID": true
 }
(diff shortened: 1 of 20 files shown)


src/Pagination/NativeQueryAdapter.php
```diff
@@ -49,6 +46,40 @@ public function __construct(
...
+                $item->expiresAfter(new \DateInterval('PT10M'));
+            }
+
+            return $count;
+        });
+    }
+
+    private function calculateNumOfResults(string $sql, array $parameters): int
+    {
+        $sql2 = 'SELECT COUNT(*) as cnt FROM ('.$sql.') sub';
```
melroy89, line 73: This `COUNT(`) might be very slow on very large tables.. Just be aware of that.

melroy89 (approved):

BentiGorlich:
@melroy89 I changed the base branch because this one was based on `new/combined-front-page` from PR #1696

src/Repository/ContentRepository.php
```diff
@@ -302,7 +302,7 @@ public function findByCriteria(Criteria $criteria): PagerfantaInterface
         $this->logger->debug('Rewritten to: {s} | {p}', ['p' => $rewritten['parameters'], 's' => $rewritten['sql']]);
 
         $numResults = null;
-        if (!$criteria->magazine && !$criteria->moderated && !$criteria->favourite && Criteria::TIME_ALL === $criteria->time && Criteria::AP_ALL === $criteria->federation && 'all' === $criteria->type) {
+        if ('test' !== $this->kernel->getEnvironment() && !$criteria->magazine && !$criteria->moderated && !$criteria->favourite && Criteria::TIME_ALL === $criteria->time && Criteria::AP_ALL === $criteria->federation && 'all' === $criteria->type) {
```
melroy89, line 305: Ideally you don't have test code or switches in production code. So the way to do this is most likely stubbing or mocking the interface.

BentiGorlich: Ideally yes, but this is a performance shortcut which is only relevant in production, as it does not report the actual number of results anymore. How would I be making this an interface that shares the same code but not this performance workaround, without duplicating most of the code?

PR #1698, opened 2025-08-11
Add instance actor test and remove url field

BentiGorlich:
After observing a 401 error from mastodon like this:
```json
{
    "error":"Verification failed for gehirneimer.de@gehirneimer.de https://gehirneimer.de/i/actor"
}
```

I decided to add a test for the instance actor. I suspect that mastodon actually tried to pull the `url` field of the actor which would result in a 404 causing mastodon to spit out this error. Not 100% sure, but having a pointer to an address we do not serve is pointless, so I removed it

### DIFF
--- a/src/Factory/ActivityPub/InstanceFactory.php
+++ b/src/Factory/ActivityPub/InstanceFactory.php
@@ -36,7 +36,6 @@ class InstanceFactory
                 'owner' => $actor,
                 'publicKeyPem' => $this->client->getInstancePublicKey(),
             ],
-            'url' => 'https://'.$this->kbinDomain.'/instance-actor',
         ];
     }
 }
--- a/tests/Functional/Controller/WebfingerControllerTest.php
+++ b/tests/Functional/Controller/WebfingerControllerTest.php
@@ -1,0 +1,43 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Tests\Functional\Controller;
+
+use App\Tests\WebTestCase;
+
+class WebfingerControllerTest extends WebTestCase
+{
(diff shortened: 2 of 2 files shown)


melroy89:
I think we can still keep the `url`. But point to the correct link. I believe we did had some generic Mbin about page (static). This one is perfect: https://kbin.melroy.org/agent

src/Factory/ActivityPub/InstanceFactory.php
```diff
@@ -36,7 +36,6 @@ public function create(): array
                 'owner' => $actor,
                 'publicKeyPem' => $this->client->getInstancePublicKey(),
             ],
-            'url' => 'https://'.$this->kbinDomain.'/instance-actor',
```
melroy89, line 39: Could we maybe just point to `/agent`? Like: https://kbin.melroy.org/agent

BentiGorlich: That could be something, yes :thinking: 
I have no idea whether mastodon expects an activity pub page at this URL though :thinking: Maybe we have to redirect an activity pub request to the correct id :thinking:

melroy89: yeah I just deleted my initial comment. But that was saying something like: I tried to search the documentation about instance actor activitypub protocol body content. But its very hard to find proper documentation about this. Crazy.

BentiGorlich: I think we should for now just leave it out. Maybe we can reach out to the Mastodon devs and ask them how they expect this instance actor to behave...

PR #1700, opened 2025-08-12
Federate bans

### LINKED ISSUE #943: Federate Bans
When a lemmy user is banned magazine or instance wide, it does not translate to mbin at all. That means that a bad user and their content will stay publicly visible on mbin instances

Federate bans and understand the lemmy activities

none

**Additional context**
https://join-lemmy.org/docs/contributors/05-federation.html#block-user

BentiGorlich:
- bans are now federated, both incoming and outgoing and both magazine and instance-bans
- to support that, add a new field to the activity table to reference magazine bans
- also add a ban reason field for users. Mostly in advance, but lemmy instances will probably provide one
- add tests for all of that

Closes #943

### DIFF
--- a/migrations/Version20250723183702.php
+++ b/migrations/Version20250723183702.php
@@ -1,0 +1,30 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20250723183702 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add magazine ban as an object to the activity table';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE activity ADD object_magazine_ban_id INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE activity ADD CONSTRAINT FK_AC74095AE490E490 FOREIGN KEY (object_magazine_ban_id) REFERENCES magazine_ban (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
+        $this->addSql('CREATE INDEX IDX_AC74095AE490E490 ON activity (object_magazine_ban_id)');
+    }
+
+    public function down(Schema $schema): void
+    {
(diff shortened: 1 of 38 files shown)


BentiGorlich:
@melroy89 for some reason the new BlockHandler's messages do not show up in my prod log, only in the supervisor log. Do you have an idea why that might be?

melroy89:
> @melroy89 for some reason the new BlockHandler's messages do not show up in my prod log, only in the supervisor log. Do you have an idea why that might be?

Most likely because that code is then only ran via the Symfony messenger jobs (aka supervisord) and never during normal website traffic php end points.

BentiGorlich:
I have all the other log messages of the messengers in there though...



Am 2. September 2025 22:15:46 MESZ schrieb Melroy van den Berg ***@***.***>:

melroy89:
ow.. so https://github.com/MbinOrg/mbin/pull/1700/commits/3f76085a54d7302c15bd459988f5ba80b9514884 solved the issue I guess regarding logging ;)?

BentiGorlich:
not sure, but maybe :)

melroy89:
> not sure, but maybe :)

hehe, I do not think it will solve the logging.. but who knows. since there is no `return`, it believe the code should continue after for example a `$this->userManager->unban()` call

melroy89 (dismissed):

melroy89 (dismissed):

BentiGorlich:
@melroy89 remember the timestamp delta thingy you added? I added ut also but you set a delta of 5 ms and I set one of 1000ms. I was not sure about the unit (thought it might be seconds) thats why I went with your value 😁 I will fix it tomorrow

melroy89:
5ms? I thought it was in seconds..

BentiGorlich:
It says it is a unix timestamp which would make it in seconds :thinking:

melroy89:
> It says it is a unix timestamp which would make it in seconds 🤔

Exactly. So everything is now set to 5.0 seconds, the same value to make it consistent, which should be really sufficient enough.. right? 

https://github.com/search?q=repo%3AMbinOrg%2Fmbin%20assertEqualsWithDelta&type=code

If you think that we need 10 seconds, then maybe update all reference to 10.0?

BentiGorlich:
The test failure was with 5 seconds so I suppose it was not enough. I'll wait for the test results and then update the other places

melroy89 (approved):

PR #1707, opened 2025-08-16
Improve marking instances as dead

BentiGorlich:
The `DeliverHandler` previously only increased the counter for failed delivers on instance entities already present in the db, it did not create one. This commit changes that, so that this logic can mark instances as dead even if there is not an instance entity for them

### DIFF
--- a/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
+++ b/src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
@@ -138,7 +138,7 @@ class DeliverHandler extends MbinMessageHandler
 
         try {
             $this->client->post($message->apInboxUrl, $actor, $message->payload);
-            if ($instance->getLastSuccessfulDeliver() < new \DateTime('now - 5 minutes')) {
+            if ($instance->getLastSuccessfulDeliver() < new \DateTimeImmutable('now - 5 minutes')) {
                 $instance->setLastSuccessfulDeliver();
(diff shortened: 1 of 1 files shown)


melroy89 (dismissed):

src/MessageHandler/ActivityPub/Outbox/DeliverHandler.php
```diff
@@ -138,17 +138,16 @@ public function doWork(MessageInterface $message): void
 
         try {
             $this->client->post($message->apInboxUrl, $actor, $message->payload);
-            if ($instance && $instance->getLastSuccessfulDeliver() < new \DateTime('now - 5 minutes')) {
+            if ($instance->getLastSuccessfulDeliver() < new \DateTime('now - 5 minutes')) {
```
melroy89: Maybe a small remark, unrelated to your change. But I think `DateTime` can become `DateTimeImmutable`

BentiGorlich: It could, but since the value is not saved anyways it does not make a difference here...

melroy89: yeah true, but its now more aligned with the Inbox handler.

BentiGorlich: also true, I added it :D

melroy89 (approved):

PR #1709, opened 2025-08-27
Fix symfony generating http URLs with anubis

BentiGorlich:
I noticed that threads and comments did not get created on some Lemmy servers. After looking in the logs I found this response from lemmy:
```json
{
    "error": "unknown",
    "message": "URL failed verification: Http urls are only allowed in debug mode"
}
```

The fix is lying to symfony that the incoming request is actually a https request. Other fixes like changes to the `services.yaml` or other config files did not work for me in this setup (maybe because of the use of UNIX sockets?)


BentiGorlich:
Resources: 
- https://symfony.com/doc/current/routing.html#forcing-https-on-generated-urls
- https://symfony.com/doc/current/deployment/proxies.html#overriding-configuration-behind-hidden-ssl-termination (the one that actually did it)

melroy89 (approved):

docs/02-admin/03-optional-features/07-anubis.md
```diff
@@ -339,6 +344,8 @@ server {
         include fastcgi_params;
         fastcgi_param SCRIPT_FILENAME $realpath_root$fastcgi_script_name;
         fastcgi_param DOCUMENT_ROOT $realpath_root;
+        fastcgi_param SERVER_PORT "443";
```
melroy89, line 351: Maybe also add a comment here in the Nginx code block: "Lie to Symfony about HTTPS, to generate the correct URL.."

docs/02-admin/03-optional-features/07-anubis.md
```diff
@@ -180,12 +180,17 @@ server {
     location / {
         # try to serve file directly, fallback to index.php
         try_files $uri /index.php$is_args$args;
+        fastcgi_param SERVER_PORT "443";
```
melroy89, line 185: Same here?

PR #1714, opened 2025-09-02
Add a type filter to the modlogs

BentiGorlich:
- both the API and the regular site and both the instance wide one and the magazine one
- add tests with the new filters
- unify magazine modlog and instance modlog into the same twig file and the same controller

### DIFF
--- a/config/packages/framework.yaml
+++ b/config/packages/framework.yaml
@@ -33,7 +33,7 @@ framework:
     http_client:
         default_options:
             headers:
-                'User-Agent': 'Mbin/1.8.3 (+https://%kbin_domain%/agent)'
+                'User-Agent': 'Mbin/1.8.4 (+https://%kbin_domain%/agent)'
 
     #esi: true
     #fragments: true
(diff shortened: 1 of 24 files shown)


templates/modlog/front.html.twig
```diff
@@ -29,12 +31,29 @@
...
+    </div>
+    {{ form_end(form) }}
+
     {% for log in logs %}
         <div class="section section--small log">
             <div>
                 {%- with {
                     log: log,
-                    showMagazine: true,
+                    showMagazine: not hasMagazine,
```
melroy89, line 56: When reading this, it doesn't makes any sense.. You have `not` here? So you set `showMagazine` to **false** when `hasMagazine` is **true**!?

BentiGorlich: hasMagazine means it is in a magazine, so don't show the magazine if we are in one. We only need to know which magazine this belongs to, if if we are in a list with multiple magazines

melroy89: Ah yes that makes sense.

melroy89 (approved):

PR #1715, opened 2025-09-05
Fix user outbox

BentiGorlich:


melroy89:
I went from `"totalItems": 1912,` to `"totalItems": 1057,` is that expected?

https://browser.pub/https%3A%2F%2Fkbin.melroy.org%2Fu%2Fmelroy%2Foutbox%3Fpage%3D1

melroy89:
Also, are the APIs fine only AP end-point was effected correct?

BentiGorlich:
> I went from `"totalItems": 1912,` to `"totalItems": 1057,` is that expected?
> 
> https://browser.pub/https%3A%2F%2Fkbin.melroy.org%2Fu%2Fmelroy%2Foutbox%3Fpage%3D1

Yes that is to be expected. Maybe we should create a command that creates a `Create` activity for all the posts all the local users every created, but we have not done that yet. Previously it looked at all your posts, now it looks for all the `Create` activities

BentiGorlich:
> Also, are the APIs fine only AP end-point was effected correct?

Correct, only the AP endpoint is affected

PR #1724, opened 2025-09-21
Properly logout user after changing the password or 2fa

BentiGorlich:
- Previously only the active session was logged out when the password or 2FA changed, now every active session gets logged out when that happens. The way symfony does this is by checking whether the serialized version of a user saved in the session object is equal to the freshly fetched one. Previously the `User::isEqualTo` function did not really check anything, other than whether the user is banned. To properly check that we need to add some more information to the serialized object, so do that in the `User::__serialize` function
- To also invalidate the remember me cookies they will be signed with some properties of the user now, see `security.yaml`
- Change the password change dialog so that the user also has to put in a 2fa code if they have that enabled
- Change the 2fa change dialog (remove and regenerate backup codes) so that a user has to insert their password and 2fa code if they want to change that
- create a dto for generating 2fa codes for the setup so the user object does not get changed. With the new logout mechanism one would get logged out as soon as one accesses the 2fa setup page (also it is just better this way)

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -53,6 +53,8 @@ security:
                 path: /
                 token_provider:
                     doctrine: true
+                # see https://symfony.com/doc/current/security/remember_me.html#using-signed-remember-me-tokens
+                signature_properties: ['username', 'password', 'totpSecret', 'isBanned', 'isDeleted', 'markedForDeletionAt', 'visibility']
             two_factor:
                 auth_form_path: 2fa_login
                 check_path: 2fa_login_check
--- a/src/Controller/User/Profile/User2FAController.php
+++ b/src/Controller/User/Profile/User2FAController.php
@@ -5,8 +5,11 @@ declare(strict_types=1);
 namespace App\Controller\User\Profile;
 
 use App\Controller\AbstractController;
+use App\DTO\Temp2FADto;
 use App\DTO\UserDto;
 use App\Entity\User;
+use App\Form\UserDisable2FAType;
+use App\Form\UserRegenerate2FABackupType;
 use App\Form\UserTwoFactorType;
 use App\Service\TwoFactorManager;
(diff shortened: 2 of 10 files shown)


BentiGorlich:
Maybe including the `visibility` property is not needed (in `User::isEqualTo` and the remember me signing) because suspended users are still able to login, they are just not able to post anything (I think)

melroy89:
Just to confirm: this will log-out the user on all sessions right? Not just the current computer? Because the cookie becomes invalid on all machines..

config/packages/security.yaml
```diff
@@ -53,6 +53,8 @@ security:
                 path: /
                 token_provider:
                     doctrine: true
+                # see https://symfony.com/doc/current/security/remember_me.html#using-signed-remember-me-tokens
+                signature_properties: ['username', 'password', 'totpSecret', 'isBanned', 'isDeleted', 'markedForDeletionAt', 'visibility']
```
melroy89: This will also log the user out if the `visibility` has changed by the admin.. is that really necessary?

BentiGorlich: Like I said :)

> Maybe including the visibility property is not needed (in User::isEqualTo and the remember me signing) because suspended users are still able to login, they are just not able to post anything (I think)

BentiGorlich:
> Just to confirm: this will log-out the user on all sessions right? Not just the current computer? Because the cookie becomes invalid on all machines..

Correct. As it does not do that without these changes

config/packages/security.yaml
melroy89: Yeah so let's remove that attribute here. So the user doesn't notice anything.

BentiGorlich: done

melroy89 (approved):

melroy89:
Great PR! Import security fixes IMO

PR #1725, opened 2025-09-21
Add command to rotate private keys

BentiGorlich:
- add a command to rotate the private keys for one user or all local users
- add fields for storing the old public- and private-keys
- add option to deliver a message by signing it with the old private key
- when updating a user we also update their public key now, also log a message when a public key was changed


docs/02-admin/04-running-mbin/05-cli.md
```diff
@@ -121,6 +121,24 @@ Options:
 > [!NOTE] 
 > If neither `--activate` nor `--deactivate` are provided, the current verification status will be returned
 
+
+### Rotate users private keys
+
+> [!WARNING]
+> You should inform your users about this when you are running this command
```
melroy89: Maybe explain why you should tell it to the users. What is the impact, etc. Do they need to take some actions?..

BentiGorlich: The impact is stated in the final message of the command, so yeah I should add it to the docs: 

> It might take up to 24 hours for other software to get the new public keys

PR #1743, opened 2025-10-02
Fix missing delete activity when normal user deletes their content

BentiGorlich:
- Add tests for a local user deleting their content
- fix the `DeleteService::announceIfNecessary` not taking into account local users without being moderators


BentiGorlich:
@jwr1 FYI it was a bug

PR #1744, opened 2025-10-02
Add dropdown for who can send you a private message

### LINKED ISSUE #1439: Setting to disable DMs / user messages
Currently everyone can DM everyone else, but I've seen some people say that they would prefer to simply not be DM'd at all and there is currently no user setting for this.

Add a new user setting to turn off the ability to receive DMs (possibly excluding those sent by instance admins? other exclusions like mod-to-mod too? idk).

I suppose the _ideal_ solution would include some kind of UI feedback so that if user Alfed tries to message user Beatrice (who has DMs disabled), Alfred would be aware that Beatrice will never receive the message (perhaps most ideally, the ability to send messages to Beatrice would be disabled on Alfred's end so that he doesn't waste time writing a message that will never be received anyway).

There isn't really a clear alternative that is scoped specifically to messages without affecting other things (like blocking all other users).

**Additional context**
Related to #1423 and the user request for this feature was prompted by the same circumstances in 1423.

BentiGorlich:
- add user setting `directMessageSetting`
- add a new enum for the new setting
  - `EDirectMessageSettings::Everyone` everyone can send you a message (default)
  - `EDirectMessageSettings::FollowersOnly` only your followers can send you a message
  -  `EDirectMessageSettings::Nobody` nobody can send you a message
- add a new exception type `UserCannotReceiveDirectMessage`

closes #1439

### DIFF
--- a/README.md
+++ b/README.md
@@ -284,14 +284,21 @@ Unofficial magazines:
                 </a>
             </td>
             <td align="center">
+                <a href="https://github.com/e-michalak">
+                    <img src="https://avatars.githubusercontent.com/u/236532325?v=4" width="100;" alt="e-michalak"/>
+                    <br />
+                    <sub><b>e-michalak</b></sub>
+                </a>
+            </td>
+		</tr>
+		<tr>
+            <td align="center">
                 <a href="https://github.com/MHLoppy">
                     <img src="https://avatars.githubusercontent.com/u/12670674?v=4" width="100;" alt="MHLoppy"/>
                     <br />
                     <sub><b>Mark Heath</b></sub>
                 </a>
             </td>
-		</tr>
-		<tr>
             <td align="center">
                 <a href="https://github.com/robertolopezlopez">
                     <img src="https://avatars.githubusercontent.com/u/1519467?v=4" width="100;" alt="robertolopezlopez"/>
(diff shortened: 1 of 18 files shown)


BentiGorlich:
This does setting does not yet federate. If that is required for it, then it will take a bit longer.
Additionally I am not so sure whether `Followers Only` is the best middleground as lemmy users cannot follow mbin users, so maybe a `Followers and Followings` setting might be more useful...

melroy89:
> This does setting does not yet federate. If that is required for it, then it will take a bit longer. Additionally I am not so sure whether `Followers Only` is the best middleground as lemmy users cannot follow mbin users, so maybe a `Followers and Followings` setting might be more useful...

What do you mean with setting does not federate? This is just a local personal configuration anyway, right? What do you want to federate about it?

BentiGorlich:
So that other people are not confused when they do not get a reply. Ideally they would not be able to send a message in the first place or gwt notified that the message could not be delivered (maybe indicated by a `Reject` activity)

melroy89:
> So that other people are not confused when they do not get a reply. Ideally they would not be able to send a message in the first place or gwt notified that the message could not be delivered (maybe indicated by a `Reject` activity)

Yeah indeed. A default response might be an idea. However this can be misused to ddos the instance.

PR #1749, opened 2025-10-10
unify forms for creating an entry (article, link, photo) into one

blued-gear:


src/Service/EntryManager.php
```diff
@@ -151,30 +154,25 @@ public function create(EntryDto $dto, User $user, bool $rateLimit = true, bool $
...
-        }
-
-        if ($dto->url) {
-            $entry->type = Entry::ENTRY_TYPE_LINK;
-
-            return $entry;
-        }
-
-        if ($dto->body) {
+        } else if ($dto->url) {
```
blued-gear: Before my change entries with a URL would be of type *link* if the URL does not point to an image even if an image was attached to the entry.
If this changed behavior is not wanted, I will revert it.

BentiGorlich: Fine by me :)

src/Controller/Api/Entry/MagazineEntryCreateApi.php
```diff
@@ -368,4 +372,124 @@ public function uploadImage(
...
+                    ],
+                )
+            ),
+            encoding: [
+                'imageUpload' => [
+                    'contentType' => ImageManager::IMAGE_MIMETYPE_STR,
+                ],
+            ]
+        ),
+        description: '"uploadImage" is optional'
```
blued-gear: In Swagger `uploadImage` is marked as *required* and I can't find an option to change this.

src/Controller/Api/Entry/MagazineEntryCreateApi.php
```diff
@@ -368,4 +372,124 @@ public function uploadImage(
...
+                        Entry::ENTRY_TYPE_ARTICLE,
+                        Entry::ENTRY_TYPE_LINK,
+                        Entry::ENTRY_TYPE_VIDEO,
+                        Entry::ENTRY_TYPE_IMAGE,
+                        'common',
+                    ],
+                )
+            ),
+            encoding: [
+                'imageUpload' => [
```
blued-gear, line 439: shouldn't this be `uploadImage`?

BentiGorlich: The constant you're referencing in the `groups` property above `ImageUploadDto::IMAGE_UPLOAD` is set to `imageUpload`. I guess that is why it is using that name

src/DTO/EntryRequestDto.php
```diff
@@ -9,7 +9,7 @@
 use OpenApi\Attributes as OA;
 use Symfony\Component\Serializer\Annotation\Groups;
 
-#[OA\Schema()]
+#[OA\Schema(required: ['title'])]
```
BentiGorlich, line 12: I would just add the `required: true` to the indiviual properties, in this case only `title`

src/Controller/Api/Entry/MagazineEntryCreateApi.php
BentiGorlich: I think you have to change the property `$uploadImage` in `ImageUploadDto` and add a `required: false` to the `OA\Property` attribute. `EntryRequestDto` extends `ContentRequestDto` which extends `ImageUploadDto`

src/Controller/Api/Entry/MagazineEntryCreateApi.php
```diff
@@ -368,4 +372,126 @@ public function uploadImage(
...
+                'uploadImage' => [
+                    'contentType' => ImageManager::IMAGE_MIMETYPE_STR,
+                ],
+            ]
+        )
+    )]
+    #[OA\Tag(name: 'magazine')]
+    #[Security(name: 'oauth2', scopes: ['entry:create'])]
+    #[IsGranted('ROLE_OAUTH2_ENTRY:CREATE')]
+    public function entry(
```
BentiGorlich, line 451: Just a note: if you were to add phpdoc to this method the comment would show up in the open API doc like this:

```suggestion
    /**
     * Some comment text for the docs
     */
    public function entry(
```

Image

PR #1751, opened 2025-10-14
filter crosspost candidates by image and write tests for crossposts

blued-gear:
I noticed that many entries are marked as crossposts even if they aren't, just because they share the same title (very prominent in *irl* communities).

This PR fixes this by including the image of an entry as a filter in the crosspost query.\
Also we now have some tests for the crosspost detection.

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -6,12 +6,14 @@ doctrine:
             enumApplicationStatus: App\DoctrineExtensions\DBAL\Types\EnumApplicationStatus
             enumNotificationStatus: App\DoctrineExtensions\DBAL\Types\EnumNotificationStatus
             enumSortOptions: App\DoctrineExtensions\DBAL\Types\EnumSortOptions
+            enumDirectMessageSettings: App\DoctrineExtensions\DBAL\Types\EnumDirectMessageSettings
         mapping_types:
             user_type: string
             citext: citext
             enumApplicationStatus: string
             enumNotificationStatus: string
             enumSortOptions: string
+            enumDirectMessageSettings: string
 
         # IMPORTANT: You MUST configure your server version,
         # either here or in the DATABASE_URL env var (see .env file)
--- a/migrations/Version20250924105525.php
+++ b/migrations/Version20250924105525.php
@@ -1,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20250924105525 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Add column direct_message_setting to "user"';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('CREATE TYPE enumDirectMessageSettings AS ENUM(\'everyone\', \'followers_only\', \'nobody\')');
+        $this->addSql('ALTER TABLE "user" ADD direct_message_setting enumDirectMessageSettings DEFAULT \'everyone\' NOT NULL');
+    }
+
+    public function down(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE "user" DROP direct_message_setting');
+        $this->addSql('DROP TYPE enumDirectMessageSettings');
+    }
+}
--- a/src/DTO/UserSettingsDto.php
+++ b/src/DTO/UserSettingsDto.php
@@ -5,6 +5,7 @@ declare(strict_types=1);
 namespace App\DTO;
 
 use App\Entity\User;
+use App\Enums\EDirectMessageSettings;
 use App\PageView\EntryCommentPageView;
 use App\PageView\EntryPageView;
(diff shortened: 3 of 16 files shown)


src/Repository/EntryRepository.php
```diff
@@ -496,14 +496,21 @@ public function findCross(Entry $entry): array
                 ->setParameter('title', $entry->title);
         }
 
+        if($entry->image) {
+            $qb->leftJoin('e.image', 'i')
+                ->andWhere('e.image is not null')
```
BentiGorlich: This is redundant. SQL does not have null exceptions :)

src/Repository/EntryRepository.php
```diff
@@ -496,14 +496,21 @@ public function findCross(Entry $entry): array
...
         $qb->andWhere('e.id != :id')
             ->andWhere('m.visibility = :visibility')
             ->andWhere('e.visibility = :visibility')
             ->andWhere('u.isDeleted = false')
             ->innerJoin('e.user', 'u')
-            ->join('e.magazine', 'm')
-            ->setParameter('visibility', VisibilityInterface::VISIBILITY_VISIBLE)
+            ->innerJoin('e.magazine', 'm')
             ->setParameter('id', $entry->getId())
+            ->setParameter('visibility', VisibilityInterface::VISIBILITY_VISIBLE)
```
BentiGorlich, line 512: Why these changes? especially replacing the `join` with the `innerJoin`?

tests/FactoryTrait.php
```diff
@@ -596,14 +595,15 @@ public function register($active = false): KernelBrowser
...
+    public function getKibbyImageDto(string $suffix = ''): ImageDto
     {
         $imageRepository = $this->imageRepository;
         $imageFactory = $this->imageFactory;
 
         // Uploading a file appears to delete the file at the given path, so make a copy before upload
         $tmpPath = bin2hex(random_bytes(32));
-        copy($this->kibbyPath, $tmpPath.'.png');
+        $srcPath = dirname($this->kibbyPath).'/'.basename($this->kibbyPath, '.png').$suffix.'.png';
+        copy($srcPath, $tmpPath.'.png');
```
BentiGorlich, line 617: What is this for? :sweat_smile:

tests/assets/kibby_emoji_flipped.png
BentiGorlich, line 1: Was this added by mistake?

tests/Functional/Misc/Entry/CrosspostDetectionTest.php
```diff
@@ -0,0 +1,143 @@
...
+
+        $this->checkCrossposts([$entry1, $entry2, $entry3]);
+        $this->checkCrossposts([$entry4]);
+    }
+
+    public function testCrosspostsByTitleWithImageFilter(): void
+    {
+        $user = $this->getUserByUsername('JohnDoe');
+        $img1 = $this->getKibbyImageDto();
+        $img2 = $this->getKibbyImageDto('_2');
```
BentiGorlich: If the only purpose of the added image is to have a second dto here, just create a file with random bytes in it and call it a png one :thinking: 
I think that this might actually not work anyways, since our code matches the sha1 hash of image files and only creates a new one if the hash does not already exist

src/Repository/EntryRepository.php
blued-gear: `join` is an alias for `innerJoin` (if you look at the function definition of `join` you will see it just calls `innerJoin`) so I wanted to make it consistent.

tests/Functional/Misc/Entry/CrosspostDetectionTest.php
blued-gear: The images are different (number two is mirrored), so their hashes are too.

This is also why I changed `getKibbyImageDto` to easily use the two different images.

tests/assets/kibby_emoji_flipped.png
blued-gear: nope, it is a mirrored version of kibby to test posts with different images.

tests/FactoryTrait.php
blued-gear: Maybe a bit of laziness from my side. I added a changed version of kibby and to avoid duplicating mist of this function, I just added a parameter to append a suffix to the asset filename (before the file extension).

tests/FactoryTrait.php
```diff
@@ -596,14 +596,15 @@ public function register($active = false): KernelBrowser
         return $this->client;
     }
 
-    public function getKibbyImageDto(): ImageDto
+    public function getKibbyImageDto(string $suffix = ''): ImageDto
```
BentiGorlich: I think I would prefer it if one would not be able to pass a suffix through. Maybe this could be an intern method and then you'd have one method for the flipped image and one for the regular one. This way it is not as confusing :D

tests/assets/kibby_emoji_flipped.png
BentiGorlich: Could you then rename it to `kibby_emofi_flipped.png` or something, so that it is obvious what this is?

BentiGorlich (changes requested):

BentiGorlich (approved):
Very nice!

PR #1756, opened 2025-10-22
Fix mentions from mastodon

### LINKED ISSUE #538: If multiple persons use the same name, they get each others notifications.
People can register with the same name on multiple instances. If multiple persons use the same name, they get each others notifications

BentiGorlich:
Mentions from mastodon have the "problem" that the link text is only the username without the domain, so not the full handle. For that Mastodon uses the `tag` property in which they put mentions (`type == Mention`) with the profile URL (`href`) and the full handle (`name`) (which we are doing the same way, see `MentionsWrapper::build`). The `MarkdownConverter` (the one for html -> Markdown) now has a new parameter for this `tag` array. There was already some logic for replacing the link text, but it did not work, as it expected a webfinger handle which just is not the case. As a fallback to the tag array we now search for an actor (which accepts the profile URL)

And using `assertEqualsWithDelta` on timestamp compare tests to make the tests more stable.

Fixes #538

### DIFF
--- a/tests/Functional/Controller/Api/Magazine/Moderate/MagazineBanApiTest.php
+++ b/tests/Functional/Controller/Api/Magazine/Moderate/MagazineBanApiTest.php
@@ -161,8 +161,8 @@ class MagazineBanApiTest extends WebTestCase
         self::assertSame($user->getId(), $jsonData['bannedBy']['userId']);
 
         $actualExpiry = \DateTimeImmutable::createFromFormat(\DateTimeImmutable::ATOM, $jsonData['expiredAt']);
-        // Hopefully the API responds fast enough that there is only a max delta of 1 second between these two timestamps
-        self::assertEqualsWithDelta($expiredAt->getTimestamp(), $actualExpiry->getTimestamp(), 1.0);
+        // Hopefully the API responds fast enough that there is only a max delta of 5 second between these two timestamps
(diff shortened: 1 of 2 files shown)


melroy89:
This is a great fix I believe! Thank you.

We don't have this issue with other fediverse software like lemmy?

And if it's just a handle without domain name, then will we assume a local user?

Ps. The test is failing.

BentiGorlich:
> And if it's just a handle without domain name, then will we assume a local user?

Yes correct.

> We don't have this issue with other fediverse software like lemmy?

The issue could exist for everything that does not supply a markdown source. That excludes lemmy. 

> Ps. The test is failing.

Should be fixed now :)

melroy89:
Uuh oops?

```
1) App\Tests\Functional\Controller\Api\Magazine\Moderate\MagazineActionReportsApiTest::testApiCanRejectReport
Failed asserting that 1761216.0 is identical to 1761217.0.
```

BentiGorlich:
Yeah that is a flaky test because it assumes that creating something returns the same timestamp in seconds... I restarted the workflow, lets see

BentiGorlich:
Well I know that is bad practice, it is not my test

melroy89:
OK. I fixed it (I hope).

Its also not my test, but anyhow, I like to have stable tests :)

melroy89 (dismissed):

melroy89 (approved):

PR #1762, opened 2025-10-23
Upgrade to trixie Docker images

melroy89:
- Upgrade docker images from bookworm to trixie (latest stable Debian release)  https://www.debian.org/releases/
- Both `composer.yaml` and `compose.dev.yaml`

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,7 +2,7 @@
 
 
 # Base FrankenPHP image
-FROM dunglas/frankenphp:1-php8.4-bookworm AS base
+FROM dunglas/frankenphp:1-php8.4-trixie AS base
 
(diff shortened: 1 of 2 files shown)


melroy89:
You can first test it locally on you setup, just to be fully sure it will work just fine.

With Docker you still might need to do a docker pull, otherwise nothing happens. 

For example: `docker compose pull <container-name>` where the `container-name` is the name defined in the yaml file (also know as the service name).

That pull needs to be repeated for each service you might want to upgrade....

jwr1:
I can try testing it out sometime tomorrow.

melroy89:
> I can try testing it out sometime tomorrow.

Sure! If you have questions or problems; of course let me know.

PR #1769, opened 2025-10-29
Revert all

melroy89:
I still don't understand why this flaky integration test is failing. But We can just revert all back to PHP 8.3

### DIFF
--- a/ci/Dockerfile
+++ b/ci/Dockerfile
@@ -5,11 +5,11 @@ ARG DEBIAN_FRONTEND=noninteractive
 
 RUN apt update
 RUN apt upgrade -y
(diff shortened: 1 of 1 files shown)


melroy89:
This will be a force merge.. Since main is broken now on PHP-CS-Fixer.. and apparently also broken on this single integration test.. A test I don't fully understand to be honest.

melroy89:
We now also have two tagged docker images, so we can easily switch between them (instead of only having `latest`):

- 8.3 (Ubuntu 24.04):  https://github.com/MbinOrg/mbin/pkgs/container/mbin-pipeline-image/560081536?tag=8.3
- 8.4 (Ubuntu 25.05): https://github.com/MbinOrg/mbin/pkgs/container/mbin-pipeline-image/559785975?tag=8.4

PR #1770, opened 2025-10-30
Allow global mods to manage signups and receive notifications

jwr1:
This PR will allow global mods to manage (view, approve, and deny) account signups. Additionally, global mods can now receive signup notifications, which is, of course, configurable in the notification settings (same as admins). There is a new option in the global app menu (only for global mods) that will link to the signup requests page.

I was able to test almost everything in the PR and am using it on kbin.earth, but I did not get a chance to verify that the changes to the API routes work. Futhermore, this is my first change to the PHP code and do not have PHP experience, so let me know if I did something wrong :)


BentiGorlich:
First of all: awesome that you implemented it already :)

Second: you have to execute the linter ([docs](https://docs.joinmbin.org/contributing/linting#php-code)) so the CI does not fail

To test the changes it would be best for you to add test cases for the changes. But I honestly don't know if I added test cases when I added this feature, so this might be too much too ask

jwr1:
> To test the changes it would be best for you to add test cases for the changes. But I honestly don't know if I added test cases when I added this feature, so this might be too much too ask

I unfortunately have no idea how to add a testcase, but I'll see if I can figure it out. Would it just be a test case to see if the new link shows up in the menu for global mods, or would I need to test something else?

jwr1:
Can you help me with the error in my test?

BentiGorlich:
Exactly. You'd test that the link is there and that a global mod can access the page. You cpuld alsp test the functionality, so create a user with signup approval on and test accepting and rejecting this request

BentiGorlich:
What it says is that there is no `Signup Requests` link it can click. Could have multiple causes:
- settings don't really work (though it should)
- User not getting the global moderstor role (though it should)
- something wrong with the selector in general. If you can debug it than I'd make multiple variables filtering down and looking at the different html values in there to check whether you just missed something

jwr1:
> so create a user with signup approval on and test accepting and rejecting this request

What would be the best way to create a new user? Should I just copy the way it's done in RegisterControllerTest.php?

Also, if I create two separate tests in the same file, do the changes I make persist between them? For instance, if I set `MBIN_NEW_USERS_NEED_APPROVAL` to true in one test, do I still need to do it in all the other tests?

BentiGorlich:
In theory the way you did it should work, but it does probably not trigger notifications. I am not sure if it sets the applicationStatus correctly, but it should.

If you have 2 separate tests the db gets reset between the execution of the tests. PHPUnit has a `setUp` method that gets called before each test case. You should overwrite that for shared boiler plate code

src/Service/Notification/SignupNotificationManager.php
```diff
@@ -22,7 +22,9 @@ public function __construct(
 
     public function sendNewSignupNotification(User $newUser): void
     {
-        $receivers = $this->userRepository->findAllAdmins();
+        $receiver_admins = $this->userRepository->findAllAdmins();
+        $receiver_moderators = $this->userRepository->findAllModerators();
+        $receivers = array_merge($receiver_admins, $receiver_moderators);
```
BentiGorlich: Maybe you should make a `array_unique` call here as well? In case that an admin is also a moderator?

tests/Functional/Controller/Moderator/ModeratorSignupRequestsControllerTest.php
```diff
@@ -0,0 +1,22 @@
...
+    public function testModeratorCanViewSignupRequests(): void
+    {
+        $this->settingsManager->set('MBIN_NEW_USERS_NEED_APPROVAL', true);
+
+        $this->client->loginUser($this->getUserByUsername('moderator', isModerator: true));
+
+        $crawler = $this->client->request('GET', '/');
+        $this->client->click($crawler->filter('#header menu')->selectLink('Signup requests')->link());
+
+        $this->assertSelectorTextContains('#main h3', 'Signup Requests');
```
BentiGorlich, line 21: before this you should make a `assertResponseIsSuccessful` call

melroy89:
@jwr1 maybe we should have documented this here for example? https://docs.joinmbin.org/user/user_guide (both how to configure global mods, as well as how mods can use this new feature)

Issue #248, opened 2023-11-11
Correct propagation of mod actions


BentiGorlich:
**Is your feature request related to a problem? Please describe.**
Modertion actions are not beeing federated right now. That includes:
- adding/removing a moderator
- reporting anything

**Describe the solution you'd like**
add federation to those events

**Describe alternatives you've considered**
there are none

**Additional context**
Lemmy docs on how this is implemented:
- https://join-lemmy.org/docs/contributors/05-federation.html#report-post-comment-or-private-message
- https://join-lemmy.org/docs/contributors/05-federation.html#moderation
- https://github.com/LemmyNet/lemmy/blob/main/crates/apub/assets/lemmy/activities/community/report_page.json

BentiGorlich:
Actions that definitely need to be investigated/added:
- [x] Pinning/unpinning of posts via a moderator: #831
- [ ] Changing the language of a Post by a moderator
- [ ] Marking a post NSFW by a moderator
- [x] Banning/unbanning a user: #943 #1700
- [x] Deletion of a post
- [ ] Transferring the post to another magazine

BentiGorlich:
For federating the transfer of a post to another magazine:
Piefed did implement it and an example activity looks like this ([see post](https://piefed.social/comment/9645163) by wjs018):

```json
{  
  "actor": "https://crust.piefed.social/u/wjs018",  
  "cc": [  
    "https://piefed.wjs018.xyz/c/testing"  
  ],  
  "id": "https://crust.piefed.social/activities/move/l7NaOQha12PQyYv",  
  "object": "https://crust.piefed.social/c/testing/p/108274/this-post-will-be-moved",  
  "origin": "https://piefed.wjs018.xyz/c/testing",  
  "target": "https://crust.piefed.social/c/test",  
  "to": [  
    "https://www.w3.org/ns/activitystreams#Public"  
  ],  
  "type": "Move"  
}  
```
> For context, this was moved from [!testing@piefed.wjs018.xyz](https://gehirneimer.de/search?search%5Bq%5D=testing@piefed.wjs018.xyz) to [!test@crust.piefed.social](https://gehirneimer.de/search?search%5Bq%5D=test@crust.piefed.social). The author, wjs018 , is a moderator of the community that the post was initially posted to. I did find a bug in this process, the post flair, which should be community-specific, was moved as well when it should have been stripped. So, I created an issue for tracking that fix.

julianlam:
[...]

PR #1773, opened 2025-10-31
Introducing thread/microblog locking

BentiGorlich:
- Threads and microblogs can now be locked so comments cannot be added anymore. This includes:
  - API support (new scopes)
  - ActivityPub support (`Lock` activity, post editing, etc.)
- Authors and moderators can lock threads and microblogs
- Add modlog items for the new lock and unlock events
- Add tests for basically everything

Related to #248

### DIFF
--- a/src/Service/ActivityPub/MarkdownConverter.php
+++ b/src/Service/ActivityPub/MarkdownConverter.php
@@ -33,7 +33,7 @@ class MarkdownConverter
 
         foreach ($matches as $match) {
             if ($this->mentionManager->extract($match[1])) {
-                $mentionFromTag = array_filter($apTags, fn ($tag) => 'Mention' === ($tag['type'] ?? '') && $match[2] === ($tag['href'] ?? ''));
+                $mentionFromTag = $this->findMentionFromTag($match, $apTags);
                 if (\count($mentionFromTag)) {
                     $mentionFromTagObj = $mentionFromTag[array_key_first($mentionFromTag)];
                     $mentioned = null;
@@ -51,7 +51,8 @@ class MarkdownConverter
                 } else {
                     try {
                         $actor = $this->activityPubManager->findActorOrCreate($match[2]);
-                        $replace = '@'.($actor instanceof User ? $actor->username : $actor->name);
+                        $username = $actor instanceof User ? $actor->username : $actor->name;
+                        $replace = $this->mentionManager->getUsername($username, true);
                     } catch (\Throwable) {
                         $replace = $match[1];
                     }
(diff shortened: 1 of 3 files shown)


melroy89:
Is still a draft or not?

BentiGorlich:
Nope, ready to be reviewed :)

src/Service/ActivityPub/ActivityJsonBuilder.php
```diff
@@ -478,6 +479,31 @@ private function buildBlockFromActivity(Activity $activity): array
...
+            '@context' => $this->contextsProvider->referencedContexts(),
+            'id' => $this->urlGenerator->generate('ap_object', ['id' => $activity->uuid], UrlGeneratorInterface::ABSOLUTE_URL),
+            'type' => 'Lock',
+            'actor' => $this->personFactory->getActivityPubId($activity->userActor),
+            'to' => [ActivityPubActivityInterface::PUBLIC_URL],
+            'cc' => [
+                $this->groupFactory->getActivityPubId($object->magazine),
+                $this->urlGenerator->generate('ap_user_followers', ['username' => $activity->userActor->username], UrlGeneratorInterface::ABSOLUTE_URL),
+            ],
+            'object' => $objectUrl,
```
melroy89, line 515: Maybe still good to check if `$objectUrl` is no null/empty?

src/MessageHandler/ActivityPub/Inbox/CreateHandler.php
```diff
@@ -113,6 +115,8 @@ public function doWork(MessageInterface $message): void
             $this->logger->info('[CreateHandler::doWork] Did not create the post, because the user\'s instance is banned');
         } catch (UserBlockedException $e) {
```
melroy89, line 114: Unrelated to your change, but I think ` $e` can be removed here as well. Since its not used.

melroy89:
I notice you recently made (yesterday) some discoveries in this PR.. maybe its a good idea to wait with large PR after the 1.9.0 release..? Do you agree?

BentiGorlich:
I only resolved the merge conflict. But I don't need to have this in 1.9.0

src/Service/ActivityPub/ActivityJsonBuilder.php
BentiGorlich: It can't be null or empty because in the if/else above there is no case in which it will not be assigned

migrations/Version20251031174052.php
```diff
@@ -0,0 +1,28 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20251031174052 extends AbstractMigration
```
melroy89, line 10: due to the fact we might introduced other migrations.. it might be wise to upgrade the file name and class name to reflect the current date again..

BentiGorlich: In theory yes, but I have already been running this migration locally and on gehirneimer, so it would complicate things there... If you do not insist on it, I'd prefer to leave it as is

melroy89 (dismissed):

migrations/Version20251031174052.php
melroy89: OK let's hope then it won't case any issues and symfony migration is smart enough

BentiGorlich:
Sorry, I had to fix conflicts :)

melroy89 (approved):

PR #1775, opened 2025-11-03
Add revert option to rotate keys command

BentiGorlich:
- add a revert option that switches the old and new private and public key

### DIFF
--- a/src/Command/UserRotatePrivateKeys.php
+++ b/src/Command/UserRotatePrivateKeys.php
@@ -75,6 +75,13 @@ class UserRotatePrivateKeys extends Command
         }
 
         $userCount = \count($users);
+        $action = $revert ? 'reverted' : 'rotated';
+
+        $io->confirm("This command will $action the private and public key of $userCount users. "
+            .'After running this command it can take up to 24 hours for other instances to update their stored public keys. '
+            .'In this timeframe federation might be impacted by this, as those services cannot successfully verify the identity of your users. '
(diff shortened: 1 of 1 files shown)


src/Entity/Traits/ActivityPubActorTrait.php
```diff
@@ -79,12 +79,24 @@ public function getPublicKey(): ?string
...
-        // set new private and public key
-        KeysGenerator::generate($this);
+        if (!$revert) {
+            $this->oldPrivateKey = $this->privateKey;
+            $this->oldPublicKey = $this->publicKey;
+            // set new private and public key
+            KeysGenerator::generate($this);
+        } else {
+            if (null === $this->oldPrivateKey || null === $this->oldPublicKey) {
+                throw new \InvalidArgumentException('you cannot revert if there is no old key');
```
melroy89, line 91: `oldPrivateKey` & `oldPublicKey` now only lives in memory right? 

So a simple PHP-FPM restart will get rid of the old keys. So the message should be more about that you can't revert the keys in case you restarted PHP or the server? Since after all the data will be lost?

BentiGorlich: No that is not the case. This is a trait that the `User` and `Magazine` entity use. These properties are saved to the db as soon as they are assigned. This exception only prevents the private or public key from becoming `null`

melroy89: owh ok then its working as expected I guess.

melroy89:
This option was not documented.

PR #1777, opened 2025-11-04
sort list of languages for entry form alphabetically

blued-gear:
The options for the language of a new entry are now sorted.

### DIFF
--- a/src/Form/Type/LanguageType.php
+++ b/src/Form/Type/LanguageType.php
@@ -49,6 +49,7 @@ class LanguageType extends AbstractType
                         }
 
                         natcasesort($choices);
+
                         return array_flip($choices);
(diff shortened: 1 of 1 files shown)


blued-gear:
`UserSettingsType` is using `LanguageType` from Symfony instead of the class defined in the project. Is this intended or canit be replaced to use the sorted list there too?

BentiGorlich:
> `UserSettingsType` is using `LanguageType` from Symfony instead of the class defined in the project. Is this intended or canit be replaced to use the sorted list there too?

I am not sure. I have not touched that part of the code I think, or at least I cannot remember it. 

An idea I had was that the dropdown should be limited to the languages a user has selected (if they have selected any) in their settings (filter thread and microblogs by language), but that can be done in another PR :)

PR #1780, opened 2025-11-04
Introducing federation allow list and refactoring

### LINKED ISSUE #1482: Allowlist for federation
It would really make the decission of enabling federation on the instance easier if it would be possible to hand pick the servers that I want to federate with.

1. I would love to be able to just put together a list of instances that I want to federate with, even if it's just a text file on server
  1.1. It would be perfect it it could be available as a setting from the control panel at some point

It would be possible to just blacklist servers as needed, I suppose, but this requires way more energy and labour put into the moderation, and usually it's done on a volunteer basis.

**Additional context**
It was discussed on Matrix, and @BentiGorlich not only agreed it's a cool idea, but also said that it would be relatively easy to implement.

BentiGorlich:
- refactor our banned instance setting. Banning an instance is now just noted in the instance table. The migration for this has a complete up and down procedure, so it is non-destructive
- Add another federation setting `USE_FEDERATION_ALLOW_LIST` which switches from a ban list to an allow list, meaning a more restrictive federation where one only federates with the instances explicitly allowed
- Redo the admin interface for defederating from an instance. There is now a list of instances with a button next to them to ban/unban them or (in case of the allowlist) allow/deny federation with them grafik grafik
- defederating from an instance now shows the connections from the instance to ours and requires the admin to confirm the defederation Bildschirmfoto vom 2025-11-04 11-47-42

Closes #1482


melroy89:
I would still like to have a simple list of managing my ban list, like we have now:

image


You didn't remove that part I hope? Or is the only way of managing the bans of instances by scrolling through 7000 instances in my case...

We currently have a merge conflict that needs to be resolved first, so I couldn't easily test it.

BentiGorlich:
> You didn't remove that part I hope?

I didn't remove that part, but I changed it to look a lot different. It looks like this now: 
grafik


> We currently have a merge conflict that needs to be resolved first, so I couldn't easily test it.

I don't know why that would keep you from testing, but sure that will be resolved.

melroy89:
> I don't know why that would keep you from testing, but sure that will be resolved.

now I think about it.. I don't know either 😆 .. I was afraid I would break my server when trying to switch between branches with conflicts.. but thats just not true.

melroy89:
If I ban the an instance that is listed at the top of my instance table. After that the same instance is no longer at the top of the instance table? 

Before:

image


After the ban:

image


I thought that you are ordering by `id`? I can confirm that the ID in the table keeps the same, only `is_banned` changes.. which makes sense of course. 

image

How can my instance table now have a different order then??

---

Finally, I unban again the same instance.. And its still not at the top of the instance list:

image

image

If we are ordering by `id`.. it shouldn't matter what happened to the data.. The order of the instances should always be the same.

e-t:
There appears to be a SQL syntax error in Version20250813132233.php.
The INSERT statement defines more columns than there are VALUES to fill them.

INSERT INTO instance(id, domain, created_at, failed_delivers, updated_at, is_banned)
VALUES (NEXTVAL('instance_id_seq'), current_timestamp(0), 0, current_timestamp(0), true);

There is no value to populate the 'domain' column. Should it be tempRow.value?

melroy89:
> There appears to be a SQL syntax error in Version20250813132233.php. The INSERT statement defines more columns than there are VALUES to fill them.
> 
> INSERT INTO instance(id, domain, created_at, failed_delivers, updated_at, is_banned) VALUES (NEXTVAL('instance_id_seq'), current_timestamp(0), 0, current_timestamp(0), true);
> 
> There is no value to populate the 'domain' column. Should it be tempRow.value?

I think you are right. We will look into this. Not sure why we didn't spot this earlier, you would expect that the migration would fail on this.

e-t:
> you would expect that the migration would fail on this.

The migration did fail for me when I updated my instance yesterday.

melroy89:
> > you would expect that the migration would fail on this.
> 
> The migration did fail for me when I updated my instance yesterday.

Problem should now be resolved on the latest `main` branch. Or if you are using docker, in that case on the `main` tag (not to be confused with `latest` tag)

melroy89:
Today we will also release a new stable release, which will contain the migration fix as well.

PR #1781, opened 2025-11-04
Greatly reduce table padding

BentiGorlich:
| Before | After |
| ------ | ----- |
| grafik | Bildschirmfoto vom 2025-11-04 12-01-39  | 
| Bildschirmfoto vom 2025-11-04 12-02-48 | Bildschirmfoto vom 2025-11-04 12-03-12 |

### DIFF
--- a/assets/controllers/subject_controller.js
+++ b/assets/controllers/subject_controller.js
@@ -37,22 +37,29 @@ export default class extends Controller {
         if (!this.element.classList.contains('isSingle')) {
             this.element.querySelector('.content')?.addEventListener('click', (e) => {
                 if (e.defaultPrevented) {
-                    return
+                    return;
                 }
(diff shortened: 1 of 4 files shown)


melroy89:
Actually I just notice.. this change also impacts the user facing webpages; like magazine listing eg: https://kbin.melroy.org/magazines

Which is no longer mobile friendly anymore (eg. for touch screen use), causing accessibility issues. 

@BentiGorlich For the admin panels I don't care about the spacing, in fact I also prefer less space. But for the pages facing to the user, some padding is required for accessibility reasons. 

Could we create a dedicated CSS class that will be used in the admin panels, separate from the user pages maybe?

image

BentiGorlich:
I prefer it there as well. If you check the mobile view of this page it looks very different. For desktop I also prefer this to be honest. Mobil looks like this:
grafik

PR #1784, opened 2025-11-04
add some margin around the expand button

blued-gear:
This affects the expand button for comments and posts, but not entry bodies.
Please verify that the marging on all placed doesn't look bad.


melroy89:
Thanks for the PR! For UI changes, often a "before" and "after" screenshot might be desirable to easily see what the difference is.

blued-gear:
before

after

PR #1786, opened 2025-11-04
Table td padding

melroy89:
- Create just a little bit more "~~breeding~~ breathing room" in the table without claustrophobia :)
  - Especially because user facing pages are also effected
- A very small increase in height to _only_ 0.5rem, but still **far from** the 1.5 rem what is used to be...
- Vertical padding kept the same at 0.25rem (like its now in the `main`)

| Before | After |
| ------ | ----- |
| image | image  | 

Previous PR: https://github.com/MbinOrg/mbin/pull/1781

### DIFF
--- a/assets/styles/layout/_layout.scss
+++ b/assets/styles/layout/_layout.scss
@@ -167,7 +167,7 @@ table th a{
 
 table th,
 table td {
-  padding: .25rem;
+  padding: 0.5rem 0.25rem;
   position: relative;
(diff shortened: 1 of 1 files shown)


melroy89:
> Just FYI `breeding room` sounds like a sex cave :D I think you meant `breathing room` right? :D

Uh lol.. Yes I meant breathing room. Although, I do not limit the use of Mbin for anybody 😆

PR #1788, opened 2025-11-04
Crossposts

blued-gear:
This PR adds a new menu item for Entries to create crosspost by opening the `new_entry` form with prefilled values.

The maintainers should check if this works with federated content (especially images) as my local instance is not federated.

### DIFF
--- a/README.md
+++ b/README.md
@@ -175,6 +175,13 @@ Unofficial magazines:
                 </a>
             </td>
             <td align="center">
+                <a href="https://github.com/blued-gear">
+                    <img src="https://avatars.githubusercontent.com/u/164888202?v=4" width="100;" alt="blued-gear"/>
+                    <br />
+                    <sub><b>blued_gear</b></sub>
+                </a>
+            </td>
+            <td align="center">
                 <a href="https://github.com/GauthierPLM">
                     <img src="https://avatars.githubusercontent.com/u/2579741?v=4" width="100;" alt="GauthierPLM"/>
                     <br />
@@ -186,13 +193,6 @@ Unofficial magazines:
                     <img src="https://avatars.githubusercontent.com/u/40727284?v=4" width="100;" alt="Ahrotahn"/>
                     <br />
                     <sub><b>Ahrotahn</b></sub>
-                </a>
-            </td>
-            <td align="center">
-                <a href="https://github.com/blued-gear">
-                    <img src="https://avatars.githubusercontent.com/u/164888202?v=4" width="100;" alt="blued-gear"/>
-                    <br />
-                    <sub><b>blued_gear</b></sub>
                 </a>
             </td>
             <td align="center">
--- a/assets/controllers/collapsable_controller.js
+++ b/assets/controllers/collapsable_controller.js
@@ -1,0 +1,89 @@
+import { Controller } from '@hotwired/stimulus';
+import debounce from '../utils/debounce';
+
+// use some buffer-space so that the expand-button won't be included if just a couple of lines would be hidden
+const MAX_COLLAPSED_HEIGHT_REM = 25;
+const MAX_FULL_HEIGHT_REM = 28;
+
+/* stimulusFetch: 'lazy' */
+export default class extends Controller {
+
+    static targets = ['content', 'button'];
+
+    maxCollapsedHeightPx = 0;
+    maxFullHeightPx = 0;
+
+    isActive = false;
+    isExpanded = true;
+    button = null;
+    buttonIcon = null;
+
+    connect() {
+        const remConvert = parseFloat(getComputedStyle(document.documentElement).fontSize);
+        this.maxCollapsedHeightPx = MAX_COLLAPSED_HEIGHT_REM * remConvert;
+        this.maxFullHeightPx = MAX_FULL_HEIGHT_REM * remConvert;
+
+        this.setup();
+
+        const observerDebounced = debounce(200, () => {
+            this.setup();
+        });
+        const observer = new ResizeObserver(observerDebounced);
+        observer.observe(this.contentTarget);
+    }
+
+    setup() {
+        const activate = this.checkSize();
+        if (activate === this.isActive) {
+            return;
+        }
+
+        if (activate) {
+            this.setupButton();
+            this.setExpanded(false, true);
+        } else {
+            this.contentTarget.style.maxHeight = null;
+            this.button.remove();
+        }
+
+        this.isActive = activate;
+    }
+
+    checkSize() {
+        const elem = this.contentTarget;
+        return elem.scrollHeight - 30 > this.maxFullHeightPx || elem.scrollWidth > elem.clientWidth;
+    }
+
+    setupButton() {
+        this.buttonIcon = document.createElement('i');
+        this.buttonIcon.classList.add('fa-solid', 'fa-angles-down');
+
+        this.button = document.createElement('div');
+        this.button.classList.add('more');
+        this.button.appendChild(this.buttonIcon);
+
+        this.button.addEventListener('click', () => {
+            this.setExpanded(!this.isExpanded, false);
+        });
+
+        this.buttonTarget.appendChild(this.button);
+    }
+
+    setExpanded(expanded, skipEffects) {
+        if (expanded) {
+            this.contentTarget.style.maxHeight = null;
+            this.buttonIcon.classList.remove('fa-angles-down');
+            this.buttonIcon.classList.add('fa-angles-up');
+        } else {
+            this.contentTarget.style.maxHeight = `${MAX_COLLAPSED_HEIGHT_REM}rem`;
+            this.buttonIcon.classList.remove('fa-angles-up');
+            this.buttonIcon.classList.add('fa-angles-down');
+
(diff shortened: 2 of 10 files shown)


blued-gear:
img

blued-gear:
> It sounds pretty unintuitive to me to fill out the form by JS and not in the backend.
> 
> You should be able to prefill values by setting the fields in the `EntryDto` and then create the form like in the `EntryCreateController`.

The reason I did it this way is then we then would have an additional feature: other websites can provide "share" links with preset values for their users. As an example a blog has a "Share on Mbin" button, which then asks the visitor for their instance domain and redirect them with fitting *prefill* parameters.\
Then I build the crossposts upon this mechanic to prevent two logic paths for the same result.

BentiGorlich:
That sounds like a great idea, but it would also be doable in php. We have the policy of doing everything we can on the server side so people having JS disabled still get the best experience using mbin we can provide. 

You can easily get query parameters via `#[MapQueryParameter] ?string $title` etc. in the function declaration.

blued-gear:
> We have the policy of doing everything we can on the server side

ahh; I have quite the opposite mantra so that the server is mostly just a REST API. But for non-JS users this really makes sense.

I will rewrite the approach.\
Only concern I have there is that long bodies might  hit the limit of URL length.

src/Controller/CrosspostController.php
```diff
@@ -0,0 +1,83 @@
...
+            ['magazine_name' => $entry->magazine->name, 'entry_id' => $entry->getId()],
+            UrlGeneratorInterface::ABSOLUTE_URL
+        );
+        $body = 'Crossposted from ['.$entryUrl.']('.$entryUrl.')';
+        if (null !== $entry->body && '' !== $entry->body) {
+            $body = $body."\n\n".$entry->body;
+        }
+        $query = $query.'&body='.urlencode($body);
+
+        return $this->redirect('/new_entry?'.$query);
```
BentiGorlich: You should use the `redirectToRoute` method, or ,if that is not applicable because of the query string, generate the route string with a `UrlGeneratorInterface`.

Maybe it works this way ([article](https://yarnaudov.com/symfony-redirecttoroute-with-parameters.html)):
> Symfony redirectToRoute with query parameters?
> 
> If you want to use redirectToRoute with query parameters, you can to pass them in the second argument like the example above, Symfony will detect that they are not part of the route and will convert them to query parameters.

src/Controller/Entry/EntryCreateController.php
```diff
@@ -46,11 +47,41 @@ public function __construct(
...
+        #[MapQueryParameter]
+        ?array $tags,
+        Request $request,
+    ): Response {
         $dto = new EntryDto();
         $dto->magazine = $magazine;
+        $dto->title = $title;
+        $dto->url = $url;
+        $dto->body = $body;
+        $dto->imageUrl = $imageUrl;
```
BentiGorlich: Do we need an `$imageId` here so we can crossposting an image posts works?

templates/entry/_form_entry.html.twig
```diff
@@ -18,7 +18,10 @@
         {% endset %}
         {{ form_label(form.url, label, {'label_html': true}) }}
         {{ form_errors(form.url) }}
-        {{ form_widget(form.url, {attr: {'data-action': 'entry-link-create#fetchLink', 'data-entry-link-create-target': 'url'}}) }}
+        {{ form_widget(form.url, {attr: {
+            'data-action': 'entry-link-create#fetchLink',
+            'data-entry-link-create-target': 'url'
```
BentiGorlich: is this still needed?

src/Controller/CrosspostController.php
```diff
@@ -0,0 +1,83 @@
...
+
+        $entryUrl = $this->urlGenerator->generate(
+            'ap_entry',
+            ['magazine_name' => $entry->magazine->name, 'entry_id' => $entry->getId()],
+            UrlGeneratorInterface::ABSOLUTE_URL
+        );
+        $body = 'Crossposted from ['.$entryUrl.']('.$entryUrl.')';
+        if (null !== $entry->body && '' !== $entry->body) {
+            $body = $body."\n\n".$entry->body;
+        }
```
BentiGorlich, line 66: Actually each line of the original body should be prefixed with a `> ` so it is a quote

src/Controller/CrosspostController.php
```diff
@@ -0,0 +1,83 @@
...
+
+        if (null !== $entry->image) {
+            $imgUrl = $this->getImageUrl($entry->image);
+            if (null !== $imgUrl) {
+                if (null !== $entry->url && '' !== $entry->url) {
+                    $query = $query.'&imageUrl='.urlencode($imgUrl);
+                } else {
+                    $query = $query.'&url='.urlencode($imgUrl);
+                }
+            }
```
BentiGorlich: I am torn about this, as this is definitely the easiest way and it does work. However it kinda feels wrong implementing it this way. If you think that it is fine this way, we'll keep it, but it bothers me a little 😅

BentiGorlich (changes requested):

src/Controller/CrosspostController.php
blued-gear: Should a crosspost be seen as a citation? If I remember right I never saw quote style on Lemmy crossposts.

BentiGorlich: I thought lemmy did it this way :thinking:

blued-gear: Oh, then I misremembered.

src/Controller/Entry/EntryCreateController.php
```diff
@@ -40,17 +43,55 @@ public function __construct(
...
+        $dto->url = $url;
+        $dto->body = $body;
+        $dto->imageAlt = $imageAlt;
+        $dto->isAdult = '1' === $isNsfw;
+        $dto->isOc = '1' === $isOc;
+        $dto->tags = $tags;
+
+        if (null !== $imageHash) {
+            $img = $this->imageRepository->findOneBySha256(hex2bin($imageHash));
+            if (null !== $img) {
```
blued-gear, line 95: can `null === $img` be silently ignored or should I throw an exception?

BentiGorlich: Hmm tough decision :thinking: ... I'd say a user should know when something didn't work

tests/Functional/Controller/Entry/EntryCreateControllerTest.php
BentiGorlich, line 1: Can you add a test with an image hash?

translations/messages.en.yaml
```diff
@@ -812,6 +812,7 @@ position_top: Top
 pending: Pending
 flash_thread_new_error: Thread could not be created. Something went wrong.
 flash_thread_tag_banned_error: Thread could not be created. The content is not 
+flash_thread_ref_image_not_found: The image referenced by 'imageHash' could not be found.
   allowed.
```
BentiGorlich: You inserted it between lines. The full error message at the moment will be this: "The image referenced by \'imageHash\' could not be found. allowed." which does not make the test fail as it only checks for "...TextContains"

BentiGorlich (approved):
Pretty nice work 👍 
I want to point out that there is not a test for crossposting an entry with an image attached (there is one for prefilling the form by image hash). But I am fine without it.

melroy89:
My audit check is also failing in the other PR.. 

Ow we have an issue:

```md
Run composer audit
Found 1 security vulnerability advisory affecting 1 package:
+-------------------+----------------------------------------------------------------------------------+
| Package           | symfony/http-foundation                                                          |
| Severity          |                                                                                  |
| CVE               | CVE-2025-64500                                                                   |
| Title             | CVE-2025-64500: Incorrect parsing of PATH_INFO can lead to limited authorization |
|                   | bypass                                                                           |
| URL               | https://symfony.com/blog/cve-2025-64500-incorrect-parsing-of-path-info-can-lead- |
|                   | to-limited-authorization-bypass                                                  |
| Affected versions | >=2.0.0,<3.0.0|>=3.0.0,<4.0.0|>=4.0.0,<5.0.0|>=5.0.0,<5.1.0|>=5.1.0,<5.2.0|>=5.2 |
|                   | .0,<5.3.0|>=5.3.0,<5.4.0|>=5.4.0,<5.4.50|>=6.0.0,<6.1.0|>=6.1.0,<6.2.0|>=6.2.0,< |
|                   | 6.3.0|>=6.3.0,<6.4.0|>=6.4.0,<6.4.29|>=7.0.0,<7.1.0|>=7.1.0,<7.2.0|>=7.2.0,<7.3. |
|                   | 0|>=7.3.0,<7.3.7                                                                 |
| Reported at       | 2025-11-12T11:09:14+00:00                                                        |
+-------------------+----------------------------------------------------------------------------------+
```

EDIT: And `composer update "symfony/*"` didn't solve it ;(

melroy89:
> EDIT: And `composer update "symfony/*"` didn't solve it ;(

OK it needed a version upgrade to v7.3.x, see separate PR: https://github.com/MbinOrg/mbin/pull/1818/

PR #1790, opened 2025-11-04
Fix frontend linting and run it in CI

BentiGorlich:
- run eslint in CI
- adjust the rules a bit to better fit our code
- fix all lint problems

### DIFF
--- a/assets/controllers/rich_textarea_controller.js
+++ b/assets/controllers/rich_textarea_controller.js
@@ -1,5 +1,5 @@
 import { Controller } from '@hotwired/stimulus';
-import {fetch} from "../utils/http";
+import { fetch } from '../utils/http';
 
 /* stimulusFetch: 'lazy' */
 export default class extends Controller {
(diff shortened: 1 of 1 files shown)


.github/workflows/action.yaml
```diff
@@ -211,6 +211,48 @@ jobs:
...
+
+      - name: Composer install
+        run: composer install --no-scripts --no-progress
+
+      - run: cp .env.example .env
+
+      - name: NPM install
+        run: npm ci --include=dev
+        env:
+          NODE_ENV: production
```
melroy89: is there a reason to set the `NODE_ENV` to production? While you do actually want to install dev dependencies as well..?

Since without setting NODE_ENV to production, `npm ci` will actually install both production and development dependencies.

BentiGorlich: well eslint is still working, so it is not a problem 🤔

melroy89: I mean to say, you can remove line 250 and 251 fully. And the run command can just be `npm ci`

melroy89 (approved):

PR #1795, opened 2025-11-05
Fix flaky test

BentiGorlich:


melroy89 (approved):
A sleep is far from ideal.
But okay it is what it is for now.

BentiGorlich:
It is only one second and that is due to the datetime field not having milliseconds... I think it is the only working thing and not really a problem

melroy89:
True, and on 20 minutes total test duration you won't notice the difference anyways 😅🤣..

BentiGorlich:
Well tbh a one second delay is never noticable :sweat_smile: 

I gotta say the 20 minutes are sometimes annoying but I have to live with a 90 minute pipeline at work so it is a farcry from terrible :D

melroy89:
> but I have to live with a 90 minute pipeline at work so it is a farcry from terrible :D

Sorry to hear that. My work sometimes also used to have 4+ hours integration tests with test benches running in VMs. Just the boot up time plus driver init will take 20 minutes. Assuming no failures during the driver init part 😅.
Then again, these are the examples I rather not follow haha. 

> Well tbh a one second delay is never noticable 😅

Yes it is. With good unit tests running fully in memory and mocked correctly. Those can run within ms.

PR #1797, opened 2025-11-06
Use amqproxy in Docker setup

jwr1:
- Add amqproxy to the Docker setup in order to reduce the number of RabbitMQ connections down to 1.
- Change RabbitMQ healthcheck to ensure the port is accepting connections.
- Set amqproxy to wait till RabbitMQ is healthy, that way PHP won't log a ton of errors trying to connect to RabbitMQ when it's not even up yet. This does have the side effect of the whole stack taking a little longer to start up, but I think it's a good practice.

A note will need to be added in the next release to update your `.env` file. In `MESSENGER_TRANSPORT_DSN`, the host needs to be changed from `rabbitmq:5672` to `amqproxy:5673`.

### DIFF
--- a/.github/workflows/action.yaml
+++ b/.github/workflows/action.yaml
@@ -211,6 +211,46 @@ jobs:
       - name: PHP CS Fixer to PR Annotations
         run: tools/vendor/bin/php-cs-fixer fix --dry-run --format=checkstyle --show-progress=none | cs2pr
 
+  frontend-lint:
+    runs-on: ubuntu-latest
+    container:
+      image: ghcr.io/mbinorg/mbin-pipeline-image:latest
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Get Composer Cache Directory
+        id: composer-cache
+        run: |
+          echo "dir=$(composer config cache-files-dir)" >> $GITHUB_OUTPUT
+
+      - name: Get NPM cache directory path
+        id: npm-cache-dir-path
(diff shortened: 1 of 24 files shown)


compose.dev.yaml
```diff
@@ -55,6 +55,12 @@ services:
             # Change to 1 to enable federation
             replicas: 0
```
melroy89, line 56: do admins now need to change two values in order to enable federation?  both here and on line 62?

Or is it not redundant?

melroy89:
Do we want to add something about AMQProxy in the docker docs? https://docs.joinmbin.org/admin/installation/docker

compose.dev.yaml
jwr1: Actually 3 values, but yes. One for `messenger`, one for `amqproxy`, and one for `rabbitmq`. All it's doing is changing the number of containers for each service from 0 to 1 if federation is desired.

jwr1:
> Do we want to add something about AMQProxy in the docker docs? [docs.joinmbin.org/admin/installation/docker](https://docs.joinmbin.org/admin/installation/docker)

I'm not really sure it's needed. Literally the only references to RabbitMQ are the management UI page, setting the password, and creating the storage directories, none of which are needed for AMQProxy.

I do think it would be a good idea to have general information about what AMQProxy is, what it's doing, and why it's useful, but I think that would be a better fit for the FAQ page. Maybe you could include that when you update the bare metal install page?

PR #1800, opened 2025-11-07
Don't match microblogs by magazine name hashtag anymore

BentiGorlich:
- previously we matched microblog post by their used hashtags and searched for magazines that define that hashtag or had a matching name. Change that to only use the defined hashtags from a magazine
- matching of magazine using the handle name will still work 
- to keep the current behaviour, add a migration that will add the magazine name to the tags array for all local magazines

New create form:
grafik

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,12 +3,12 @@
 {
     "name": "PHP",
     // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-    "image": "mcr.microsoft.com/devcontainers/php:2.0.1-8.4-trixie",
+    "image": "mcr.microsoft.com/devcontainers/php:3.0.3-8.4-trixie",
 
     // Features to add to the dev container. More info: https://containers.dev/features.
     "features": {
-        "ghcr.io/opencodeco/devcontainers/install-php-extensions:0": {
-            "extensions": "amqp apcu bcmath exif gd intl opcache pcntl pdo_pgsql pgsql redis xsl"
+        "ghcr.io/yassinedoghri/devcontainers/php-extensions-installer:1": {
+            "extensions": "amqp apcu bcmath exif gd intl opcache pcntl pdo_pgsql pgsql redis"
         },
         "ghcr.io/devcontainers/features/github-cli:1": {},
         "ghcr.io/devcontainers/features/node:1": {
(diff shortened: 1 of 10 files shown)


melroy89:
Uhm I don't now.. This means all microblogs will be empty by default under all the magazines..?

Is my assume true?

blued-gear:
> Uhm I don't now.. This means all microblogs will be empty by default under all the magazines..?

If that is true, then I'm a bit against it. \
The magazine-name tag could be added to newly created magazines automatically and show up it the settings (so that it is opt-out).

jwr1:
> The magazine-name tag could be added to newly created magazines automatically and show up it the settings (so that it is opt-out).

I think that's a great idea!

BentiGorlich:
> This means all microblogs will be empty by default under all the magazines

If I am correctly understanding the question, then no that that is not correct. As far as I know putting microblogs in magazines by their hashtag is not used widely (at least not intentionally). The way it is done usually in microblogs is by mentioning the magazine (which we also support). Making it opt out would mean that it will stay active in most cases as people are just not aware of it (I think). I am not entirely against it, I just think that it may miss the point... If all of youbare for it I have no problem doing it

jwr1:
> Making it opt out would mean that it will stay active in most cases as people are just not aware of it (I think).

That's a good point, I'm not sure how many people are actually aware of it.

blued-gear:
> That's a good point, I'm not sure how many people are actually aware of it.

Then why not try to make them more aware? Like mentioning it it the docs or even better: add a checkbox in the *create magazine* form to en- or disable the adding of the magazine-name-hashtag at creation.

melroy89:
> > This means all microblogs will be empty by default under all the magazines
> 
> If I am correctly understanding the question, then no that that is not correct. As far as I know putting microblogs in magazines by their hashtag is not used widely (at least not intentionally). The way it is done usually in microblogs is by mentioning the magazine (which we also support). Making it opt out would mean that it will stay active in most cases as people are just not aware of it (I think). I am not entirely against it, I just think that it may miss the point... If all of youbare for it I have no problem doing it

OK clear. I updated the title and description a bit to make this more clear.

In that case I'm not fully against the change. But like other said, we could make it opt out per magazine or even maybe per instance.

melroy89:
Since we "merged" the threads together with the microblog, here is an example where my `updates` magazine gets pollute by a random post:

image

BentiGorlich:
Yeah I think I'll add a migration that adds the magazine name as a tag in all local magazines. Then it is opt out and we control it.
And also add a checkbox on magazine creation like mention here :+1:

blued-gear (approved):

PR #1803, opened 2025-11-07
Improve collapse button of entry / post / comment bodies

blued-gear:
This PR moves the functionality into its own controller.
It also adds some margin to the button (even more on touch devices).

Additionally the subject_controller functions were cleaned up a little bit.\
By the nature of this JS codebase I could not 100% verify that referenced functions got disrupted by this change, **so please have a close look at this**.

Do you think the margins are okay?


blued-gear:
moved to #1810

PR #1809, opened 2025-11-09
Add paths-ignore for pull requests to action.yaml

melroy89:
- Add `paths-ignore` to the `action.yaml` on paths that shouldn't need to trigger our workflow
  - Running or not running the workflow will not make any difference to these file changes.
- Save server resources, time and save the planet 🌎  my adding some paths ignores (can always be extended or reduced in the future).

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -139,6 +139,7 @@ DATABASE_URL="postgresql://${POSTGRES_USER}:${POSTGRES_PASSWORD}@127.0.0.1:5432/
 ###> symfony/messenger ###
 # Choose one of the transports below
 RABBITMQ_PASSWORD=!ChangeThisRabbitPass!
+# Change port 5672 to 5673 if you want to use AMQProxy (recommended)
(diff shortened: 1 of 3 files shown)


.github/workflows/action.yaml
```diff
@@ -1,6 +1,18 @@
...
+      - '.github/ISSUE_TEMPLATE/**'
+      - '.github/PULL_REQUEST_TEMPLATE/**'
+      # A separate workflow is used to build and publish the pipeline image
+      - 'ci/**'
+      # Documentation changes are not relevant to the workflow and are taking care of by mbin-docs repository
+      - 'docs/**'
+      # Other files are not relevant to the workflow
+      - 'LICENSES/**'
+      - '*.md'
+      - 'public/robots.txt'
```
blued-gear, line 17: I think `.devcontainer` can also be excluded since nothing is build there.

melroy89: True

jwr1 (approved):

PR #1810, opened 2025-11-09
Fix/more margin for comment expand btn

blued-gear:
This PR moves the functionality into its own controller.
It also adds some margin to the button (even more on touch devices).

Additionally the subject_controller functions were cleaned up a little bit.
By the nature of this JS codebase I could not 100% verify that referenced functions got disrupted by this change, so please have a close look at this.

Do you think the margins are okay?

### DIFF
--- a/src/Controller/Api/Notification/NotificationBaseApi.php
+++ b/src/Controller/Api/Notification/NotificationBaseApi.php
@@ -139,7 +139,7 @@ class NotificationBaseApi extends BaseApi
             case 'new_signup':
                 /** @var NewSignupNotification $n */
                 $n = $dto;
-                $toReturn['subject'] = $this->userFactory->createDto($n->getSubject()->getId());
+                $toReturn['subject'] = $this->userFactory->createDto($n->getSubject());
                 break;
         }
 


melroy89:
I did check the code in a real-world example(s). The expended state seems to work much better now indeed.

However, now the issue is when the text is _still collapsed_, look:

image

and here:

image

And the text seems to hide behind the `margin-top` on `.subject .more` but the `margin` on `.more` CSS classes.

This is a bit weird to me since it should not be `position: absolute`.. but here we are??


**EDIT:** I tested it within the Developer Tools, but it seems that both `margin` and `padding` is causing this issue that text will disappear behind the button..

blued-gear:
I think some parts of the text will always vanish behind the button (depending on the clients text size). Techniaclly it is not hiiden *by the button* but *cut of* as the div of the conent has a limited height. And if this height is not a multiple of a lines height, then text will be cut of.

Using a line-height depending calculation is difficult (at least with my CSS skills...). But what do you think if the bottom of a collapsed content box has a slight blur effect?

melroy89:
So the line break can happen indeed anyway in the text.. or could we play with maybe a value of: `max-height: 27rem`..  (`MAX_COLLAPSED_HEIGHT` of 27rem)?

I dunno why.. but that seems to work out better now, with the combination of font family, font size  and line height we are having..?

EDIT: I also like the blurry effect indeed.

melroy89:
> EDIT: I also like the blurry effect indeed.

That being said, we do not always know if the user has GPU acceleration on their side, slowing the website down for them..

melroy89:
> max-height: 27rem

both 27rem but also `max-height: 24rem` seems to works better.. Granted, this is only a very small number of use-cases I test it on.

melroy89:
I actually just found a bigger issue now, in the case that there is no expand button is shown, yet the content disappears at the bottom?? This used to be no problem...

image

Source: https://kbin.melroy.org/m/opensource@lemmy.ml/t/1305645/My-thoughts-on-GPL-vs-permissive-licenses

If we should change to `max-height: 27rem` in CSS and  the problem is gone.. but I don't know if this is a real solution.

blued-gear:
> Using a line-height depending calculation is difficult (at least with my CSS skills...)

Ok... I should have put a minute into research. There is the [`lh`](https://developer.mozilla.org/en-US/docs/Web/CSS/Reference/Values/length#lh) unit, which [should be broadly usable](https://caniuse.com/mdn-css_types_length_lh).

blued-gear:
> Source: https://kbin.melroy.org/m/opensource@lemmy.ml/t/1305645/My-thoughts-on-GPL-vs-permissive-licenses

This is not visible without login.\
Could you send the text you used?

melroy89:
> > Source: https://kbin.melroy.org/m/opensource@lemmy.ml/t/1305645/My-thoughts-on-GPL-vs-permissive-licenses
> 
> This is not visible without login. Could you send the text you used?

In markdown the content is:

```markdown
I'm a very much pro free software person and I used to think that GPL is basically the only possible option when it comes to benefits for free software (and not commercial use), but I've recently realised this question is actually much more ambiguous.

I think there are two sides to this issue:
- GPL forces all contributions to stay open-source which prevents commercialisation* of FOSS projects, but also causes possible interference of corporate software design philosophy and all kinds of commercial decisions, if contributions come from companies.
- MIT-like permissive licenses, on the other hand, easily allow for making proprietary forks, which, however, separates commercial work from the rest of the project, therefore making the project more likely to stay free both of corporate influence and in general.

So it boils down to the fact, that in my opinion what makes free software free is not only the way it's distributed but also the whole philosophy behind it: centralisation vs. decentralisation, passive consumer vs. co-developper role of the user etc. And this is where things start to be a bit controversial.

What do you think?

*UPD: wrong word. I mean close-sourcing and turning into a profitable product instead of something that fulfils your needs
```

In HTML:

```html
                            <p>I'm a very much pro free software person and I used to think that GPL is basically the only possible option when it comes to benefits for free software (and not commercial use), but I've recently realised this question is actually much more ambiguous.</p>
<p>I think there are two sides to this issue:</p>
<ul>
<li>GPL forces all contributions to stay open-source which prevents commercialisation* of FOSS projects, but also causes possible interference of corporate software design philosophy and all kinds of commercial decisions, if contributions come from companies.</li>
<li>MIT-like permissive licenses, on the other hand, easily allow for making proprietary forks, which, however, separates commercial work from the rest of the project, therefore making the project more likely to stay free both of corporate influence and in general.</li>
</ul>
<p>So it boils down to the fact, that in my opinion what makes free software free is not only the way it's distributed but also the whole philosophy behind it: centralisation vs. decentralisation, passive consumer vs. co-developper role of the user etc. And this is where things start to be a bit controversial.</p>
<p>What do you think?</p>
<p>*UPD: wrong word. I mean close-sourcing and turning into a profitable product instead of something that fulfils your needs</p>
```

melroy89 (dismissed):

melroy89 (approved):
I believe now with the additional max full height/collapsed height plus ResizeObserver (and debounce) everything works as expected now.

An the other previous fixes (mentioned via PM and already fixed in this PR as well). I approve it now.

assets/utils/debounce.js
```diff
@@ -0,0 +1,7 @@
+export default function debounce(delay, handler) {
+    let timer = 0;
```
BentiGorlich, line 2: Isn't this variable always 0. Why have it if I am correct in assuming this?

assets/controllers/collapsable_controller.js
```diff
@@ -0,0 +1,89 @@
...
+        this.maxCollapsedHeightPx = MAX_COLLAPSED_HEIGHT_REM * remConvert;
+        this.maxFullHeightPx = MAX_FULL_HEIGHT_REM * remConvert;
+
+        this.setup();
+
+        const observerDebounced = debounce(200, () => {
+            this.setup();
+        });
+        const observer = new ResizeObserver(observerDebounced);
+        observer.observe(this.contentTarget);
```
BentiGorlich, line 32: Could you explain why the timeout is necessary? In most cases it is an indicator that something else is wrong...

assets/utils/debounce.js
melroy89: This code is actually based on https://stackoverflow.com/questions/69469814/how-can-i-throttle-a-resizeobserver

timer is set on line 5. Its used as a debounce. Hence the name `debounce` :D

assets/controllers/collapsable_controller.js
melroy89: Its on purpose, I asked mf to add it, its called a `debounce`, so instead of using [`ResizeObserver`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver) directly every millisecond after a change event, we added a debounce on purpose.. So it will be more performant. 

So its not really a time-out in a traditional sense. Its like this: https://www.freecodecamp.org/news/javascript-debounce-example/. But in our case we use it in combination with `ResizeObserver` of course. It will increase the performance.

assets/utils/debounce.js
BentiGorlich: I would have thought, since this timer is in a local scope, that the next time this method is called it will again be `0`, but I am obviously missing some magic here. But ok, understood the purpose :+1:

melroy89: Magic of async behavior in js 🪄

blued-gear: it is set at `timer = setTimeout(handler, delay);` as it is captured in the lambda.

assets/controllers/collapsable_controller.js
blued-gear: (edit: I only saw the new comments on your comment after I wrote mine.)

It is not a classic timeout, but a *debounce*. So quick calls by the observer (for example when the browser window is resized by dragging an edge) get buffered so that only the last call is executed. This saves compute resources.

assets/utils/debounce.js
blued-gear: > that the next time this method is called it will again be 0

not magic :) `debounce` returns a function which you call to debounce the passed handler. So it is like a factory which only needs to be called once per handler.

BentiGorlich: Ahhhh ok .... it returns a lambda, that explains it 👍

PR #1811, opened 2025-11-10
Remove entry component from edit thread + fix top-margin

melroy89:
- Removed entry component from edit thread page, seems like an old-school copy/paste error (it also caused JS errors)
- Add top margin via `section--top` class
- Fix integration test
- Solve conflict within the `_section.scss` file trying to enforce a `margin-top: 0 !important;`, which doesn't allow us to set an actual top-margin on `.section--top`, which are defined in [_layout.scss ](https://github.com/MbinOrg/mbin/blob/main/assets/styles/layout/_layout.scss#L201) but was never getting applied correctly ....

The latter fix with margin-top also solves issues on other pages like the microblog post will now have finally the correct top margin getting applied (which is an long standing issue):

| Before | After |
|--------|--------|
| image | image | 
| image | image | 

But the main change of course is:

| Before | After |
|--------|--------|
| image | image | 

Basically similar to other edit pages like the magazine edit page:

image

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,17 +161,17 @@ Unofficial magazines:
 		</tr>
 		<tr>
             <td align="center">
-                <a href="https://github.com/garrettw">
-                    <img src="https://avatars.githubusercontent.com/u/84885?v=4" width="100;" alt="garrettw"/>
-                    <br />
-                    <sub><b>Garrett W.</b></sub>
-                </a>
-            </td>
-            <td align="center">
                 <a href="https://github.com/blued-gear">
                     <img src="https://avatars.githubusercontent.com/u/164888202?v=4" width="100;" alt="blued-gear"/>
                     <br />
                     <sub><b>blued_gear</b></sub>
+                </a>
+            </td>
+            <td align="center">
+                <a href="https://github.com/garrettw">
+                    <img src="https://avatars.githubusercontent.com/u/84885?v=4" width="100;" alt="garrettw"/>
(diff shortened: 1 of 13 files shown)


melroy89:
(yes I cancelled the workflow on purpose.. I will need to rebase again again.. so I the update rebase will trigger again a new workflow run..)

assets/styles/layout/_section.scss
```diff
@@ -37,7 +37,6 @@
 
   &--top {
     border-top: 0 !important;
-    margin-top: 0 !important;
```
BentiGorlich, line 40: `section--top` is used in a lot of places... Did you look at every one of them?

BentiGorlich: I think I aggree with you on this one, I was just unsure :)

melroy89: I will run kbin.melroy.org on this branch now.

melroy89: but to answer your question, yes section--top is indeed used a lot.. And currently not working as intending.. This PR fixes all that margin top issues as well:

image

image


image

melroy89: I will now switch my branch on kbin.melroy.org again to `upgrade_symfony_7_3`

melroy89: and now I move back again*

PR #1822, opened 2025-11-12
No-dev and --abandoned flag on composer audit

melroy89:
Using the `--no-dev` and `--abandoned` parameters on the `composer audit` command: https://getcomposer.org/doc/03-cli.md#audit

Making the audit similar to the `npm audit`


melroy89:
~~ok great, another thing we first need to update, use a new composer binary, see PR: https://github.com/MbinOrg/mbin/pull/1823~~ merged

PR #1827, opened 2025-11-14
Make comment not nullable in EntryCommentViewController

melroy89:
Multiple instances are sufferings 500 internal server errors suddenly..? 

Seen on kbin.earth according to @jwr1 but @blued-gear  also mentioned fedia.io (keep in mind, fedia.io didn't update recently at all.. Jerry is still running on the latest stable release). 

**Problem:** `$comment->root` is used, without checking if `$comment` is not null, `$comment` can also be null.

**Solution:** Make `?EntryComment $comment` not nullable anymore

```json
{
    "message": "Warning: Attempt to read property \"root\" on null",
    "context": {
        "exception": {
            "class": "ErrorException",
            "message": "Warning: Attempt to read property \"root\" on null",
            "code": 0,
            "file": "/app/src/Controller/Entry/Comment/EntryCommentViewController.php:52"
        }
    },
    "level": 400,
    "level_name": "ERROR",
    "channel": "php",
    "datetime": "2025-11-13T21:23:00.305241+00:00",
    "extra": {}
}
```

Code line that caused issue was introduced in PR: https://github.com/MbinOrg/mbin/pull/1563

### DIFF
--- a/config/routes/dev/framework.yaml
+++ b/config/routes/dev/framework.yaml
@@ -1,3 +1,3 @@
 _errors:
-    resource: "@FrameworkBundle/Resources/config/routing/errors.xml"
+    resource: '@FrameworkBundle/Resources/config/routing/errors.php'
     prefix: /_error
--- a/config/routes/framework.yaml
+++ b/config/routes/framework.yaml
@@ -1,3 +1,3 @@
 _errors:
-  resource: '@FrameworkBundle/Resources/config/routing/errors.php'
-  prefix: /_error
+    resource: '@FrameworkBundle/Resources/config/routing/errors.php'
+    prefix: /_error


src/Controller/Entry/Comment/EntryCommentViewController.php
```diff
@@ -49,7 +49,8 @@ public function __invoke(
         $this->dispatcher->dispatch(new EntryHasBeenSeenEvent($entry));
 
         $this->entryRepository->hydrate($entry);
-        if (null !== $comment->root) {
+        // Comment can be null
+        if (null !== $comment && null !== $comment->root) {
```
blued-gear: Does it make sense to just continue if comment is `null`? The render function consumes comment so null might cause other problems. I think a 404 makes more sense in this case.

melroy89: Maybe. Why is the comment optional to begin with `?EntryComment $comment,` in the constructor.. Maybe @BentiGorlich  knows this.?

EDIT: I notice something similar (although that variable is called `$parent`) in another class https://github.com/MbinOrg/mbin/blob/main/src/Controller/Entry/Comment/EntryCommentResponseTrait.php#L27

BentiGorlich: I have no idea why $comment is allowed to be null.... I would just remove the nullability and then it would throw the 404 automatically I think 🤔

melroy89: I hope so.. I did made the change.

melroy89:
It seems the default comment_id to be set to `null` as well here: https://github.com/MbinOrg/mbin/blob/75f19ecd5ebb49bb5f0fcd88c162f06153cf9d49/config/mbin_routes/entry.yaml#L12

melroy89:
> It seems the default comment_id to be set to `null` as well here:
> 
> https://github.com/MbinOrg/mbin/blob/75f19ecd5ebb49bb5f0fcd88c162f06153cf9d49/config/mbin_routes/entry.yaml#L12

I also removed that default now.

src/Controller/Entry/Comment/EntryCommentViewController.php
melroy89: This approach does work:

image


However, we no longer see the magazine anymore. So that is the reason the `$comment` was nullable. To allow to display the magazine and just show an error message of "comment not found" below the magazine. 

Hence.. I do not think we want this, I think we do want to show the magazine in this case..?

melroy89:
follow up https://github.com/MbinOrg/mbin/pull/1829

PR #1828, opened 2025-11-14
update swagger-php and use new attribute for encoding of file upload

blued-gear:


melroy89 (approved):

melroy89:
We just may or may not want to wait for a response from the developer: https://github.com/zircote/swagger-php/issues/1841

melroy89:
> We just may or may not want to wait for a response from the developer: [zircote/swagger-php#1841](https://github.com/zircote/swagger-php/issues/1841)

Developer replied. I don't think he will roll back. But maybe fix the breaking change.

@blued-gear either way, we can just continue. And merge.

PR #1829, opened 2025-11-14
Fix Attempt to read property "root" on null

melroy89:
Fixing:

```json
{
    "message": "Warning: Attempt to read property \"root\" on null",
    "context": {
        "exception": {
            "class": "ErrorException",
            "message": "Warning: Attempt to read property \"root\" on null",
            "code": 0,
            "file": "/app/src/Controller/Entry/Comment/EntryCommentViewController.php:52"
        }
    },
    "level": 400,
    "level_name": "ERROR",
    "channel": "php",
    "datetime": "2025-11-13T21:23:00.305241+00:00",
    "extra": {}
}
```

Keep the rest of the current existing code as is.
I like the error message shown while still seeing the magazine, this code works fine. We don't need to clean up the existing **working** code: https://github.com/search?q=repo%3AMbinOrg%2Fmbin%20%22Comment%20not%20found%22&type=code.

Code line that caused issue was introduced in PR: https://github.com/MbinOrg/mbin/pull/1563

### DIFF
--- a/README.md
+++ b/README.md
@@ -161,17 +161,17 @@ Unofficial magazines:
 		</tr>
 		<tr>
             <td align="center">
-                <a href="https://github.com/garrettw">
-                    <img src="https://avatars.githubusercontent.com/u/84885?v=4" width="100;" alt="garrettw"/>
-                    <br />
-                    <sub><b>Garrett W.</b></sub>
-                </a>
-            </td>
-            <td align="center">
                 <a href="https://github.com/blued-gear">
                     <img src="https://avatars.githubusercontent.com/u/164888202?v=4" width="100;" alt="blued-gear"/>
                     <br />
                     <sub><b>blued_gear</b></sub>
+                </a>
+            </td>
+            <td align="center">
+                <a href="https://github.com/garrettw">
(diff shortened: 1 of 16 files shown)


src/Controller/Entry/Comment/EntryCommentViewController.php
```diff
@@ -49,7 +49,8 @@ public function __invoke(
         $this->dispatcher->dispatch(new EntryHasBeenSeenEvent($entry));
 
         $this->entryRepository->hydrate($entry);
-        if (null !== $comment->root) {
+        // Both comment and root comment can be null
+        if (null !== $comment && null !== $comment->root) {
```
BentiGorlich: If you want Ou can make it "cleaner" by checking `null !== $comment?->root`, but thats totally optional

BentiGorlich (approved):

src/Controller/Entry/Comment/EntryCommentViewController.php
melroy89:
```suggestion
        if (null !== $comment?->root) {
```

PR #1835, opened 2025-11-15
Improve Anubis docs

melroy89:
Fix spelling mistakes / typos, markdown layout fixes,rephrase sentences and other readability improvements to the Anubis documentation page.

I went through all the PRs for creating the release notes, but came across this documentation page. Hence it would be nice to get this also merged before the new release.

### DIFF
--- a/docs/02-admin/03-optional-features/07-anubis.md
+++ b/docs/02-admin/03-optional-features/07-anubis.md
@@ -2,9 +2,9 @@
 
 ### Why?
 
-Anubis is a bot programm trying to block bots by presenting proof-of-work challengers to the user. A normal browser will just solve them (provided JavaScript is enabled on it) whilst AI scrapers will most likely just accept the challenge as the response.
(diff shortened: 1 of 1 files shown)


docs/02-admin/03-optional-features/07-anubis.md
```diff
@@ -116,30 +118,34 @@ thresholds:
       report_as: 4
 ```
 
-The default config includes a few snippets that require a subscription. To avoid warn messages you should comment out everything that "Requires a subscription to Thoth to use" (just search for it in the file).
+The default config includes a few snippets that requires a subscription. To avoid warn messages you should comment-out everything that "Requires a subscription to Thoth to use" (just search for it in the file).
+
+For Anubis to be able to access the socket, that we will use later, we will have to change the service file (`/usr/lib/systemd/system/anubis@.service`) and set the user anubis is being executed by to `www-data`:
```
melroy89: `www-data` user or `www-data` group? Or both? Please explain.

melroy89: I just assume both www-data user and www-data group.

jwr1 (approved):

PR #1837, opened 2025-11-16
Missing options and optional username argument

melroy89:
- `<username>` is actually optional in case of `-a`
- Explain all the options for this command

### DIFF
--- a/docs/02-admin/03-optional-features/07-anubis.md
+++ b/docs/02-admin/03-optional-features/07-anubis.md
@@ -1,27 +1,29 @@
-# Anubis setup for mbin
+# Anubis setup for Mbin
 
 ### Why?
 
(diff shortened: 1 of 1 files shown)


docs/02-admin/04-running-mbin/05-cli.md
```diff
@@ -136,11 +136,15 @@ If private keys have been leaked you should rotate the private keys to avoid the
 Usage:
 
 ```bash
-php bin/console mbin:user:private-keys:rotate [-a|--all-local-users] <username>
+php bin/console mbin:user:private-keys:rotate [-a|--all-local-users] [-r|--revert] [<username>]
 ```
 
 Arguments:
-- `username`: the single user for which this command should be executed
+- `username`: the single user for which this command should be executed (optional)
```
blued-gear: Instead of "optional" I think you could write "not required if `--all-local-users` is used".

melroy89: done.

PR #1838, opened 2025-11-16
Use repository-dispatch to dispatch update-docs event to mbin-docs repo

melroy89:
Instead using a cronjob, I will leverage the https://github.com/peter-evans/repository-dispatch action to trigger a repository dispatch event from this repo to the mbin-docs repository.

Secret token is already in place, in order to have get this working: https://github.com/MbinOrg/mbin/settings/secrets/actions

See also PR in the documentation repo: https://github.com/MbinOrg/mbin-docs/pull/16

### DIFF
--- a/config/packages/security.yaml
+++ b/config/packages/security.yaml
@@ -121,7 +121,7 @@ security:
         - { path: ^/2fa, role: IS_AUTHENTICATED_2FA_IN_PROGRESS }
 
         - { path: ^/admin, roles: [ROLE_ADMIN, ROLE_MODERATOR] }
-        - { path: ^/authorize, roles: PUBLIC_ACCESS }
+        - { path: ^/authorize, roles: IS_AUTHENTICATED_REMEMBERED }
         - { path: ^/token, roles: PUBLIC_ACCESS }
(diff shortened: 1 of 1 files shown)


.github/workflows/action.yaml
```diff
@@ -300,3 +300,12 @@ jobs:
           push: ${{ github.event_name != 'pull_request' }}
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
+
+      # We will also use this job to dispatch an event to the mbin-docs repository to trigger documentation build and publish
+      - name: Trigger mbin-docs workflow dispatch
+        if: github.event_name != 'pull_request'
```
jwr1, line 306: Is there a way to check that the push is only to the main branch? Since this step will currently run if the push is to `main` or `dev/new_features`.

melroy89: Yes via `if: github.ref == 'refs/heads/main'`

But then it's also not triggered anymore on new release tags.

Well the mbin-docs repo just only looks at this main branch anyways. So triggering maybe one time too often isn't the problem..?

It's better then the current cron job.

BentiGorlich:
nice

PR #1850, opened 2025-11-18
Improve ban notifications

BentiGorlich:
- rewrite the notification texts
- add a new notification for being unbanned
- use the correct method which dispatches a `NotificationCreatedEvent`

grafik


src/Entity/MagazineUnBanNotification.php
```diff
@@ -0,0 +1,46 @@
...
+    }
+
+    public function getType(): string
+    {
+        return 'magazine_unban_notification';
+    }
+
+    public function getMessage(TranslatorInterface $trans, string $locale, UrlGeneratorInterface $urlGenerator): PushNotification
+    {
+        $message = $trans->trans('you_are_no_longer_banned_from_magazine', ['%m' => $this->ban->magazine->name]);
```
blued-gear: You use `$locale` in line 44 but not here. Is this correct?

templates/notifications/_blocks.html.twig
```diff
@@ -97,12 +97,30 @@
...
-    {{ component('user_inline', {user: notification.ban.bannedBy, showNewIcon: true}) }} {{ 'banned'|trans|lower }}. {{ 'ban_expired'|trans }}
-    {%- if notification.ban.expiredAt is not same as null -%}
-        {{ component('date', {date: notification.ban.expiredAt}) }}
-    {%- endif -%}. {{ notification.ban.reason }}
+    {% if notification.ban.expiredAt is not same as null -%}
+        {{ 'you_have_been_banned_from_magazine'|trans({'%m': component('magazine_inline', {'magazine': notification.ban.magazine})})|raw }}
+    {% else -%}
+        {{ 'you_have_been_banned_from_magazine_permanently'|trans({'%m': component('magazine_inline', {'magazine': notification.ban.magazine})})|raw }}
+    {% endif -%}
+    {% if notification.ban.expiredAt is not same as null -%}
```
blued-gear: why not merge this part with the first `if notification.ban.expiredAt is not same as null`?

src/Entity/MagazineUnBanNotification.php
BentiGorlich: No it is probably not, good catch 👍

blued-gear: I see you updated to use the locale in `MagazineBanNotification`, but not here. Did you miss it?

BentiGorlich: Yes I did miss it. Updated it now

templates/notifications/_blocks.html.twig
melroy89: @BentiGorlich if you have resolved it my advice would be to close the conversation (see button below: "Resolve conversation"). Which makes it easier for yourself, but also other reviewers to not spend time on this comment, since its resolved. :)

BentiGorlich: I go by the principle that I thumbs up when I implemented it. The reviewer can then resolve the conversation if they think it is actually done.

src/Entity/MagazineUnBanNotification.php
blued-gear: > No it is probably not, good catch 👍

Line `$trans->trans('you_are_no_longer_banned_from_magazine', [...]` still does not use `$locale`, or am I missing something? \
Same for MagazineBanNotification.php every `$trans->trans()` call.

BentiGorlich: I did indeed miss one, but only one. Maybe you're looking at an old commit?

blued-gear: I used the 'Files changed' button in GH, so I don't think. But with your latest commit every call got its locale, so it is fine now.

PR #1856, opened 2025-11-18
2nd attempt at fixing the notifications from other users

BentiGorlich:
- the `name` property of a mention does not always contain the full handle as thought previously, but can also only contain `@name`. Adjust the test case accordingly
- In the `MarkdownConverter`: don't trust the `name` property. Instead, search for the `href` which is the profile URL via the `ActivityPubManager` 

Previous PR #1756 
Previous Issue: #538

### DIFF
--- a/src/Entity/MagazineUnBanNotification.php
+++ b/src/Entity/MagazineUnBanNotification.php
@@ -12,7 +12,7 @@ use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 #[Entity]
-class MagazineBanNotification extends Notification
+class MagazineUnBanNotification extends Notification
 {
     #[ManyToOne(targetEntity: MagazineBan::class)]
     #[JoinColumn(nullable: true, onDelete: 'CASCADE')]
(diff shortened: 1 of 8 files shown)


BentiGorlich:
The first commit expects a test failure. I will push the second commit which fixes this after the first pipeline finished

tests/Functional/ActivityPub/Inbox/CreateHandlerTest.php
melroy89, line 177: Is it maybe a good idea to introduce a new test for local users notification as well?

BentiGorlich: What exactly do you mean?

melroy89: I know jerry was complaining about getting notifications from another remote user with his username right? I thought maybe its a good idea to also write a test (good and bad weather) where or not the notification is triggered correctly?

BentiGorlich: I added a test for basically that. A remote user that is mentioned like a local user still looks like a remote user afterwards, so this cannot happen anymore

PR #1862, opened 2025-11-18
Adapt the AP presentation of an entry with a URL and an image

BentiGorlich:
- Create a test for an entry with both an image and a URL
- adapt the way our code handles this

### DIFF
--- a/src/Entity/MagazineUnBanNotification.php
+++ b/src/Entity/MagazineUnBanNotification.php
@@ -12,7 +12,7 @@ use Symfony\Component\Routing\Generator\UrlGeneratorInterface;
 use Symfony\Contracts\Translation\TranslatorInterface;
 
 #[Entity]
-class MagazineBanNotification extends Notification
+class MagazineUnBanNotification extends Notification
 {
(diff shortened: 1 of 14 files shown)


melroy89 (commented):
Let me test this..

melroy89:
> Let me test this..

My thread no longer federate anymore ;( .. but comments do. 

image

After waiting for 10 minutes the thread is still not here:

https://fedia.io/m/testing@kbin.melroy.org/all/newest

I also tested a comment, and that comment was posted within 2 seconds: https://fedia.io/m/testing@kbin.melroy.org/t/2990125/Testing-thread-in-my-own-test-magazine/comment/13113373#entry-comment-13113373 so federation works, except for threads now

melroy89:
I can confirm its this PR that is indeed brake federation.. You could test it yourself by using a thread with an URL. And then also attach your own image.

It won't federate anymore.. But sure to run `./bin/post-upgrade` and restart PHP-fpm. And ALSO.. very importantly restart the messenger jobs `sudo supervisorctl restart all`. After which federation is no longer working on threads with URLs + images.

melroy89 (approved):

PR #1863, opened 2025-11-19
Improve Performance in the `ContentRepository`

BentiGorlich:
- Change `UNION` to `UNION ALL` -> Difference on gehirneimer: 2700ms -> 700ms
- Don't query for empty id arrays in the `ContentPopulationTransformer`
- Don't hydrate favourites and votes in the `PostRepository`, `EntryRepository`, `PostCommentRepository` and the `EntryCommentRepository`


melroy89:
> * Change `UNION` to `UNION ALL`

Union will remove duplicates. Union all will basically do the opposite and will actually include all duplicates.

BentiGorlich:
> > * Change `UNION` to `UNION ALL`
> 
> Union will remove duplicates. Union all will basically do the opposite and will actually include all duplicates.

That is perfectly fine as it is just impossible to get duplicates when unioning the entry and post tables :D  
I just didn't know that `UNION` is basically the same as `UNION DISTINCT`

melroy89 (approved):

melroy89:
I took some rough measurements (with this PR included).. Each page I tested twice.

`/` homepage (~850ms):

image

image


And `/threads` (~750ms):

image

image

All these where performed with empty pool cache and redis cache flushed (each time less then 1 second before I refreshed the page). And browser cache is also disabled. 

Conclusion: On my server I have around 100ms difference.

PR #1864, opened 2025-11-19
Use the `NativeQueryAdapter` instead of fetching everything

BentiGorlich:
- the user profile was quite expensive to call if a user had posted a lot. Fix that by using the existing `NativeQueryAdapter`
- same for the trashed content view in magazines and in the reputation tab in a users profile

This should fix the OOM exception on fedia.io


BentiGorlich:
Should first try to confirm that it is indeed fixing the fedia issue? @melroy89 @Jerry-infosecexchange

PR #1366, opened 2025-01-13
Add twig component rendering to our markdown rendering

BentiGorlich:
- new components for rendering inline entries, posts and their comments
- use components for inline users and magazines
- add new appearance setting for displaying the domains of users and magazines
- add tags to clear the cache when a URL is used within markdown, but not yet known to the server

| Before | After |
| ------ | ----- |
| grafik | grafik |

PR #1866, opened 2025-11-19
Fix Undefined array key "path"

melroy89:
Fix the following issue (you can only see it in `dev mode`):

Before:

image

After:

image

Introduced in: https://github.com/MbinOrg/mbin/pull/1366

### DIFF
--- a/src/Pagination/Transformation/ContentPopulationTransformer.php
+++ b/src/Pagination/Transformation/ContentPopulationTransformer.php
@@ -6,62 +6,61 @@ namespace App\Pagination\Transformation;
 
 use App\Entity\Entry;
 use App\Entity\EntryComment;
+use App\Entity\Magazine;
 use App\Entity\Post;
 use App\Entity\PostComment;
-use App\Repository\EntryCommentRepository;
-use App\Repository\EntryRepository;
-use App\Repository\MagazineRepository;
-use App\Repository\PostCommentRepository;
-use App\Repository\PostRepository;
-use App\Repository\UserRepository;
+use App\Entity\User;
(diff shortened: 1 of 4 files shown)


src/Repository/ApActivityRepository.php
```diff
@@ -90,7 +90,7 @@ public function findLocalByApId(string $apId): ?array
             return null;
         }
 
-        if ($parsed['host'] === $this->settingsManager->get('KBIN_DOMAIN') && null !== $parsed['path'] && '' !== $parsed['path']) {
+        if ($parsed['host'] === $this->settingsManager->get('KBIN_DOMAIN') && isset($parsed['path']) && '' !== $parsed['path']) {
```
blued-gear: maybe even [`empty()`](https://www.php.net/manual/en/function.empty.php) can be used here.
(`''` is equal to `false` in PHP, isn't it?)

melroy89: sort of.. but no. In php check on `''` is a loose equality.. so like '', 0 and: `'' == false` is true xD and even '' == null is also true.

While `empty()` is indeed true for '', null, false, 0 '0' and [] and undefined

blued-gear: > A variable is considered empty if it does not exist [...]

So it should be able to replace the `isset()` too

melroy89: We use both now.. `isset()` and `!empty()` we can not be more safe than that.

melroy89: ok yes.. `!empty()` should be sufficient.

BentiGorlich (approved):

PR #1870, opened 2025-11-20
Improve English default view

melroy89:
- Default content doesn't sound correct. I believe it should be called the default view.
- Also the default value, just call it what it is: "Default view (Threads)"

Currently looks like this:

image
After:

image

### DIFF
--- a/translations/messages.en.yaml
+++ b/translations/messages.en.yaml
@@ -1091,7 +1091,7 @@ federation_page_use_allowlist_help: If an allow list is used, this instance will
   only federate with the explicitly allowed instances. Otherwise this instance 
   will federate with every instance, except those that are banned.
 front_default_content: Front default view
-default_content_default: Default view (Threads)
+default_content_default: Server default (Threads)
 default_content_all: Threads + Microblog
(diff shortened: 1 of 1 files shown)


jwr1 (approved):
It looks good to me

translations/messages.en.yaml
```diff
@@ -1090,8 +1090,8 @@ allow_instance: Allow instance
 federation_page_use_allowlist_help: If an allow list is used, this instance will
   only federate with the explicitly allowed instances. Otherwise this instance 
   will federate with every instance, except those that are banned.
-front_default_content: Front default content
-default_content_default: The default, currently Threads
+front_default_content: Front default view
+default_content_default: Default view (Threads)
```
blued-gear: This would read as "Front default view: Default view [...]" which I find odd. I propose 'Server default (Threads)'.

melroy89: well its looks like this, since its a dropdown:


image

melroy89: Anyhow good suggestion. I applied it.

PR #1876, opened 2025-11-23
Increase shared memory size postgresql container

melroy89:
- Increase the `shm_size` to 2gb of the `postgresql` service (the default is only 64MB)
- Added to docs
- Slight increase in `shm_size` in the test docker setup yaml file as well

Fixes fatal errors like (in Docker setups):

> Disk full: 7 ERROR:  could not resize shared memory segment \"/PostgreSQL.1006932470\" to 16777216 bytes: No space left on device

### DIFF
--- a/docs/02-admin/01-installation/02-docker.md
+++ b/docs/02-admin/01-installation/02-docker.md
@@ -208,7 +208,7 @@ If you run a larger Mbin instance, its recommended to increase the `shm_size` va
 
 The following step is **optional** and also depends on how much RAM you have left as well as how many parallel workers, table sizes, expected concurrent users and more. You can first use the default `2gb`, which should be sufficient, however below is explained how to further increase this number.
 
-1. In `compose.override.yaml`, add `shm_size` to the `postgres` service (`4gb` is an example here):
+In `compose.override.yaml`, add `shm_size` to the `postgres` service (`4gb` is an example here):
 
 ```yaml
 services:


docs/02-admin/01-installation/02-docker.md
```diff
@@ -202,6 +202,20 @@ In this example, port `8080` will connect to your Mbin server.
 > [!TIP]
 > In order to verify your server is correctly detecting it's public protocol (HTTP vs HTTPS), visit `/.well-known/nodeinfo` and look at which protocol is being used in the `href` fields. A public server should always be using HTTPS and not contain port numbers (i.e., `https://DOMAINHERE/`).
 
+### Additional configuration (Optional)
+
+If you run a larger Mbin instance, its recommended to increase the `shm_size` value of the `postgres` service (first try the default `2gb`!). Although you can also decrease the number if you wish on smaller instances. `shm_size` sets the size of the shared memory (`/dev/shm`) and used for dynamic memory allocation. PostgreSQL is using this for buffering the write-ahead logs (also known as "WALL buffer").
+
+The following step is **optional** and also depends on how much RAM you have left as well as how many parallel workers, table sizes, expected concurrent users and more. You can first use the default `2gb`, which should be sufficient, however below is explained how to further increase this number.
+
+1. In `compose.override.yaml`, add `shm_size` to the `postgres` service (`4gb` is an example here):
```
jwr1:
```suggestion
In `compose.override.yaml`, add `shm_size` to the `postgres` service (`4gb` is an example here):
```

If there's not more than one "step", then it doesn't make much sense to use a list.

melroy89: Yes.. for now..... 🥹

Issue #1868, opened 2025-11-19
Performance: "Random threads" query


melroy89:
**Describe the bug**

In the sidebar we have something called "Random threads" that only show 4 threads: 

<img height="300" alt="Image" src="https://github.com/user-attachments/assets/1c678345-0bad-4e2a-847c-9368c731e340" />

And the random threads (also known as the `related_entries` twig template) is using the following method used called `findLast`: 

https://github.com/MbinOrg/mbin/blob/7aa330f00487104bcf222357a0dd2b7336026532/src/Twig/Components/RelatedEntriesComponent.php#L71

Which is defined here: 

https://github.com/MbinOrg/mbin/blob/58910f41f11cb969479692965182efbd8796a2a6/src/Repository/EntryRepository.php#L412

This results in the following query (example):

```sql
[...]

PR #1878, opened 2025-11-26
Add 2 settings for the sidebar sections + give performance warning about the random section + fix max_image_bytes

melroy89:
**Background story:**

After a very long time of debugging, we came to the conclusion that using the query part: `magazine.ap_id IS NULL` (in WHERE) is very problematic when using together in a JOIN query together with other tables. That is because PostgreSQL no longer leverage the `ap_id` index on that. Causing a scan index across 160k+ rows on my instance and possibly more rows on large instances.

**Done in this PR:**

1. This is only a problem with our "Random threads" / "Random posts" queries. Luckily the default is `false`. However, we should really inform the admins about the performance impact when this option is enabled.
2. I added a link in the comments to the issue for future reference and better traceability...
3. I create 2 different settings now (so the setting is spitted into two options): `MBIN_SIDEBAR_SECTIONS_RANDOM_LOCAL_ONLY` and `MBIN_SIDEBAR_SECTIONS_USERS_LOCAL_ONLY`
  This allows admins to enable `MBIN_SIDEBAR_SECTIONS_USERS_LOCAL_ONLY` without any performance impact. 
4. Updated the .env files (example + docker)
  - Add the missing options like the `MBIN_SIDEBAR_SECTIONS*` that was previously fully missing to the .env files
  - And rearranged all the variables, so all our own settings are at the top (this was the intent from day 1)
6. Add the missing `MBIN_` prefix to our custom `MAX_IMAGES_BYTES` option as well (defaults to `6000000` bytes). This was missed when introducing this option.
7. Updated `.devcontainer` (incl. a major upgrade of postgresql from v13 to v18).

We have multiple breaking changes, but no worries 1.9.0 has many breaking change. I will add this all to the release notes. 

See related issue: https://github.com/MbinOrg/mbin/issues/1868

The PR that introduced the generic local only setting `MBIN_SIDEBAR_SECTIONS_LOCAL_ONLY`: https://github.com/MbinOrg/mbin/pull/435

PR that wrongly introduced the `MAX_IMAGE_BYTES`: https://github.com/MbinOrg/mbin/pull/987

---

Although we can't easily solve the SQL query, we at least can give the admin a warning when this setting is enabled and saved, which looks like this now:

image


src/Service/SettingsManager.php
```diff
@@ -98,11 +98,12 @@ public function __construct(
                 $this->find($results, 'MBIN_SSO_ONLY_MODE', FILTER_VALIDATE_BOOLEAN) ?? $this->mbinSsoOnlyMode,
                 $this->find($results, 'MBIN_PRIVATE_INSTANCE', FILTER_VALIDATE_BOOLEAN) ?? false,
                 $this->find($results, 'KBIN_FEDERATED_SEARCH_ONLY_LOGGEDIN', FILTER_VALIDATE_BOOLEAN) ?? true,
-                $this->find($results, 'MBIN_SIDEBAR_SECTIONS_LOCAL_ONLY', FILTER_VALIDATE_BOOLEAN) ?? false,
+                $this->find($results, 'MBIN_SIDEBAR_SECTIONS_RANDOM_LOCAL_ONLY', FILTER_VALIDATE_BOOLEAN) ?? false,
+                $this->find($results, 'MBIN_SIDEBAR_SECTIONS_USERS_LOCAL_ONLY', FILTER_VALIDATE_BOOLEAN) ?? false,
```
melroy89, line 102: We could make the default `true` here (for `MBIN_SIDEBAR_SECTIONS_USERS_LOCAL_ONLY`). That would more embrace our own local community / users. Without impacting performance.

.env.example
```diff
@@ -31,29 +31,37 @@ KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
...
 KBIN_FEDERATION_PAGE_ENABLED=true
 MBIN_DEFAULT_THEME=default
-
-# If you are running Mbin behind a reverse proxy, uncomment the line below and adjust the proxy address/range below
-# to your server's IP address if it does not already fall within the private IP spaces specified.
-#TRUSTED_PROXIES=::1,127.0.0.1,10.0.0.0/8,172.16.0.0/12,192.168.0.0/16
-TRUSTED_PROXIES=
-
-# Max image filesize (in bytes)
+# Set the max image file size (in bytes)
```
blued-gear, line 34: Please sync these changes also to `.devcontainer/.env.devcontainer`

src/Service/SettingsManager.php
```diff
@@ -252,7 +253,7 @@ public function getLocale(): string
 
     public function getMaxImageByteString(): string
     {
-        $megaBytes = round($this->maxImageBytes / 1024 / 1024, 2);
+        $megaBytes = round($this->mbinMaxImageBytes / 1024 / 1024, 2);
 
         return $megaBytes.'MB';
```
blued-gear, line 258: as this uses power of 2, it should be `MiB`

translations/messages.en.yaml
```diff
@@ -1096,3 +1094,6 @@ default_content_all: Threads + Microblog
 default_content_threads: Threads
 default_content_microblog: Microblog
 combined: Combined
+sidebar_sections_random_local_only: Restrict "Random X" sidebar sections to local only
```
blued-gear: maybe "Random threads / posts" would fit better

blued-gear (changes requested):

.env.example
melroy89: Thanks. I will!

src/Service/SettingsManager.php
melroy89: It does return a value in MB see line below. I didn't change the algorithm. The calculations works fine.

Ow you mean it's mib instead of mb? It could be that the consumer of this string value only support "MB". I shouldn't change it. It's working good enough I think.

translations/messages.en.yaml
melroy89: haha yea. before people think its Twitter / X

.env.example
melroy89: I also updated several versions within the dev container it was really lagging behind as well..

src/Service/SettingsManager.php
melroy89: Or not.. to keep it consistent.

src/Service/SettingsManager.php
melroy89: it seems to mainly affect user view. so the in the frontend.. but still. people generally have no clue what MiB is :P

melroy89: let me change the calculation instead so it reflect MB instead of MiB

blued-gear (approved):

PR #1879, opened 2025-11-26
Add user admin panel to more then just the overview.html page

melroy89:
Currently the admin panel for user management is only added to the "User Overview" page.

However, we have many more pages/ HTML Twig templates like threads, comments, replies, etc. It would be nice to keep showing this admin panel on the right side in those cases.

1. I moved the twig code to a dedicated local include template.
2. Reuse and include this template on multiple user pages now (incl overview, but also the rest).

Before:

image

After:

image

(screenshot is just only showing the "comments" page.. but you get my point)


blued-gear:
Did you forget "Boosts"?

melroy89:
> Did you forget "Boosts"?

Nope. Boost is also using overview.html.

blued-gear (approved):

PR #1883, opened 2025-11-27
Fix broken combined (homepage) + Fix API endpoints

melroy89:
Currently the combined path is conflicting with the already established [Homepage option](https://github.com/MbinOrg/mbin/blob/main/src/Controller/Entry/EntryFrontController.php#L236) (also called 'all').

This is also why the 'all' path is not working on the homepage, nor the menu item will become active, see: 

image

After:

image

---

This PR actually resolves three things.

1. I renamed the combined path just `/combined`. The same as we already have for `microblog`, `threads`, `people` and alike. And renamed the enum and translation etc. etc.
2. Make the API names in the `content_api.yaml` file unique (it was not showing up in the swagger and was conflicting with the post APIs). And... `api/posts`  is the wrong path name.
3. I also renamed this content API to combined API.
4. I also needed to remove the tags to avoid repeating the endpoints too often in the swagger docs.. And now it shows up:

image

Introduced in PR: https://github.com/MbinOrg/mbin/pull/1696


config/mbin_routes/content_api.yaml
```diff
@@ -1,13 +1,13 @@
-api_posts_subscribed:
-  controller: App\Controller\Api\Content\ContentRetrieveApi::userCollection
-  path: /api/posts/{contentType}
-  requirements:
-    contentType: sub|fav|mod
+api_content_collection:
+  controller: App\Controller\Api\Content\ContentRetrieveApi::collection
+  path: /api/content
```
melroy89: Everything is now called `combined`, like our menu button and the path. Even in Interstellar they call it combined. 

Maybe the API should also be called `combined` instead of `content`?

jwr1: Sure, that sounds good

melroy89: OK let me change that then, before you go start using it 😊

blued-gear: +1 for calling it `combined`.

But as other paths (like the entry API) use names like `favourited` and `moderated` for the filter-subpaths, I think we should do it here too (instead of the short versions).

melroy89: I agree with you as well here. I will fix it now.

melroy89: Fixed now: 
image

blued-gear:
You renamed the message `default_content_combined`, but I can't find in the PR where the usages of this message are renamed.

melroy89:
> You renamed the message `default_content_combined`, but I can't find in the PR where the usages of this message are renamed.

It's a magic combination of two string dynamically. Try to search for `default_content_`

blued-gear (approved):

BentiGorlich:
This PR breaks saving the user settings if a user wants to have the combined view as their default. The psql types do not include `combined` as a value.
@melroy89

melroy89:
> This PR breaks saving the user settings if a user wants to have the combined view as their default. The psql types do not include `combined` as a value. @melroy89

See PR: https://github.com/MbinOrg/mbin/pull/1891

PR #1885, opened 2025-11-28
Image preview for URL threads

blued-gear:
- Threads with URLs will get a small "link" icon.
- Threads with URLs to images will have the image as a preview
- Threads with URLs to images but also an image attached will show the attached image

example

### DIFF
--- a/migrations/Version20251129140919.php
+++ b/migrations/Version20251129140919.php
@@ -1,0 +1,26 @@
+<?php
+
+declare(strict_types=1);
+
+namespace DoctrineMigrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+
+final class Version20251129140919 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Rename value all to combined in front_default_content enum.';
+    }
+
+    public function up(Schema $schema): void
+    {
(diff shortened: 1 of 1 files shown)


blued-gear:
@melroy89 I use a *blur* effect for url images. I hope this is not too bad for performance.

blued-gear:
I also updated the Devcontainer config and docs for the updated Postgres

melroy89:
> @melroy89 I use a _blur_ effect for url images. I hope this is not too bad for performance.

I see. I have no clue how bad this impact performance on devices that are not GPU accelerated. I can test it on some mobile devices (Android), which are actually GPU accelerated. But more than that I can't test myself.

.devcontainer/.env.devcontainer
```diff
@@ -136,7 +136,7 @@ APP_SECRET=!ChangeSecret!
 POSTGRES_HOST=127.0.0.1:5432
 POSTGRES_DB=postgres
 POSTGRES_USER=postgres
-POSTGRES_PASSWORD=
```
melroy89, line 139: thanks for fixing.

blued-gear:
> I have no clue how bad this impact performance on devices that are not GPU accelerated

The alternative would be, that the background is just transparent. But this would look different from normal image-threads.

melroy89 (dismissed):

melroy89:
I see broken image thumbnails sometimes like here: 
image

image

This is because you are trying to include mp4 as an image source?

image

PR #1888, opened 2025-11-29
Harmonize posts and threads in combined view

blued-gear:
## Posts on the *Microblog* page:

- move vote buttons to the left
- image in expandable ares

example-1

example-2

## Posts on *Combined* page:

- similar to threads

example

### Current problems:

- [X] no title -> no link to click to open the post

### DIFF
--- a/templates/components/post_combined.html.twig
+++ b/templates/components/post_combined.html.twig
@@ -25,21 +25,30 @@
             id="post-{{ post.id }}"
             data-controller="subject preview mentions html-refresh"
             data-action="notifications:Notification@window->subject#notification">
-            <header>
-                {% if post.visibility in ['visible', 'private'] or (post.visibility is same as 'trashed' and this.canSeeTrashed) %}
-                    <h2>
-                        {% if post.isAdult %}<small class="badge danger">18+</small>{% endif %}
+            {% with %}
+                {% set hasTitle = post.visibility in ['visible', 'private'] or (post.visibility is same as 'trashed' and this.canSeeTrashed) %}
+                {% set isAdult = post.isAdult %}
+                {% set hasLang = post.lang is not same as app.request.locale and post.lang is not same as kbin_default_lang() %}
+                {% set isModDeleted = post.visibility is same as 'trashed' %}
+                {% set isUserDeleted = post.visibility is same as 'soft_deleted' %}
+                {% set needsHeader = (hasTitle and (isAdult or hasLang)) or isModDeleted or isUserDeleted %}
 
-                        {% if post.lang is not same as app.request.locale and post.lang is not same as kbin_default_lang() %}
-                            <small class="badge-lang kbin-bg">{{ post.lang|language_name }}</small>
-                        {% endif %}
-                    </h2>
-                {% elseif(post.visibility is same as 'trashed') %}
-                    <p class="text-muted">&lsqb;<i>{{ 'deleted_by_moderator'|trans }}</i>&rsqb;</p>
-                {% elseif(post.visibility is same as 'soft_deleted') %}
-                    <p class="text-muted">&lsqb;<i>{{ 'deleted_by_author'|trans }}</i>&rsqb;</p>
-                {% endif %}
-            </header>
+                <header {% if not needsHeader %} style="margin: 0 0 var(--kbin-entry-element-spacing) 0;" {% endif %}>
+                    {% if hasTitle %}
+                        <h2>
+                            {% if isAdult %}<small class="badge danger">18+</small>{% endif %}
+
+                            {% if hasLang %}
+                                <small class="badge-lang kbin-bg">{{ post.lang|language_name }}</small>
+                            {% endif %}
+                        </h2>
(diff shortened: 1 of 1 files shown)


melroy89:
Could we also please move the images to a thumbnail format as well? Especially when its listed in the overview?

blued-gear:
> Could we also please move the images to a thumbnail format as well?

I'm not sure about that. It wouldn't work well with a longer text block beside or below it.
To make it works posts would need to be similarly designed as threads, which has its own problems (like that posts have no title line). \
I like that posts are this distinguishable to threads, but that is just my personal opinion.

melroy89:
> > Could we also please move the images to a thumbnail format as well?
> 
> I'm not sure about that. It wouldn't work well with a longer text block beside or below it. To make it works posts would need to be similarly designed as threads, which has its own problems (like that posts have no title line). I like that posts are this distinguishable to threads, but that is just my personal opinion.

See also community feedback: https://fedia.io/m/mbinmeta@gehirneimer.de/t/3026365/-/comment/13104842

So as a user with the combined view, my feedback would be this:

image

Or this:

image


As a user this annoyed me a lot, since I'm just scrolling through some content and suddenly 80%-100% of my screen is filled by a **single** post. The Interstellar app does **not** have this issue.

melroy89:
Also the test is failing.. :)

melroy89:
> * no title -> no link to click to open the post

Both Mastodon and PieFed solves this by making the card clickable. Be aware that when there is a link **within** the card text, that URL should also still work if you click on that.

And you could argue to make that consistent when threads as well then, meaning both threads and posts will have a clickable content.. 

But again, just keep in mind that you can still have URLs within the text itself that has priority in terms of "mouse clicking" . So clicking on these links should still work, well just look how Mastodon solves it.

melroy89:
Its running now at kbin.melroy.org

melroy89:
Why does the post has a down vote button here?
image


But not here:

image


Or I should ask it in reverse; should we show a down-vote in the combined view for posts?

melroy89:
Also the text limit of the post is now much smaller then a thread.. (top is a thread, at the bottom of the image is a post)

image

I actually suspected the post to have a longer char limit before you cut off. right?

blued-gear:
> Or I should ask it in reverse; should we show a down-vote in the combined view for posts?

You're right, the down-vote button should not be there in the first place.

blued-gear:
> Also the text limit of the post is now much smaller then a thread.

This is odd as both use the same code for the body preview (`{{ get_short_sentence(post.body|markdown|raw, striptags = true) }}`)

melroy89:
> > Also the text limit of the post is now much smaller then a thread.
> 
> This is odd as both use the same code for the body preview (`{{ get_short_sentence(post.body|markdown|raw, striptags = true) }}`)

it seems that it stops right after the first sentence / paragraph. It looks like that people are often creating a huge paragraph in a thread.. but in posts people use multiple smaller paragraphs.. Anyhow, for posts we should use different function maybe that should not short the text after the first paragraph?

blued-gear:
> Anyhow, for posts we should use different function maybe that should not short the text after the first paragraph?

Should I create a special function case for posts in combined or would it be better to modify `get_short_sentence` in general to stop at a fixed char count (+- if a sentence-end is near) in general?

melroy89:
> > Anyhow, for posts we should use different function maybe that should not short the text after the first paragraph?
> 
> Should I create a special function case for posts in combined or would it be better to modify `get_short_sentence` in general to stop at a fixed char count (+- if a sentence-end is near) in general?

I personally believe we should treat a post differently, and do not stop at a new line, and also with a larger char limit etc.

blued-gear:
> I personally believe we should treat a post differently, and do not stop at a new line, and also with a larger char limit etc.

Done (except I reused the char-limit of threads so that both previews have a similar height).

We now even have tests for this shortening.

melroy89:
> one (except I reused the char-limit of threads so that both previews have a similar height).

Great! I still notice its removing new lines / markup language though. Is that something we want..? I mean if fine with either removing the new lines or showing them. If you compare it with like a Mastodon feed, I could imagine people would like to see the same markup text in the overview as if they would click on it.

blued-gear:
> I still notice its removing new lines / markup language though.

While not impossible, it would be way harder to calculate the resulting string length while keeping all the HTML tags needed for formatting.

I think for a preview the current approach is sufficient. Also perfect is the enemy of good; we can update the shortening function at a later time if needed.

BentiGorlich:
I know I am late to the party and for the most part I would like to stay out of this. I only noticed that on kbin.melroy.org posts do have an empty header which makes the spacing look weird.  
grafik

blued-gear:
> I only noticed that on kbin.melroy.org posts do have an empty header which makes the spacing look weird.

Yea, that is not optimal. But I don't really know what to put there, and I can't remove the row completely as it hosts the NSFW and language badge.

melroy89 (approved):

PR #1891, opened 2025-11-29
Add missing migration for front_default_content DB enum

melroy89:
I forgot to include the migration after PR: https://github.com/MbinOrg/mbin/pull/1883

We just need to rename `all` to `combined`.

Now you can change the setting again without errors.

### DIFF
--- a/migrations/Version20251129140919.php
+++ b/migrations/Version20251129140919.php
@@ -16,11 +16,11 @@ final class Version20251129140919 extends AbstractMigration
 
     public function up(Schema $schema): void
     {
-        $this->addSql('ALTER TYPE enumfrontcontentoptions RENAME VALUE \'all\' TO \'combined\'');
+        $this->addSql('ALTER TYPE enumFrontContentOptions RENAME VALUE \'all\' TO \'combined\'');
     }
 
     public function down(Schema $schema): void
     {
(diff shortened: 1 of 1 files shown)


melroy89:
Ow maybe `ALTER TYPE name RENAME VALUE existing_enum_value TO new_enum_value` is the correct one?

migrations/Version20251129140919.php
```diff
@@ -0,0 +1,26 @@
...
+final class Version20251129140919 extends AbstractMigration
+{
+    public function getDescription(): string
+    {
+        return 'Rename value all to combined in front_default_content enum.';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TYPE enumfrontcontentoptions RENAME VALUE \'all\' TO \'combined\'');
```
BentiGorlich: I know SQL is not case sensitive, but it should be this:

```suggestion
        $this->addSql('ALTER TYPE enumFrontContentOptions RENAME VALUE \'all\' TO \'combined\'');
```
Same in the `down` method

melroy89: I noticed that in postgresql is stored with all lower caps.. That is why I used it like that.

BentiGorlich (approved):

PR #1892, opened 2025-11-29
Improve performance of the main content query

BentiGorlich:
- don't join the domain anymore, only the id is needed and that is in the entry table (unless we are looking at a domain)
- only join the "user" table in an outer query. This avoids querying that table twice
- only have a `UNION ALL` in the inner query if we are actually querying both tables. This actually massively speeds up the threads and microblog views as this is using their indexes instead of sequentially scanning the tables (example [here](https://explain.dalibo.com/plan/c212d7b2g2023a3g))
- Use the criteria constants and introduce one for `combined`

| Before | After |
| ------ | ----- |
|  Execution time: 672ms, Planning time: 5,96ms, JIT: 223ms | Execution time: 500ms, Planning time: 5,95ms, JIT: 180ms |
| grafik | grafik |
| https://explain.dalibo.com/plan/7h7f8d366784aadd | https://explain.dalibo.com/plan/52b3a530a9c780a6 |

### DIFF
--- a/src/Service/ActivityPub/MarkdownConverter.php
+++ b/src/Service/ActivityPub/MarkdownConverter.php
@@ -34,12 +34,17 @@ class MarkdownConverter
             if ($this->mentionManager->extract($match[1])) {
                 $mentionFromTag = array_filter($apTags, fn ($tag) => 'Mention' === $tag['type'] ?? '' && $match[2] === $tag['href'] ?? '');
                 if (\count($mentionFromTag)) {
-                    $replace = $mentionFromTag[array_key_first($mentionFromTag)]['name'] ?? $match[1];
+                    $mentionFromTagObj = $mentionFromTag[array_key_first($mentionFromTag)];
+                    try {
+                        $mentioned = $this->activityPubManager->findActorOrCreate($mentionFromTagObj['href']);
+                    } catch (\Throwable) {
+                    }
+                    $replace = $mentioned?->username ?? $mentioned?->name ?? $mentionFromTagObj['name'] ?? $match[1];
                 } else {
                     try {
                         $actor = $this->activityPubManager->findActorOrCreate($match[2]);
                         $replace = '@'.($actor instanceof User ? $actor->username : $actor->name);
-                    } catch (\Throwable $e) {
+                    } catch (\Throwable) {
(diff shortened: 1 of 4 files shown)


BentiGorlich:
The main problem with the combined view is that postgresql is not using any indexes for that because the cost and row estimations are way off... I am researching why that is, but in the meantime this improves things quite a bit by itself (especially for the thread and microblog views)

melroy89:
> The main problem with the combined view is that postgresql is not using any indexes for that because the cost and row estimations are way off... I am researching why that is, but in the meantime this improves things quite a bit by itself (especially for the thread and microblog views)

You could try to delete slowly more and more from the query (like some WHERE sections here and there) and run it directly on the postgres database. And keep an eye on the runtime execution time.. eventually you will find which sub-query is causing the issue. Not the solution, but at least you know where to look for.

That is how I found: https://github.com/MbinOrg/mbin/issues/1868 (we went from 200ms down to just 2ms, like how it should be)

melroy89:
> * (example [here](https://explain.dalibo.com/plan/c212d7b2g2023a3g))

Query is missing in the upload? I don't know what this DB query was.

blued-gear:
The biggest issue is still the Sequential Scan. With bigger instances this query *will* get slower and slower. \
I experimented with a lot of different tweaks to the query, but I can't get Postgres to use a relevant index to produce presorted results. It will rather load almost the full table in memory and then sort it; WTF.

The only way I got Postgres to do it efficiently is to use ORDER BY and LIMIT inside the parts of the UNION. \
A problem might be that Pagerfanta manages the values of OFFSET and LIMIT. But our `NativeQueryAdapter` always uses the same SQL params for LIMIT and OFFSET, so we can use them in the subparts too (so a subpart ends with the following: `ORDER BY <required sorting> LIMIT (:offset + :limit)`).

BentiGorlich:
> > * (example [here](https://explain.dalibo.com/plan/c212d7b2g2023a3g))
> 
> Query is missing in the upload? I don't know what this DB query was.

Just the same as the new query, just with half the query removed, so only one of the SELECTs from entry or post

melroy89 (approved):

PR #1900, opened 2025-12-02
Fix multiple mentions, fix magazine mentions

### LINKED ISSUE #1898: Federated microblog replies have incorrect mentions
**Describe the bug**

When viewing a mircoblog reply federated from Mastodon which starts with two different mentions, Mbin replaces the second mention with a copy of the first mention.

fedia.io

1.9.0-rc2

**To Reproduce**

1. View the Mastodon post at <https://mastodon.social/@lemgandi/115645515479942600>
2. View the same post on Fedia at <https://fedia.io/m/random/p/4303327/FreeSoftwareAdvent-I-m-not-sure-if-I-will-keep-this-up#post-comment-3797258>
3. Observe that the Mastodon post begins with `@david_chisnall @neil`, but the version on Fedia begins with `@david_chisnall @david_chisnall`

**Expected behavior**

The post should be the same on both platforms.

**Screenshots**

 - OS: Debian GNU/Linux
 - Browser: Firefox
 - Browser Version: 145.0b7

BentiGorlich:
- magazine mentions were not working, because their name in the DB does not contain an `@` in front of them like users do
- fix mentioning multiple people. The missing brackets messed the filter up, so that all mentions were in this array and it always used only the first one
- fix local users not being found in testing
- fix local users and magazines not being mentioned correctly (you were right @melroy89 :) )
- add tests for that, which can also be easily expanded

Former PR: #1856 
Fixes #1898 

Note that this really only applies to posts which do not contain a markdown source. If there are platforms that add mentions for local users like this: `[@localUser](https://some.instance.tld/u/localUser)` instead of just as `@localUser@some.instance.tld` then they will not work. (they did also not work previously, this is just a reminder)

### DIFF
--- a/src/Service/ActivityPub/MarkdownConverter.php
+++ b/src/Service/ActivityPub/MarkdownConverter.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace App\Service\ActivityPub;
 
+use App\Entity\Magazine;
 use App\Entity\User;
 use App\Service\ActivityPubManager;
 use App\Service\MentionManager;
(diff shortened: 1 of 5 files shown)


BentiGorlich:
Because of this we should recommend clearing the redis cache in 1.9.0 if someone upgraded from one of the release candidates

BentiGorlich:
When looking at gehirneimer I noticed that clearing the cache is actually not the solution. The problem is that we only save the extracted markdown to the DB. So since there was this buggy code, we save the wrong extracted markdown to the DB. Currently there is no way of fixing this, so we might just have to accept that...

PR #1902, opened 2025-12-02
Improve findByActivity

melroy89:
- Small improvements to the find by activity magazine SQL query.

Currently on fedia.io you will see this "top section" at the "[people](https://fedia.io/people)" overview for example:

image

Not only is this query not accurate. It also show invalid magazines here now, like 'porn' or [NonPoliticalTwitter](https://fedia.io/m/NonPoliticalTwitter/people), since are both showing 404s. 

Let see if this PR improvement solves all that.

I got it running now on my instance: https://kbin.melroy.org/people

### DIFF
--- a/translations/messages.de.yaml
+++ b/translations/messages.de.yaml
@@ -545,7 +545,7 @@ last_active: Zuletzt Aktiv
 purge_content: Lösche Inhalt endgültig
 oauth2.grant.domain.subscribe: Abonniere oder Deabonniere Domains und zeige die 
   Domains die du abonniert hast.
-magazine_theme_appearance_icon: Eigenes Icon für das Magazin.
+magazine_theme_appearance_icon: Benutzerdefiniertes Icon für das Magazin.
 oauth2.grant.moderate.magazine.ban.create: Sperre Nutzer in deinen moderierten 
   Magazinen.
 oauth2.grant.admin.user.delete: Lösche Nutzer von deiner Instanz.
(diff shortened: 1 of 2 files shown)


src/Repository/MagazineRepository.php
```diff
@@ -380,8 +380,13 @@ public function findByActivity()
...
             ->andWhere('m.postCount > 0')
             ->orWhere('m.entryCount > 0')
-            ->orderBy('m.postCount', 'DESC')
+            ->andWhere('m.lastActive >= :date')
+            ->andWhere('m.isAdult = false')
+            ->andWhere('m.visibility = :visibility')
             ->setMaxResults(50)
+            ->setParameter('date', new \DateTime('-5 months'))
+            ->setParameter('visibility', VisibilityInterface::VISIBILITY_VISIBLE)
+            ->orderBy('m.entryCount', 'DESC')
```
blued-gear, line 389: Would you move `orderBy()` just below the last `andWhere()`?

I think it fits better as it represents a SQL statement, while all lines below `setMaxResults()` are about building / running the query setup.

melroy89: I dunno. It will end up being the same sql query no matter what here xd

melroy89: In the end it will create exactly the same query.

But normally the order by is later in the query. Like first you have select. Then the joins, then the where, then the order by, finally the limit.

Ignore the setParameter, that's just for the where as well.

blued-gear: > In the end it will create exactly the same query.

Exactly, so the code can be arranged for better readability without harm :p

> Like first you have select. Then the joins, then the where, then the order by, finally the limit.

And currently it is WHERE (line 381 - 385), LIMIT (line 386) and after that ORDER BY (line 389).

blued-gear: I don't want to be too nitpicky, so I am going to approve the PR. You can change the lineorder as you see fit.

PR #1905, opened 2025-12-04
retry update activity if entity was not found

blued-gear:


melroy89:
This could fix https://fedia.io/m/fedia/t/2868838/Some-threads-of-my-kbin-earth-account-are-missing-their-bodies/comment/13230470#entry-comment-13230470 problems right?

blued-gear:
Maybe, but I'm not sure, as Jerry didn't find the logline that the update was discarded.
But this might still fix some edge cases.

I will also look into how updates are sent outwards to check if the update activity might not be sent correctly in the linked scenario.

src/Exception/RetryMessageException.php
```diff
@@ -0,0 +1,19 @@
...
+declare(strict_types=1);
+
+namespace App\Exception;
+
+use JetBrains\PhpStorm\Pure;
+
+/**
+ * Thrown when something is not as expected but might succeed when retried later.
+ */
+final class RetryMessageException extends \Exception
```
melroy89: We already have many custom exceptions. Could we not just reuse a simple existing error class?

Anything will do (except this special error class to avoid retries). Just look which errors we throw in other message handlers.

blued-gear: Maybe the Java Dev in me is talking, but: you call that "many"? :D The more, the merrier and more specific the errors (also for debugging logs).

But ok, I will see what fits best.

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
melroy89: while you at this, shouldn't we also give this a description as well 😕 ?

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
```diff
@@ -108,7 +104,7 @@ public function doWork(MessageInterface $message): void
             return;
         }
 
-        $this->logger->warning("[UpdateHandler::doWork] didn't know what to do with the update activity concerning '{id}'. We don't have a local object that has this id", ['id' => $payload['object']['id']]);
+        throw new \LogicException('Don\'t know what to do with the update activity concerning \''.$payload['object']['id'].'\'. We didn\'t have a local object that has this id.');
```
melroy89, line 107: besides an error, should we still also log  warning. At least that is what Benti was saying. I'm not saying we should also log a warning still.. Just asking what is best to do.

And quote @BentiGorlich :

> Yeah you could just throw after the warning log

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
blued-gear: Should add descriptions to all other such occurrences?

melroy89: I think it can't hurt right? It might actually help us debug issues better in the future..? Can it hurt?

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
blued-gear: The error should be logged when the message fails, so I think logging a warning would be an unnecessary doubling.\
Or are there codepaths which might not print the thrown error?

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
blued-gear: It won't hurt. Just asked as it extends the scope of this PR.

BentiGorlich: Please don't bother 😅
This is a dead code entry but I could not make an interface around it with the transaction logic otherwise, as PHP does not support generics, yet.

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
BentiGorlich: Maybe a NotFoundException better fits this case? But I dpnt really care that much... I'd still log the warning as it will be burried in "Exception occured during the processing of ........" but I am fine with removing the warning as well 🤷

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
blued-gear: > Please don't bother

Too late :D

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
blued-gear: `EntityNotFoundException` is the only existing I found which would fit.

I choose `LogicException` as other handlers uses it when something is not as expected. But I'm fine with using an exception which expresses "not found (yet)" more.

melroy89: Its its getting logged already (eventually), in that case I would **not** log it again. We already have GBs of log file sizes. Redundant logging is not ideal and actually error log pollution in that case.

Is the `Exception occurred during the processing of` error also logging the error message of the class we can improve that if we need to.

Keep in mind  that `EntityNotFoundException` is a special class again, catched here: https://github.com/MbinOrg/mbin/blob/fix/retry_update_activity/src/MessageHandler/ActivityPub/Inbox/ChainActivityHandler.php#L156

This might or might not work.. If we would reuse `EntityNotFoundException`, I would consider renaming it.

blued-gear: Then I would prefer just keeping `LogicException`

melroy89: @BentiGorlich I'm actually more concerned whether these exceptions are logged at all.. https://github.com/search?q=repo%3AMbinOrg%2Fmbin+LogicException

I searched for many of these exceptions. Like "actor must not be null if the object" or "Only magazine and user actors supported" or "could not find a user actor on url" and I grepped many more strings..

I tried to `grep` in all my production logs from the past few days, 100's of MBs of logging.. And no single result. Are we sure these exceptions are even logged in your production logs? Or are all the exceptions we have never reached (by chance, which could be true)?

I also remembered that you where missing log messages, right? I think its because its not part of the info/warning/error logging in the try/catch we have here: https://github.com/MbinOrg/mbin/blob/fix/retry_update_activity/src/MessageHandler/ActivityPub/Inbox/ChainActivityHandler.php#L158 (at least regarding the inbox)

Maybe I'm wrong.. but maybe not 😆

BentiGorlich: Most of these exceptions are unreachable. These are just invalid AP things that should never occurr...

I'd think that all the error messages are getting logged correctly, but either way the transaction should not affect the logging at all

PR #1907, opened 2025-12-06
Do not use settings table for max images bytes (Cleanup both MAX_IMAGE_BYTES and MBIN_MAX_IMAGE_BYTES from table)

melroy89:
Since we now have introduced `MBIN_MAX_IMAGE_BYTES` (with the correct prefix), we can now clean-up the old `MAX_IMAGE_BYTES` setting name from the table. Since that value is not used anymore in the code.

**Later** I found a bug that already exists for years actually, but new max image byte value are not taken into account when setting this value in the `.env`. And because we do not have any admin panel to change the max image bytes, it was impossible to change the default after it was set (just once).

1. Remove the MBIN_MAX_IMAGE_BYTES option from the Settings DTO
2. Introduce a simple getting: `getMaxImageBytes()`
3. Only use the value from `.env` (which also already has a fallback value set): https://github.com/MbinOrg/mbin/blob/b7fb7aee4f09d4f402591acc81bd22fb15c2884c/config/services.yaml#L110
4. Clean-up the settings table

No more out of sync issues between the `.env` setting and the DB setting, since we do not use the settings table for this option (there is no need for it). And we also do not need to expose it on the admin panel, `.env` should therefor be always leading.

### DIFF
--- a/tests/Unit/Service/SettingsManagerTest.php
+++ b/tests/Unit/Service/SettingsManagerTest.php
@@ -31,6 +31,7 @@ class SettingsManagerTest extends WebTestCase
         $setMaxImagesBytes = 1500000;
 
         $settingsRepository = $this->createStub(SettingsRepository::class);
+        $settingsRepository->method('findAll')->willReturn([]);
         $entityManager = $this->createStub(EntityManagerInterface::class);
         $requestStack = $this->createStub(RequestStack::class);
         $kernel = $this->createStub(KernelInterface::class);
@@ -81,6 +82,7 @@ class SettingsManagerTest extends WebTestCase
         $setMaxImagesBytes = 1572864;
 
         $settingsRepository = $this->createStub(SettingsRepository::class);
+        $settingsRepository->method('findAll')->willReturn([]);
         $entityManager = $this->createStub(EntityManagerInterface::class);
         $requestStack = $this->createStub(RequestStack::class);
         $kernel = $this->createStub(KernelInterface::class);


melroy89:
But we might even want to remove the new `MBIN_MAX_IMAGE_BYTES` also from the settings table actually.. Since storing this value to the settings might sounds like a good idea, but it will give out-of-sync issues with the .env configuration.. 

Also the max image bytes can **NOT** be changed in the **admin panel**, and a new value will not be overwritten in the DB config. Meaning it will only be set once. This is a bug.

On Matrix Mbin issues room we are now discussing how to solve this once and for all. I think this setting should not even be stored in the `settings` table. And we should always rely on the `.env` config file called `$this->mbinMaxImageBytes` see https://github.com/MbinOrg/mbin/blob/817ff3741284df1545e8bc708d638404195e72b0/src/Service/SettingsManager.php#L60 

(this already also has a default fallback value, in case the admin didn't provide this .env variable)

melroy89:
Tested and running perfectly on https://kbin.melroy.org. Adapting the setting the `.env` and then running our `./bin/post-upgrade` script works now as expected. Finally.

You can confirm this yourself by opening the image pop-up previewer eg. on the Add new thread form page and look at the "Maximum file size": 

image

Changing this max file size is just as easy as just changing the value in your `.env` and then running the post-upgrade script like I said.

BentiGorlich:
Maybe we have to rethink our settings a bit. We have values like these that are only editable in the `.env`, but we have a lot of settings that are "prefilled" from the `.env` the first time someone clicks save on the admin settings... All of these will then not be changed if one changes their value in the `.env` file. We should think about separating the settings that come from the `.env` and those that are set in the settings table.

So what I mean is:
1. Have some settings in the `.env` file, that are not present in the settings table
2. Have some settings in the settings table that are not present in the `.env` file

melroy89:
> Maybe we have to rethink our settings a bit. We have values like these that are only editable in the `.env`, but we have a lot of settings that are "prefilled" from the `.env` the first time someone clicks save on the admin settings... All of these will then not be changed if one changes their value in the `.env` file. We should think about separating the settings that come from the `.env` and those that are set in the settings table.
> 
> So what I mean is:
> 
>     1. Have some settings in the `.env` file, that are not present in the settings table
> 
>     2. Have some settings in the settings table that are not present in the `.env` file

Yes you are right in that regard. But for now I'm just doing a single last bug fix in 1.9.0 actually. I hope you understand.

But definitely something to look into in 1.9.1.

blued-gear (approved):

blued-gear:
I agree with @BentiGorlich , there should be a clear and consistent separation. This PR can still be merged as is, but we must not forget to deal with the other settings later.

BentiGorlich:
> Yes you are right in that regard. But for now I'm just doing a single last bug fix in 1.9.0 actually. I hope you understand.

Yeah of course, I did mean we should fix it in the future :)

melroy89:
> I agree with @BentiGorlich , there should be a clear and consistent separation. This PR can still be merged as is, but we must not forget to deal with the other settings later.

Thanks for the approval.

PR #1926, opened 2025-12-23
fix post_controller comment element handling

blued-gear:
The `post_controller` uses hardcoded relative positions of elements to place and remove comments. This can lead to broken layouts when the post component is used in new contexts. This PR contains all comment related content into a `div`, which the controller will manipulate.


melroy89:
Failing audit is fixed here: https://github.com/MbinOrg/mbin/pull/1928

PR #1927, opened 2025-12-23
use bigger PHP limits for devcontainer

blued-gear:


BentiGorlich:
Isn't 2 GB way too much? I am pretty sure my servers are set to "only" 512 MB...

blued-gear:
> Isn't 2 GB way too much?

Maybe. But this is only the upper limit and I wanted to make sure my dev environment does not get any errors anymore. And this limit shouldn't hurt I think.

BentiGorlich:
If the limit is different from the recommended server config it could lead to more and more errors in the production environment which we didn't see coming. I would say that these errors you encounter should be fixed by improving memory management and not by raising the memory limit...

blued-gear:
> I would say that these errors you encounter should be fixed by improving memory management and not by raising the memory limit...

The errors on my side are in the main branch features. The default limit in the devcontainer image is just way to small.
What is the limit on your prod server? I can use that to keep it realistic.

melroy89:
I also recommend increasing the max execution time out while you at it:

see also: https://docs.joinmbin.org/contributing/getting_started#prepare-php

```ini
max_execution_time = 120
```

melroy89:
> > I would say that these errors you encounter should be fixed by improving memory management and not by raising the memory limit...
> 
> The errors on my side are in the main branch features. The default limit in the devcontainer image is just way to small. What is the limit on your prod server? I can use that to keep it realistic.

In my case:

```ini
memory_limit = 1G
```

And see also my comment above this one xD

BentiGorlich:
> What is the limit on your prod server? I can use that to keep it realistic.

On gehirneimer: 512 MB
On thebrainbin: 512 MB

@melroy89 my production servers have a max execution time of 30s, just FYI

BentiGorlich:
@blued-gear [our docs](https://docs.joinmbin.org/admin/installation/bare_metal#php) also suggest 512 MB

melroy89:
Just keep in mind that the max execution time might be too low (and possible also the memory limit) if you want to run `doctrine:fixtures:load` now  in a dev container. its a dev container after all.. so I don't know why we would use production values..

Same applies for unit tests and integration tests.. In production you are not running fixtures or integrations tests..

**EDIT:** I mean fixtures:load, not migration of course.

melroy89:
> LGTM, but maybe edit the title :)

What do you think about increasing the values so `doctrine:fixtures:load` won't crash?

melroy89:
> @melroy89 my production servers have a max execution time of 30s, just FYI

but on production servers you are not running integration/unit tests (I hope?) and you are also not running `doctrine:fixtures:load`.. doctrine:fixtures:load can take a long time..

blued-gear:
But sides are valid. In dev more memory for things like the fixtures are useful. On the other hand to large values can hide memory-mismanagement until the version is run on a prod instance.

melroy89:
> On the other hand to large values can hide memory-mismanagement until the version is run on a prod instance.

For that they invented acceptance instances. Which are servers that mimic production setup, but used as acceptance testing, before rolling out to production. Acc is often used to validate the latest release or testing out new functionality or proof of concepts.

However, currently we do not host any acceptance instance. But we could of course setup one. Which should automatically update to the latest main branch for example (unless you tell it otherwise). Such an acceptance server can also help us in bridging the gap between development and production. So yes, as the name implies dev containers are meant for development. Dev containers should not need to be aligned with production values, since it's not acceptance or production.

blued-gear:
Ok then. I increased the limits again (even if I did not revert to 2G). This PR can be merged once the automatic checks with the new skips are on main, or you manually allow this merge.

melroy89 (approved):

PR #1930, opened 2026-01-03
Fix MessageManager#createMessage to properly access 'to' & 'cc' properties

grahhnt:
Allows `createMessage` to correctly pull 'to' & 'cc' properties from the AP object to follow json-ld spec

Similar change is present elsewhere (like ActivityPubManager#createInboxesFromCC)

*From a [similar contribution](https://codeberg.org/rimu/pyfedi/pulls/1531):*
> JSON-LD allows for arrays to be specified by just their first value
>
> I cannot find the specification at the moment, but the [json-ld.org playground](https://json-ld.org/playground/next/) (linked from w3.org) shows this
>
> [Activity with solo To object (playground)](https://json-ld.org/playground/next/#json-ld=%7B%22%40context%22%3A%22https%3A%2F%2Fwww.w3.org%2Fns%2Factivitystreams%22%2C%22to%22%3A%22https%3A%2F%2Fexample.com%22%7D&startTab=tab-expanded) vs [Activity with array of a single To value (playground)](https://json-ld.org/playground/next/#json-ld=%7B%22%40context%22%3A%22https%3A%2F%2Fwww.w3.org%2Fns%2Factivitystreams%22%2C%22to%22%3A%5B%22https%3A%2F%2Fexample.com%22%5D%7D&startTab=tab-expanded)
>
> This change will allow for implementations that use [npm jsonld](https://www.npmjs.com/package/jsonld) (like [fedify](https://fedify.dev)) to correctly interact with Piefed

### DIFF
--- a/src/Service/MessageManager.php
+++ b/src/Service/MessageManager.php
@@ -51,9 +51,7 @@ class MessageManager
         }
         $message = new Message($thread, $sender, $dto->body, $dto->apId);
 
-        foreach ($thread->participants as
-        /** @var User $participant */
-        $participant) {
+        foreach ($thread->participants as $participant) {
             if ($sender->getId() !== $participant->getId()) {
                 if ($participant->isBlocked($sender)) {
                     throw new UserBlockedException();
@@ -123,7 +121,6 @@ class MessageManager
     public function createMessage(array $object): Message|MessageThread
     {
         $this->logger->debug('creating message from {o}', ['o' => $object]);
-        // Possible improvement: convert this to a utility function for getting JSON-LD value or default of empty array
         $obj_to = \App\Utils\JsonldUtils::getArrayValue($object, 'to');
         $obj_cc = \App\Utils\JsonldUtils::getArrayValue($object, 'cc');
         $participantIds = array_merge($obj_to, $obj_cc);


BentiGorlich:
Thanks for bringing this to our attention :)

I think this change would work, but we should implement it in all the places and not just adjust one specific one. If you search the project for reads on this property via `['to']` you'll find a lot of those. The `to` and `cc` properties are the most prominent arrays that might contain only one value, but there are more.

However if it is too much work for you, then we can close this PR and open an issue, so we can work on it when we have time :)

grahhnt:
I have gone through and found the common occurrences of 'to' & 'cc' being accessed and replaced them with a new utility class & method, `JsonldUtils::getArrayValue`

I have also ran php-cs-fixer to resolve that failed test

src/Service/MessageManager.php
```diff
@@ -51,7 +51,9 @@ public function toMessage(MessageDto $dto, MessageThread $thread, User $sender):
         }
         $message = new Message($thread, $sender, $dto->body, $dto->apId);
 
-        foreach ($thread->participants as /** @var User $participant */ $participant) {
+        foreach ($thread->participants as
+        /** @var User $participant */
+        $participant) {
```
BentiGorlich: Please revert this format 😅

src/Service/MessageManager.php
```diff
@@ -121,7 +123,10 @@ public function canUserEditMessage(Message $message, User $user): bool
     public function createMessage(array $object): Message|MessageThread
     {
         $this->logger->debug('creating message from {o}', ['o' => $object]);
-        $participantIds = array_merge($object['to'] ?? [], $object['cc'] ?? []);
+        // Possible improvement: convert this to a utility function for getting JSON-LD value or default of empty array
```
BentiGorlich: This comment can now be removed

PR #1919, opened 2025-12-16
skip tests on non-code paths but keep the GitHub action

blued-gear:

PR #1935, opened 2026-01-05
skip tests on non-code paths but keep the GitHub action - vol 2

blued-gear:
Let's try #1919 again.

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -11826,16 +11826,16 @@
         },
         {
             "name": "symfony/process",
-            "version": "v7.3.4",
+            "version": "v7.3.11",
             "source": {
                 "type": "git",
                 "url": "https://github.com/symfony/process.git",
-                "reference": "f24f8f316367b30810810d4eb30c543d7003ff3b"
+                "reference": "81fe4ea2c3b8677fa2adfd8e48ba42374ede0e3b"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/symfony/process/zipball/f24f8f316367b30810810d4eb30c543d7003ff3b",
-                "reference": "f24f8f316367b30810810d4eb30c543d7003ff3b",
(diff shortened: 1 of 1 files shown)


ci/skipOnExcluded.sh
```diff
@@ -0,0 +1,40 @@
+#!/usr/bin/env bash
+
+set -eu
+
+# Necessary in the GitHub Action environment
+git config --global --add safe.directory "$(realpath "$GITHUB_WORKSPACE")"
+
+ignoredPatterns="$(cat "$GITHUB_WORKSPACE"/ci/ignoredPaths.txt)"
+if [[ "${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" == 'main' ]]; then
+    git fetch origin main --depth 2
```
blued-gear, line 10: I changed this line which was likely the reason for the fail on main branch. But I still have no 100% realistic environment to verify this. \
Do you have an idea?

melroy89: Yes, so what you could try is just temporarily disable the `if` checks in the workflow file. Maybe not all, but some at least, which means that job will get started in your PR.. Meaning you can test if your PR is actually working correctly across the other jobs as well.

melroy89: Or in this case.. maybe updating the `branches:` section.. to include your branch for testing?

blued-gear: On a non-main branch the script works (I saw it on the pipelines here; if it works for one job it will work for all). My question is about the section which gathers the diff to check. On other branches it works by fetching remote main comparing against it. But on main it pulls an older commit to diff. This is the location where the last version failed.

I tried the new script on a local repo copy with main checked out. Not the environment the CI uses, but I think it is realistic enough.

melroy89: well we can try it again.

melroy89:
Its still workingggg 😸

PR #1941, opened 2026-01-12
Improve our RSS feed

BentiGorlich:
- Make it use the new `ContentRepository`
- Add new option for the content. The generated link uses the criteria if present (the set filters)
- Add an image attachment if present
- Add Post "serialization"
- Add the user or magazine selected to the title of the feed
- Set the URL according to the user or magazine selected
- Make the author always the full handle (previously local users were just the plain username)
- Make the content the rendered markdown (HTML) instead of the raw markdown

### DIFF
--- a/README.md
+++ b/README.md
@@ -152,19 +152,19 @@ Unofficial magazines:
                 </a>
             </td>
             <td align="center">
-                <a href="https://github.com/danielpervan">
-                    <img src="https://avatars.githubusercontent.com/u/5121830?v=4" width="100;" alt="danielpervan"/>
+                <a href="https://github.com/blued-gear">
+                    <img src="https://avatars.githubusercontent.com/u/164888202?v=4" width="100;" alt="blued-gear"/>
                     <br />
-                    <sub><b>Daniel Pervan</b></sub>
+                    <sub><b>blued_gear</b></sub>
                 </a>
             </td>
 		</tr>
 		<tr>
             <td align="center">
-                <a href="https://github.com/blued-gear">
-                    <img src="https://avatars.githubusercontent.com/u/164888202?v=4" width="100;" alt="blued-gear"/>
+                <a href="https://github.com/danielpervan">
+                    <img src="https://avatars.githubusercontent.com/u/5121830?v=4" width="100;" alt="danielpervan"/>
                     <br />
-                    <sub><b>blued_gear</b></sub>
+                    <sub><b>Daniel Pervan</b></sub>
                 </a>
             </td>
             <td align="center">
--- a/docs/05-fediverse_developers/README.md
+++ b/docs/05-fediverse_developers/README.md
@@ -218,6 +218,12 @@ If a user boosts content:
 %activity_user_delete%
 ```
 
+### Delete own account
+
+```json
+%activity_user_delete_account%
+```
+
 ## Moderator Activities
 
 ### Add or Remove moderator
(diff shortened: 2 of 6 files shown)


src/Service/FeedManager.php
```diff
@@ -20,63 +22,33 @@
...
-        }
-
-        $items = $this->entryRepository->findByCriteria($criteria);
-
-        if (0 === $items->count()) {
-            throw new NotFoundHttpException();
-        }
-
-        $items = $this->getEntries($items->getCurrentPageResults());
+        $items = $this->getItems($items->getCurrentPageResults());
```
blued-gear: Could you use here or on the variable before different variable-names? It is confusing to have a variable which is reassigned with a different type and use.

src/Service/FeedManager.php
```diff
@@ -125,4 +132,56 @@ public function getEntries(\ArrayIterator $entries): \Generator
...
+        if ($content && \in_array($content, Criteria::CONTENT_OPTIONS, true)) {
+            $criteria->setContent($content);
+        } else {
+            $criteria->setContent(Criteria::CONTENT_THREADS);
+        }
+
+        if ($magazine = $request->get('magazine')) {
+            $magazine = $this->magazineRepository->findOneBy(['name' => $magazine]);
+            if (!$magazine) {
+                throw new NotFoundHttpException();
```
blued-gear: If possible, add a short description what could not be found in the response.

src/Service/FeedManager.php
```diff
@@ -125,4 +132,56 @@ public function getEntries(\ArrayIterator $entries): \Generator
...
+            }
+            $criteria->magazine = $magazine;
+        }
+
+        if ($user = $request->get('user')) {
+            $user = $this->userRepository->findOneByUsername($user);
+            if (!$user) {
+                throw new NotFoundHttpException();
```
blued-gear: If possible, add a short description what could not be found in the response.

src/Service/FeedManager.php
```diff
@@ -85,37 +57,72 @@ public function getFeed(Request $request): FeedInterface
...
     }
 
-    private function createFeed(): Feed
+    private function createFeed(Criteria $criteria): Feed
     {
         $feed = new Feed();
-        $feed->setTitle($this->settings->get('KBIN_META_TITLE'));
+        if ($criteria->magazine) {
+            $title = "{$criteria->magazine->title} - {$this->settings->get('KBIN_META_TITLE')}";
+            $url = $this->urlGenerator->generate('front_magazine', ['name' => $criteria->magazine->name, 'content' => $criteria->content], UrlGeneratorInterface::ABSOLUTE_URL);
```
blued-gear, line 63: In line 142 the content option is validated, here it is not. This might lead to inconsistent results.

src/Service/FeedManager.php
```diff
@@ -85,37 +57,72 @@ public function getFeed(Request $request): FeedInterface
...
+        $feed->setTitle($title);
         $feed->setDescription($this->settings->get('KBIN_META_DESCRIPTION'));
-        $feed->setUrl($this->settings->get('KBIN_DOMAIN'));
+        $feed->setUrl($url);
 
         return $feed;
     }
 
-    public function getEntries(\ArrayIterator $entries): \Generator
+    public function getItems(iterable $content): \Generator
```
blued-gear, line 84: As this is a public function, please add the type of the generator content and the type of the iterable content to a doc-string.

src/Service/FeedManager.php
BentiGorlich: This code uses the criteria, which did validate the content back in line 142 :)

PR #1944, opened 2026-01-13
Like Activities - use magazine id instead of followers collection

BentiGorlich:
It does not make sense to reference the followers collection of a magazine in `Like` activities -> switch to the magazine id


BentiGorlich:
I've looked through our snapshots of the outgoing activities and noticed this. I did not see any other occurrances that are not justified. All `Announce` activities reference this magazine followers collection, but whenever the magazine is the actor (which it is for `Announce` activities) this makes sense

blued-gear (approved):
I will approve this. But we should monitor of likes are correctly federated over multiple instances once this is rolled out.

And once more I'm missing a central specification for the AP stuff where one can look up what should be expected in these fields...

BentiGorlich:
Understandable :D

PR #1946, opened 2026-01-13
Add Indexable setting for users and magazines

BentiGorlich:
- Add an option in the general settings for users and magazines to be indexable. Since it was basically the case until now, the default for local users is true.
- Add a `noindex` header if the magazine a thread or microblog belongs to, or the author of the thread or microblog are not indexable. (Note: this does not apply to comments, since they are always rendered on the thread or microblog page)
- Add the missing storage url to the `.env.test` file. This resulted in a few snapshot changes
- Add the testing services to the DI and add the arguments to only use the testing services when the environment is `test`
- improve the help text rendering underneath checkboxes and apply it to the other checkbox help text in the admin federation page

### DIFF
--- a/.github/workflows/action.yaml
+++ b/.github/workflows/action.yaml
@@ -1,20 +1,6 @@
 name: Mbin Workflow
 on:
   pull_request:
-    paths-ignore:
-      # Dev container is not relevant to the workflow
-      - '.devcontainer/**'
-      # Issue and pull request templates are not relevant to the workflow
-      - '.github/ISSUE_TEMPLATE/**'
-      - '.github/PULL_REQUEST_TEMPLATE/**'
-      # A separate workflow is used to build and publish the pipeline image
-      - 'ci/**'
-      # Documentation changes are not relevant to the workflow and are taking care of by mbin-docs repository
-      - 'docs/**'
-      # Other files are not relevant to the workflow
-      - 'LICENSES/**'
-      - '*.md'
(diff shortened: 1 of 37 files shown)


BentiGorlich:
I honestly do not know why the build step fails. Even the file paths mentioned there look fine (although the path includes `mbin` twice, but it seems to be correct anyway, since the services.yaml also has it in its path)

blued-gear:
The error in the build should be fixable with a `composer dump-autoload`. But shouldn't this be included in the preceding steps?

BentiGorlich:
I think the dump autoload happens also when you just run composer install

BentiGorlich:
Funny thing: the `--no-dev` part of this command `composer install --no-dev --no-progress` is the issue. Without it I have no problems. I tried requiring `symfony/dependency-injection` in the normal, non-dev part of the composer.json, but that didn't change anything...

BentiGorlich:
And I do get this error locally as well

blued-gear:
But why would it break specifically for this PR? If it is just the flag it should have failed in the past too.

BentiGorlich:
Nono it is this code. But I could not replicate it. I just mean that a normal `composer install` is working, but as soon as you add that `--no-dev` it breaks. I started a discussion in the symfony repo about it: https://github.com/symfony/symfony/discussions/63120

BentiGorlich:
New PR for the DI thingies: #1973

PR #1947, opened 2026-01-13
Add discoverable setting for users and magazines

BentiGorlich:
- Add an option in the general settings for users and magazines to be discoverable. Since it was basically the case until now, it is set to true by default.
- Remove non-discoverable magazines and users from all the random panels and search results


blued-gear:
Wouldn't it be better to hide non-discoverable content also from the /all feed?

BentiGorlich:
That is a discussion to be had. Probably makes sense, yeah. @melroy89 @jwr1 what do you think?

jwr1:
> That is a discussion to be had. Probably makes sense, yeah. @melroy89 @jwr1 what do you think?

Yeah, I think I would agree.

I was actually hoping for a feature where admins could white/blacklist the content that shows up in /all (for instance, if I don't want people to see a bunch of memes on my server's front page). Would this be a good way to do that, or should that be a separate feature?

BentiGorlich:
> I was actually hoping for a feature where admins could white/blacklist the content that shows up in /all (for instance, if I don't want people to see a bunch of memes on my server's front page). Would this be a good way to do that, or should that be a separate feature?

No that is not doable with this. Magazines and Users set this value, so admins do not really have a say in that other than on their local magazines... I absolutely understand this feature idea, but it needs to be its own issue

jwr1:
> No that is not doable with this. Magazines and Users set this value, so admins do not really have a say in that other than on their local magazines... I absolutely understand this feature idea, but it needs to be its own issue

Ok, I understand. I think I would maybe still agree, a magazine or user with discoverable disabled shouln't show in /all.

melroy89:
> That is a discussion to be had. Probably makes sense, yeah. @melroy89 @jwr1 what do you think?

My 2 cents are its fine as long as it won't cause impact on the query execution times.. After all, we are trying to reduce this. It would be a shame if we got 2 steps forward and 3 steps back in terms of performance I mean ;P

PR #1949, opened 2026-01-13
Improve search

BentiGorlich:
- dont use ILIKE to search for activities, it has to be a full URL anyway and the other operator is a lot faster
- trim the search query
- use the message bus and force the sync queue instead of manually calling the CreateHandler which is a lot cleaner


blued-gear:
Is the URL for the DB-query normalized before it is used? For example when the user pastes an AP URL, but it has a fragment on it or they somehow left out the https://, this would not match.

BentiGorlich:
It would not even trigger this branch of the search controller, as it will not be a valid URL anymore. See https://github.com/MbinOrg/mbin/blob/main/src/Controller/SearchController.php#L59

PR #1957, opened 2026-01-18
Improve MagazinePeopleFrontController

BentiGorlich:
- the query took too much time in larger magazines -> use the time limit if the magazine has more than 1000 threads/microblogs/comments
- the repo had a hard coded limit of 35 even though it has a limit parameter -> use that limit and set it 28 in the controller, same as PeopleFrontController
- use `UNION ALL` in the query, it is both more performant and more correct in this case
- actually sort the users by content count (after adding the numbers up)

### DIFF
--- a/assets/controllers/subject_controller.js
+++ b/assets/controllers/subject_controller.js
@@ -390,23 +390,21 @@ export default class extends Controller {
     }
 
     wireTouchEvent() {
+        if (this.isOnCombined()) {
+            this.wireTouchEventCombined();
+        } else {
+            this.wireTouchEventRegular();
+        }
+    }
+
+    wireTouchEventRegular() {
         // if in a list and the click is made via touch, open the post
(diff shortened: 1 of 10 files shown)


src/Repository/UserRepository.php
```diff
@@ -581,8 +585,7 @@ public function findUsersForMagazine(Magazine $magazine, ?bool $federated = fals
 
         if (null !== $federated) {
             if ($federated) {
-                $qb->andWhere('u.apId IS NOT NULL')
-                    ->andWhere('u.apDiscoverable = true');
+                $qb->andWhere('u.apId IS NOT NULL');
```
blued-gear, line 588: Why did you remove the `u.apDiscoverable = true` condition?

BentiGorlich: Because it is already part of the where a few lines above this

PR #1960, opened 2026-01-20
Introducing monitoring

BentiGorlich:
- add monitoring capabilities for troubleshooting purposes. This cannot be enabled or disabled from the UI, you have to go into the `.env` files for this
- Add monitoring capabilities for the biggest factors: curl requests (AP), twig rendering (frontend) and of course database querying
- 2 sources can start an execution context: an incoming request and a started message
- added an admin UI for inspecting requests and their response times, with an overview grouped by route name

The main goal of this PR is to add monitoring capabilities, so we can see what our messengers are doing in the background. Additionally I think the graph in the overview (even though it is from a dev instance) shows pretty well that we have to tackle our twig rendering. My guess is that that is a big problem on larger instances, like fedia.io.

<details>
  <summary>Some screenshots:</summary>

grafik
grafik
grafik
grafik
grafik
</details>

### DIFF
--- a/docs/02-admin/04-running-mbin/05-cli.md
+++ b/docs/02-admin/04-running-mbin/05-cli.md
@@ -351,6 +351,18 @@ Options:
 - `--noActivity`: delete images that doesn't have recorded activity. Like comments, updates and/or boosts. (default: `false`)
 - `--batchSize`: the number of images to delete for each type at a time. (default: `800`)
 
+### Remove orphaned media
+
+This command iterates over your media filesystem and deletes all files that do not appear in the database.
+
+```bash
+php bin/console mbin:images:remove-orphaned
+```
+
+Options:
+- `--ignored-paths=IGNORED-PATHS`: A comma seperated list of paths to be ignored in this process. If the path starts with one of the supplied strings it will be skipped. e.g. "/cache" [default: ""]
+- `--dry-run`: Dry run, don't delete anything
+
 ### Rebuild image cache
 
 This command allows you to rebuild image thumbnail cache.
--- a/assets/controllers/subject_controller.js
+++ b/assets/controllers/subject_controller.js
@@ -10,6 +10,7 @@ export default class extends Controller {
     static targets = ['loader', 'more', 'container', 'commentsCounter', 'favCounter', 'upvoteCounter', 'downvoteCounter'];
     static values = {
         loading: Boolean,
+        isOnCombined: Boolean,
     };
     static sendBtnLabel = null;
 
@@ -390,7 +391,7 @@ export default class extends Controller {
     }
 
     wireTouchEvent() {
-        if (this.isOnCombined()) {
+        if (this.isOnCombinedValue) {
             this.wireTouchEventCombined();
         } else {
             this.wireTouchEventRegular();
@@ -434,10 +435,6 @@ export default class extends Controller {
                 }
             }
         });
-    }
-
-    isOnCombined() {
-        return location.pathname.endsWith('/combined') || location.pathname.includes('/combined/');
     }
 
     filterClickEvent(e) {
--- a/composer.lock
+++ b/composer.lock
@@ -6438,16 +6438,16 @@
         },
         {
             "name": "predis/predis",
-            "version": "v3.2.0",
+            "version": "v3.4.1",
             "source": {
                 "type": "git",
                 "url": "https://github.com/predis/predis.git",
-                "reference": "9e9deec4dfd3ebf65d32eb368f498c646ba2ecd8"
+                "reference": "0850f2f36ee179f0ff96c92c750e1366c6cd754c"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/predis/predis/zipball/9e9deec4dfd3ebf65d32eb368f498c646ba2ecd8",
-                "reference": "9e9deec4dfd3ebf65d32eb368f498c646ba2ecd8",
+                "url": "https://api.github.com/repos/predis/predis/zipball/0850f2f36ee179f0ff96c92c750e1366c6cd754c",
+                "reference": "0850f2f36ee179f0ff96c92c750e1366c6cd754c",
                 "shasum": ""
             },
             "require": {
@@ -6489,7 +6489,7 @@
             ],
             "support": {
                 "issues": "https://github.com/predis/predis/issues",
-                "source": "https://github.com/predis/predis/tree/v3.2.0"
+                "source": "https://github.com/predis/predis/tree/v3.4.1"
             },
             "funding": [
                 {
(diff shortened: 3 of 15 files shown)


BentiGorlich:
Just a warning: In about 2 days the gathered stats take up about 16GB on my relatively small server that mainly gets a lot of AP requests, but not that much user requests

|                          table_name                          | table_size | indexes_size | total_size |
| - | - | - | - |
"public"."monitoring_query"                                  | 14 GB      | 599 MB       | 14 GB
"public"."monitoring_twig_render"                            | 1338 MB    | 294 MB       | 1633 MB
"public"."monitoring_execution_context"                      | 345 MB     | 52 MB        | 396 MB
"public"."monitoring_curl_request"                           | 14 MB      | 4792 kB      | 18 MB

blued-gear:
Maybe the stats then should get their own DB connection configuration. This way admins can store the data somewhere where the fill up is not much of a problem or even choose a DB which allows compression.

melroy89:
This is really exciting !

But yeah, you definitely want this is a separate DB.. Or maybe even in a timeseries optimized DB. Especially due to the sizes.

BentiGorlich:
I think I can bring down the size a lot if I separate the query string out to another table and only save a reference to that in the "instances" of the query. Also I want to make some things optional. At the moment it does record the parameters of each query as well, which might be overkill or just unnecessary, so I want it to be optional

BentiGorlich:
And it is quite a lot of data, so you might not want to monitor it for that long anyway. Otherwise you just don't get through it :D

melroy89:
> And it is quite a lot of data, so you might not want to monitor it for that long anyway. Otherwise you just don't get through it :D

We need monitoring of the monitoring, that is monitoring. 😅

BentiGorlich:
Ok, so I pushed a few changes:
- to save space in the DB:
  - make parameter storing optional via env var
  - similar queries/statements will not take up space for each call, but will be saved to a separate table and referenced by hash
- Make the overview chart also display aggregated stats based on the current filter
- Add a dropdown to switch between the total time and the mean time in the overview chart
- Add a gradient to the twig renders either based on the percentage of the total duration or the parent duration

I have it live on gehirneimer, **without** saving the parameters to the DB. Lets see how much space it is taking up in 2 days :)

BentiGorlich:
Now after 3,5 days we have this:
|                          table_name                          | table_size | indexes_size | total_size  |
|--------------------------------------------------------------|------------|--------------|------------ |
| "public"."monitoring_query"                                  | 5397 MB    | 1518 MB      | 6915 MB |
| "public"."monitoring_twig_render"                            | 1906 MB    | 418 MB       | 2324 MB |
| "public"."monitoring_execution_context"                      | 706 MB     | 106 MB       | 812 MB |
| "public"."monitoring_curl_request"                           | 27 MB      | 9472 kB      | 36 MB |
| "public"."monitoring_query_string"                           | 19 MB      | 888 kB       | 20 MB |

still quite a bit and I have parameter logging disabled

BentiGorlich:
I think making query logging and twig render logging optional this is a lot more viable for just use in general, instead of a developer tool.

blued-gear:
Would you also add some markdown documentation about this feature?

docs/02-admin/03-optional-features/08-monitoring.md
```diff
@@ -0,0 +1,41 @@
+# Internal Mbin Monitoring
+
+We have a few environment variables that can enable monitoring of the mbin server, 
+specifically the executed database queries, rendered html components and requested web resources during the execution of
+an HTTP request or a message handler (a background job).
```
blued-gear, line 5: Maybe mention that this monitoring will produce statistics the admin can use to profile server performance.

docs/02-admin/03-optional-features/08-monitoring.md
```diff
@@ -0,0 +1,41 @@
...
+
+Enabling monitoring on your server will have a performance impact. It is not necessarily noticeable for your users, 
+but it will increase the resource consumption .
```
blued-gear: aaaaaaa, space before period

docs/02-admin/03-optional-features/08-monitoring.md
```diff
@@ -0,0 +1,41 @@
+# Internal Mbin Monitoring
+
+We have a few environment variables that can enable monitoring of the mbin server, 
+specifically the executed database queries, rendered html components and requested web resources during the execution of
+an HTTP request or a message handler (a background job).
+
+Enabling monitoring on your server will have a performance impact. It is not necessarily noticeable for your users, 
+but it will increase the resource consumption .  
+During an execution context (request or messenger) we will collect monitoring information according to your settings.
```
blued-gear: 'we' -> 'the server' (else it sounds that the metrics will be sent to the devs)

blued-gear:
I checked out the branch and ran the migrations. The monitoring page shows ` Query monitoring is enabled, but only the execution time. Twig rendering is enabled, but only the execution time. Curl request monitoring is enabled, but only the execution time. `. I did not update the environment yet, so I thought it should be disabled. Also, no data is recorded to the DB.

blued-gear:
It would be useful to have a button for deleting all recorded monitoringdata.

BentiGorlich:
> It would be useful to have a button for deleting all recorded monitoringdata.

I thought about that, but decided against it, because if the monitoring ran for a while, then the request will probably time out. It takes a while to delete all this data. I just forgot to add a command deleting it :sweat_smile:

BentiGorlich:
> I checked out the branch and ran the migrations. The monitoring page shows `Query monitoring is enabled, but only the execution time. Twig rendering is enabled, but only the execution time. Curl request monitoring is enabled, but only the execution time.`. I did not update the environment yet, so I thought it should be disabled. Also, no data is recorded to the DB.

The monitoring in general is disabled, so you should not have seen a link in the admin options, but like mentioned in the docs and the `.env` some settings default to true

blued-gear:
> onitoring ran for a while, then the request will probably time out. It takes a while to delete all this data. I just forgot to add a command deleting it 😅

A command is fine too.

blued-gear:
> > I checked out the branch and ran the migrations. The monitoring page shows `Query monitoring is enabled, but only the execution time. Twig rendering is enabled, but only the execution time. Curl request monitoring is enabled, but only the execution time.`. I did not update the environment yet, so I thought it should be disabled. Also, no data is recorded to the DB.
> 
> The monitoring in general is disabled, so you should not have seen a link in the admin options, but like mentioned in the docs and the `.env` some settings default to true

I noticed it afterward. The link is indeed not shown in the adminpanel. But as the docs reference the URL, it would be good to show on the page that monitoring is generally disabled in such a case.

blued-gear:
The new command needs a quick documentation in our list of commands.

docs/02-admin/04-running-mbin/05-cli.md
```diff
@@ -326,6 +326,32 @@ Arguments:
 
 ## Images
 
+### Delete monitoring data
```
blued-gear: The text is good. But it would fit better under `Miscellaneous` than `Images`.

BentiGorlich: Thats where I wanted to put it :sweat_smile:

PR #1970, opened 2026-01-27
Remove duplicate indexes

BentiGorlich:
I just tested pghero and it showed me 27 duplicate indexes. My guess is that there are a few more than that, but for this first PR I will keep it at these 27 indexes. I did not notice a performance hit when deleting these.

For ease of use I added a lot of comments and also `IF EXISTS`/`IF NOT EXISTS` to all the queries

### DIFF
--- a/public/robots.txt
+++ b/public/robots.txt
@@ -1438,3 +1438,23 @@ Disallow: /newMagazine
 Disallow: /terms
 Disallow: /privacy-policy
 
+# Voting, Boosting, Reporting
+Disallow /er/*
+Disallow /ev/*
+Disallow /ef/*
+Disallow /eb/*
+Disallow /ecr/*
+Disallow /ecv/*
+Disallow /ecf/*
+Disallow /ecb/*
+Disallow /pr/*
+Disallow /pv/*
+Disallow /pf/*
+Disallow /pb/*
+Disallow /pcr/*
+Disallow /pcv/*
+Disallow /pcf/*
+Disallow /pcb/*
+
+# Crossposting
+Disallow /crosspost/*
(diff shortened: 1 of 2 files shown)


migrations/Version20260127111110.php
```diff
@@ -0,0 +1,78 @@
...
+    public function down(Schema $schema): void
+    {
+        $this->addSql('CREATE INDEX IF NOT EXISTS IDX_DA62921D3DAE168B ON bookmark (list_id)');
+        $this->addSql('CREATE INDEX IF NOT EXISTS IDX_A650C0C4A76ED395 ON bookmark_list (user_id)');
+        $this->addSql('CREATE INDEX IF NOT EXISTS IDX_5060BFF4A76ED395 ON domain_block (user_id)');
+        $this->addSql('CREATE INDEX IF NOT EXISTS IDX_3AC9125EA76ED395 ON domain_subscription (user_id)');
+        $this->addSql('CREATE INDEX IF NOT EXISTS entry_visibility_idx ON entry (visibility)');
+        $this->addSql('CREATE INDEX IF NOT EXISTS IDX_9E561267A76ED395 ON entry_comment_vote (user_id)');
+        $this->addSql('CREATE INDEX IF NOT EXISTS IDX_62A2CA1960C33421 ON favourite (entry_comment_id)');
+        $this->addSql('CREATE INDEX IF NOT EXISTS IDX_62A2CA1960C33421 ON favourite (entry_comment_id)');
```
blued-gear: This line is duplicate

blued-gear:
You removed some of the indexes from the Entities; do you think there are more such cases?

BentiGorlich:
> You removed some of the indexes from the Entities; do you think there are more such cases?

No. I did a full text search on everyone of them and the removed ones are the only ones that were explicitly defined in the entities. The others are implicitly defined through the relations with other entities. I also let doctrine auto generate a migration to see whether it wants to re-add them, but it did not.

PR #1972, opened 2026-01-27
Add missing oauth user consent and translation

BentiGorlich:
- in #1773 an consent key was missing (noticed through test warnings) -> add that
- also add translation for the lock consent, for both entries and posts


BentiGorlich:
Oh this has already been merged, sorry :see_no_evil:

PR #1973, opened 2026-01-27
Improve DI for testing

BentiGorlich:
If we're in the test environment we should never load `ApHttpClient` or `ImageManager` because these classes actually request things from other servers, which we do not want while testing. Instead we want to automatically get our testing services. This PR autowires the testing services when we're in the test environment


melroy89:
I agree. the test is however now failing:

```
In DefinitionErrorExceptionPass.php line 48:
                                                                               
  Cannot autowire service "App\Command\ApImportObject": argument "$client" of  
   method "__construct()" references interface "App\Service\ActivityPub\ApHtt  
  pClientInterface" but no such service exists. You should maybe alias this i  
  nterface to the existing "App\Tests\Service\TestingApHttpClient" service.    
                                                                               
```

BentiGorlich:
Yeah I know, but I don't know how to fix it. See https://github.com/symfony/symfony/discussions/63120
I linked it in the old PR, but forgot to mention it here

PR #1980, opened 2026-01-30
preselect "random" magazine for post-create-form (microblogs) when not in a magazine

blued-gear:


src/Controller/Post/PostCreateController.php
```diff
@@ -21,14 +23,22 @@ class PostCreateController extends AbstractController
...
         private readonly IpResolver $ipResolver,
     ) {
     }
 
     #[IsGranted('ROLE_USER')]
     public function __invoke(Request $request): Response
     {
-        $form = $this->createForm(PostType::class);
+        $dto = new PostDto();
+        // check if teh "random" magazine exists and if so, use it
```
melroy89:
```suggestion
        // check if the "random" magazine exists and if so, use it
```

Small typo.

melroy89:
But overall, the random magazine shouldn't ideally be used to post anything in it..

@BentiGorlich what do you think? I don't think we should do this change.

jwr1:
It is actually preselected in Interstellar when creating a microblog as well. If people don't know how microblogs work, they likely won't know random is a special magazine to be used for uncategorized microblogs.

Really, what I did was not require a magazine to be selected, and then if they didn't specify one, then `random` would be picked automatically. There is a brief message underneath the magazine picker explaining it to them.

blued-gear:
> But overall, the random magazine shouldn't ideally be used to post anything in it..

The post form requires a destination magazine to be selected and the purpose of `random` is to catch all content which is not associated with any magazine, so I think this fits.

src/Controller/Entry/EntryFrontController.php
```diff
@@ -210,9 +212,17 @@ private function renderResponse(Request $request, Criteria $criteria, array $dat
 
         if ('microblog' === $criteria->content) {
             $dto = new PostDto();
+
             if (isset($data['magazine'])) {
                 $dto->magazine = $data['magazine'];
+            } else {
+                // check if teh "random" magazine exists and if so, use it
```
BentiGorlich:
```suggestion
                // check if the "random" magazine exists and if so, use it
```

BentiGorlich:
I am with @blued-gear and @jwr1 . I have been asked before where microblogs should be sent to and this would fix it. The ideal state (from which we are far away) would be that a magazine is not required (imo)

src/Controller/Post/PostCreateController.php
```diff
@@ -21,14 +23,22 @@ class PostCreateController extends AbstractController
...
     {
-        $form = $this->createForm(PostType::class);
+        $dto = new PostDto();
+        // check if the "random" magazine exists and if so, use it
+        $randomMagazine = $this->magazineRepository->findOneByName('random');
+        if (null !== $randomMagazine) {
+            $dto->magazine = $randomMagazine;
+        }
+
+        $form = $this->createForm(PostType::class)->setData($dto);
```
BentiGorlich:
```suggestion
        $form = $this->createForm(PostType::class, $dto);
```

melroy89:
Ah I see this is only for microblog?

blued-gear:
> Ah I see this is only for microblog?

Yes. Affecting the forms at `https://mbin.tld/microblog/create` and `https://mbin.tld/microblog`.

PR #1987, opened 2026-02-01
Deprecate the magazine rules field

BentiGorlich:
- remove the field from the form if it is empty -> cannot set new rules on creation or if the rules are already empty
- add a help text informing users of the deprecation
- only try to render the form field in the edit form if it exists
- don't render the rules field upon magazine creation


blued-gear:
Rules are not federated? But I know at least one magazine where the rules show up on Lemmy.

BentiGorlich:
Right, I added that some time back... It is federated, but not in a good way (see `GroupFactory`). Need to adjust the help text

blued-gear:
You should add the deprecation also to the Magazine API endpoints.

PR #1991, opened 2026-02-02
Fix SearchController URL actor search

blued-gear:
Now we can search again via URL: users, entries, posts, magazines, entry comments, post replies.

Caveats:
- comments and replies will not be fetched recursively; so the result will be empty if the parent does not already exist in the DB

(the branch name does not fit anymore as I changed the scope of this fix)

### DIFF
--- a/src/Controller/SearchController.php
+++ b/src/Controller/SearchController.php
@@ -15,6 +15,7 @@ use App\Service\ActivityPubManager;
 use App\Service\SearchManager;
 use App\Service\SettingsManager;
 use Psr\Log\LoggerInterface;
+use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\HttpFoundation\Response;
 use Symfony\Component\Messenger\MessageBusInterface;
@@ -66,6 +67,14 @@ class SearchController extends AbstractController
                 $res = $this->manager->findPaginated($user, $query, $this->getPageNb($request), authorId: $dto->user?->getId(), magazineId: $dto->magazine?->getId(), specificType: $dto->type, sinceDate: $dto->since);
 
                 $this->logger->debug('results: {num}', ['num' => $res->count()]);
+
+                if ($request->isXmlHttpRequest()) {
+                    return new JsonResponse([
+                        'html' => $this->renderView('search/_list.html.twig', [
+                            'results' => $res,
+                        ]),
+                    ]);
+                }
 
                 return $this->render(
                     'search/front.html.twig',
--- a/assets/controllers/infinite_scroll_controller.js
+++ b/assets/controllers/infinite_scroll_controller.js
@@ -58,7 +58,7 @@ export default class extends Controller {
         const elements = div.querySelectorAll('[data-controller="subject-list"] > *');
         for (let i = 0; i < elements.length; i++) {
             const element = elements[i];
-            if (element.id && null === document.getElementById(element.id)) {
+            if ((element.id && null === document.getElementById(element.id)) || element.classList.contains('user-box-inline') || element.classList.contains('magazine') || element.classList.contains('post-container')) {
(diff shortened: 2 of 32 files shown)


src/Repository/SearchRepository.php
blued-gear, line 226: entry_comment (and likely post_comment) can not be searched, as their ap_id is NULL even after they were fetched by the CreateHandler

BentiGorlich: I have not really a clue what you mean by that. I have not a single entry_comment or post_comment in my DB where the user has an ap_id, but their comment does not. If you mean local comments cannot be found by their URL, then that is true for for all content. To find local content by URL (which I think is not necessary) we could use our `ApActivityRepository`

blued-gear: I have to take my comment back. Yesterday I tried to query a remote thread comment via URL and it did not work. To write the reproducing steps I tried it again just now and it works as expected (you enter the URL and the comment gets fetched and shows up in the results).

What still is missing: fetching while comment threads. The search will fail when querying the URL to a comment where the parent is not already present in the DB.

BentiGorlich: > What still is missing: fetching while comment threads. The search will fail when querying the URL to a comment where the parent is not already present in the DB.

I am not sure this is true :thinking:  
What should be happening is that it dispatches a `ChainActivityMessage` which will not be worked on synchronously, so it will not show up after the search is done, however it should be fetched by the messengers in the background. (the sync stamp is not transferred to new dispatched messages if I am not mistaken)

blued-gear: Is the `ChainActivityMessage` created by the `CreateHandler`? If so, then good. I don't have messangers running (I think) in my devcontainer, so I would not notice.

But the immediate result for the user will be the same: empty results at first search. But IMO we can ignore this until a user has the real need to fetch chains via search without rety.

src/Controller/SearchController.php
```diff
@@ -151,4 +136,62 @@ private function lookupHandle(ActorHandle $handle): array
...
+            } else {
+                $type = 'subject';
+            }
+
+            return [
+                'type' => $type,
+                'object' => $object,
+            ];
+        }, $objects);
+    }
```
BentiGorlich, line 205: What is this mapping for? It is not used anywhere as far as I can tell...

src/Repository/SearchRepository.php
```diff
@@ -228,6 +228,10 @@ public function findByApId($url): array
         SELECT id, created_at, 'post' AS type FROM post WHERE ap_id = :url
         UNION ALL
         SELECT id, created_at, 'post_comment' AS type FROM post_comment WHERE ap_id = :url
+        UNION ALL
+        SELECT id, created_at, 'user' AS type FROM \"user\" WHERE ap_profile_id = :url
+        UNION ALL
+        SELECT id, created_at, 'magazine' AS type FROM magazine WHERE ap_profile_id = :url
```
BentiGorlich: Maybe both of these should get a `OR ap_public_url = :url`? (again Mastodon has 2 different URLs for profiles)  
I just tested it and it seems that postgresql is still utilizing the indexes like it should: https://explain.dalibo.com/plan/c5bf8bbe23706a0a

src/Repository/SearchRepository.php
BentiGorlich: Yes, correct. The ChainActivitiyMessage is created by the CreateHandler. We could in theory get the stamps of the message and if it contains a specfic transport also use that for newly dispatched messages. That way we could make it synchronous for the User. But I also think that it is a separate issue

src/Controller/SearchController.php
blued-gear: Line 177. SearchManager::findByApId() returns a different shape than SearchController::lookupHandle() (and both are passed as `objects` to the template), so this mapping makes the former compatible with the latter shape.

PR #1994, opened 2026-02-04
Add check for duplicates command (guided tour)

melroy89:
- Add a new command: `./bin/console mbin:check:duplicates-users-magazines`

This allow the admin to easily check for either duplicated users or magazines. In a guided tour.

Eg.

```sh
./bin/console mbin:check:duplicates-users-magazine

Duplicate Users and Magazines Checker
=====================================

 What would you like to check for duplicates? [Users]:
  [users    ] Users
  [magazines] Magazines
 > 

```

If there are duplicates. You can then use the following command to remove one or more records from the DB:

```sh
 What would you like to check for duplicates? [Users]:
  [users    ] Users
  [magazines] Magazines
 > magazines

Duplicate Magaziness Found
--------------------------

 
==============================
 Duplicate Group: https://piefed.social/c/photography
 ------- --------------------------- --------------------- --------------------- 
  ID      Name                        Created At            Last Active          
 ------- --------------------------- --------------------- --------------------- 
  44255   photography@pixelfed.de     2024-07-15 14:56:17   N/A                  
  22557   photography@piefed.social   2024-07-15 14:56:17   2026-01-17 21:33:21  
 ------- --------------------------- --------------------- --------------------- 

 
Total duplicate Magaziness: 2

 Would you like to delete any of these duplicates? (yes/no) [no]:
 > 

```

Continue with `yes`:

```sh
 Would you like to delete any of these duplicates? (yes/no) [no]:
 > yes

 Enter the IDs to delete (comma-separated, e.g., 1,2,3):
 > 44255
```

### DIFF
--- a/src/Command/CheckDuplicatesUsersMagazines.php
+++ b/src/Command/CheckDuplicatesUsersMagazines.php
@@ -4,6 +4,7 @@ declare(strict_types=1);
 
 namespace App\Command;
 
+use App\Entity\Magazine;
 use App\Entity\User;
 use App\Service\MagazineManager;
 use App\Service\UserManager;
@@ -174,7 +175,7 @@ class CheckDuplicatesUsersMagazines extends Command
                     $io->success("Deleted user: {$existingUser->getUsername()} (ID: $id)");
                 } else { // magazines
                     // Check if magazine exists first
-                    $magazine = $this->entityManager->getRepository(\App\Entity\Magazine::class)->find($id);
+                    $magazine = $this->entityManager->getRepository(Magazine::class)->find($id);
                     if (!$magazine) {
                         $io->warning("Magazine with ID $id not found, skipping...");
                         continue;


src/Command/CheckDuplicatesUsersMagazines.php
```diff
@@ -0,0 +1,208 @@
...
+                $stmt = $stmt->executeQuery(['id' => $id]);
+                $item = $stmt->fetchAssociative();
+
+                if (!$item) {
+                    $io->warning("{$entityName} with ID $id not found, skipping...");
+                    continue;
+                }
+
+                // Delete the entity
+                $deleteSql = "DELETE FROM {$tableName} WHERE id = :id";
```
blued-gear: I think this can fail if there are other entities (like entries) associated via foreign key.

melroy89: yea true.. we are missing actually "ON DELETE" in our foreign keys it seems 😢 .. 
image

That would actually also do a cascade delete of foreign keys... 

Same is true for a magazine btw.

melroy89: anyway. fixed now.

BentiGorlich:
I think it would be better if you just had a guided tour through all duplicate magazines. So you check for duplicates and then you loop over all the `ap_public_url`s and for each of them you list all matching users/magazines and ask which one should be kept. Then you delete the others and go to the next. That way the admins only have to execute one command. What do you think?

melroy89:
> and ask which one should be kept

Its easier to ask which to delete. Rather then which IDs to keep...

---

Anyways, yes I created a guided tour from it now. Works great imo, I also updated the description (see top of PR).

BentiGorlich:
> Its easier to ask which to delete. Rather then which IDs to keep...

Ok, but what do you do if you have 3 magazine matching one public url?

blued-gear (approved):

blued-gear:
> Ok, but what do you do if you have 3 magazine matching one public url?

While I agree with Bento that for the admin it is easier to select just the IDs to keep rather than to IDs to delete, I think we do not need to spend too much time polishing the UX of this command as it is probably use-once only and the admins of Mbin instances are likely capable to handle this TUI without issue.

If Melroy wants to implement the suggestion, I won't stop him of course :D

melroy89:
> > Its easier to ask which to delete. Rather then which IDs to keep...
> 
> Ok, but what do you do if you have 3 magazine matching one public url?

Well.. The guided tour is clear right?

> Enter the IDs to delete (comma-separated, e.g., 1,2,3):

Meaning you enter the id's you want to remove. If you have 3, then you most likely want to remove two. You can provide the 2 ids, as said, comma separated.

PR #1966, opened 2026-01-23
Update symfony to 7.4

BentiGorlich:
- Update symfony to 7.4
- update all recipes

PR #2005, opened 2026-02-07
Use Debian Trixie as our Docker CI/CD image

melroy89:
- Use PHP 8.4 via Debian Trixie based Docker image (slim)
- Fixes issues like we are facing right now during upgrading towards Symfony 7.4: https://github.com/MbinOrg/mbin/pull/1966#issuecomment-3791190103

---

Proof of new Redis v6.2 (inside the Docker image):

```sh
dpkg -l php8.4-redis
Desired=Unknown/Install/Remove/Purge/Hold
| Status=Not/Inst/Conf-files/Unpacked/halF-conf/Half-inst/trig-aWait/Trig-pend
|/ Err?=(none)/Reinst-required (Status,Err: uppercase=bad)
||/ Name           Version      Architecture Description
+++-==============-============-============-========================================
ii  php8.4-redis   6.2.0-1      amd64        PHP extension for interfacing with Redis
```

Related: https://github.com/symfony/symfony/issues/63087

### DIFF
--- a/assets/styles/components/_post.scss
+++ b/assets/styles/components/_post.scss
@@ -106,11 +106,6 @@ blockquote.post {
     display: flex;
     gap: .5rem;
     justify-content: flex-end;
-
-    button {
-      height: 1.2rem;
-      width: 4rem;
-    }
   }
 
   footer {
--- a/config/mbin_routes/user_api.yaml
+++ b/config/mbin_routes/user_api.yaml
@@ -187,3 +187,21 @@ api_user_post_comments_retrieve:
   path: /api/users/{user_id}/post-comments
   methods: [ GET ]
   format: json
+
+api_user_content_retrieve:
+    controller: App\Controller\Api\User\UserContentApi::getUserContent
+    path: /api/users/{user_id}/content
+    methods: [ GET ]
+    format: json
+
+api_user_boosts_retrieve:
+    controller: App\Controller\Api\User\UserContentApi::getBoostedContent
+    path: /api/users/{user_id}/boosts
+    methods: [ GET ]
+    format: json
+
+api_user_moderated_retrieve:
+    controller: App\Controller\Api\User\UserModeratesApi
+    path: /api/users/{user_id}/moderatedMagazines
+    methods: [ GET ]
+    format: json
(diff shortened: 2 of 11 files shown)


blued-gear:
Instead of relying on a PPA what is preventing us from using a base-image of an OS which has more modern packages?

melroy89:
> Instead of relying on a PPA what is preventing us from using a base-image of an OS which has more modern packages?

If that was possible, I would have done that already. But this is exactly the same problem that was also stated here by Ostico: https://github.com/symfony/symfony/issues/63087#issuecomment-3840598546

melroy89:
OK, I see what you mean now. You mean we will move away from Ubuntu.. maybe using Debian?

blued-gear:
> me: [...] OS which has more modern packages
> Melroy: [...] maybe using Debian

Lol. I was thinking about Alpine, Fedora or Open Suse.

melroy89:
Debian 13 (Trixie) comes with PHP v8.4 (so not 8.5), **but** also has PHP Redis version v6, meaning this will also fix the issue. (And is ironically more modern than Ubuntu)

Hence I can use the following base image: `debian:13-slim`.

The rationale why I rather will go for Debian based, is basically all our bare metal installations are running on a Debian based distro, not Alpine, not Fedora, not OpenSuse. So Debian will reflect more the real-world scenario from production deployments.

ci/Dockerfile
```diff
@@ -1,17 +1,35 @@
...
+    php8.4-pgsql \
+    php8.4-gd \
+    php8.4-curl \
+    php8.4-xml \
+    php8.4-redis \
+    php8.4-mbstring \
+    php8.4-zip \
+    php8.4-bz2 \
+    php8.4-intl \
+    php8.4-bcmath
```
BentiGorlich: `php8.4-bcmath` is duplicated and also was before

melroy89: solved..

melroy89:
Should we test it..?

blued-gear (approved):

melroy89:
I believe its still working :)

PR #2007, opened 2026-02-09
Add documentation for the check duplicates command

BentiGorlich:
- Add documentation for mbin:check:duplicates-users-magazines
- Command introduced in #1994


BentiGorlich:
@blued-gear seems like skipping the actions works like a charm :+1:

PR #2009, opened 2026-02-10
make the vote button of posts the same size as on entries

blued-gear:
This affects all post elements (microblog view, combined view, profile).

Before:
before

After:
after


blued-gear:
The vote button for comments (and so post replies) is still on the right. On the one hand this creates left-right alternating buttons when scrolling through posts. On the other hand I like to keep the replies vote buttons consistent with the comments of entries.

melroy89:
And suddenly the after is white themed instead of dark theme🤣..

blued-gear:
White theme is my dev instance. I already implemented the fix and didn't have the motivation to rebuild the CSS and so used a public instance as the example.

melroy89:
> White theme is my dev instance. I already implemented the fix and didn't have the motivation to rebuild the CSS and so used a public instance as the example.

yea totally, I understand. I was just kidding around. I approved it.

PR #2013, opened 2026-02-12
use new AP resolve code from SearchController in SearchRetrieveApi

blued-gear:
This PR modifies the search API endpoint to
- use the new functionality of SearchManager
- return EntryComment, PostComment, Magazine and user additionally to the previous result types
- use a DTO like in the ContentApi to differentiate result item types
- clean up search by handle and URL

**This change is breaking to the current REST-API contract.**

New API for #1001

### DIFF
--- a/composer.lock
+++ b/composer.lock
@@ -834,16 +834,16 @@
         },
         {
             "name": "doctrine/dbal",
-            "version": "4.4.2",
+            "version": "4.4.3",
             "source": {
                 "type": "git",
                 "url": "https://github.com/doctrine/dbal.git",
-                "reference": "476f7f0fa6ea4aa5364926db7fabdf6049075722"
+                "reference": "61e730f1658814821a85f2402c945f3883407dec"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/doctrine/dbal/zipball/476f7f0fa6ea4aa5364926db7fabdf6049075722",
-                "reference": "476f7f0fa6ea4aa5364926db7fabdf6049075722",
+                "url": "https://api.github.com/repos/doctrine/dbal/zipball/61e730f1658814821a85f2402c945f3883407dec",
+                "reference": "61e730f1658814821a85f2402c945f3883407dec",
                 "shasum": ""
             },
             "require": {
@@ -920,7 +920,7 @@
             ],
             "support": {
                 "issues": "https://github.com/doctrine/dbal/issues",
-                "source": "https://github.com/doctrine/dbal/tree/4.4.2"
+                "source": "https://github.com/doctrine/dbal/tree/4.4.3"
             },
             "funding": [
                 {
@@ -936,7 +936,7 @@
                     "type": "tidelift"
                 }
             ],
-            "time": "2026-02-26T12:12:19+00:00"
+            "time": "2026-03-20T08:52:12+00:00"
         },
         {
             "name": "doctrine/deprecations",
@@ -7132,16 +7132,16 @@
         },
         {
             "name": "scienta/doctrine-json-functions",
-            "version": "6.3.0",
+            "version": "6.5.0",
             "source": {
                 "type": "git",
                 "url": "https://github.com/ScientaNL/DoctrineJsonFunctions.git",
-                "reference": "554b2fd281e976a791501fc4753ffd4c5891ec62"
+                "reference": "d19d67a40734fd81025bfffdf40ff26d5ace16ea"
             },
             "dist": {
                 "type": "zip",
-                "url": "https://api.github.com/repos/ScientaNL/DoctrineJsonFunctions/zipball/554b2fd281e976a791501fc4753ffd4c5891ec62",
-                "reference": "554b2fd281e976a791501fc4753ffd4c5891ec62",
+                "url": "https://api.github.com/repos/ScientaNL/DoctrineJsonFunctions/zipball/d19d67a40734fd81025bfffdf40ff26d5ace16ea",
+                "reference": "d19d67a40734fd81025bfffdf40ff26d5ace16ea",
                 "shasum": ""
             },
             "require": {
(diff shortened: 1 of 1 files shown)


src/Controller/Api/Search/SearchRetrieveApi.php
```diff
@@ -155,46 +159,104 @@ public function __invoke(
         $items = $manager->findPaginated($this->getUser(), $q, $page, $perPage, authorId: $authorId, magazineId: $magazineId, specificType: $type);
         $dtos = [];
         foreach ($items->getCurrentPageResults() as $value) {
-            \assert($value instanceof ContentInterface);
-            array_push($dtos, $this->serializeContentInterface($value));
+            // TODO here we have two options
```
blued-gear: Please give me some feedback

jwr1: I'm not sure I understand what the change to the API is here. Doesn't the search API already return federated users and magazines?

blued-gear: Oddly enough no, not in the normal result items. Before my change it asserted, that the items can only be of `ContentInterface` and the OpenAPI schema lists these contents:
```
new OA\Schema(ref: new Model(type: EntryResponseDto::class)),
        new OA\Schema(ref: new Model(type: EntryCommentResponseDto::class)),
        new OA\Schema(ref: new Model(type: PostResponseDto::class)),
        new OA\Schema(ref: new Model(type: PostCommentResponseDto::class)),
```

jwr1: In the API docs, it looks like there's an `apActors` property returned from `/search`, which can be either a user or a magazine. Then it looks like there's an `apObjects` property which can be a thread, thread comment, microblog, or microblog comment.



I'm obviously not looking at the code though, so maybe the API docs are incorrect?

blued-gear: They are correct about the `items` not containing users or magazines. But the content of the returned object is a bit weird. As far as I understood it from the code the `items` contain *content* which matched by text, `apActors` contain users and magazines fetched via searched handle or AP url and `apObjects` contain *content* fetched via AP url.

Honestly, I'm in favor to rework the whole search API to make it more consistent and easy to understand, but breaking an unversioned API is probably not the best idea.

jwr1: Ah, gotcha. Well, I'm fine keeping it how it is or changing it. It would be nice if the Mbin API were versioned. Considering it's not in wide use (that I know of), you could just make the breaking change, and we can fix Interstellar to work with the change afterwards.

blued-gear: I changed the API to make it more consistent with other newer APIs we have. I still needed to keep one extra prop in the response for results of handle and url searches, as these are not paginated and so must be kept separately from `items`.

src/Controller/Api/Search/SearchRetrieveApi.php
```diff
@@ -152,49 +141,82 @@ public function __invoke(
             throw new BadRequestHttpException();
         }
 
+        /** @var ?SearchResponseDto $searchResults */
```
BentiGorlich:
```suggestion
        /** @var ?SearchResponseDto[] $searchResults */
```

src/Controller/Api/Search/SearchRetrieveApi.php
```diff
@@ -152,49 +141,82 @@ public function __invoke(
...
-        $objects = [];
-        if (!$settingsManager->get('KBIN_FEDERATED_SEARCH_ONLY_LOGGEDIN') || $this->getUser()) {
-            $actors = $manager->findActivityPubActorsByUsername($q);
-            $objects = $manager->findActivityPubObjectsByURL($q);
-        }
+        /** @var ?SearchResponseDto $apResults */
+        $apResults = [];
+        if ($this->federatedSearchAllowed()) {
+            if ($handle = ActorHandle::parse($q)) {
+                $actors = $manager->findActivityPubActorsByUsername($handle);
```
BentiGorlich: This is making a webfinger request right away. I think we should be searching in the DB first. I know this is code you just moved, but maybe it should be fixed anyways

src/Service/SearchManager.php
```diff
@@ -71,64 +74,131 @@ public function findRelated(string $query): array
...
     /**
-     * @param string $query a string that may or may not be a URL
+     * Will dispatch a getActivityObject request if a valid URL was provided but no item was found locally.
      *
-     * @return array A list of objects found by the given query, or an empty array if none were found.
-     *               Will dispatch a getActivityObject request if a valid URL was provided but no item was found
-     *               locally.
+     * @param string $url a string that may or may not be a URL
+     *
+     * @return array array ['results' => ['type' => 'magazine'|'user'|'subject', 'object' => Magazine|User|ContentInterface][], 'errors' => Exception[]]
```
BentiGorlich: This is not working for me in PHPStorm

```suggestion
     * @return array{'results': array{'type': 'magazine'|'user'|'subject', 'object': Magazine|User|ContentInterface}, 'errors': \Exception[]}
```

src/Service/SearchManager.php
```diff
@@ -71,64 +74,131 @@ public function findRelated(string $query): array
     }
 
     /**
-     * @param string $query One or more canonical ActivityPub usernames, such as kbinMeta@kbin.social or @ernest@kbin.social (anything that matches RegPatterns::AP_USER)
+     * @param ActorHandle $handle a valid handle (can be obtained from string via ActorHandle::parse())
      *
-     * @return array a list of magazines or users that were found using the given identifiers, empty if none were found or no @ is in the query
+     * @return array ['type' => 'magazine'|'user', 'object' => Magazine|User][]
```
BentiGorlich: This is not working for me in PHPStorm

```suggestion
     * @return array{'type': 'magazine'|'user', 'object': Magazine|User}[]
```

tests/Service/ApHttpClientProxy.php
BentiGorlich, line 1: What is this for exactly? Because we already have two classes distinguished for testing and prod. Imo we should never call remote things in testing. We can simulate it with the testing client, like I do in the AP Tests

blued-gear: Ok, I will try to create some json mocks.

src/Controller/Api/Search/SearchRetrieveApi.php
blued-gear: ActivityPubManager::findActorOrCreate() can consume handles and look users (and only users) up in the DB first (see lines 153 and following). So I might just drop the webfinger and for-loop and use only the manager function, if it does not create a problem ignoring possible >1 webfinger url results.

If this is ok, then I will also add a magazine lookup in findActorOrCreate().

src/Controller/Api/Search/SearchRetrieveApi.php
```diff
@@ -197,4 +204,191 @@ public function __invoke(
...
+            throw new BadRequestHttpException();
+        }
+
+        $page = $this->getPageNb($request);
+        $perPage = self::constrainPerPage($request->get('perPage', SearchRepository::PER_PAGE));
+        $authorIdRaw = $request->get('authorId');
+        $authorId = null === $authorIdRaw ? null : \intval($authorIdRaw);
+        $magazineIdRaw = $request->get('magazineId');
+        $magazineId = null === $magazineIdRaw ? null : \intval($magazineIdRaw);
+        $type = $request->get('type');
```
BentiGorlich: You can get all these values by using `#[MapQueryParameter]` and putting it in the method definition

src/Controller/Api/Search/SearchRetrieveApi.php
```diff
@@ -298,23 +301,22 @@ public function searchV2(
...
         RateLimiterFactoryInterface $apiReadLimiter,
         RateLimiterFactoryInterface $anonymousApiReadLimiter,
+        #[MapQueryParameter(validationFailedStatusCode: Response::HTTP_BAD_REQUEST)]
+        string $q,
+        #[MapQueryParameter(validationFailedStatusCode: Response::HTTP_BAD_REQUEST)]
+        int $perPage = SearchRepository::PER_PAGE,
+        #[MapQueryParameter('authorId', validationFailedStatusCode: Response::HTTP_BAD_REQUEST)]
+        ?int $authorId = null,
+        #[MapQueryParameter('magazineId', validationFailedStatusCode: Response::HTTP_BAD_REQUEST)]
+        ?int $magazineId = null,
```
BentiGorlich, line 311: Is this working for nullable parameters if omitted?

blued-gear: according to https://symfony.com/blog/new-in-symfony-6-3-query-parameters-mapper yes

PR #2017, opened 2026-02-16
use DateType instead of DateTimeType in search and label the field

blued-gear:
**Before:**

before-desktop
before-mobile

**After:**

after

after

### DIFF
--- a/assets/app.js
+++ b/assets/app.js
@@ -1,4 +1,4 @@
-import './bootstrap.js';
+import './stimulus_bootstrap.js';
 import './styles/app.scss';
 import './utils/popover.js';
 import '@github/markdown-toolbar-element';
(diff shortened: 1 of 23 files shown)


BentiGorlich:
I am actually not sure about this change... I think it should default to the 00:00 time, but I think it is quite useful to also be able to set the time

blued-gear:
I also think it is useful. The problem is, that the browsers date-picker doesn't even show a picker for the time which results in bad UX as the form is marked as invalid.

I don't think I can set a default for only the time component. How about we set it to 01.01.1970 00:00:00?

melroy89:
In Firefox this is indeed now broken for me:

image


image

blued-gear:
> In Firefox this is indeed now broken for me
> approved PR

?? \
Or do you mean that it is broken on main?

melroy89:
> > In Firefox this is indeed now broken for me
> > approved PR
> 
> ?? Or do you mean that it is broken on main?

yes for me it is. 

So selecting a date in the pop-up window and then pressing enter/search will result in:

image

blued-gear:
Ok, the 'now' could also have been interpreted as this current branch is broken.

melroy89:
> Ok, the 'now' could also have been interpreted as this current branch is broken.

Sorry that I confused you. Now I mean currently on main branch.

PR #2018, opened 2026-02-16
Improve performance of random panels

BentiGorlich:
- Random entries, posts and magazines now utilize cached user- and magazine-blocks
- the active user component is no longer locale specific -> fewer cache misses
- improve the code for the active user fetch so that
  - the DB is doing more of the heavy lifting
  - avoid doing unnecessary queries (previously we always fetched 3 times from the DB now only 2: fill cache with ids, query users by ids)
- Move the code to get cached user specific things to the sql helpers, so more repositories can utilize it without risking circular dependencies

### DIFF
--- a/src/Repository/UserRepository.php
+++ b/src/Repository/UserRepository.php
@@ -162,7 +162,7 @@ class UserRepository extends ServiceEntityRepository implements UserLoaderInterf
 
     public function findAllInactivePaginated(int $page, bool $onlyLocal = true, ?string $searchTerm = null, ?OrderBy $orderBy = null): PagerfantaInterface
     {
-        $builder = $this->createBasicQueryBuilder($onlyLocal, $searchTerm);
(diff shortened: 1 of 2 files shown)


blued-gear:
Did you run these changes on an instance used regularly (or on a dev instance where you checked touched features)? While I don't see an obvious error in the code, as the changes are somewhat large I would feel safer if it was tested in reality-like conditions.

BentiGorlich:
It has been running on gehirneimer for a few days I think (you can check in git log)

blued-gear:
If there were no issues on gehirneimer so far, you will get the approval after fixing the merge conflict :)

BentiGorlich:
I had no issues with it so far


Am 20. Februar 2026 22:14:37 MEZ schrieb blued_gear ***@***.***>:

blued-gear (approved):

PR #2020, opened 2026-02-16
fix layout of posts on startpage if combined mode is used

blued-gear:


templates/layout/_subject.html.twig
```diff
@@ -20,7 +20,8 @@
     {% set userAttributes = {} %}
 {% endif %}
 
-{% set forCombined = route_param_exists('content') and get_route_param('content') is same as 'combined' %}
+{% set forCombined = (route_param_exists('content') and get_route_param('content') is same as 'combined')
+    or (criteria is defined and criteria.getOption('content') == 'combined') %}
```
BentiGorlich:
```suggestion
    or (criteria is defined and criteria.getOption('content') is same as 'combined') %}
```

BentiGorlich (approved):

assets/controllers/subject_controller.js
```diff
@@ -437,7 +438,7 @@ export default class extends Controller {
     }
 
     isOnCombined() {
-        return location.pathname.endsWith('/combined') || location.pathname.includes('/combined/');
+        return this.forceCombinedValue || location.pathname.endsWith('/combined') || location.pathname.includes('/combined/');
```
BentiGorlich: Is the location check still necessary then?

blued-gear: Not really, but if the component is ever used outside the `subject`template, it is useful again. And the check is short-circuited, so there is no impact in leaving it in. \
But in such a case the issue with taking the mode form criteria stays, so I can remove the path check if you wish.

BentiGorlich:
Yeah I think we should remove it 🤔


Am 20. Februar 2026 22:11:25 MEZ schrieb blued_gear ***@***.***>:

BentiGorlich (approved):

PR #2021, opened 2026-02-17
Abbreviate numbers

BentiGorlich:
- abbreviate all numbers:
  - votes
  - boosts
  - activity numbers
  - subscription and follower-counts
  - instance statistics
- add a unit test for this abbreviation. The implementation is kept simple, because we don't need much and this way it is really fast

### DIFF
--- a/README.md
+++ b/README.md
@@ -145,17 +145,17 @@ Unofficial magazines:
                 </a>
             </td>
             <td align="center">
-                <a href="https://github.com/TheVillageGuy">
-                    <img src="https://avatars.githubusercontent.com/u/47496248?v=4" width="100;" alt="TheVillageGuy"/>
-                    <br />
-                    <sub><b>TheVillageGuy</b></sub>
-                </a>
-            </td>
-            <td align="center">
                 <a href="https://github.com/blued-gear">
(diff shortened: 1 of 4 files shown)


src/Twig/Runtime/FormattingExtensionRuntime.php
```diff
@@ -66,4 +66,20 @@ private static function strrposMulti(string $haystack, array $needle, int $offse
...
+        // this implementation is offly simple, but therefore fast
+        if ($value < 1000) {
+            return ''.$value;
+        } elseif ($value < 1000000) {
+            return round($value / 1000, 2).'K';
+        } elseif ($value < 1000000000) {
+            return round($value / 1000000, 2).'M';
+        } elseif ($value < 1000000000000) {
+            return round($value / 1000000000, 2).'B';
+        } else {
```
blued-gear, line 79: With this very big numbers would not be abbreviated. I think it would be better to use the *B* case as the else.

BentiGorlich: Sure, but we will never ever see numbers that big :sweat_smile:

BentiGorlich: done

PR #2024, opened 2026-02-18
Fix the inactive user list

BentiGorlich:
- the inactive user list was always empty because it basically checked `isVerified = true AND isVerified = false`, which can never be true

### DIFF
--- a/src/Repository/UserRepository.php
+++ b/src/Repository/UserRepository.php
@@ -153,7 +153,6 @@ class UserRepository extends ServiceEntityRepository implements UserLoaderInterf
             ->andWhere('u.isDeleted = false')
             ->andWhere('u.isBanned = false')
             ->andWhere('u.applicationStatus = :status')
-            ->andWhere('u.isVerified = true')
             ->setParameter('status', EApplicationStatus::Approved->value)
             ->setParameter('visibility', VisibilityInterface::VISIBILITY_VISIBLE);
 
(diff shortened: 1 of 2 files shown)


BentiGorlich:
The audit check fails because there is a vulnerability in `firebase/php-jwt`. We have 2 packages depending on it and one of them does not yet work with the newly patched version:

- [stevenmaguire/oauth2-keycloak](https://packagist.org/packages/stevenmaguire/oauth2-keycloak) (not working with 7.0.0 yet), I created an issue: https://github.com/stevenmaguire/oauth2-keycloak/issues/92
- [thenetworg/oauth2-azure](https://packagist.org/packages/thenetworg/oauth2-azure)

src/Repository/UserRepository.php
```diff
@@ -208,8 +212,7 @@ private function createBasicQueryBuilder(bool $onlyLocal, ?string $searchTerm):
     {
         $builder = $this->createQueryBuilder('u');
         if ($onlyLocal) {
-            $builder->where('u.apId IS NULL')
-                ->andWhere('u.isVerified = true');
+            $builder->where('u.apId IS NULL');
```
blued-gear, line 214: Removing the `u.isVerified = true` from a commonly used builder might be error-prone as later a developer might forget to add the condition on their own query.

How about adding a new parameter `onlyVerified` to the function which is true per default and the inactive user check sets it to false?

BentiGorlich: Sure, we can do that

PR #2027, opened 2026-02-20
Fix composer audit: firebase/php-jwt use v7.0.2 as 6.11.1

BentiGorlich:


blued-gear:
As this aliases a more recent major version as 6.x, wouldn't that create potential incompatibilities with the consuming libraries?

BentiGorlich:
Yes it absolutely can. The probability depends on the amount of changes. I didn't look into it, but someone in the issue of the other repo suggested this as a quick fix. So it should work?...


Am 20. Februar 2026 22:08:05 MEZ schrieb blued_gear ***@***.***>:

blued-gear (approved):
Ok, let's try this. If we get reports of something broken, we know where to look first.

PR #2028, opened 2026-02-22
Send delete to all known instances

BentiGorlich:
Previously we only send the delete activity to all instances a user interacted with. That poses a risk, because when creating content on a remote instances this content gets sent to lots of instances the user might have never interacted with. Instead of that, send the delete activity to all instances we know.


blued-gear:
Is it enough to send it to every user, or should magazine inboxes also be included?

PR #2032, opened 2026-02-24
Fix responding with html to ActivityStreams profile

### LINKED ISSUE #2030: AP JSON not provided on required (by spec) accept header (since 1.9.1)
**Describe the bug**
The ActivityPub spec requires that servers return the JSON representation of an object when they receive the full `application/ld+json; profile="https://www.w3.org/ns/activitystreams"` accept header, including the profile. Currently, Mbin only returns it on just `application/ld+json`, the profile isn't supported.

I tried objects on both kbin.earth and thebrainbin.org.
It seems to work correctly on rimworld.gallery, which at the time of writing this is running 1.9.0.

1.9.1 for both the instances with the problem
1.9.0 and earlier seems to not have this issue

**To Reproduce**
1. Make a GET request to `https://kbin.earth/u/Pamasich` with an `Accept` header of `application/ld+json; profile="https://www.w3.org/ns/activitystreams"`
2. The response will be HTML
3. Make the same request to `https://rimworld.gallery/m/4DMiner` instead, the result will be JSON

**Expected behavior**
JSON is returned, not HTML.

> Servers MAY use HTTP content negotiation as defined in [[RFC7231](https://www.w3.org/TR/activitypub/#bib-RFC7231)] to select the type of data to return in response to a request, but ***MUST present the ActivityStreams object representation in response to application/ld+json; profile="https://www.w3.org/ns/activitystreams"***, and SHOULD also present the ActivityStreams representation in response to application/activity+json as well. The client MUST specify an Accept header with the application/ld+json; profile="https://www.w3.org/ns/activitystreams" media type in order to retrieve the activity.

The emphasized portion of the quote above is the expected behavior.

In isolation:

> MUST present the ActivityStreams object representation in response to application/ld+json; profile="https://www.w3.org/ns/activitystreams"

**Screenshots**
Correct behavior (1.9.0):
Image

Incorrect behavior (1.9.1):
Image

BentiGorlich:
Fixes #2030

### DIFF
--- a/src/Entity/Instance.php
+++ b/src/Entity/Instance.php
@@ -19,7 +19,7 @@ class Instance
         CreatedAtTrait::__construct as createdAtTraitConstruct;
     }
     use UpdatedAtTrait;
-    public const NUMBER_OF_FAILED_DELIVERS_UNTIL_DEAD = 10;
+    public const int NUMBER_OF_FAILED_DELIVERS_UNTIL_DEAD = 10;
 
     public static function getDateBeforeDead(): \DateTimeImmutable
     {
--- a/src/Factory/ActivityPub/EntryPageFactory.php
+++ b/src/Factory/ActivityPub/EntryPageFactory.php
@@ -85,16 +85,18 @@ class EntryPageFactory
             $entry->lang => $page['content'],
         ];
 
-        if ($entry->image) {
-            $page = $this->imageWrapper->build($page, $entry->image, $entry->title);
-        }
-
         if ($entry->url) {
             $page['source'] = $entry->url;
             $page['attachment'][] = [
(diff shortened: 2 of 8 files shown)


tests/Functional/Controller/ActivityPub/GeneralAPTest.php
```diff
@@ -0,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Tests\Functional\Controller\ActivityPub;
+
+use App\Tests\WebTestCase;
+
+class GeneralAPTest extends WebTestCase
```
blued-gear, line 10: Now that there is a test, it should cover all content-type cases :)

PR #2033, opened 2026-02-24
Add a command to delete orphaned media

BentiGorlich:


docs/02-admin/04-running-mbin/05-cli.md
```diff
@@ -350,6 +350,19 @@ Options:
...
+### Remove orphaned media
+
+This command iterates over your media filesystem and deletes all files that do not appear in the database.
+
+```bash
+php bin/console mbin:images:remove-orphaned
+```
+
+Options:
+- `--ignored-paths=IGNORED-PATHS`: A comma seperated list of paths to be ignored in this process. If the path starts with one of the supplied string it will be skipped. e.g. "/cache" [default: ""]
```
blued-gear:
```suggestion
- `--ignored-paths=IGNORED-PATHS`: A comma seperated list of paths to be ignored in this process. If the path starts with one of the supplied strings it will be skipped. e.g. "/cache" [default: ""]
```

src/Service/ImageManager.php
```diff
@@ -210,4 +214,143 @@ public function getMimetype(\App\Entity\Image $image): string
...
+                foreach ($this->deleteOrphanedFilesIntern($repository, $dryRun, $deleteEmptyDirectories, $ignoredPaths, $content->path()) as $deletedPath) {
+                    yield $deletedPath;
+                }
+            }
+        }
+
+        if ($deleteEmptyDirectories) {
+            $contents = $this->publicUploadsFilesystem->listContents($path);
+            $length = 0;
+            foreach ($contents as $content) {
```
blued-gear: this loop can be canceled after the first item or replaced with whatever "check for empty" functions the type of $contents supports

src/Service/ImageManager.php
```diff
@@ -210,4 +214,143 @@ public function getMimetype(\App\Entity\Image $image): string
...
+                ++$length;
+            }
+            if (0 === $length) {
```
blued-gear: will this work for nested empty directories?

src/Service/ImageManager.php
```diff
@@ -210,4 +214,143 @@ public function getMimetype(\App\Entity\Image $image): string
...
+                    break;
+                }
+            }
+        }
+        $parts = explode('/', $path);
+
+        return [$path, end($parts)];
+    }
+
+    private function shouldNodeBeIgnored(array $ignoredPaths, StorageAttributes $content): bool
```
blued-gear: this might be extracted to a util class and then covered by a unit test

src/Service/ImageManager.php
BentiGorlich: No it will probably not. Also for s3 servers it would be a lot faster if I'd remove the default delimiter on the list operation, which would probably lead to no directories being listed anymore (I think)

tests/Unit/Utils/GeneralUtilTest.php
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    public static function providePaths(): array
+    {
+        // our paths never start with a '/'
+        return [
+            [['/cache'], 'ca/fe/asdoihsd.png', false],
+            [['/cache'], 'cache/ca/fe/asdoihsd.png', true],
+            [['cache'], 'cache/ca/fe/asdoihsd.png', true],
+            [['/cache'], 'cache/ca/fe/asdoihsd.png', true],
```
blued-gear: this case is a duplicate; did you mean `cache/`?

BentiGorlich: no I didn't, this is just a dupe, but this is a good idea :)

PR #2034, opened 2026-02-25
Add modlogs for purges by moderators

BentiGorlich:
- when a moderator purges content this should also be reflected in the modlog -> add new modlog types with a title property, as the entity is purged, so no longer referencable


migrations/Version20260225095315.php
```diff
@@ -0,0 +1,33 @@
...
+    public function getDescription(): string
+    {
+        return 'Add magazine log columns for the purged classes';
+    }
+
+    public function up(Schema $schema): void
+    {
+        $this->addSql('ALTER TABLE magazine_log ADD author_id INT DEFAULT NULL');
+        $this->addSql('ALTER TABLE magazine_log ADD title VARCHAR(255) DEFAULT NULL');
+        $this->addSql('ALTER TABLE magazine_log ADD CONSTRAINT FK_87D3D4C5F675F31B FOREIGN KEY (author_id) REFERENCES "user" (id) ON DELETE CASCADE NOT DEFERRABLE INITIALLY IMMEDIATE');
```
blued-gear, line 21: please name the constraint and index with a meaningful name

blued-gear:
I see a potential issue with persisting the title: purging is a drastic action, so there must be reasons with it. What if the mod purged content because there was illegal content (potentially also in the title). Then it would continue existing on the server and UI which could get the server admin into legal trouble.

migrations/Version20260225095315.php
BentiGorlich: These are the doctrine generated ones. Afaik if I rename them to something else, doctrine will try to regenerate them in every diff... I can try if that it true

BentiGorlich:
> What if the mod purged content because there was illegal content (potentially also in the title). Then it would continue existing on the server and UI which could get the server admin into legal trouble.

Yes in theory that is a problem. I could not think of a reason why you would then delete a post, but not the user. If the user was deleted the modlog would also be deleted. Maybe this has to be postponed until the moderate actions have a useful page of their own so that we can add a checkbox there whether the modlog has to be hidden (not created)

blued-gear:
> Maybe this has to be postponed until the moderate actions have a useful page of their own

Yea; we also have the mod-action federation issue. So we need to plan some restructuring in general. \
I can approve this PR (after you checked the index name thing) and the feature can be adjusted later.

PR #2036, opened 2026-02-25
Implement support for custom usernames

### LINKED ISSUE #459: Let users specify display names
Similar to how magazines have display names, I think users would like to be able to customize a display name with spaces or different capitalization from their username.

blued-gear:
(I hope I found all places in the UI this should be displayed.)

The user can now set a custom string to be displayed instead of just the name-part of the handle.
This is federated inward and outward.

closes #459


src/Factory/ActivityPub/PersonFactory.php
```diff
@@ -31,7 +31,7 @@ public function create(User $user, bool $context = true): array
             $person ?? [], [
                 'id' => $this->getActivityPubId($user),
                 'type' => $user->type,
-                'name' => $user->username,
+                'name' => $user->getDisplayname() ?: $user->getUsername(),
```
BentiGorlich: Everywhere else you use the `??` operator instead of this one. Either also use the `??` here as well (which I'd prefer) or switch everywhere else to `?:`

src/Entity/User.php
```diff
@@ -102,6 +103,8 @@ class User implements UserInterface, PasswordAuthenticatedUserInterface, Visibil
     public string $email;
     #[Column(type: 'string', unique: true, nullable: false)]
     public string $username;
+    #[Column(type: 'string', unique: false, nullable: true)]
+    private ?string $displayname = null;
```
BentiGorlich: Why private and not public like every other property?

src/Factory/ActivityPub/PersonFactory.php
blued-gear: Whops, this typed the Kotlin dev in me 😅

blued-gear:
> there it is title, but it could be moved in the ActivityPubActorTrait

Making it consistent sounds good to me. But I want to make sure: will there be most likely never a type of actor which does not support a custom title?

src/Entity/User.php
blued-gear: *takes deep breath in*

Because our current code-design allows property access and modification to be messy. The `username` prop is the best example, as it is returned by 3 functions with very different names but the same value + the prop gets directly accessed at other locations in the codebase. \
With the getter/setter paradigm it is easy to track access, modifications and future transformations of the value can be implemented transparently without affecting all the consuming locations.

It is inconsistent with our current style, yes. But I wanted that at least my submitted code is easier to maintain and debug (refactoring all would be too much for me / now).

*brethes out*

BentiGorlich: Imo the `username` property is the worst example, because multiple interfaces we need to implement for our framework require such methods. I don't think getter and setter methods are easier to track than just look for references to the property itself. My IDE has no problem showing me all write accesses to some property. If you'd wanted it to be read-only, then sure. That is what we do with most of the `id` properties. But with normal entity properties I don't like that notation as it has no benefits for me, but takes up 8 lines instead of 2 (if you count the attribute). What is the benefit for having a public setter and getter instead of a public property? The modification can still be done by everything, the references can still be gathered, but now you have to look for the references of 2 things instead of just 1... I don't see it...

BentiGorlich:
I don't know if we will implement a new type of actor any time soon :sweat_smile: 
I think every actor can have a custom displayname though

src/Entity/User.php
blued-gear: > but now you have to look for the references of 2 things instead of just 1

As it is private you still have to only search for the references of one element - the getter (assuming nothing happens in the entity itself, but this would be easy to see).

> I don't think getter and setter methods are easier to track than just look for references to the property itself.

It helps when one want to set breakpoints or temporary logging for value-changes, as these have to be added to only the setter.
My IDE can show me all the references too (as long as they are not accessed by something like `$user->$theVarNameVar`), but when trying to find out why and when a particular value was set a breakpoint / logpoint (which I don't have to attach on every occurrence) is very useful.

> but takes up 8 lines instead of 2

I agree, it is more verbose.

Also, let me give you another example for the use of getters (do not have to apply to this specific instance): we have a nullable property somewhere. Then we learn that this property can have a default which some AP objects set. Without getter all consumers have to be adjusted with a `??`; with getter this change only needs to happen in the getter.

BentiGorlich: Let's just agree to disagree :sweat_smile: 

Since this is again a styling choice we should ask @melroy89 what he thinks. I am not a fan of doing it in exactly one place. I agree that, if we settle on getters and setters, we don't have to it everywhere at once, but I think we should at least do it class by class, so that the code we're looking at is consistent.

melroy89: Hi :). Its me again. 

Let me try to fallback again to the official Symfony documentation to keep this more factual. https://symfony.com/doc/current/doctrine.html#creating-an-entity-class

Which apparently has a command for it even to help out (`php bin/console make:entity`), that generate also private members and then introduces getters/setters. 

It also can generate getters/setters for you apparently? I didn't test it.. `php bin/console make:entity --regenerate --overwrite`.

blued-gear: > Let's just agree to disagree 😅
> [..]
> I agree that, if we settle on getters and setters, we don't have to it everywhere at once, but I think we should at least do it class by class, so that the code we're looking at is consistent.

Yea, that sounds good to me. Especially if the command Melroy found helps us out. \
For now this is settled as I had to revert to public property when moving the prop into the trait to have it in sync with Magazine.

translations/messages.en.yaml
BentiGorlich, line 1: Oh you have the same problem with PHP Storm that I have. You always have to uncheck all whitespace changes. It is quite annoying, but afaik weblate wants to have these present

translations/messages.en.yaml
```diff
@@ -91,9 +91,10 @@ to: to
 in: in
 from: from
 username: Username
+displayname: Custom name
```
BentiGorlich: Why not just call it `Displayname`?

translations/messages.en.yaml
blued-gear: I think I might have found a solution: Settings -> Editor -> Code Style -> [Tab] Formatter -> Do not format -> `*.yaml`

translations/messages.en.yaml
blued-gear: I thought `Custom name` is more clear in saying "this is a name which you can customize whenever you want". `Displayname` is fine too; I'm unsure myself.

melroy89: `Display name` right? Although I'm not native English speaker.

blued-gear: Spellcheck and dictionary says you are right.

translations/messages.en.yaml
melroy89: In vscode/or forks  you have the option to "Save without formatting" .. but that is just me.

PR #2039, opened 2026-02-27
change DTO of signup notifications

blued-gear:
This is a breaking API change, but it should be fine.


melroy89 (approved):
@jwr1 could you double check if you will need to update your app, since it will return different data?

jwr1 (approved):
I don't think this should cause any issues

PR #2041, opened 2026-03-04
Improve image handling

### LINKED ISSUE #511: Further Improve Handling of huge remote images
Follow-up issue for #501 

Main problem after #501: if there is an entry without a local image (because it was too large) we do not have a thumb_image of that.

Solution: compress the image and safe it afterwards. We add a new column is_compressed, so we can display the original full res image from the source url when it is needed and we know that we don't have the original full res image at hand

BentiGorlich:
- add `created_at` and `downloaded_at` timestamps to the image entity
- add `local_size` and `original_size` properties to the image entity. Will be 0 per default and will stay like this for a while
- explicitly pull in the `imagine` library already used via `liip/imagine-bundle` to compress images
- new setting for the compression quality, it defaults to `-1` which disables it. If enabled it will compress images to stay under the max image bytes setting
- add command to remove cached remote images without removing the reference, that way they can be redownloaded if needed
- redownload not-cached images when opening the thread it is attached to (if it was a comment then the thread is also the relevant thing)
- delete all cached variants (meaning the liip/imagine filters: avatar_thumb, entry_thumb, etc.) of an image when deleting it

Closes #511

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -16,6 +16,7 @@ RUN apt-get update && apt-get install -y --no-install-recommends \
 	gettext \
 	git \
 	gosu \
+    procps \
 	&& rm -rf /var/lib/apt/lists/*
 
 RUN set -eux; \
--- a/src/Service/ActivityPub/ContextsProvider.php
+++ b/src/Service/ActivityPub/ContextsProvider.php
@@ -17,8 +17,6 @@ class ContextsProvider
     public static function embeddedContexts(): array
     {
         return [
-            ActivityPubActivityInterface::CONTEXT_URL,
-            ActivityPubActivityInterface::SECURITY_URL,
             [
                 ...ActivityPubActivityInterface::ADDITIONAL_CONTEXTS,
             ],
@@ -28,6 +26,8 @@ class ContextsProvider
     public function referencedContexts(): array
     {
         return [
+            ActivityPubActivityInterface::CONTEXT_URL,
+            ActivityPubActivityInterface::SECURITY_URL,
             $this->urlGenerator->generate('ap_contexts', [], UrlGeneratorInterface::ABSOLUTE_URL),
         ];
     }
--- a/tests/Unit/ActivityPub/Outbox/JsonSnapshots/AddHandlerTest__testAddModerator__1.json
+++ b/tests/Unit/ActivityPub/Outbox/JsonSnapshots/AddHandlerTest__testAddModerator__1.json
@@ -1,5 +1,7 @@
 {
     "@context": [
+        "https://www.w3.org/ns/activitystreams",
+        "https://w3id.org/security/v1",
         "https://kbin.test/contexts"
     ],
     "id": "SCRUBBED_ID",
--- a/tests/Unit/ActivityPub/Outbox/JsonSnapshots/AddHandlerTest__testAddPinnedPost__1.json
+++ b/tests/Unit/ActivityPub/Outbox/JsonSnapshots/AddHandlerTest__testAddPinnedPost__1.json
@@ -1,5 +1,7 @@
 {
     "@context": [
+        "https://www.w3.org/ns/activitystreams",
+        "https://w3id.org/security/v1",
         "https://kbin.test/contexts"
     ],
     "id": "SCRUBBED_ID",
--- a/tests/Unit/ActivityPub/Outbox/JsonSnapshots/AddHandlerTest__testRemoveModerator__1.json
+++ b/tests/Unit/ActivityPub/Outbox/JsonSnapshots/AddHandlerTest__testRemoveModerator__1.json
@@ -1,5 +1,7 @@
 {
     "@context": [
+        "https://www.w3.org/ns/activitystreams",
+        "https://w3id.org/security/v1",
         "https://kbin.test/contexts"
     ],
     "id": "SCRUBBED_ID",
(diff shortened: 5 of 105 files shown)


docs/02-admin/04-running-mbin/05-cli.md
```diff
@@ -326,6 +326,30 @@ Arguments:
...
+This command allows you to remove the cached file of remote media, **without** deleting the reference.
+You can run this command as a cron job to only keep cached media from the last 30 days for example.
+
+> [!TIP]
+> If a thread or microblog is opened without a local cache of the attached image existing, the image will be downloaded again.
+> Once an image is downloaded again, it will not get deleted for the number of days you set as a parameter.
+
+> [!NOTE]
+> User avatars and covers and magazine icons and banners are not affected by this command, 
+> only images from threads, microblogs and comments
```
blued-gear:
```suggestion
> only images from threads, microblogs and comments.
```

src/Controller/Entry/EntrySingleController.php
```diff
@@ -28,17 +31,24 @@ class EntrySingleController extends AbstractController
 {
     use PrivateContentTrait;
 
+    public function __construct(
+        private readonly Security $security,
+        private readonly ImageRepository $imageRepository,
+        private readonly EntryCommentRepository $repository,
```
blued-gear: maybe rename this to `$commentRepository` as my first association with `$repository` inside a class named `EntrySingleController` is the entry repo (same for the post controller)

src/Controller/Post/PostSingleController.php
```diff
@@ -39,6 +40,7 @@ public function __invoke(
         MentionManager $mentionManager,
         Request $request,
         Security $security,
+        ImageRepository $imageRepository,
```
blued-gear: why did you not move some of the injectable objects into the constructor like in EntrySingleController?

src/Service/ImageManager.php
```diff
@@ -79,6 +88,61 @@ public function store(string $source, string $filePath): bool
...
+        $compressed = false;
+        $quality = 0.9;
+        if (0.1 <= $this->imageCompressionQuality && 1 > $this->imageCompressionQuality) {
+            $quality = $this->imageCompressionQuality;
+        }
+        while ($bytes > $maxBytes && $quality > 0.1) {
+            $this->logger->debug('[ImageManager::compressUntilSize] Trying to compress "{path}" with {q}% quality', ['path' => $tempPath, 'q' => $quality * 100]);
+            $image->save($tempPath, [
+                'jpeg_quality' => $quality * 100, // jpeg max value is 100
+                'png_compression_level' => \intval((1 - $quality) * 9), // png max is 9, but it is not quality, but compression
```
blued-gear: As PNG is lossless it is useless to use any other value than max compression.

I think PNGs should be converted to JPEG or skipped by this function.

src/Service/ImageManager.php
```diff
@@ -79,6 +88,61 @@ public function store(string $source, string $filePath): bool
...
+                    'bytesBefore' => $this->formattingExtensionRuntime->abbreviateNumber($initialBytes).'B',
+                    'bytesNow' => $this->formattingExtensionRuntime->abbreviateNumber($bytes).'B',
+                ]);
+            }
+        }
+        if (file_exists($tempPath)) {
+            unlink($tempPath);
+        }
+
+        return $compressed && $copied;
```
blued-gear: If $copied is true, then $compressed has to be true. So this && is redundant.

BentiGorlich: no it does not. The copy operation could fail, then you would not make into the if in line 130 and `$copied` would be `false`

src/Service/ImageManager.php
BentiGorlich: yeah we could do that

src/Service/ImageManager.php
blued-gear: Right. My argument still holds. As `$compressed && $copied` can be shortened to `$copied`.

```
$compressed === false -> $copied === false === ($compressed && $copied)
$compressed === true ; <copy fails> -> $copied === false === ($compressed && $copied)
$$compressed === true ; <copy sucessful> -> $copied === true === ($compressed && $copied)
```

docs/02-admin/03-optional-features/09-image-compression.md
```diff
@@ -0,0 +1,34 @@
...
+
+When image compression is enabled you can reduce your maximum image size to, lets say 1MB. 
+Without the compression this might not be suitable, because too many images exceed that size,
+and you don't want to risk compatibility problems, 
+but with it enabled the images will just be compressed, saving space.
+
+## A better user experience
+
+Normally there is a maximum image size your users must adhere to, but if image compression is enabled,
+instead of showing your user an error that the image exceeds that size, the upload goes through and the image is compressed.
```
blued-gear, line 34: What happens if a user uploads a PNG which is over the size limit. As it can not be further compressed, will it be rejected or be stored despite being larger than allowed?

melroy89: I think it should then still get rejected imo

BentiGorlich: Everything above the size limit will still be rejected

src/Repository/ImageRepository.php
```diff
@@ -172,4 +187,128 @@ public function blurhash(string $filePath): ?string
...
+    /**
+     * @param int $limit use a high limit, as this query takes a few seconds and the limit does not affect that, so we are using as high a number as we can -> we're limited by memory
+     *
+     * @return Pagerfanta<Image>
+     *
+     * @throws Exception
+     */
+    public function findOldRemoteMediaPaginated(int $olderThanDays, int $limit = 10000): Pagerfanta
+    {
+        $query = $this->createQueryBuilder('i')
```
blued-gear: this variable seems to be unused

PR #2047, opened 2026-03-10
Add command to fix user duplicates

BentiGorlich:
- make the `ActivityPubManager` update the username as well
- Add a command to fix duplicate usernames, either by refreshing the remote actors or by merging the users together

### DIFF
--- a/assets/controllers/infinite_scroll_controller.js
+++ b/assets/controllers/infinite_scroll_controller.js
@@ -22,9 +22,19 @@ export default class extends Controller {
         try {
             this.loadingValue = true;
 
-            const paginationElem = this.paginationTarget.getElementsByClassName('pagination__item--current-page')[0].nextElementSibling;
-            if (paginationElem.classList.contains('pagination__item--disabled')) {
-                throw new Error('No more pages');
+            const cursorPaginationElement = this.paginationTarget.getElementsByClassName('cursor-pagination');
+            let paginationElem = null;
+            if (cursorPaginationElement.length) {
+                const button = cursorPaginationElement[0].getElementsByClassName('next');
+                if (!button.length) {
+                    throw new Error('No more pages');
+                }
+                paginationElem = button[0];
+            } else {
+                paginationElem = this.paginationTarget.getElementsByClassName('pagination__item--current-page')[0].nextElementSibling;
+                if (paginationElem.classList.contains('pagination__item--disabled')) {
+                    throw new Error('No more pages');
+                }
             }
 
             if (window.infiniteScrollUrls.includes(paginationElem.href)) {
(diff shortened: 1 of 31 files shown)


melroy89:
We already have a command that checks for user duplicates. I know that that command didn't found these issues.. Nevertheless, it would be better for the end user to just have 1 command to once and for all fix user duplicates. 😊

BentiGorlich:
True, I forgot about that. I will try to consolidate the 2 commands :+1:

blued-gear:
What happens with mentions when a user changed their handle? I expect in the current situation they will just be broken, right?

BentiGorlich:
> What happens with mentions when a user changed their handle? I expect in the current situation they will just be broken, right?

Yes it would just be broken

src/Service/ActivityPubManager.php
```diff
@@ -384,6 +384,7 @@ private function updateUser(string $actorUrl): ?User
             $user->type = $actor['type'] ?? 'Person';
             $user->apInboxUrl = $actor['endpoints']['sharedInbox'] ?? $actor['inbox'];
             $user->apDomain = parse_url($actor['id'], PHP_URL_HOST);
+            $user->username = '@'.($actor['preferredUsername'] ?? $user->username).'@'.$user->apDomain;
```
blued-gear: I think we should add a log when the handle changed. This way we can create some statistics to estimate if we need to fix possible broken mentions.

src/Command/CheckDuplicatesUsersMagazines.php
```diff
@@ -196,4 +219,333 @@ private function deleteEntities(SymfonyStyle $io, string $entity, array $ids): i
...
+            }
+
+            return [];
+        }
+
+        $io->writeln("fetching remote object for '$user->username' ($user->apProfileId)");
+        $actorObject = $this->apHttpClient->getActorObject($url);
+        if (!$actorObject) {
+            if (!$dryRun) {
+                $this->purgeUser($user);
```
blued-gear, line 327: This and the next purge should get a log what and why the user got purged.

src/Command/CheckDuplicatesUsersMagazines.php
```diff
@@ -196,4 +219,333 @@ private function deleteEntities(SymfonyStyle $io, string $entity, array $ids): i
...
+                continue;
+            }
+
+            if ($dryRun) {
+                $io->writeln("Would have merged '$user->username' ('$user->apProfileId') into '$mainUser->username' ('$mainUser->apProfileId')");
+                continue;
+            }
+
+            $io->writeln("Merging '$user->username' ('$user->apProfileId') into '$mainUser->username' ('$mainUser->apProfileId')");
+            $conn = $this->entityManager->getConnection();
```
blued-gear, line 412: I think this should be wrapped in a transaction.

PR #2053, opened 2026-03-15
Supress some logs

blued-gear:
The deprecation `User Deprecated: Since symfony/http-foundation 7.4: Request::get() is deprecated` can be safely ignored as the function is not terminally deprecated (for now) and also not malfunctioning. The message about this deprecation is spamming the logs, especially on dev as it is triggered for every usage in every request.

Migrating away from the function would be time-consuming, as we use it a lot and kinda rely on it current behavior.


melroy89:
The changelog says indeed: 

> Deprecate method `Request::get()`, use properties `->attributes`, `query` or `request` directly instead

We could change the `get` called with `input` maybe as well..? I like to consider fixing in the correct way.

blued-gear:
What do you mean with `input`?

The problem is that at some locations our API is not clearly stating if the expected parameters should be in the query, body, ... . So migrating away from get(), which checks all possible sources, might be hard.

melroy89:
> What do you mean with `input`?

see: https://github.com/laravel/framework/issues/57962#issuecomment-3598076755 :)

But that might only works with Laravel too bad :( 

---

But yea,  I get your point, we now need to migrate to, which isn't always easy to know what to pick from the code:

```php

    // use this if the value comes from route placeholders or custom attributes
    $value = $request->attributes->get('some_key');

    // use this for GET query parameters
    $value = $request->query->get('some_key');

    // use this for POST-submitted data
    $value = $request->request->get('some_key');
```

blued-gear:
> But that might only works with Laravel too bad :(

Yep, there is no `input` function in Symfonys request.

PR #2054, opened 2026-03-15
show boosts for followed users

blued-gear:
This PR adds the option for a user to get also content (threads, posts, thread comments, post comments) which was boosted by users they follow in their feed (/sub) and API responses.

**Important:** some API changes are breaking \
(@jwr1)


config/mbin_routes/combined_api.yaml
```diff
@@ -6,8 +6,8 @@ api_combined:
 
 api_combined_user:
   controller: App\Controller\Api\Combined\CombinedRetrieveApi::userCollection
-  path: /api/combined/{contentType}
+  path: /api/combined/{collectionType}
   requirements:
-    contentType: subscribed|moderated|favourited
-  methods: [ PUT ]
+    collectionType: subscribed|moderated|favourited
+  methods: [ GET ]
```
melroy89, line 12: yup I agree here..

BentiGorlich:
Since your change is a path parameter, I don't think that it is a breaking change

blued-gear:
> Since your change is a path parameter, I don't think that it is a breaking change

The breaking part is the change from `PUT` to `GET`.

And partly the edit of `ContentResponseDto` as some deserializers don't like new unknown fields.

PR #2063, opened 2026-03-24
Restore compatibility with Mastodon

BentiGorlich:
We didn't really do anything wrong, our actors and activities are compliant, however, Mastodon determines whether it does support an actor/activity by the default `https://www.w3.org/ns/activitystreams` being present in the `@context` variable. In order to fix this quickly and still keep the json-ld compliance, I just moved the 2 default URLs into the referenced contexts.

The error in the log looks like this (I couldn't find it in my production log, only on my dev instance): 

> Truncated content: {"error":"Unsupported JSON-LD context for document URL_HERE"}.

This incompatibility was introduced in #1997

### DIFF
--- a/src/Service/ActivityPub/ContextsProvider.php
+++ b/src/Service/ActivityPub/ContextsProvider.php
@@ -17,8 +17,6 @@ class ContextsProvider
     public static function embeddedContexts(): array
     {
         return [
-            ActivityPubActivityInterface::CONTEXT_URL,
-            ActivityPubActivityInterface::SECURITY_URL,
             [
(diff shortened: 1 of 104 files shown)


blued-gear:
Did you also create an Issue at Mastodons repo?

BentiGorlich:
No I didn't because I think it is highly unlikely that they'll fix it. It seems to me like very intended behaviour. Should I do it anyway?

blued-gear:
It doesn't hurt and costs nothing even if it just lays there unanswered (except for a few minutes). I can do it too if you don't have time currently.

PR #2064, opened 2026-03-25
Improve performance of RemoveRemoteMediaCommand

BentiGorlich:
- for some reason doctrine takes a long time to compute the changeset in this case. Use a native statement instead that is a lot faster

### DIFF
--- a/assets/controllers/infinite_scroll_controller.js
+++ b/assets/controllers/infinite_scroll_controller.js
@@ -22,9 +22,19 @@ export default class extends Controller {
         try {
             this.loadingValue = true;
 
-            const paginationElem = this.paginationTarget.getElementsByClassName('pagination__item--current-page')[0].nextElementSibling;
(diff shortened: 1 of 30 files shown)


src/Service/ImageManager.php
```diff
@@ -388,10 +388,8 @@ public function removeCachedImage(MbinImage $image): bool
         try {
             $this->publicUploadsFilesystem->delete($image->filePath);
             $this->imagineCacheManager->remove($image->filePath);
-            $image->filePath = null;
-            $image->downloadedAt = null;
-            $this->entityManager->persist($image);
-            $this->entityManager->flush();
+            $sql = 'UPDATE image SET file_path = NULL, downloaded_at = NULL WHERE id = :id';
```
blued-gear, line 391: the two properties should still be set to null, in case code somewhere else uses them for checks or so

PR #2073, opened 2026-03-31
Add ability to filter feeds, comments and the profile by key-words

### LINKED ISSUE #1210: Filter titles against user keywords
An often requested user feature is keyword filtering. This would be an attractive addition to users deciding on a platform to join. 

Filtering threads/posts by title might be a good place to start.  

This would require a user-definable value in Settings to store keywords. Perhaps a comma separated list.
 A sane max count of words should exist. 

Modification to page load queries to exclude matches to keywords in titles.

BentiGorlich:
- filtering in ContentRepository, EntryCommentRepository, EntryComment (for nested filtering), PostCommentRepository, PostComment (for nested filtering) and SearchRepository (users public activity)
- new UI to configure the filter lists
- new JS controller for CollectionType form fields (to dynamically add a field)

Closes #1210

The new UI looks like this:

Bildschirmfoto vom 2026-03-31 13-39-14

Bildschirmfoto vom 2026-03-31 13-39-04


assets/controllers/form_collection_controller.js
```diff
@@ -0,0 +1,19 @@
+// assets/controllers/form-collection_controller.js
```
blued-gear: Is this comment necessary?

assets/controllers/form_collection_controller.js
```diff
@@ -0,0 +1,19 @@
...
+
+import { Controller } from '@hotwired/stimulus';
+
+export default class extends Controller {
```
blued-gear, line 4: Controllers which are needed on only some pages can be annotated with `/* stimulusFetch: 'lazy' */` (right above the `export default ...`) to be fetched for only these pages.

src/Controller/Api/User/UserFilterListApi.php
```diff
@@ -0,0 +1,222 @@
...
+            new OA\Header(header: 'X-RateLimit-Remaining', description: 'Number of requests left until you will be rate limited', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Retry-After', description: 'Unix timestamp to retry the request after', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Limit', description: 'Number of requests available', schema: new OA\Schema(type: 'integer')),
+        ],
+        content: new OA\JsonContent(
+            properties: [
+                new OA\Property(
+                    property: 'items',
+                    type: 'array',
+                    items: new OA\Items(ref: new Model(type: UserFilterListDto::class))
```
blued-gear: Did you mean `UserFilterListResponseDto`?

src/Controller/Api/User/UserFilterListApi.php
```diff
@@ -0,0 +1,222 @@
...
+
+    #[OA\Response(
+        response: 200,
+        description: 'Filter list created',
+        headers: [
+            new OA\Header(header: 'X-RateLimit-Remaining', description: 'Number of requests left until you will be rate limited', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Retry-After', description: 'Unix timestamp to retry the request after', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Limit', description: 'Number of requests available', schema: new OA\Schema(type: 'integer')),
+        ],
+        content: new Model(type: UserFilterListDto::class)
```
blued-gear: Did you mean `UserFilterListResponseDto`?

src/Controller/Api/User/UserFilterListApi.php
```diff
@@ -0,0 +1,222 @@
...
+        content: new OA\JsonContent(ref: new Model(type: TooManyRequestsErrorSchema::class))
+    )]
+    #[OA\RequestBody(content: new Model(type: UserFilterListDto::class))]
+    #[OA\Tag(name: 'user')]
+    #[Security(name: 'oauth2', scopes: ['user:profile:edit'])]
+    #[IsGranted('ROLE_OAUTH2_USER:PROFILE:EDIT')]
+    #[IsGranted(FilterListVoter::EDIT, 'list')]
+    public function edit(
+        RateLimiterFactoryInterface $apiUpdateLimiter,
+        #[MapEntity(class: UserFilterList::class)] UserFilterList $list,
```
blued-gear: Why is the `class:` specified? Shouldn't Symfony deduce it from the parameter type?

src/Controller/Api/User/UserFilterListApi.php
```diff
@@ -0,0 +1,222 @@
...
+        $freshList = $this->entityManager->getRepository(UserFilterList::class)->find($list->getId());
+
+        return new JsonResponse(
+            $this->serializeFilterList($freshList),
+            headers: $headers
+        );
+    }
+
+    #[OA\Response(
+        response: 200,
```
blued-gear: As this doesn't respond with any content the code should be 204.

src/Security/Voter/FilterListVoter.php
```diff
@@ -0,0 +1,49 @@
...
+                [self::EDIT, self::DELETE],
+                true
+            );
+    }
+
+    protected function voteOnAttribute(string $attribute, $subject, TokenInterface $token): bool
+    {
+        $user = $token->getUser();
+
+        if (!$user instanceof User) {
```
blued-gear, line 31:
```suggestion
        if (!($user instanceof User)) {
```

The `!$user` would convert $user to a boolean before the instanceof.

blued-gear: ... now I'm wondering how many bugs of this kind we might have in our codebase

src/DTO/UserFilterListResponseDto.php
```diff
@@ -0,0 +1,51 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\DTO;
+
+use App\Entity\UserFilterList;
+
+class UserFilterListResponseDto implements \JsonSerializable
```
blued-gear, line 11: This misses `#[OA\Schema()]`

src/Entity/EntryComment.php
```diff
@@ -267,7 +267,36 @@ public function containsBannedHashtags(): bool
...
+                continue;
+            }
+
+            foreach ($list->words as $word) {
+                if ($word['exactMatch']) {
+                    if (str_contains($this->body, $word['word'])) {
+                        return true;
+                    }
+                } else {
+                    if (str_contains(strtolower($this->body), strtolower($word['word']))) {
```
blued-gear: This creates two temporary strings for every comparison. If comments x filters is small, this is negligible. If not, then you might consider evaluating if the check can be replaced with [mb_stripos](https://www.php.net/manual/en/function.mb-stripos.php) (or `stripos`).

src/Entity/EntryComment.php
```diff
@@ -267,7 +267,36 @@ public function containsBannedHashtags(): bool
         return false;
     }
 
-    public function getChildrenByCriteria(MbinCriteria $entryCommentCriteria, DownvotesMode $downvoteMode): array
+    /**
+     * @param 'profile'|'comments' $filterRealm
+     */
+    public function containsFilteredWords(User $loggedInUser, string $filterRealm): bool
```
blued-gear, line 273: The code is fine. But personally I'm not a fan of putting even more functionality into an entity (IMO entities should be just used for storing data). Maybe creating a service for list checks and other management operations is a better fit.

src/Entity/EntryComment.php
```diff
@@ -267,7 +267,36 @@ public function containsBannedHashtags(): bool
         return false;
     }
 
-    public function getChildrenByCriteria(MbinCriteria $entryCommentCriteria, DownvotesMode $downvoteMode): array
+    /**
+     * @param 'profile'|'comments' $filterRealm
```
blued-gear, line 271: Why can the realm be `profile` if just the comment body is checked?

src/Entity/EntryComment.php
```diff
@@ -307,11 +336,11 @@ public function getChildrenByCriteria(MbinCriteria $entryCommentCriteria, Downvo
 
                 break;
             case MbinCriteria::SORT_OLD:
-                uasort($children, fn (EntryComment $a, EntryComment $b) => ArrayUtils::numCompareDescending($a->createdAt->getTimestamp(), $b->createdAt->getTimestamp()));
```
blued-gear, line 310: I think I remember an issue saying that the order of some of Mbins sortings is wrong. Is this change because of this?

src/Entity/UserFilterList.php
```diff
@@ -0,0 +1,72 @@
...
+
+    public function getId(): int
+    {
+        return $this->id;
+    }
+
+    /**
+     * @return string[]
+     */
+    public function getLocationStrings(): array
```
blued-gear: As the values are referenced in other functions as "realm", this should be kept consistent.

templates/user/settings/filter_lists_form.html.twig
```diff
@@ -0,0 +1,30 @@
...
+{{ form_row(form.name, {label: 'name'}) }}
+<div class="form-control">
+    {{ form_label(form.expirationDate, 'expiration_date') }}
+    {{ form_widget(form.expirationDate, {'attr': {'class': 'form-control', 'style': 'padding: 1rem .5rem;', 'title': 'expiration_date'|trans}}) }}
+</div>
+
+<div>{{ 'filter_lists_where_to_filter'|trans }}:</div>
+{{ form_row(form.feeds) }}
+{{ form_row(form.comments) }}
+{{ form_row(form.profile) }}
```
blued-gear: Initially I thought the 'profile' realm will filter content from profiles which had certain words in their profile-description. Then I read the SearchRepository. So I think this option needs a help text.

And maybe 'feed' too (hinting that it affects entries and posts).

src/Repository/PostCommentRepository.php
```diff
@@ -190,6 +195,32 @@ private function filter(QueryBuilder $qb, Criteria $criteria)
...
+            if (!$list->comments) {
+                continue;
+            }
+
+            foreach ($list->words as $word) {
+                if ($word['exactMatch']) {
+                    $qb->andWhere("NOT (c.body LIKE :word$i) or c.user = :filterUser")
+                        ->setParameter("word$i", '%'.$word['word'].'%');
+                } else {
+                    $qb->andWhere("NOT (lower(c.body) LIKE lower(:word$i)) or c.user = :filterUser")
```
blued-gear, line 211: in the SearchRepository you used `ILIKE`, why not here?

src/Repository/ContentRepository.php
```diff
@@ -304,6 +305,54 @@ private function getQueryAndParameters(Criteria $criteria, bool $addCursor): arr
...
+                        }
+                    }
+                    if ($filterList->feeds || ($filterList->comments && ($includeEntryComments || $includePostComments))) {
+                        $parameters["word$i"] = '%'.$word.'%';
+                    }
+                    ++$i;
+                }
+            }
+            if (\sizeof($listOrsEntry) > 0) {
+                $filterClauseEntry = 'NOT ('.implode(' OR ', $listOrsEntry).') OR c.user_id = :loggedInUser';
```
blued-gear, line 346: Maybe the `OR c.user_id = :loggedInUser` can be extracted as a general condition for every content select, as I think the users own content should be always visible (except when comments are generally not requested and such).

templates/components/filter_list.html.twig
```diff
@@ -0,0 +1,33 @@
+<div class="filter-list">
+    <div class="section">
```
blued-gear, line 2: If it is not to much work, an indicator for expired lists would be nice.

src/Security/Voter/FilterListVoter.php
BentiGorlich: are you sure this behaves this way?

src/Entity/EntryComment.php
BentiGorlich: Yes exactly. My test failed and then I discovered this

src/Entity/EntryComment.php
BentiGorlich: Because 'profile' means on the profile page. And comments are there too. Currently it is not used on profile. I just thought that this method will never be used when viewing feeds, but it might be used on the other realms

src/Repository/PostCommentRepository.php
BentiGorlich: because doctrine does not accept it...

src/Repository/ContentRepository.php
BentiGorlich: That would be a lot of work. Maybe create an issue for that and we can do that separately from this PR

src/Security/Voter/FilterListVoter.php
BentiGorlich: So I have tried it out on https://php-play.dev/ and as far as I can tell it works without the brackets as well:

```php
<?php
class User {}

$u = new stdClass();

if (!$u instanceof User){
    echo "no";
} else {
    echo "yes";
}
```

outputs "no"

And 
```php
<?php
class User {}

$u = new User();

if (!$u instanceof User){
    echo "no";
} else {
    echo "yes";
}
```

outputs "yes"

src/Entity/EntryComment.php
BentiGorlich: Since it is called from within EntryComment there is not really a way to pull only this method out of the Entity. I could try to pull the whole `getChildrenByCriteria` method out of the entity, but that would increase the scope of this issue...

src/Entity/EntryComment.php
blued-gear: Then you should mention the other issue in this PR so that it will be closed too.

src/Security/Voter/FilterListVoter.php
blued-gear: Right, my bad. I think I messed something up when trying it out in an online PHP interpreter.

blued-gear (approved):

PR #2074, opened 2026-03-31
fix error when searching for a non-fedi URL

blued-gear:


src/Service/SearchManager.php
```diff
@@ -132,8 +132,12 @@ public function findActivityPubActorsOrObjects(string $handleOrUrl): array
             // the url could resolve to a different id.
             try {
                 $body = $this->apHttpClient->getActivityObject($url);
-                $apId = $body['id'];
-                $objects = $this->findByApId($apId);
+                if (null !== $body) {
+                    $apId = $body['id'];
```
BentiGorlich, line 136: is this guaranteed to be set when body is not null? Or does this need another isset check or something?

blued-gear: The `id` *should* be present in every AP object, but making sure will not hurt.

BentiGorlich: I thought that maybe the answer might just be a random json and then you'd get an array but it is not clear what is in this array. If it is indeed an AP object, then yes it has to have an id, but you don't know

PR #2077, opened 2026-04-03
fix: handle array-typed url field in ActivityPub actor payload

melroy89:
Some ActivityPub servers (e.g. [PeerTube](https://joinpeertube.org/)) return the 'url' field as an array of Link objects rather than a plain string, which caused a TypeError when assigning to `User::$apPublicUrl (?string)`.

~~When 'url' is an array, filter to only 'text/html' typed links (falling back to all links if none match), then pick the **last** entry's 'href' as the canonical profile URL.~~

Get the shortest path urls in case the `url` is an array as the canonical profile URL.

A final `is_string()` guard ensures a string is always assigned, which was also discussed already on Matrix.

Example payload from the remote server:

```json
[
  {
    "type": "Link",
    "mediaType": "text/html",
    "href": "https://tinkerbetter.tube/a/peertube/video-channels"
  },
  {
    "type": "Link",
    "mediaType": "text/html",
    "href": "https://tinkerbetter.tube/a/peertube"
  },
  {
    "type": "Link",
    "mediaType": "text/html",
    "href": "https://tinkerbetter.tube/accounts/peertube"
  }
]
```


blued-gear:
> filter to only 'text/html' typed links

I don't think this is the right way, as the media type can also be one of the AP mediatypes.

> then pick the last

This is only the fitting link for Peertube, but here a more universal approach is needed. How about picking the URL with the fewest path elements?

melroy89:
> > filter to only 'text/html' typed links
> I don't think this is the right way, as the media type can also be one of the AP mediatypes.

This is in user update function.. (part of ` private function updateUser(string $actorUrl): ?User`), if media type is present, I rather want to retrieve the url link (aka `text/html`) in case of an user link. Instead of any other mediatype.

>> then pick the last
> This is only the fitting link for Peertube, but here a more universal approach is needed. How about picking the URL with the fewest path elements?

As far as I know, Peertube is in fact the only one that use an array in the first place here. But also picking the url with the fewest path elements is an assumption that might also be incorrect. I dunno what the best thing to do here.

Also the shortest path in this case is getting the Peertube client url.

melroy89:
Anyhow.. Pushed.

blued-gear (approved):

PR #2084, opened 2026-04-04
Update npm package minor/patch versions only

melroy89:
- Update npm package minor/patch releases only
- Only ran: `npm update`
- **NO MAJOR** version updates.

Running & tested at kbin.melroy.org


melroy89:
CI/CD is failing due to github cache I think. Since `npm ci` works fine locally as well as on my server.

image

melroy89:
> CI/CD is failing due to github cache I think. Since `npm ci` works fine locally as well as on my server.
> image

It was down-grading some packages, my bad. I fixed it now!

PR #2095, opened 2026-04-11
Fix `Create` -> `Announce` -> `Deliver` circle

BentiGorlich:
It seems like we have a bug where a magazine can infinitely re-`Announce` a `Create`-activity in some specific circumstances. To avoid that this PR introduces a new check for existing `Announce` activities of the same object by the same actor.

### DIFF
--- a/src/MessageHandler/ActivityPub/Outbox/AnnounceHandler.php
+++ b/src/MessageHandler/ActivityPub/Outbox/AnnounceHandler.php
@@ -24,6 +24,7 @@ use App\Service\ActivityPubManager;
 use App\Service\DeliverManager;
 use App\Service\SettingsManager;
 use Doctrine\ORM\EntityManagerInterface;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\Messenger\Attribute\AsMessageHandler;
 use Symfony\Component\Messenger\Exception\UnrecoverableMessageHandlingException;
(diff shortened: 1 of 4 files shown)


src/MessageHandler/ActivityPub/Outbox/AnnounceHandler.php
```diff
@@ -81,6 +83,16 @@ public function doWork(MessageInterface $message): void
...
             }
+            $alreadySentActivities = $this->activityRepository->findAllActivitiesByTypeObjectAndActor('Announce', $object, $actor);
+            if (!$message->removeAnnounce && \sizeof($alreadySentActivities) > 0) {
+                $this->logger->info('[AnnounceHandler::doWork] not sending announcing {object}, because it is not an Undo and the same actor (magazine {magazine}) already announced the Create activity {create}', [
+                    'object' => \get_class($object)." \"{$object->getShortTitle()}\"",
+                    'magazine' => $actor->name,
+                    'create' => $createActivity->uuid,
+                ]);
+
+                return;
```
melroy89, line 94: return or `UnrecoverableMessageHandlingException`?

BentiGorlich: Imo the "this was successful" state fits with having announced it already

melroy89: I mean a return could mean it will do a retry. Later.

BentiGorlich: No it will not. The retry logic is triggered when an Exception is thrown (with the exception of the `UnrecoverableMessageHandlingException`). A return is the same as if the method would just run to the end and means a success

blued-gear (approved):

PR #2102, opened 2026-04-20
fix handling of special unicode in user title

blued-gear:
Filter out all Unicode special chars so that https://discuss.online/u/m_f does not break the UI anymore.

### DIFF
--- a/src/Twig/Extension/UserExtension.php
+++ b/src/Twig/Extension/UserExtension.php
@@ -20,6 +20,7 @@ class UserExtension extends AbstractExtension
             new TwigFunction('get_instance_of_user', [UserExtensionRuntime::class, 'getInstanceOfUser']),
             new TwigFunction('get_user_attitude', [UserExtensionRuntime::class, 'getUserAttitude']),
             new TwigFunction('is_instance_of_user_banned', [UserExtensionRuntime::class, 'isInstanceOfUserBanned']),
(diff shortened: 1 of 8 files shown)


src/Twig/Extension/UserExtension.php
```diff
@@ -20,6 +20,7 @@ public function getFunctions(): array
             new TwigFunction('get_instance_of_user', [UserExtensionRuntime::class, 'getInstanceOfUser']),
             new TwigFunction('get_user_attitude', [UserExtensionRuntime::class, 'getUserAttitude']),
             new TwigFunction('is_instance_of_user_banned', [UserExtensionRuntime::class, 'isInstanceOfUserBanned']),
+            new TwigFunction('get_user_title', [UserExtensionRuntime::class, 'getTitle']),
```
BentiGorlich, line 23: I think it would be better if it was a filter and not a function 🤔

blued-gear: At first, I designed it as a filter for `user.title` but Twig has no (easy) concept for `someNullableVar|filter ?? alternativeVar`. \
Also the function approach has the advantage that it makes the used text consistent on all occurrences (I saw for example that on some location the apPreferredNamse got forgotten) and easy to change in the future as the logic is at a central location.

BentiGorlich (approved):

PR #2105, opened 2026-04-21
only enable includeBoosts if sorting is set to newest

blued-gear:
As boost might include comments, this makes it incompatible with cursor pagination for some sortings, so restrict it to `Newest` as it only makes sense mostly there anyway.

### DIFF
--- a/src/Controller/Api/Post/PostsRetrieveApi.php
+++ b/src/Controller/Api/Post/PostsRetrieveApi.php
@@ -359,7 +359,7 @@ class PostsRetrieveApi extends PostsBaseApi
 
     #[OA\Response(
         response: 200,
-        description: 'A paginated list of posts from user\'s subscribed magazines',
+        description: 'A paginated list of posts from user\'s subscribed magazines (boosted content will only be included if sort is set to "newest")',
         content: new OA\JsonContent(
             type: 'object',
             properties: [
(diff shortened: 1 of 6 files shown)


src/Controller/Api/Post/PostsRetrieveApi.php
```diff
@@ -442,12 +442,6 @@ public function subscribed(
         in: 'query',
         schema: new OA\Schema(type: 'string', default: Criteria::AP_ALL, enum: Criteria::AP_OPTIONS)
     )]
-    #[OA\Parameter(
-        name: 'includeBoosts',
-        description: 'if true then boosted content from followed users are included',
-        in: 'query',
-        schema: new OA\Schema(type: 'boolean', default: false)
-    )]
```
BentiGorlich, line 450: Shouldn't the parameter still stay? Like it is with the other entry points?

blued-gear: The function does not have this parameter and the endpoint is `subscribedWithBoosts`, so it would make no sense to have an option to disable the inclusion.

BentiGorlich: Ah ok, yeah that makes sense

BentiGorlich (approved):

PR #2106, opened 2026-04-21
update PHP and JS dependencies

blued-gear:
I just update all dependencies which were possible.

As our tests do not cover everything (especially not the JS stuff), please review this PR and maybe run a test deployment on your instance.

### DIFF
--- a/config/packages/doctrine.yaml
+++ b/config/packages/doctrine.yaml
@@ -31,6 +31,7 @@ doctrine:
         identity_generation_preferences:
             Doctrine\DBAL\Platforms\PostgreSQLPlatform: identity
         auto_mapping: false
+        enable_native_lazy_objects: true
         controller_resolver:
             auto_mapping: false
         mappings:
--- a/composer.json
+++ b/composer.json
@@ -125,7 +125,7 @@
         "fakerphp/faker": "^1.23.1",
         "justinrainbow/json-schema": "^6.0.0",
         "phpstan/phpstan": "^2.0.2",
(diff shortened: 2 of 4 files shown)


config/packages/doctrine.yaml
```diff
@@ -31,6 +31,7 @@ doctrine:
         identity_generation_preferences:
             Doctrine\DBAL\Platforms\PostgreSQLPlatform: identity
         auto_mapping: false
+        enable_native_lazy_objects: true
```
blued-gear, line 34: was there a reason this setting had not been enabled when Doctrine got updated?

blued-gear:
As we use PHP 8.4 in our Docker images, should we set 8.4 in the composer file too?

blued-gear:
The security scan is failing because the Docker image uses a PHP version which is too old.

@melroy89 please update https://github.com/MbinOrg/psalm-github-security-scan to use the current PHP version (and maybe also what was committed upstream).

melroy89:
> @melroy89 please update https://github.com/MbinOrg/psalm-github-security-scan to use the current PHP version (and maybe also what was committed upstream).

Its now using PHP 8.5, right?


I re-enabled the watch workflow (again): https://github.com/MbinOrg/psalm-github-security-scan/actions/workflows/watch.yml. latest docker image is now 6.16.1: https://github.com/MbinOrg/psalm-github-security-scan/pkgs/container/psalm-security-scan

blued-gear:
Superseded by #2162

PR #2108, opened 2026-04-23
Fix type error on mention extraction from tags

### LINKED ISSUE #2107: Bug: ActivityPub processing fails with `as:Public` and string `to`/`cc`
**Describe the bug**

bin.pivox.net

1.10.0rc1

**To Reproduce**

**Expected behavior**

**Screenshots**
N/A (log-based issue)

 - OS: OS: Ubuntu24.04LTS
 - Browser: Firefox
 - Browser Version: 149

**Additional context**
Errors observed:

```text id="q8k1pd"
```

```text id="v3kz7a"
```

BentiGorlich:
- ignore something in the tag array if it is not an array or is missing one of the required fields `type`, `name` or `href`

#2107

### DIFF
--- a/src/Service/ActivityPub/MarkdownConverter.php
+++ b/src/Service/ActivityPub/MarkdownConverter.php
@@ -11,6 +11,7 @@ use App\Service\MentionManager;
 use App\Service\TagExtractor;
 use League\HTMLToMarkdown\Converter\TableConverter;
 use League\HTMLToMarkdown\HtmlConverter;
+use Psr\Log\LoggerInterface;
 
 class MarkdownConverter
 {
@@ -18,6 +19,7 @@ class MarkdownConverter
         private readonly TagExtractor $tagExtractor,
         private readonly MentionManager $mentionManager,
         private readonly ActivityPubManager $activityPubManager,
+        private readonly LoggerInterface $logger,
     ) {
     }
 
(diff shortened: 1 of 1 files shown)


src/Service/ActivityPub/MarkdownConverter.php
```diff
@@ -72,6 +72,9 @@ private function findMentionFromTag(array $match, array $apTags): array
     {
         $res = [];
         foreach ($apTags as $tag) {
+            if (!\is_array($tag) || !isset($tag['type']) || !isset($tag['name']) || !isset($tag['href'])) {
+                continue;
```
blued-gear, line 80: Instead of just skipping the item, we should log the content (as a warning) to notice if a new type of tag goes around or so.

BentiGorlich: will do :+1:

src/Service/ActivityPub/MarkdownConverter.php
```diff
@@ -72,6 +74,11 @@ private function findMentionFromTag(array $match, array $apTags): array
     {
         $res = [];
         foreach ($apTags as $tag) {
+            if (!\is_array($tag) || !isset($tag['type']) || !isset($tag['name']) || !isset($tag['href'])) {
+                $this->logger->warning('Ignoring tag (not an array or missing "name", "type" or "href": {t}', ['t' => $tag]);
```
blued-gear:
```suggestion
                $this->logger->warning('Ignoring tag (not an array or missing "name", "type" or "href"): {t}', ['t' => $tag]);
```

blued-gear (approved):

PR #2109, opened 2026-04-23
Treat `as:Public` or just `Public` correctly -> the same as `https://www.w3.org/ns/activitystreams#Public`

### LINKED ISSUE #2107: Bug: ActivityPub processing fails with `as:Public` and string `to`/`cc`

BentiGorlich:
- according to https://www.w3.org/TR/activitypub/#x5-6-public-addressing `as:Public` and `Public` should be treated the same as `https://www.w3.org/ns/activitystreams#Public`
- add constants for the other public URL string
- ignore the new constants the same as the PUBLIC_URL constant
- treat the new constants the same as PUBLIC_URL in getting the visibility

#2107

### DIFF
--- a/src/Entity/Contracts/ActivityPubActivityInterface.php
+++ b/src/Entity/Contracts/ActivityPubActivityInterface.php
@@ -8,14 +8,16 @@ use App\Entity\User;
 
 interface ActivityPubActivityInterface
 {
-    public const FOLLOWERS = 'followers';
-    public const FOLLOWING = 'following';
-    public const INBOX = 'inbox';
-    public const OUTBOX = 'outbox';
-    public const CONTEXT = 'context';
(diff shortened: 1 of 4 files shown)


blued-gear (approved):

blued-gear:
If I'm not mistaken then `as:Public` uses a namespace identifier `as` which expands to `https://www.w3.org/ns/activitystreams#`. So if a similar issue comes up for a different URL in the future, we should try to implement proper namespace expansion.

PR #2111, opened 2026-04-23
Fix support for `image` property on `Page` type, add support for string

### LINKED ISSUE #2107: Bug: ActivityPub processing fails with `as:Public` and string `to`/`cc`

BentiGorlich:
- When a `Page` ActivityPub object contains only an `image` property, but no `attachment` property our code threw an error
- Add support for either `attachment` or `image` property to contain only a string (Ghost does this for example)

With this change an AP-Object like this: https://www.shoptechblog.com/.ghost/activitypub/article/38919605-5b90-4b0b-ad3c-5991a889855c
will properly work and get the image attached

Also connection to #2107

### DIFF
--- a/src/Service/ActivityPub/Page.php
+++ b/src/Service/ActivityPub/Page.php
@@ -91,7 +91,7 @@ class Page extends ActivityPubContent
             $dto->title = $object['name'];
             $dto->apId = $object['id'];
 
-            if ((isset($object['attachment']) || isset($object['image'])) && $image = $this->activityPubManager->handleImages($object['attachment'])) {
+            if ((isset($object['attachment']) || isset($object['image'])) && $image = $this->activityPubManager->handleImages($object['attachment'] ?? $object['image'])) {
                 $this->logger->debug("adding image to entry '{title}', {image}", ['title' => $dto->title, 'image' => $image->getId()]);
                 $dto->image = $this->imageFactory->createDto($image);
             }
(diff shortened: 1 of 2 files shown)


src/Service/ActivityPub/Page.php
```diff
@@ -91,7 +91,7 @@ public function create(array $object, bool $stickyIt = false): Entry
             $dto->title = $object['name'];
             $dto->apId = $object['id'];
 
-            if ((isset($object['attachment']) || isset($object['image'])) && $image = $this->activityPubManager->handleImages($object['attachment'])) {
+            if ((isset($object['attachment']) || isset($object['image'])) && $image = $this->activityPubManager->handleImages($object['attachment'] ?? $object['image'])) {
```
blued-gear, line 94: This should also be applied to `Note`

BentiGorlich: In Note we do not support the `image` property. I could add that, but I am not sure that we should...

blued-gear: Ah, alright. I wasn't sure if it was forgotten to handle `image` in the Note code.

blued-gear (approved):

PR #2113, opened 2026-04-23
Initial poll implementation

### LINKED ISSUE #647: Implement polls like they have at mastodon
This example says it all

https://kzoo.to/@jeze/112170632611266384

This is what we get on mbin

https://rimworld.gallery/m/debug/t/787/Elley-Smith-vote-at-jeze-at-kzoo-to

BentiGorlich:
- add the ability to add a poll to Entry, EntryComment, Post and PostComment
- Polls:
  - can be multiple choice
  - can contain an unlimited number of choices
  - can expire
- Add a hidden form for the poll to all create forms, the form appears when a checkbox is checked
- The ActivityPub implementation is the same as on Mastodon: https://docs.joinmastodon.org/spec/activitypub/#Question
- Add tests for the ActivityPub implementation and the Api controller
- The api had to be shifted a bit -> the `serialize...` functions in the base api take only the entity classes now instead of their dto variant -> touched a lot of api classes because of this

I will add images when this moves out of the draft phase :)

Closes #647

### DIFF
--- a/assets/styles/components/_poll.scss
+++ b/assets/styles/components/_poll.scss
@@ -1,0 +1,24 @@
+.poll {
+  margin: auto;
+
+  form {
+    margin: auto;
+    max-width: 400px;
+  }
+
+  .poll-result {
+    margin: .5rem 0;
+  }
+
+  .choice-result {
+    margin-bottom: .5rem;
+
+    .choice-bar {
+      background-color: var(--kbin-button-primary-bg);
+      color: var(--kbin-button-primary-text-color);
+      height: .5rem;
+      border: var(--kbin-button-primary-border);
+      border-radius: var(--kbin-rounded-edges-radius);
+    }
+  }
+}
--- a/config/mbin_routes/poll.yaml
+++ b/config/mbin_routes/poll.yaml
@@ -1,0 +1,4 @@
+poll_vote:
+  controller: App\Controller\PollVoteController::vote
+  path: /poll/{id}/vote
+  methods: [GET]
--- a/config/mbin_routes/poll.yaml
+++ b/config/mbin_routes/poll.yaml
@@ -2,3 +2,8 @@ poll_vote:
   controller: App\Controller\PollVoteController::vote
   path: /poll/{id}/vote
   methods: [GET]
+
+poll_refresh:
+    controller: App\Controller\PollVoteController::refreshVoteCounts
+    path: /poll/{id}/refresh
+    methods: [GET]
--- a/src/Command/DebugCommand.php
+++ b/src/Command/DebugCommand.php
@@ -1,0 +1,33 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Command;
+
+use App\Repository\PollRepository;
+use Symfony\Component\Console\Attribute\AsCommand;
+use Symfony\Component\Console\Command\Command;
+use Symfony\Component\Console\Input\InputInterface;
+use Symfony\Component\Console\Output\OutputInterface;
+
+#[AsCommand('mbin:debug')]
+class DebugCommand extends Command
+{
+    public function __construct(
+        private readonly PollRepository $pollRepository,
+    ) {
+        parent::__construct();
+    }
+
+    protected function execute(InputInterface $input, OutputInterface $output): int
+    {
+        foreach ($this->pollRepository->getAllEndedPollsToSentNotifications() as $poll) {
+            $output->writeln("poll {$poll->getId()}");
+            foreach ($this->pollRepository->getAllLocalVotersOfPoll($poll) as $voter) {
+                $output->writeln("Voter $voter->username in poll {$poll->getId()}");
+            }
+        }
+
+        return 0;
+    }
+}
--- a/src/Controller/Api/Entry/Comments/EntryCommentsCreateApi.php
+++ b/src/Controller/Api/Entry/Comments/EntryCommentsCreateApi.php
@@ -118,11 +118,10 @@ class EntryCommentsCreateApi extends EntriesBaseApi
 
         // Rate limiting already taken care of
         $comment = $manager->create($dto, $this->getUserOrThrow(), rateLimit: false);
-        $dto = $factory->createDto($comment);
         $dto->parent = $parent;
 
         return new JsonResponse(
-            $this->serializeEntryComment($dto, $this->tagLinkRepository->getTagsOfContent($comment)),
+            $this->serializeEntryComment($comment, $this->tagLinkRepository->getTagsOfContent($comment)),
             status: 201,
             headers: $headers
         );
@@ -227,11 +226,10 @@ class EntryCommentsCreateApi extends EntriesBaseApi
 
         // Rate limiting already taken care of
         $comment = $manager->create($dto, $this->getUserOrThrow(), rateLimit: false);
-        $dto = $factory->createDto($comment);
         $dto->parent = $parent;
 
         return new JsonResponse(
-            $this->serializeEntryComment($dto, $this->tagLinkRepository->getTagsOfContent($comment)),
+            $this->serializeEntryComment($comment, $this->tagLinkRepository->getTagsOfContent($comment)),
             status: 201,
             headers: $headers
         );
--- a/src/Controller/Api/Poll/PollVoteController.php
+++ b/src/Controller/Api/Poll/PollVoteController.php
@@ -1,0 +1,302 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Controller\Api\Poll;
+
+use App\Controller\Api\BaseApi;
+use App\DTO\PollResponseDto;
+use App\Entity\Entry;
+use App\Entity\EntryComment;
+use App\Entity\Poll;
+use App\Entity\Post;
+use App\Entity\PostComment;
+use App\Entity\User;
+use App\Schema\Errors\BadRequestErrorSchema;
+use App\Schema\Errors\NotFoundErrorSchema;
+use App\Schema\Errors\TooManyRequestsErrorSchema;
+use App\Schema\Errors\UnauthorizedErrorSchema;
+use App\Service\PollManager;
+use Nelmio\ApiDocBundle\Attribute\Model;
+use Nelmio\ApiDocBundle\Attribute\Security;
+use OpenApi\Attributes as OA;
+use Symfony\Bridge\Doctrine\Attribute\MapEntity;
+use Symfony\Component\HttpFoundation\JsonResponse;
+use Symfony\Component\HttpFoundation\Request;
+use Symfony\Component\HttpKernel\Exception\BadRequestHttpException;
+use Symfony\Component\HttpKernel\Exception\NotFoundHttpException;
+use Symfony\Component\RateLimiter\RateLimiterFactoryInterface;
+use Symfony\Component\Security\Http\Attribute\IsGranted;
+
+class PollVoteController extends BaseApi
+{
+    #[OA\Response(
+        response: 200,
+        description: 'voted on poll',
+        headers: [
+            new OA\Header(header: 'X-RateLimit-Remaining', description: 'Number of requests left until you will be rate limited', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Retry-After', description: 'Unix timestamp to retry the request after', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Limit', description: 'Number of requests available', schema: new OA\Schema(type: 'integer')),
+        ],
+        content: new Model(type: PollResponseDto::class)
+    )]
+    #[OA\Response(
+        response: 400,
+        description: 'Poll was not valid. Possibly: poll already ended, choices do not exist, user already voted',
+        content: new OA\JsonContent(ref: new Model(type: BadRequestErrorSchema::class))
+    )]
+    #[OA\Response(
+        response: 401,
+        description: 'Permission denied due to missing or expired token',
+        content: new OA\JsonContent(ref: new Model(type: UnauthorizedErrorSchema::class))
+    )]
+    #[OA\Response(
+        response: 404,
+        description: 'Poll not found',
+        content: new OA\JsonContent(ref: new Model(type: NotFoundErrorSchema::class))
+    )]
+    #[OA\Response(
+        response: 429,
+        description: 'You are being rate limited',
+        headers: [
+            new OA\Header(header: 'X-RateLimit-Remaining', description: 'Number of requests left until you will be rate limited', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Retry-After', description: 'Unix timestamp to retry the request after', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Limit', description: 'Number of requests available', schema: new OA\Schema(type: 'integer')),
+        ],
+        content: new OA\JsonContent(ref: new Model(type: TooManyRequestsErrorSchema::class))
+    )]
+    #[OA\Parameter(
+        name: 'choices',
+        description: 'The user\'s voting choices',
+        in: 'query',
+        schema: new OA\Schema(type: 'array', items: new OA\Items(type: 'string')),
+    )]
+    #[OA\Tag(name: 'entry')]
+    #[OA\Tag(name: 'poll')]
+    #[Security(name: 'oauth2', scopes: ['entry:vote'])]
+    #[IsGranted('ROLE_OAUTH2_ENTRY:VOTE')]
+    public function voteOnEntry(
+        #[MapEntity(mapping: ['entryId' => 'id'])] Entry $entry,
+        Request $request,
+        PollManager $pollManager,
+        RateLimiterFactoryInterface $apiVoteLimiter,
+    ): JsonResponse {
+        $headers = $this->rateLimit($apiVoteLimiter);
+        $poll = $entry->poll;
+        if (null === $poll) {
+            throw new NotFoundHttpException();
+        }
+
+        return $this->voteOnPoll($request, $pollManager, $this->getUserOrThrow(), $entry, $poll, $headers);
+    }
+
+    #[OA\Response(
+        response: 200,
+        description: 'voted on poll',
+        headers: [
+            new OA\Header(header: 'X-RateLimit-Remaining', description: 'Number of requests left until you will be rate limited', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Retry-After', description: 'Unix timestamp to retry the request after', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Limit', description: 'Number of requests available', schema: new OA\Schema(type: 'integer')),
+        ],
+        content: new Model(type: PollResponseDto::class)
+    )]
+    #[OA\Response(
+        response: 400,
+        description: 'Poll was not valid. Possibly: poll already ended, choices do not exist, user already voted',
+        content: new OA\JsonContent(ref: new Model(type: BadRequestErrorSchema::class))
+    )]
+    #[OA\Response(
+        response: 401,
+        description: 'Permission denied due to missing or expired token',
+        content: new OA\JsonContent(ref: new Model(type: UnauthorizedErrorSchema::class))
+    )]
+    #[OA\Response(
+        response: 404,
+        description: 'Poll not found',
+        content: new OA\JsonContent(ref: new Model(type: NotFoundErrorSchema::class))
+    )]
+    #[OA\Response(
+        response: 429,
+        description: 'You are being rate limited',
+        headers: [
+            new OA\Header(header: 'X-RateLimit-Remaining', description: 'Number of requests left until you will be rate limited', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Retry-After', description: 'Unix timestamp to retry the request after', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Limit', description: 'Number of requests available', schema: new OA\Schema(type: 'integer')),
+        ],
+        content: new OA\JsonContent(ref: new Model(type: TooManyRequestsErrorSchema::class))
+    )]
+    #[OA\Parameter(
+        name: 'choices',
+        description: 'The user\'s voting choices',
+        in: 'query',
+        schema: new OA\Schema(type: 'array', items: new OA\Items(type: 'string')),
+    )]
+    #[OA\Tag(name: 'entry_comment')]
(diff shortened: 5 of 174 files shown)


BentiGorlich:
Open things:
- This does not yet send an `Update` activity upon a new vote, which I though Mastodon also doesn't do, but there is an fep for this that was also mentioned in the issue and it says that an implementation should do it: https://codeberg.org/silverpill/feps/src/branch/main/9967/fep-9967.md
- it does not yet use the `closed` property, which it should do if `endTime` is not provided
- the Poll AP object does not yet have an `updated` field. 
- All in all some AP stuff has to be done
- one cannot yet manually refresh the results
- one cannot yet jump to the results without voting

BentiGorlich:
As I implemented all the remaining things a while back and now tested it on my server in combination with mastodon, I think this is ready

src/Factory/EntryFactory.php
```diff
@@ -42,10 +45,18 @@ public function createFromDto(EntryDto $dto, User $user): Entry
...
+    public function createResponseDto(Entry $entry, array $tags, ?array $crosspostedEntries = null): EntryResponseDto
     {
-        $dto = $entry instanceof Entry ? $this->createDto($entry) : $entry;
+        $dto = $this->createDto($entry);
         $badges = $dto->badges ? array_map(fn (Badge $badge) => $this->badgeFactory->createDto($badge), $dto->badges->toArray()) : null;
+        $pollDto = null;
+        if ($dto->addPoll) {
+            $user = $this->security->getUser();
+            $pollDto = new PollResponseDto();
+            $pollDto->voterCount = $entry->poll->voterCount;
```
melroy89, line 56: The response fills in the voter count and choices, but never sets `endDate`. API clients therefore get `null` for a poll that has an expiry in the database. Please copy the poll end date into `pollDto` here, as the post and comment responses do.

src/Service/PollManager.php
```diff
@@ -0,0 +1,264 @@
...
+
+        $choices = $object['anyOf'] ?? $object['oneOf'];
+        $choiceNames = [];
+        foreach ($choices as $choice) {
+            if (array_find($choiceNames, fn (string $choiceName) => $choiceName === $choice['name'])) {
+                // do not create duplicate choices
+                continue;
+            }
+            $pollChoice = new PollChoice();
+            $pollChoice->poll = $poll;
```
melroy89, line 188: This sets the owning side, but the in-memory `$poll->choices` collection is never updated. The poll is used immediately after import, so it can render with no choices until a reload. Also, `$choiceNames` is never appended to, so the duplicate check never works. Please add each choice to the collection and record its name.

melroy89:
Just some small minor remarks;

- The `refresh()` workaround adds a DB round-trip and only populates `choices` lazily while the EM is open; directly appending to the collection would be more robust (note line 80 already uses `$poll->choices[] in createPoll`).
- `tests/Unit/Service/PollManagerTest.php` boots the kernel/DB, it's a functional test under `Unit/`...

assets/styles/components/_poll.scss
```diff
@@ -0,0 +1,24 @@
+.poll {
+  margin: auto;
```
blued-gear, line 2: Why centered instead of left-aligned?

src/Command/DebugCommand.php
```diff
@@ -0,0 +1,33 @@
+<?php
```
blued-gear, line 1: This looks like dev code which should not be in prod.

src/Controller/Api/Entry/Comments/EntryCommentsCreateApi.php
```diff
@@ -118,11 +118,10 @@ public function __invoke(
 
         // Rate limiting already taken care of
         $comment = $manager->create($dto, $this->getUserOrThrow(), rateLimit: false);
-        $dto = $factory->createDto($comment);
         $dto->parent = $parent;
```
blued-gear, line 121: Whats up with this line? You removed the creation of the `dto` variable, but still use it here.

src/Controller/Api/Entry/Comments/EntryCommentsCreateApi.php
```diff
@@ -227,11 +226,10 @@ public function uploadImage(
...
```
blued-gear, line 229: same here

src/Controller/Api/Poll/PollVoteController.php
```diff
@@ -0,0 +1,302 @@
...
+        headers: [
+            new OA\Header(header: 'X-RateLimit-Remaining', description: 'Number of requests left until you will be rate limited', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Retry-After', description: 'Unix timestamp to retry the request after', schema: new OA\Schema(type: 'integer')),
+            new OA\Header(header: 'X-RateLimit-Limit', description: 'Number of requests available', schema: new OA\Schema(type: 'integer')),
+        ],
+        content: new Model(type: PollResponseDto::class)
+    )]
+    #[OA\Response(
+        response: 400,
+        description: 'Poll was not valid. Possibly: poll already ended, choices do not exist, user already voted',
```
blued-gear, line 45: > user already voted

maybe return an HTTP "conflict" status instead of "bad request"

src/Controller/Api/Poll/PollVoteController.php
```diff
@@ -0,0 +1,302 @@
...
+        }
+
+        if ($comment->post->getId() !== $post->getId()) {
+            throw new BadRequestHttpException();
+        }
+
+        return $this->voteOnPoll($request, $pollManager, $this->getUserOrThrow(), $comment, $poll, $headers);
+    }
+
+    public function voteOnPoll(Request $request, PollManager $pollManager, User $user, Entry|EntryComment|Post|PostComment $content, Poll $poll, array $headers): JsonResponse
```
blued-gear, line 283: this should be `private`

src/Controller/Api/Post/PostsBaseApi.php
```diff
@@ -106,10 +106,6 @@ protected function deserializePostCommentFromForm(?PostCommentDto $dto = null):
      */
     protected function serializePostCommentTree(?PostComment $comment, PostCommentPageView $commentPageView, ?int $depth = null): array
     {
-        if (null === $comment) {
```
blued-gear, line 109: Without this check a NullPointerException can occur. If $comment will never be null, please remove the nullable type.

src/Controller/Entry/EntryCreateController.php
```diff
@@ -50,6 +52,8 @@ public function __construct(
         private readonly ValidatorInterface $validator,
         private readonly IpResolver $ipResolver,
         private readonly Security $security,
+        private readonly PollManager $pollManager,
+        private readonly EntityManagerInterface $entityManager,
```
blued-gear, line 56: Are these two properties used in this class? (GH only shows me 3 changes: the added imports, the added properties and `$dto->addEmptyChoices();`).

Same for the other classes.

src/Controller/Post/Comment/PostCommentCreateController.php
```diff
@@ -106,6 +106,7 @@ public function __invoke(
     private function getForm(Post $post, ?PostComment $parent): FormInterface
     {
         $dto = new PostCommentDto();
+        $dto->addEmptyChoices();
```
blued-gear, line 109: As this seems to be used at every Dto creation, why not invoke it in the constructor?

src/Controller/PollVoteController.php
```diff
@@ -0,0 +1,88 @@
...
+    {
+        if (null === $poll->getSubject()->apId) {
+            throw new BadRequestHttpException('Cannot refresh the vote counts of a local poll');
+        }
+
+        $this->apHttpClient->invalidateActivityObjectCache($poll->getSubject()->apId);
+        $object = $this->apHttpClient->getActivityObject($poll->getSubject()->apId);
+        if ($this->pollManager->hasPollProperties($object)) {
+            $this->pollManager->updatePollCounts($poll, $object);
+        }
```
blued-gear, line 68: This looks more like a job for the Manager, not for a Controller.

src/DTO/ContentWithPollDto.php
```diff
@@ -0,0 +1,29 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\DTO;
+
+use App\DTO\Contracts\PollDtoTrait;
+
+class ContentWithPollDto
```
blued-gear, line 9: This should be an `abstract class`.

src/DTO/PollResponseDto.php
```diff
@@ -0,0 +1,42 @@
...
+namespace App\DTO;
+
+use App\Entity\Poll;
+use App\Entity\PollChoice;
+use App\Entity\User;
+use Nelmio\ApiDocBundle\Attribute\Model;
+use OpenApi\Attributes as OA;
+use Symfony\Component\Security\Core\User\UserInterface;
+
+class PollResponseDto implements \JsonSerializable
```
blued-gear, line 14: Maybe this needs `#[OA\Schema]`, but I'm not sure.

src/Entity/Poll.php
```diff
@@ -0,0 +1,145 @@
...
+            ];
+        }
+
+        return $result;
+    }
+
+    public function updateVoterCount(): void
+    {
+        if ($this->isRemote) {
+            ++$this->voterCount;
```
blued-gear, line 131: I'm not sure if I like this design. The function name sounds like it recounts the existing voters (idempotent), not that it increments the count on each call.

src/Entity/PollVote.php
```diff
@@ -0,0 +1,52 @@
...
+    #[ManyToOne(targetEntity: User::class)]
+    public User $voter;
+
+    #[JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    #[ManyToOne(targetEntity: PollChoice::class)]
+    public PollChoice $choice;
+
+    #[JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    #[ManyToOne(targetEntity: Poll::class)]
+    public Poll $poll;
```
blued-gear, line 41: `inversedBy` are missing

@melroy89 you found a command to validate entity definitions, can you remember it?

src/EventSubscriber/PollEventSubscriber.php
```diff
@@ -0,0 +1,65 @@
...
+    }
+
+    public function onPollVote(PollVoteEvent $event): void
+    {
+        if ($event->poll->isRemote && null === $event->voter->apId) {
+            // remote poll, local user -> send the vote
+            foreach ($event->poll->votes as $vote) {
+                $this->bus->dispatch(new PollVoteMessage($vote->uuid->toString()));
+            }
+        } elseif (!$event->poll->isRemote) {
```
blued-gear, line 42: comment says 'remote poll' but condition is "!isRemote"

src/Factory/ActivityPub/EntryPageFactory.php
```diff
@@ -48,7 +49,7 @@ public function create(Entry $entry, array $tags, bool $context = false): array
 
         $page = array_merge($page ?? [], [
             'id' => $this->getActivityPubId($entry),
-            'type' => 'Page',
+            'type' => $entry->poll ? 'Question' : 'Page',
```
blued-gear, line 52: How will this affect compatibility with other thread-oriented Fedi platforms, which have no poll support?

Is it a possibility to keep the Page type and detect an Entry with a Poll by the existence of typical poll attributes?

src/Factory/ActivityPub/PollVoteFactory.php
```diff
@@ -0,0 +1,57 @@
...
+        private readonly PostCommentRepository $postCommentRepository,
+        private readonly PersonFactory $personFactory,
+        private readonly ApActivityRepository $apActivityRepository,
+    ) {
+    }
+
+    public function build(PollVote $vote, bool $includeContext = true): array
+    {
+        $actorUrl = $this->personFactory->getActivityPubId($vote->getUser());
+        $content = $this->entryRepository->findOneBy(['poll' => $vote->poll])
```
blued-gear, line 33: why not use `$vote->poll->getSubject()`?

src/Factory/EntryCommentFactory.php
```diff
@@ -38,9 +41,18 @@ public function createFromDto(EntryCommentDto $dto, User $user): EntryComment
...
-        $dto = $comment instanceof EntryComment ? $this->createDto($comment) : $comment;
+        $dto = $this->createDto($comment);
+        $pollDto = null;
+        if ($dto->addPoll) {
+            $pollDto = new PollResponseDto();
+            $pollDto->endDate = $dto->pollEndsAt;
+            $pollDto->voterCount = $comment->poll->voterCount;
+            $user = $this->security->getUser();
+            $pollDto->currentUserHasVoted = $user instanceof User ? $comment->poll->hasUserVoted($user) : null;
+            $pollDto->choices = $comment->poll->choices ? array_map(fn (PollChoice $choice) => PollChoiceResponseDto::createFromPollChoice($choice, $user), $comment->poll->choices->toArray()) : null;
```
blued-gear, line 54: This is repetitive. As the DTOs extend `ContentWithPollDto`, this snippet can be extracted into its own factory function.

src/Factory/EntryCommentFactory.php
```diff
@@ -38,9 +41,18 @@ public function createFromDto(EntryCommentDto $dto, User $user): EntryComment
         );
     }
 
-    public function createResponseDto(EntryCommentDto|EntryComment $comment, array $tags, int $childCount = 0): EntryCommentResponseDto
+    public function createResponseDto(EntryComment $comment, array $tags, int $childCount = 0): EntryCommentResponseDto
     {
-        $dto = $comment instanceof EntryComment ? $this->createDto($comment) : $comment;
+        $dto = $this->createDto($comment);
+        $pollDto = null;
+        if ($dto->addPoll) {
```
blued-gear, line 48: I can't see where `addPoll` is set inside `createDto()`

src/Factory/PostCommentFactory.php
```diff
@@ -122,8 +136,8 @@ public function createDto(PostComment $comment): PostCommentDto
```
blued-gear, line 136: this misses the poll attributes

src/Factory/PostFactory.php
blued-gear, line 109: this misses the poll attributes

src/Form/EntryType.php
```diff
@@ -85,6 +86,30 @@ public function buildForm(FormBuilderInterface $builder, array $options): void
...
+            ])
+            ->add('choices', CollectionType::class, [
+                'entry_type' => TextType::class,
+                'allow_add' => true,
+                'allow_delete' => true,
+                'required' => false,
+                'attr' => [
+                    'class' => 'existing-collection-items',
+                ],
+                'label' => 'poll_choices',
```
blued-gear, line 111: As a suggestion, this repetitive part of the form types could be extracted into a common function, which configures a given instance.

src/Message/CheckPollEndedMessage.php
```diff
@@ -0,0 +1,11 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\Message;
+
+use App\Message\Contracts\AsyncMessageInterface;
+
+class CheckPollEndedMessage implements AsyncMessageInterface
```
blued-gear, line 9: As this process all existing polls, it should be named `CheckPollsEndedMessage`.

src/Entity/Poll.php
```diff
@@ -0,0 +1,145 @@
...
+    public bool $multipleChoice = false;
+
+    /**
+     * @var int the number of individual users voting on this poll
+     */
+    #[Column(options: ['default' => 0])]
+    public int $voterCount = 0;
+
+    #[Column(type: Types::DATETIMETZ_IMMUTABLE)]
+    public \DateTimeImmutable $endDate;
```
blued-gear: As all polls are periodically checked if they are ended, I might be good to add a column `hasEnded` + an index on it, so that the query for CheckPollEndedMessage can skip all already ended polls. Otherwise, the message processing will get progressively slower with the number of polls in the DB.

src/MessageHandler/ActivityPub/Inbox/UpdateHandler.php
```diff
@@ -220,6 +252,17 @@ private function extractChanges(EntryDto|EntryCommentDto|PostDto|PostCommentDto
         if (isset($payload['object']['commentsEnabled']) && \is_bool($payload['object']['commentsEnabled']) && ($dto instanceof EntryDto || $dto instanceof PostDto)) {
             $dto->isLocked = !$payload['object']['commentsEnabled'];
         }
+        $this->logger->debug('[UpdateHandler::extractChanges] content was the same: {x}', ['x' => $isContentSame ? 'true' : 'false']);
+
+        if ($this->pollManager->hasPollProperties($payload['object'])) {
+            $pollIsSame = $this->pollManager->extractPollChanges($payload['object'], $dto);
```
blued-gear: `extractPollChanges()` would fit better in UpdateHandler, as it is the only location where it is used and the handlers are strongly coupled with AP stuff, whereas Managers should be kept more high-level.

src/Entity/Poll.php
```diff
@@ -0,0 +1,145 @@
...
+
+    #[Column]
+    public bool $isRemote = false;
+
+    #[Column]
+    public bool $sentNotifications = false;
```
blued-gear, line 45: This column should get an index, as otherwise the search for newly ended polls will do a full-table-scan.

src/Service/ActivityPub/Note.php
```diff
@@ -291,4 +317,48 @@ private function createPostComment(array $object, ActivityPubActivityInterface $
...
+            }
+            if ($actor->isDeleted || $actor->isSoftDeleted() || $actor->isTrashed()) {
+                throw new UserDeletedException();
+            }
+
+            $choice = $object['name'];
+            $existingVotes = $poll->getUserVotes($actor);
+            $existingSameVote = array_filter($existingVotes, fn (PollVote $vote) => $vote->choice->name === $choice);
+            if (\sizeof($existingSameVote) > 0) {
+                throw new \LogicException("User $actor->username has already voted on the poll {$poll->getId()} with choice '$choice'.");
```
blued-gear, line 346: I can already see the errors of non-confirming software which resends all votes when the user changes their vote 😅 \
But this should not be an issue for now.

src/Service/NotificationManager.php
```diff
@@ -21,6 +22,7 @@ public function __construct(
         private readonly MessageNotificationManager $messageNotificationManager,
         private readonly EntityManagerInterface $entityManager,
         private readonly MagazineBanNotificationManager $magazineBanNotificationManager,
+        private readonly PollRepository $pollRepository,
```
blued-gear, line 25: Why is this property added, when it is not used?

src/Service/PollManager.php
```diff
@@ -0,0 +1,266 @@
...
+            throw new PollHasEndedException('Poll has already ended');
+        }
+
+        $voteEntities = [];
+        foreach (array_unique($choices) as $choice) {
+            $choiceEntity = $poll->findChoice($choice);
+            $vote = new PollVote();
+            $vote->voter = $user;
+            $vote->poll = $poll;
+            $vote->choice = $choiceEntity;
```
blued-gear, line 115: (personal preference) as these properties are mandatory, they should be enforced by a (static) constructor function.

src/Service/PollManager.php
```diff
@@ -0,0 +1,266 @@
...
+        if ($poll->hasUserVoted($user) && !$allowMultipleChoices) {
+            throw new \LogicException('Already voted');
+        }
+        if ($poll->hasEnded()) {
+            throw new PollHasEndedException('Poll has already ended');
+        }
+
+        $voteEntities = [];
+        foreach (array_unique($choices) as $choice) {
+            $choiceEntity = $poll->findChoice($choice);
```
blued-gear, line 111: This does not check if the choice is a valid one. As this function can be called with AP originating user content, it must be checked.

I suggest the check should throw an exception on invalid input. Then the check in the Controllers can be replaced with a try-catch which rethrow as a fitting Http-Exception.

src/Service/PollManager.php
```diff
@@ -0,0 +1,266 @@
...
+        $this->entityManager->refresh($poll);
+        $poll->updateVoterCount();
+        $this->entityManager->flush();
+
+        $this->eventDispatcher->dispatch(new PollVoteEvent($poll, $content, $user, $voteEntities));
+    }
+
+    public function getContentOfPoll(Poll $poll): Entry|EntryComment|Post|PostComment|null
+    {
+        return $this->entryRepository->findOneBy(['poll' => $poll])
```
blued-gear, line 132: why not `poll->getSubject()`?

src/Service/PollManager.php
```diff
@@ -0,0 +1,266 @@
...
+
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    /**
+     * Call PollManager::canCreateFromApObject() first, this that $object contains all the necessary information.
```
blued-gear, line 167: this <-> that

src/Service/PollManager.php
```diff
@@ -0,0 +1,266 @@
...
+            $poll->choices[] = $pollChoice;
+        }
+        $this->entityManager->persist($poll);
+        $this->entityManager->flush();
+
+        return $poll;
+    }
+
+    /**
+     * Call PollManager::canCreateFromApObject() first, this that $payload contains all the necessary information.
```
blued-gear, line 202: this <-> that

> Call PollManager::canCreateFromApObject() first, to ensure that $payload contains all the necessary information.

tests/Functional/ActivityPub/Inbox/CreateHandlerTest.php
```diff
@@ -266,6 +327,39 @@ public function testPostWithPublicShortURL(): void
...
+        self::assertNotNull($entry->poll);
+        self::assertEquals(1, $entry->poll->voterCount);
+        self::assertNotNull($entry->poll->findChoice('A'));
+        self::assertEquals(1, $entry->poll->findChoice('A')->voteCount);
+        self::assertNotNull($entry->poll->findChoice('B'));
+        self::assertEquals(1, $entry->poll->findChoice('B')->voteCount);
+        self::assertNotNull($entry->poll->findChoice('C'));
+        self::assertEquals(1, $entry->poll->findChoice('C')->voteCount);
+    }
+
```
blued-gear, line 362: Some negative tests would be nice. Like activities with invalid choices, multiple votes for the same choice by the same user.

blued-gear:
Is it planned that incoming votes to local polls will be federated out via Announces?

config/mbin_routes/poll.yaml
```diff
@@ -0,0 +1,9 @@
+poll_vote:
+  controller: App\Controller\PollVoteController::vote
+  path: /poll/{id}/vote
+  methods: [GET]
```
melroy89, line 4: Because voting is a one-time state change, exposing it as GET lets another site make a logged-in user vote by sending them to a crafted link. Please switch this route and the poll form to POST and validate a CSRF token.

PR #2115, opened 2026-04-23
Harmonized layout for thread_comments in combined view

blued-gear:
With this, comments to threads in the Combined view (e.g. when boosted) have a layout which fits the other components.

demo

Question: IMO the `to <magazine name>` and `in <entry title>` should be swapped for entry_comment_combined and entry_comment. What do you think?


BentiGorlich:
Maybe the `to <magazine name>` should be after the author like in the other things?

src/Twig/Components/EntryCommentCombinedComponent.php
```diff
@@ -0,0 +1,51 @@
...
+                $this->comment
+            )
+            && $this->canSeeTrash) {
+            return true;
+        }
+
+        $this->comment->image = null;
+
+        return false;
+    }
```
BentiGorlich: Not that relevant for this PR, but could this maybe be moved into a baseclass?

templates/components/entry_comment_combined.html.twig
```diff
@@ -0,0 +1,185 @@
...
+                {% set hasLang = comment.lang is not same as app.request.locale and comment.lang is not same as kbin_default_lang() %}
+                {% set isModDeleted = comment.visibility is same as 'trashed' %}
+                {% set isUserDeleted = comment.visibility is same as 'soft_deleted' %}
+
+                <header>
+                    {% if hasTitle %}
+                        <small>
+                            {% if isAdult %}<small class="badge danger">18+</small>{% endif %}
+
+                            {{ 'to'|trans }} {{ component('magazine_inline', {magazine: comment.magazine, fullName: SHOW_MAGAZINE_FULLNAME is same as V_TRUE}) }}
```
BentiGorlich: I would move that down to the author line as it is with entries

src/Twig/Components/EntryCommentCombinedComponent.php
blued-gear: Do we have tests for the canSeeTrashed?

BentiGorlich: I am not sure. We have a `testApiCanRetrieveMagazineTrash` for magazine things, so maybe a bit of coverage there

blued-gear: I'm not sure if properties from superclasses work as expected in components (`$this->canSeeTrash`). So I thought maybe a test will catch it.

src/Twig/Components/AbstractSubjectComponent.php
```diff
@@ -0,0 +1,49 @@
...
+use App\Entity\Entry;
+use App\Entity\EntryComment;
+use App\Entity\Post;
+use App\Entity\PostComment;
+use Symfony\Component\Security\Core\Authorization\AuthorizationCheckerInterface;
+
+abstract class AbstractSubjectComponent
+{
+    public function __construct(
+        private readonly AuthorizationCheckerInterface $authorizationChecker,
```
BentiGorlich: maybe use protected instead?

src/Twig/Components/AbstractSubjectComponent.php
```diff
@@ -0,0 +1,49 @@
...
+    ) {
+    }
+
+    private Entry|EntryComment|Post|PostComment $subject;
```
BentiGorlich: same here

composer.lock
BentiGorlich, line 1: how come you updated phpseclib? :D

BentiGorlich: ah composer audit, got it :D

BentiGorlich (approved):

PR #2121, opened 2026-04-29
Add the second cursor to the query

BentiGorlich:
- We have a second cursor to resolve "ties" in values, this one was missing from the default feeds (it is present in the magazine feeds)
- Add a different inner limit for cursored queries so the outer query can filter things away and still have enough items to fill a page

### DIFF
--- a/src/Controller/Entry/EntryFrontController.php
+++ b/src/Controller/Entry/EntryFrontController.php
@@ -44,6 +44,7 @@ class EntryFrontController extends AbstractController
         ?string $type,
         Request $request,
         #[MapQueryParameter] ?string $cursor = null,
+        #[MapQueryParameter] ?string $cursor2 = null,
     ): Response {
         $user = $this->getUser();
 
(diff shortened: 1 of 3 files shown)


blued-gear:
What is the reason for the triple inner limit?

BentiGorlich:
> I found a second possible fix: we have a nested query that has the same limit as the outer query. The outer query might filter stuff out which makes the page not have the predicted 25 items and therefore is not offering the next page...
The main reason this might happen is because a user has apDiscoverable = false which will get filtered on the /all but not in the /sub feed

Tripling is just more leeway. What would you think would be a reasonable limit? In the end the inner limit is mostly there to force postgresql to use the index, right?

blued-gear:
> In the end the inner limit is mostly there to force postgresql to use the index, right?

Yes, this was also my observation. `* 3` is a heuristic and can fail, but I can't give a better value. The (in my mind) mathematical correct inner limit would be `<sum items of pervious pages (without filter)> + <page size>`, but this would be near impossible to implement.

blued-gear (approved):

BentiGorlich:
Just for documentation purposes: yes there is still a race condition in there. That is not really solvable as long as we will keep the old pagination style around / in the same file. The problem is the filtering of the outer query, but we moved it there because it gave quite a bit of a performance boost to the query (see #1892)

PR #2127, opened 2026-05-04
Add a session handler to migrate from pdo to redis, add managed gc times

BentiGorlich:
- symfony uses the php default garbage collector probabilities, but that is set to 0 in debian and ubuntu, so manage it ourselves so the sessions get cleared out
 - add a service definition for redis, which the redis session handler needs, this includes splitting up the dsn in our .env file
 - add a `MigratingSessionHandler` that writes all sessions to the RedisSessionHandler as well, so we can move towards that

> [!WARNING]
> To run this PR the `REDIS_DNS` variable in the `.env` file must be split so there are separate variables for `REDIS_HOST` and `REDIS_PORT`


BentiGorlich:
As a note, we moved to the PdoSessionHandler in #1145. Before that we used our redis dsn as the `handler_id`. I have no idea why that worked at all, though I will be testing the redis sessions out starting tomorrow when all sessions have been duplicated in redis (that is what the `MigratingSessionHandler` does)

blued-gear:
Please don't forget the `.env.devcontainer`.

blued-gear:
As it runs well on some servers so far, do you think this is ready for review?

config/services.yaml
```diff
@@ -289,3 +289,18 @@ services:
...
+
+    Redis:
+        class: Redis
+        calls:
+            - connect:
+                - '%env(REDIS_HOST)%'
+                - '%env(int:REDIS_PORT)%'
+
+            # uncomment the following if your Redis server requires a password
+            - auth:
```
blued-gear: it says 'uncomment the following' but it is already not commented; please change the comment or comment-out the code

.env.example
```diff
@@ -89,7 +89,9 @@ TRUSTED_PROXIES=
 
 # Redis
```
blued-gear, line 90: These changes need to be applied to the .devcontainer env file too.

blued-gear (changes requested):

blued-gear (approved):

melroy89:
This change breaks my socket line, where I solely relied on `REDIS_DNS` string:

```sh
REDIS_DNS=redis:///var/run/valkey-2/valkey.sock
```

And now causing issues, since suddenly `REDIS_HOST` and `REDIS_PORT` are required!? But what if I just want to use a Linux socket.

I will try to follow this PR with: https://github.com/MbinOrg/mbin/pull/2218

PR #2133, opened 2026-05-06
configurable language for fulltext-search

### LINKED ISSUE #2043: Full-text search is hardcoded to English, breaking search for non-English instances
## Summary

The full-text search (FTS) implementation hardcodes the `'english'` PostgreSQL text search dictionary in both the tsvector generated columns (migrations) and implicitly in `SearchRepository.php`. This causes severely degraded search quality on any instance where the primary content language is not English, as stemming and stop-word removal are incorrect for the actual content.

## Current behavior

On a Spanish-language instance:

- **Stop words are wrong**: Spanish articles/prepositions (`de`, `la`, `el`, `un`, `contra`) are indexed as regular tokens instead of being filtered out, diluting relevance scores.
- **Stemming is wrong**: The English stemmer does not reduce Spanish words to their correct roots (e.g., `comunicaciones` is not stemmed to match `comunicación`; `buscando` does not match `buscar`).
- **Ranking is flat**: Because stemming produces poor tokens, `ts_rank_cd()` scores collapse — in our instance, searching for "guerra" returns 276 results where 270 have the exact same score (0.100), making the `ORDER BY rank DESC` effectively useless. Results appear in near-random order.

## Expected behavior

FTS should use a configurable language dictionary so that:
- Stop words for the instance's language are correctly filtered
- Stemming matches the content language (e.g., Spanish stemmer for Spanish content)
- `ts_rank_cd()` produces meaningful score differentiation
- Search results are ordered by actual relevance

### 1. Migrations hardcode `'english'`

All 10 tsvector generated columns use `to_tsvector('english', ...)`:

**`migrations/Version20230615203020.php`** (entry, post, comments):
```php
$this->addSql("ALTER TABLE entry ADD COLUMN title_ts tsvector GENERATED ALWAYS AS (to_tsvector('english', title)) STORED;");
$this->addSql("ALTER TABLE entry ADD COLUMN body_ts tsvector GENERATED ALWAYS AS (to_tsvector('english', body)) STORED;");
$this->addSql("ALTER TABLE post ADD COLUMN body_ts tsvector GENERATED ALWAYS AS (to_tsvector('english', body)) STORED;");
$this->addSql("ALTER TABLE post_comment ADD COLUMN body_ts tsvector GENERATED ALWAYS AS (to_tsvector('english', body)) STORED;");
$this->addSql("ALTER TABLE entry_comment ADD COLUMN body_ts tsvector GENERATED ALWAYS AS (to_tsvector('english', body)) STORED;");
```

**`migrations/Version20250706115844.php`** (magazine, user):
```php
$this->addSql('ALTER TABLE magazine ADD name_ts tsvector GENERATED ALWAYS AS (to_tsvector(\'english\', name)) STORED');
$this->addSql('ALTER TABLE magazine ADD title_ts tsvector GENERATED ALWAYS AS (to_tsvector(\'english\', title)) STORED');
$this->addSql('ALTER TABLE magazine ADD description_ts tsvector GENERATED ALWAYS AS (to_tsvector(\'english\', description)) STORED');
$this->addSql('ALTER TABLE "user" ADD username_ts tsvector GENERATED ALWAYS AS (to_tsvector(\'english\', username)) STORED');
$this->addSql('ALTER TABLE "user" ADD about_ts tsvector GENERATED ALWAYS AS (to_tsvector(\'english\', about)) STORED');
```

### 2. SearchRepository uses implicit language

**`src/Repository/SearchRepository.php`** calls `plainto_tsquery(:query)` without a language parameter:

```sql
2 * ts_rank_cd(e.title_ts, plainto_tsquery(:query)) + ts_rank_cd(e.body_ts, plainto_tsquery(:query)) as rank
...
WHERE (e.body_ts @@ plainto_tsquery(:query) = true OR e.title_ts @@ plainto_tsquery(:query) = true)
```

This relies on PostgreSQL's `default_text_search_config` server setting, which usually happens to be `english` — so it works by coincidence, not by design. If the server config ever differs from the column dictionary, searches would silently return zero results (mismatched stems).

### 3. Both sides must match

This is the critical detail: the tsvector columns and the tsquery **must** use the same dictionary. If only one side is changed (e.g., an admin changes `default_text_search_config` to `spanish` without recreating the columns), the stems won't match and search returns nothing.

## Proposed fix

Add a configurable environment variable (e.g., `MBIN_FTS_LANGUAGE`) that controls the FTS dictionary, defaulting to `'english'` for backward compatibility.

### Migration side

Create a new migration that:
1. Reads the configured language (or a DB-level setting)
2. Drops and recreates the tsvector generated columns with the configured dictionary
3. Recreates the GIN indexes

Alternatively, use `'simple'` dictionary as a language-agnostic default (no stemming, no stop words — works for all languages but with reduced quality), or use PostgreSQL's `default_text_search_config` setting instead of hardcoding.

### Query side (SearchRepository.php)

Update all `plainto_tsquery(:query)` calls to explicitly pass the configured language:

```php
// Before:
plainto_tsquery(:query)

// After (option A — use server default explicitly):
plainto_tsquery(current_setting('default_text_search_config'), :query)

// After (option B — use env var):
plainto_tsquery('$ftsLanguage', :query)
```

### Minimal change (option C)

At minimum, the migration could use `'simple'` instead of `'english'`, and SearchRepository could use `plainto_tsquery('simple', :query)`. This removes the English bias without requiring per-instance configuration, at the cost of no stemming for any language.

## Affected tables and columns

| Table | Column | Source field |
|-------|--------|-------------|
| `entry` | `title_ts` | `title` |
| `entry` | `body_ts` | `body` |
| `entry_comment` | `body_ts` | `body` |
| `post` | `body_ts` | `body` |
| `post_comment` | `body_ts` | `body` |
| `magazine` | `name_ts` | `name` |
| `magazine` | `title_ts` | `title` |
| `magazine` | `description_ts` | `description` |
| `user` | `username_ts` | `username` |
| `user` | `about_ts` | `about` |

## Workaround

Instance admins can manually fix this at the DB level:

```sql
-- 1. Change server default
ALTER DATABASE "your_db" SET default_text_search_config = 'pg_catalog.spanish';

-- 2. Recreate each column (example for entry.title_ts)
ALTER TABLE entry DROP COLUMN title_ts;
ALTER TABLE entry ADD COLUMN title_ts tsvector GENERATED ALWAYS AS (to_tsvector('spanish', COALESCE(title, ''))) STORED;

-- 3. Recreate GIN index
CREATE INDEX entry_title_ts_idx ON entry USING GIN (title_ts);

-- Repeat for all 10 columns...
```

This workaround survives Mbin upgrades because existing migrations check for column existence before adding them (idempotent).

## Environment

- Mbin version: v1.9.0
- PostgreSQL: 17
- Instance language: Spanish
- Tested on: tardigram.com

blued-gear:
closes #2043

### DIFF
--- a/.env.example_docker
+++ b/.env.example_docker
@@ -36,6 +36,8 @@ KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 MBIN_DEFAULT_THEME=default
+# Language used to parse content and search-expressions for ranking
+MBIN_SEARCH_LANGUAGE=english
 # Set the max image file size (in bytes)
 # This should be set to <= `upload_max_filesize` and `post_max_size` in the server's php.ini file
 MBIN_MAX_IMAGE_BYTES=6000000
@@ -101,15 +103,6 @@ S3_BUCKET=
 S3_REGION=
 S3_ENDPOINT=
 S3_VERSION=
-
-# Only let admins generate oauth clients
-KBIN_ADMIN_ONLY_OAUTH_CLIENTS=false
-
-# Manually approve every new user
-MBIN_NEW_USERS_NEED_APPROVAL=false
-
-# use an allowlist instead of a ban list
-MBIN_USE_FEDERATION_ALLOW_LIST=false
 
 # oAuth (optional)
 OAUTH_AZURE_ID=
--- a/.devcontainer/.env.devcontainer
+++ b/.devcontainer/.env.devcontainer
@@ -31,6 +31,8 @@ KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 MBIN_DEFAULT_THEME=default
+# Language used to parse content and search-expressions for ranking
+MBIN_SEARCH_LANGUAGE=english
 # Set the max image file size (in bytes)
 # This should be set to <= `upload_max_filesize` and `post_max_size` in the server's php.ini file
 MBIN_MAX_IMAGE_BYTES=6000000
--- a/.env.example
+++ b/.env.example
@@ -31,6 +31,8 @@ KBIN_META_KEYWORDS="mbin, content aggregator, open source, fediverse"
 KBIN_HEADER_LOGO=false
 KBIN_FEDERATION_PAGE_ENABLED=true
 MBIN_DEFAULT_THEME=default
+# Language used to parse content and search-expressions for ranking
+MBIN_SEARCH_LANGUAGE=english
 # Set the max image file size (in bytes)
 # This should be set to <= `upload_max_filesize` and `post_max_size` in the server's php.ini file
 MBIN_MAX_IMAGE_BYTES=6000000
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -126,7 +126,6 @@ parameters:
     mbin_new_users_need_approval: '%env(bool:default::MBIN_NEW_USERS_NEED_APPROVAL)%'
     mbin_use_federation_allow_list: '%env(bool:default::MBIN_USE_FEDERATION_ALLOW_LIST)%'
 
-    #todo add in env files
     mbin_search_lang_default: 'english'
     mbin_search_lang: '%env(string:default:mbin_search_lang_default:MBIN_SEARCH_LANGUAGE)%'
 
--- a/config/services.yaml
+++ b/config/services.yaml
@@ -127,7 +127,8 @@ parameters:
     mbin_use_federation_allow_list: '%env(bool:default::MBIN_USE_FEDERATION_ALLOW_LIST)%'
 
     #todo add in env files
-    mbin_search_lang: '%env(string:default:english:MBIN_SEARCH_LANGUAGE)%'
+    mbin_search_lang_default: 'english'
+    mbin_search_lang: '%env(string:default:mbin_search_lang_default:MBIN_SEARCH_LANGUAGE)%'
 
     trueVal: 'true'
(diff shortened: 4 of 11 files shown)


BentiGorlich:
I love that you're tackling this :heart: 

I am not sure what the default should be. I guess you used English as it is the current behaviour, but would it be possible to set it to "generic" or something like that, that is not language specific?

.env.example_docker
```diff
@@ -102,15 +104,6 @@ S3_REGION=
 S3_ENDPOINT=
 S3_VERSION=
 
-# Only let admins generate oauth clients
```
melroy89, line 105: These KBIN_* options are still in use. Keep it in the example. right? Eventually we might need to rename all the configs with `MBIN_` so its all consistent. but that is unrelated to this PR.

blued-gear: They are in the file, I just deleted duplicated keys (see lines 52+).

blued-gear:
> I guess you used English as it is the current behaviour, but would it be possible to set it to "generic" or something like that, that is not language specific?

While I'm not fully against it, I prefer to have English as the default for the following reasons:
- most instances are English centered
- if set to English most instances would have to manually migrate to English search after the update to keep their current search behavior. Two such migrations can waste a lot of CPU time when the DB recreates the indexes.
- as the configuration is part of the setup docs for the admins, new instance admins won't miss the option for optimizing their search

melroy89:
> > I guess you used English as it is the current behaviour, but would it be possible to set it to "generic" or something like that, that is not language specific?
> 
> While I'm not fully against it, I prefer to have English as the default for the following reasons:
> 
>     * most instances are English centered
> 
>     * if set to English most instances would have to manually migrate to English search after the update to keep their current search behavior. Two such migrations can waste a lot of CPU time when the DB recreates the indexes.
> 
>     * as the configuration is part of the setup docs for the admins, new instance admins won't miss the option for optimizing their search

Looking at those reasons, I also think it makes sense to use English here.

melroy89 (approved):

PR #2155, opened 2026-06-08
Add ability for users to block whole instances

### LINKED ISSUE #917: Feature request: block instances

Some instances federated with either have content, users, magazines, or something one would not like to interact with. For example an extremely political instance might not be sharing any content one would like to interact with.

Block anybody and everything from an instance.

Manually block everybody and everything from an instance. However this is time consuming.

Use a client (do such clients exist?) that automatically block depending on which instance the content is coming from.

**Additional context**

There is a "block domain" feature, but that seems to only block posts with **links to instance**. For example one creates a post that has the URL https://somebodysinstance.com/some/path and one has blocked the domain "somebodysinstance.com", then that post won't show up.

blued-gear:
This PR gives users the ability to block whole instances. If blocked the Feed and Search will exclude content where the User or Magazine if from a blocked instance.

Also Admins have the option to add a block for all Users at once, as something like a "soft block".
These blocks will apply to current and future users. Users can override such a block for themselves.

Federation page:
federation

User settings with blocks
blocks

Admin Federation page:
admin

resolves #917


melroy89:
Thanks for this great PR!

garrettw (approved):

PR #2160, opened 2026-06-24
Fix devcontainer build errors

garrettw:
After #2159, I felt like I should just check that everything was working well, and... it wasn't.
But not because of the last PR; it's just some other unrelated maintenance that was needed.


blued-gear (approved):

blued-gear:
Audit fails for vulnerable dependencies. I'm currently working upgrading everything, so after that upcoming PR it should be fine.

PR #2162, opened 2026-06-28
Upgrade dependencies

blued-gear:
## Summary

This PR upgrades *all* dependencies, including PHP, Symfony and JS. Broken code was fixed, many (but not all) new deprecations cleaned and the functionality of the site restored with best effort. A summary of major version jumps is attached at the end of this text.

## I'm asking you to

verify the state of the site. There are many things I and the tests can not check; like 2FA and OAuth authenticator integration (Google, Github, ...).
We need a graceful rollout of this branch from dev to prod-like environments to ensure all broken thing (which likely still exist) get caught.

## Other fixes

While I was at it, I fixed smaller things which annoyed me ;)

- in the "create entry" page the Magazine search is now case-insensitive and clears old search-results when typing a name

## Todo

- [x] upgrade PHP in the Docker containers
- [x] revert content of `to be reverted` commits
- [ ] update docs

## Upgrades

php:
  php 8.3 -> 8.5
    - notable new features:
      - Asymmetric Visibility
      - Property hooks
      - New array_*() functions + array_first() and array_last() functions
      - Fatal Errors (such as an exceeded maximum execution time) now include a backtrace
      - Pipe Operator
      - Uri class
    - breaking changes:
      - Using null as an array offset or when calling array_key_exists() is now deprecated
  doctrine-bundle 2.18 -> 3.2
    - https://github.com/greg0ire/DoctrineBundle/blob/4.0.x/UPGRADE-3.0.md
  doctrine-migrations-bundle 3.7 -> 4.0
    - https://github.com/derrabus/DoctrineMigrationsBundle/blob/4.0.x/UPGRADE.md
    - https://github.com/doctrine/migrations/blob/3.0.x/UPGRADE.md
    - **need to run the `doctrine:migrations:sync-metadata-storage` command**
  doctrine PostgreSQL enum handling
    - replaced the custom DBAL enum types with native PHP enums mapped by `heymoon/doctrine-psql-enum`
    - the included migration converts the existing PostgreSQL enum definitions and must run as part of the normal deployment migration step
  debril/rss-atom-bundle -> php-feed-io/feed-io
    - the old bundle is abandoned and was removed
    - the RSS feed was reimplemented directly with FeedIo while keeping the existing `/feed` endpoint
  FrankenPHP runtime
    - removed `runtime/frankenphp-symfony`; FrankenPHP now uses Symfony’s standard `SymfonyRuntime` directly
  oauth2-google 4 -> 5
    - https://github.com/thephpleague/oauth2-google/blob/main/CHANGELOG.md
  reflection-docblock 5.4 -> 6.0
    - do we even use this?
  2fa-backup-code, 2fa-bundle, 2fa-totp 7.5 -> 8.6
    - https://github.com/scheb/2fa/blob/8.x/UPGRADE.md#7x-to-8x
    - Priority of the two-factor authenticator has changed from 0 to -100
  oauth2-keycloak 5.1 -> 6.1
    - https://github.com/stevenmaguire/oauth2-keycloak/releases/tag/6.0.0
    - nothing major changed
  ux-autocomplete 2.18 -> 3.2
    - https://github.com/symfony/ux-autocomplete/blob/3.x/CHANGELOG.md
    - nothing major changed for us
  ux-chartjs 2.18 -> 3.2
    - https://github.com/symfony/ux-chartjs/blob/3.x/CHANGELOG.md
    - nothing major changed for us
  ux-twig-component 2.18 -> 3.2
    - https://github.com/symfony/ux-twig-component/blob/3.x/CHANGELOG.md
    - nothing major changed for us
  webmozart/assert 1.11 -> 2.4
    - https://github.com/webmozarts/assert/blob/master/CHANGELOG.md
    - nothing major changed for us
  phpunit 12.5 -> 13.2
    - https://github.com/sebastianbergmann/phpunit/releases/tag/13.0.0
    - nothing major changed for us

js:
  babel/core 7.29 -> 8.0
    - https://babeljs.io/docs/v8-migration/
  babel/preset-env 7.29 -> 8.0
  eslint/js 9.39 -> 10.0, eslint 9.39 -> 10.5
    - https://github.com/eslint/eslint/releases/tag/v10.0.0
  fontawesome-free 6.7 -> 7.3
    - https://fontawesome.com/changelog
    - Icon canvases now render at a consistent fixed width. To use the default rendering from v6, add our new fa-width-auto class.
  stylistic/eslint-plugin 2.13 -> 5.10
    - https://github.com/eslint-stylistic/eslint-stylistic/releases/tag/v5.0.0
    - nothing major changed for us
  symfony/stimulus-bridge 3.2 -> 4.0
    - https://github.com/symfony/stimulus-bridge/releases/tag/v4.0.0
    - nothing major changed for us
  symfony/webpack-encore 5.3 -> 7.1
    - https://github.com/symfony/webpack-encore/blob/main/UPGRADE.md
    - needed changes were applied
  chart.js 3.8 -> 4.5
    - https://www.chartjs.org/docs/latest/migration/v4-migration.html
    - not directly used by us (Symfony handles it)
  globals 15.15 -> 17.7
  hotkeys-js 3.13 -> 4.0
    - https://github.com/jaywcjlove/hotkeys-js/releases/tag/v4.0.0
    - no notable changes
    - **do we even use this?**
  sass-loader 16.0 -> 17.0
    - https://github.com/webpack/sass-loader/releases/tag/v17.0.0
    - no notable changes
  simple-icons-font 14.15 -> 16.24
    - https://github.com/simple-icons/simple-icons-font/releases/tag/15.0.0
    - https://github.com/simple-icons/simple-icons/releases/tag/16.0.0
    - no notable changes
  webpack-cli 5.1 -> 7.0
    - https://github.com/webpack/webpack-cli/blob/main/CHANGELOG.md
    - only used by Symfony

GL HF


garrettw:
I probably wouldn't have done a PHP upgrade at the same time as _all_ dependencies. This is a pretty big surface area for one PR.

I would probably split it up as (in this order):
- Low-impact dep upgrades (security patches, non-BC-breaking stuff)
- High-impact dep upgrades (major versions, BC breaks ... and depending on how many packages it is and how many changes are needed to support the upgrade, maybe not all of them in the same PR)
- PHP upgrade, plus only the packages that _must_ be upgraded as part of the PHP change
- Then go back and repeat...
- Low-impact
- High-impact

It's just that this is a lot to review, and when you change so many moving pieces at once ... I just feel like there's more chances for something to go wrong. I tend to feel more confident about smaller, more incremental PRs.

Oh and I typically would separate PHP dep updates from JS ones, but that could be just me.

blued-gear:
I get what you are saying and am not totally against it, but still would prefer to get through this in one go. We did not maintain our dependencies correctly in the past which created this tech debt and splitting it up into the process described by you would create a lot of PRs with duplicated work and in sum as much for you to review as when it is done at once. I'm quite sure it would take much longer to finish the process as everyone would have to find time to go through new PRs again and again + running it on their instances (with one PR we can easily roll back to a known good state as no DB migrations are involved). Also, some of the versions (like Symfony and some JS dependencies) must be updated together.

I know this is a lot, but after this we will keep things updated in a timely manner and with narrow-scoped PRs.

blued-gear:
Test is failing because of this: https://github.com/thephpleague/oauth2-server-bundle/issues/301

assets/controllers/timeago_controller.js
```diff
@@ -30,13 +30,26 @@ export default class extends Controller {
...
+            ja: ja.default,
+            nl: nl.default,
+            pl: pl.default,
+            pt_BR: pt_BR.default,
+            ru: ru.default,
+            tr: tr.default,
+            uk: uk.default,
+            zh_TW: zh_TW.default,
+        };
+        timeago.register(lang, languages[lang]);
```
BentiGorlich: What happens if the language is not available? Is there a fallback?

assets/styles/layout/_forms.scss
```diff
@@ -1,6 +1,7 @@
 @use 'breakpoints' as b;
 @use '../mixins/mbin';
-@use '@fortawesome/fontawesome-free/scss/fontawesome' as fa;
+@use '@fortawesome/fontawesome-free/scss/fa' as fa;
+@use '@fortawesome/fontawesome-free/scss/solid' as faSolid;
```
BentiGorlich, line 4: This is imported twice now

assets/styles/layout/_forms.scss
```diff
@@ -138,16 +139,16 @@ input[type=radio] {
   -webkit-appearance: none;
   appearance: none;
   font-size: 0.9rem;
-  display: grid;
+  display: grid !important;
```
BentiGorlich: this doesn't look good, why is the !important needed?

config/packages/framework.yaml
```diff
@@ -27,9 +26,9 @@ framework:
         # previous
         #handler_id: Symfony\Component\HttpFoundation\Session\Storage\Handler\PdoSessionHandler
         # transition
-        handler_id: App\MigratePdoToRedisSessionHandler
+        #handler_id: App\MigratePdoToRedisSessionHandler
         # new
-        #handler_id: Symfony\Component\HttpFoundation\Session\Storage\Handler\RedisSessionHandler
+        handler_id: Symfony\Component\HttpFoundation\Session\Storage\Handler\RedisSessionHandler
```
BentiGorlich: This should not be committed, yet. I believe I explained before why

config/packages/http_discovery.yaml
BentiGorlich, line 1: What is this for?

config/bundles.php
```diff
@@ -23,7 +21,6 @@
     SymfonyCasts\Bundle\VerifyEmail\SymfonyCastsVerifyEmailBundle::class => ['all' => true],
     Symfony\Bundle\MercureBundle\MercureBundle::class => ['all' => true],
     Nelmio\CorsBundle\NelmioCorsBundle::class => ['all' => true],
-    Debril\RssAtomBundle\DebrilRssAtomBundle::class => ['all' => true],
```
BentiGorlich: So we don't have an rss feed anymore?

src/Utils/Polyfills.php
```diff
@@ -0,0 +1,31 @@
...
+
+class Polyfills
+{
+    /**
+     * Gets a "parameter" value from the request.
+     *
+     * This function uses the original behavior from https://github.com/symfony/symfony/blob/7.3/src/Symfony/Component/HttpFoundation/Request.php .
+     * It was deprecated and removed, but it is easier to just revive it than clean up our code.
+     */
+    public static function requestParam(Request $req, string $key, mixed $default = null): mixed
```
BentiGorlich, line 15: I'd have been fine with just forcing these parameters into the query, but I am fine with this as well :)

symfony.lock
```diff
@@ -297,6 +278,18 @@
...
+        "recipe": {
+            "repo": "github.com/symfony/recipes",
+            "branch": "main",
+            "version": "1.18",
+            "ref": "f45b5dd173a27873ab19f5e3180b2f661c21de02"
+        },
+        "files": [
+            "config/packages/http_discovery.yaml"
+        ]
+    },
```
BentiGorlich, line 292: I don't understand what this is for

webpack.config.js
```diff
@@ -55,36 +59,37 @@ Encore
     // enables hashed filenames (e.g. app.abc123.css)
     .enableVersioning(Encore.isProduction())
 
+    // Configure JS and CSS minimizers
+    .configureJsMinimizerPlugin((options, MinimizerPlugin) => {
+        options.minify = MinimizerPlugin.esbuildMinify
+    })
+    .configureCssMinimizerPlugin((options, MinimizerPlugin) => {
+        options.minify = MinimizerPlugin.lightningCssMinify;
+    })
```
BentiGorlich, line 68: This will change our js build a lot, right?

assets/controllers/tomselect_clear_on_load_controller.js
BentiGorlich, line 1: Did this come from a package?

assets/utils/popover.js
BentiGorlich, line 1: I don't undserstand these changes, but I also don't know what this is used for...

src/DTO/RemoteInstanceDto.php
BentiGorlich, line 1: why these changes?

BentiGorlich (dismissed):
I do not understand what changed with the RSS feed, can you explain that a bit?

assets/controllers/timeago_controller.js
blued-gear: I added a check. The library has a fallback to en_US if no lang arg is passed.

assets/controllers/tomselect_clear_on_load_controller.js
blued-gear: If I remember right, I wrote this with the help of https://symfony.com/bundles/ux-autocomplete/current/index.html and https://tom-select.js.org/docs/#callbacks .

assets/utils/popover.js
blued-gear: The popover is used for example when details of a user are shown as an overlay when hovered. It was quite old code not compatible with mjs, so I rewrote the original source in a more modern way.

config/packages/http_discovery.yaml
blued-gear: It was generated by a Symfony recipe. I don't know what it is for either.

config/packages/framework.yaml
blued-gear: reverted

assets/styles/layout/_forms.scss
blued-gear: Same reason as on `width`: the mixin `@include faSolid.icon(fa.$var-check);` sets rules for `width` and `display` with values we don't want.
I just checked and it seems to work now without it; I will remove it in a separate commit, so that the removal can be reverted when problems come up in a larger test.

assets/styles/layout/_forms.scss
blued-gear: I think I kept the former alias for compatibility.
I will remove it in a separate commit, so that the removal can be reverted when problems come up in a larger test.

src/Utils/Polyfills.php
blued-gear: Clearly deciding where which param should go would definitely the better solution. But finding all locations where a parameter is set would be a much bigger task as this PR is anyway.

webpack.config.js
blued-gear: Yes, but I adjusted our sources were necessary and all builds fine (for me at least).

symfony.lock
blued-gear: This was added automatically by a Symfony recipe.

src/DTO/RemoteInstanceDto.php
blued-gear: Changed in commit 0b3a196e 'fix a deprecation', so I think (I can't remember anymore) it was prompted in a deprecation warning that it should be removed.

melroy89:
Let me try to just rebase and fix conflicts, so I can run this branch on kbin.melroy.org. I like to have this PR merged.

melroy89:
> [ ] update docs

What do you wanted to update in the docs still? PHP 8.5 is already correctly documented now in the docs and README.

I had some ideas we could still improve in the docs?

  - Update docs/03-contributing/03-project-overview.md:
      - Remove the deleted` src/Feed` directory.
      - Explain that PostgreSQL enums now use native PHP enums plus heymoon/doctrine-psql-enum, if that detail is useful to contributors.
      - Mention that the RSS feed is now implemented by `src/Controller/Feed/RssFeedController.php`.
      - Fix the existing typo `php bin:console` to `php bin/console`.

  - Verify installation requirements:
      - Check whether the bare-metal installation page contains package names or commands tied to an older PHP version. I currently run ubuntu server 24.04  still ("bare metal") but with PHP 8.5 so I can test this to some degree.

blued-gear:
> What do you wanted to update in the docs still

The open checkbox is a reminder for me to go through all changes just before this is merged and update docs and upcoming release notes.

src/Enums/ESortOptions.php
```diff
@@ -4,6 +4,9 @@
 
 namespace App\Enums;
 
+use HeyMoon\DoctrinePostgresEnum\Attribute\EnumType;
+
+#[EnumType('enum_sort_options')]
```
melroy89: Doctrine does not automatically rename PostgreSQL enum types when the PHP mapping name changes. This PR changes the mapping to `enum_sort_options`, while existing databases still have `enumSortOptions` (folded by PostgreSQL to `enumsortoptions`). I confirmed the current schema diff does not request a rename, so this alone does not break the upgrade. However, the enum bundle uses the new mapping name in generated `ALTER TYPE` SQL; a later change to this enum's values can therefore target a type that does not exist. Please preserve the existing type name or add an explicit migration that renames the database type.

melroy89: Or maybe this is works as expected I think.. and automatically convert enum_sort_options to enumSortOptions?

melroy89: Doctrine does not automatically rename an existing PostgreSQL enum type when the PHP mapping name changes. Existing databases have `enumsortoptions` (created by the migration as unquoted `enumSortOptions`), while this mapping now names it `enum_sort_options`.
 
The enum bundle can generate SQL that exposes the mismatch: when it detects changed enum values, `DoctrineEnumColumnPlatform::processColumn()` reads the existing type and values, but builds the `ALTER TYPE` statement using the mapped target name. That can produce `ALTER TYPE enum_sort_options ADD VALUE ...`, which fails if only `enumsortoptions` exists. The bundle’s functional test demonstrates the enum-change path in `vendor/heymoon/doctrine-psql-enum/tests/Functional/MigrationTest.php` (another_example gains bar).

To reproduce locally, use a database upgraded through the migration that creates `enumSortOptions`, then temporarily add a case to ESortOptions and run `php bin/console doctrine:schema:update --dump-sql`. Check whether the generated SQL targets `enum_sort_options`. Try to run that SQL against the upgraded database to confirm whether it fails. Please preserve the existing PostgreSQL type name or add an explicit migration to rename it.

blued-gear: I think the best option is to just make the mappings in PHP match the current DB names.

PR #2163, opened 2026-06-28
upgrade PHP in CI Docker image

blued-gear:


garrettw:
I assume you must have some good reason for changing the distro? I mean, that makes this a little more than purely a PHP upgrade.

blued-gear:
Debian does not offer PHP 8.5 yet, so I switched to a distro which does. It would also be possible to use 3th-party repos for Debian, but IMO this would rise the attack-surface as they are not reviewed by official maintainers.

garrettw:
Well, while you're changing the base image, why not go to an official PHP image and keep the distro the same?

`php:8.5-fpm-trixie`

Or if you want a lighter distro, there are Alpine versions.

blued-gear:
I will look into them.

ci/Dockerfile
```diff
@@ -1,35 +1,36 @@
...
-FROM debian:13-slim
+FROM php:8.5-fpm-trixie
 
 COPY --from=composer/composer:latest-bin /composer /usr/bin/composer
 
 ARG DEBIAN_FRONTEND=noninteractive
 
-RUN apt update
-RUN apt upgrade -y
+RUN apt update && apt update -y
```
garrettw: I believe the second `update` should be `upgrade`

garrettw (changes requested):

garrettw (approved):
As long as it works, it looks fine to me. But I'd prefer others to approve as well.

melroy89 (approved):

melroy89:
so https://github.com/MbinOrg/mbin/actions/runs/28337112998/job/83945142612 is still failing somehow.. despite this change? Or this first needs to be merged. I guess.

blued-gear:
> so https://github.com/MbinOrg/mbin/actions/runs/28337112998/job/83945142612 is still failing somehow.. despite this change? Or this first needs to be merged. I guess.

You are looking at an old run. The latest succeeded. https://github.com/MbinOrg/mbin/actions/runs/28393820191/job/84127318482?pr=2163

blued-gear:
I have to redo this, as

-  the Node version from Debian is too old
- the Postgresql driver was not installed correctly (the `pgsql` was installed fine, but I missed that there is also a `PDO_PGSQL` module)

https://github.com/MbinOrg/mbin/actions/runs/28472511248/job/84388597761?pr=2162

PR #2164, opened 2026-07-01
Fix Node version and missing PHP extension in CI image

blued-gear:
Follow-up of #2163

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -3,7 +3,7 @@
 {
     "name": "PHP",
     // Or use a Dockerfile or Docker Compose file. More info: https://containers.dev/guide/dockerfile
-    "image": "mcr.microsoft.com/devcontainers/php:3.0.3-8.4-trixie",
+    "image": "mcr.microsoft.com/devcontainers/php:3-8.5-trixie",
 
     // Features to add to the dev container. More info: https://containers.dev/features.
     "features": {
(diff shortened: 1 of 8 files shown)


garrettw:
Looks fine, except that 8.5 breaks `composer install`. Is there any way you can go ahead and update only the dependencies required to resolve the build errors in this PR? I think it would be good to do those at the same time.

blued-gear:
> Is there any way you can go ahead and update only the dependencies required to resolve the build errors in this PR

I can do that, but don't expect it to be just an update. There is an abandoned package (the RSS stuff) which have to be replaced in order to use PHP 8.5

garrettw:
That's fine to go ahead and do it here, if you can and want to. But if it will take you some time to do, just be mindful of any other PRs that might need CI to work in the meantime.

blued-gear:
I got lucky, the RSS package was incompatible with Symfony 8, not PHP 8.5 :D

Now the linter has new rules. Should I run the fix command to change the affected sourcefiles or disable the rule so that we keep the old style?

garrettw:
I would disable it for this PR. Then you could make a new PR to re-enable it and fix the affected files.

blued-gear:
Should be ready for review now.

garrettw (approved):

garrettw:
I guess the CI failure won't be fixed until this is merged?

blued-gear:
The test pipeline needs the new image, so futures PR should work again (I hope).

melroy89:
forced merged.

PR #2167, opened 2026-07-16
update docker build

blued-gear:
- use PHP 8.5
- use Node 26
- use Postgresql 18
- expand compatibility of compose settings and image URLs to Podman

**Attention:** the change to Postgresql 18 changes the mounts of the database files and requires a DB upgrade for existing deployments.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -156,7 +156,7 @@ POSTGRES_DB=mbin
 POSTGRES_USER=mbin
 POSTGRES_PASSWORD=!ChangeThisPostgresPass!
 # IMPORTANT: You MUST configure your PostgreSQL server version!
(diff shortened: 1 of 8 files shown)


garrettw (approved):
Seems fine to me but I'd like a second approval

jwr1:
Are the `Z` volume options for Podman compatibility?

blued-gear:
To be more specific for [SELinux](https://docs.docker.com/engine/storage/bind-mounts/#configure-the-selinux-label), which is important in a rootless context to set.

jwr1 (approved):
Looks good! We just need to remember to mention the breaking change (Postgres volume mount) in the release notes.

PR #2168, opened 2026-07-17
call gc_collect_cycles() for every message

blued-gear:
This might help keeping the memory-consumption of the messengers low. It might also not help, but the worst it can do is add a few milliseconds of delay to the processing of every message.

As far as I checked it helped on some message type and on others it did not lower the overall memory consumption. If this is worth merging is up to you.

### DIFF
--- a/src/MessageHandler/MbinMessageHandler.php
+++ b/src/MessageHandler/MbinMessageHandler.php
@@ -23,18 +23,20 @@ abstract class MbinMessageHandler
      */
     public function workWrapper(MessageInterface $message): void
     {
-        gc_collect_cycles();
+        try {
+            // when we are in the test environment this would throw: ConnectionException: There is no active transaction.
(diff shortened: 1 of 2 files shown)


src/MessageHandler/MbinMessageHandler.php
```diff
@@ -23,6 +23,8 @@ public function __construct(
      */
     public function workWrapper(MessageInterface $message): void
     {
+        gc_collect_cycles();
```
BentiGorlich: wouldn't this make more sense to be called at the end? So after a message has been processed

blued-gear: It would. \
But if something goes very wrong, it might not reach this statement. As there are almost always new messages it should not make much of a difference.

garrettw: That's the use case for `finally` ;)

BentiGorlich (approved):

PR #2169, opened 2026-07-19
npm audit fix

garrettw:
Ran `npm audit fix` to get rid of Dependabot security alerts.
Also added items to `allowScripts` to ensure `npm install` completes successfully.

### DIFF
--- a/package.json
+++ b/package.json
@@ -44,5 +44,9 @@
         "build": "encore production --progress",
         "lint": "eslint .",
         "lint-fix": "eslint --fix ."
+    },
+    "allowScripts": {
+        "@parcel/watcher@2.5.6": true,
+        "core-js@3.49.0": true
     }
 }
--- a/package.json
+++ b/package.json
@@ -46,7 +46,6 @@
         "lint-fix": "eslint --fix ."
     },
     "allowScripts": {
-        "@parcel/watcher@2.5.6": true,
-        "core-js@3.49.0": true
+        "@parcel/watcher@2.5.6": true
     }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -44,5 +44,9 @@
         "build": "encore production --progress",
         "lint": "eslint .",
         "lint-fix": "eslint --fix ."
+    },
+    "allowScripts": {
+        "@parcel/watcher@2.5.6": true,
+        "core-js@3.49.0": true
```
blued-gear: I think we don't need this one as it will only show a text, see https://github.com/zloirock/core-js/blob/master/packages/core-js/postinstall.js

package.json
```diff
@@ -44,5 +44,9 @@
         "build": "encore production --progress",
         "lint": "eslint .",
         "lint-fix": "eslint --fix ."
+    },
+    "allowScripts": {
+        "@parcel/watcher@2.5.6": true,
```
blued-gear: Do you have documentation what this postinstall script does? I can't find a script declaration in watchers repo.

garrettw: It's in there.
```json
    "install": "node scripts/build-from-source.js",
```

Contents of that file:
```js
#!/usr/bin/env node

const {spawn} = require('child_process');

if (process.env.npm_config_build_from_source === 'true') {
  build();
}

function build() {
  spawn('node-gyp', ['rebuild'], { stdio: 'inherit', shell: true }).on('exit', function (code) {
    process.exit(code);
  });
}
```

blued-gear: Ah, alright. I looked only for `postinstall`.

blued-gear (approved):

PR #2172, opened 2026-07-20
Fix: video embed with unsupported media

### LINKED ISSUE #487: Media preview attempting to load unsupported video format
**Describe the bug**
The thread media preview button (or auto media preview) is attempting to load a video that is unsupported causing the browser to prompt the user to download a file "master.m3u8"

https://kbin.run/

**To Reproduce**
Go to https://kbin.run/m/news@lemmy.world/t/243728/Gen-Z-is-recording-themselves-getting-fired-in-growing-TikTok and click the media preview button (or have auto media preview enabled when viewing the thread).

**Expected behavior**
Do not preview unsupported media formats

**Screenshots**

Enable:

image

or click

image

Results in:

image

 - OS: Linux
 - Browser Firefox
 - Version 122.0.1

**Additional context**
Impacts all Mbin instances

garrettw:
Closes #487 

Before, preview embeds that contained unsupported media URLs could load when the preview button is clicked, leading to direct download of the media file.

With this change, unsupported media will return a null embed, which should solve it.

### DIFF
--- a/src/Utils/Embed.php
+++ b/src/Utils/Embed.php
@@ -9,6 +9,7 @@ use App\Event\ActivityPub\CurlRequestBeginningEvent;
 use App\Event\ActivityPub\CurlRequestFinishedEvent;
 use App\Service\ImageManager;
 use App\Service\SettingsManager;
+use App\Service\VideoManager;
 use Embed\Embed as BaseEmbed;
 use Embed\Extractor;
 use Psr\Log\LoggerInterface;
@@ -40,7 +41,7 @@ class Embed
         unset($this->dispatcher);
     }
 
-    public function fetch($url): self
+    public function fetch(string $url): self
     {
         if ($this->settings->isLocalUrl($url)) {
             if (ImageManager::isImageUrl($url)) {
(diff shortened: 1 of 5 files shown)


src/Utils/Embed.php
```diff
@@ -166,7 +205,7 @@ public function getType(): string
         }
 
         if ($this->isVideoUrl()) {
-            return Entry::ENTRY_TYPE_IMAGE;
+            return Entry::ENTRY_TYPE_VIDEO;
```
blued-gear, line 208: I don't know if the usage of `ENTRY_TYPE_IMAGE` was by mistake or intentional, as the video type is not supported everywhere in the codebase iirc.

garrettw: I checked existing usage of `ENTRY_TYPE_IMAGE` and found a couple of places that didn't have appropriate usage of `ENTRY_TYPE_VIDEO`, so I added some handling for that. I don't think it's perfect/complete, but it should be marginally better than before.

blued-gear (approved):

PR #2174, opened 2026-07-21
fix parsing of actorUrl for magazine

blued-gear:
Same fix as @melroy89 did for User for Magazine.

Please note that I adjusted the scoring algorithm to exclude URLs which have no valid path at all.

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -413,26 +413,7 @@ class ActivityPubManager
             $user->apDiscoverable = $actor['discoverable'] ?? null;
             $user->apIndexable = $actor['indexable'] ?? null;
             $user->apManuallyApprovesFollowers = $actor['manuallyApprovesFollowers'] ?? false;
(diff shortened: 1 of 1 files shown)


src/Service/ActivityPubManager.php
```diff
@@ -1320,4 +1301,33 @@ public function extractRemoteShareCount(array $apObject): ?int
 
         return null;
     }
+
+    private function findBestActorUrl(string $actorUrl, mixed $actorUrlValue): string
+    {
+        if (null === $actorUrlValue) {
+            return $actorUrl;
+        } elseif (\is_string($actorUrlValue)) {
```
garrettw: Conditionals after a `return` don't need `else`

garrettw (approved):
> Please note that I adjusted the scoring algorithm to exclude URLs which have no valid path at all.

I'll approve this as it mostly looks fine, but I can't speak to the logic of the algorithm change. Ideally I would like this PR to include new tests against URLs that are considered valid/invalid to demonstrate the changed behavior.

Issue #1452, opened 2025-02-13
Follow hashtags like in Mastodon


jgeorgeson:
**Describe the solution you'd like**
I'd like the ability to follow hashtags in the microblog UI, like we can in Mastodon

https://fedi.tips/how-do-i-follow-hashtags-on-mastodon-and-the-fediverse/

**Describe alternatives you've considered**
I haven't found an equivalent capability in MBin so far.

blued-gear:
Note to myself: when implementing this, also implement

- [x] API endpoint to get feed of a hashtag

blued-gear:
When implementing this, I should also:

- [ ] add a page and API for listing tags with some stats (number of content overall, in the last 24h)

PR #2180, opened 2026-07-26
Add ability to block and subscribe Hashtags

### LINKED ISSUE #1993: Ability to block Hashtags
Some users want to block content which fall into a specific group. The most universal tagging format the Fediverse has, are Hashtags; so we could offer the user to use them to filter their content.

- on the Hashtag page there must be a "Block / Unblock" button
- in the userprofile there should be a list of blocked Hashtags (to find it for unblocking)

Flairs or other attached metadata Fediverse software could use for filtering, but this is not standardized and not widely adopted.

blued-gear:
This PR enables users to block and subscribe to Hashtags and list their blocks / subscriptions (in UI and API equally).

Closes #1993 and #1452


src/Repository/ContentRepository.php
```diff
@@ -207,7 +220,11 @@ private function getQueryAndParameters(Criteria $criteria, bool $addCursor): arr
                     ' OR EXISTS (SELECT 1 FROM domain_subscription ds WHERE ds.domain_id = c.domain_id AND ds.user_id = :loggedInUser)' :
                     ' OR c.domain_id IN (:cachedUserSubscribedDomains)');
 
+            $subClausePost = str_replace('%hl_type%', 'post', $subClausePost);
+            $subClauseEntry = str_replace('%hl_type%', 'entry', $subClauseEntry);
+
             if ($criteria->includeBoosts) {
+                // TODO should comments with a subscribed hashtag be included too?
```
blued-gear: I would like to know your thoughts on this.

garrettw: I lean toward "no".

BentiGorlich: I am not sure... I think the default should be no, but there could be a setting for it

templates/tag/_panel.html.twig
```diff
@@ -11,6 +11,7 @@
         </header>
     </div>
 
+    {{ component('hashtag_sub', {hashtag: hashtag}) }}
```
melroy89, line 14: Could we make sure every tag page passes a `hashtag` here? I can reproduce a 500 at `/tag/blockedtag/threads/newest/all` because `TagEntryFrontController` only passes `tag`. The posts, comments and people controllers have the same issue.

src/Service/TagManager.php
```diff
@@ -177,4 +182,40 @@ public function isAnyTagBanned(?array $tags): bool
...
+        $tag->unsubscribe($user);
+
+        $this->entityManager->flush();
+
+        $this->dispatcher->dispatch(new HashtagSubscriptionChangedEvent($tag, $user, false));
+    }
+
+    public function block(User $user, Hashtag $hashtag): void
+    {
+        $user->blockHashtag($hashtag);
```
melroy89, line 210: Blocking a hashtag should also remove an existing subscription, like domains and magazines do. I reproduced subscribing and then blocking the same hashtag, and it stays subscribed. Can we unsubscribe here before adding the block? Also, subscribing removes a block directly without dispatching `HashtagBlockChangedEvent`, so the cached block list can stay stale.

src/Repository/ContentRepository.php
```diff
@@ -299,6 +318,26 @@ private function getQueryAndParameters(Criteria $criteria, bool $addCursor): arr
                 }
             }
 
+            $blockingClauseEntryComment = $blockingClausePost;
+            $blockingClausePostComment = $blockingClausePost;
+
+            // only include the subcluase if there are (/ might be) blocks
+            if (null === $criteria->cachedUserBlockedHashtags || !empty($criteria->cachedUserBlockedHashtags)) {
+                if (null === $criteria->cachedUserBlockedHashtags) {
+                    $blockingClauseEntry = $blockingClauseEntry.' AND NOT EXISTS (SELECT 1 FROM hashtag_link hl INNER JOIN hashtag_block hb ON hl.hashtag_id = hb.hashtag_id WHERE hl.entry_id = c.id AND hb.user_id = :loggedInUser)';
```
melroy89, line 343: Can we apply this blocked hashtag filtering in `EntryRepository` and `PostRepository` too? I created an entry and post containing a blocked hashtag, and both standalone repositories still returned the content. Those repositories back tag, user and API listings, so those paths currently bypass this filter.

src/Entity/HashtagSubscription.php
```diff
@@ -0,0 +1,50 @@
...
+class HashtagSubscription
+{
+    use CreatedAtTrait {
+        CreatedAtTrait::__construct as createdAtTraitConstruct;
+    }
+
+    #[ManyToOne(targetEntity: User::class, inversedBy: 'subscribedHashtags')]
+    #[JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    public ?User $user;
+    #[ManyToOne(targetEntity: Hashtag::class)]
```
melroy89: This association needs `inversedBy: 'subscriptions'` to match `Hashtag::$subscriptions`. `doctrine:schema:validate` currently reports this relationship as invalid.

src/Entity/OAuth2UserConsent.php
```diff
@@ -38,6 +38,10 @@ class OAuth2UserConsent
         'domain' => 'oauth2.grant.domain.all',
         'domain:subscribe' => 'oauth2.grant.domain.subscribe',
         'domain:block' => 'oauth2.grant.domain.block',
+        // Grants allowing applications to (un)subscribe or (un)block hashtags on behalf of the user
+        'hashtag' => 'oauth2.grant.hashtag.all',
```
melroy89, line 42: Could we add the matching English translations for these consent keys? They are missing from `messages.en.yaml`, so the OAuth consent screen will show the raw `oauth2.grant.hashtag.*` keys. The general subscribe and block descriptions should mention hashtags too.

src/Controller/Api/Tag/TagBlockApiController.php
```diff
@@ -0,0 +1,210 @@
...
+        return new JsonResponse(
+            $this->serializeHashtag($tag),
+            headers: $headers
+        );
+    }
+
+    #[OA\Response(
+        response: 200,
+        description: 'Hashtag unblocked',
+        content: new Model(type: DomainDto::class),
```
melroy89: This response documents `DomainDto`, but the endpoint serializes a `HashtagResponseDto`. The unsubscribe endpoint has the same mismatch.

src/Repository/ContentRepository.php
blued-gear: I added a setting for it.

src/Controller/Api/Tag/TagBaseApi.php
```diff
@@ -0,0 +1,37 @@
...
+abstract class TagBaseApi extends BaseApi
+{
+    private readonly HashtagFactory $factory;
+    protected readonly TagRepository $repository;
+
+    #[Required]
+    public function setFactory(HashtagFactory $factory): void
+    {
+        $this->factory = $factory;
+    }
```
BentiGorlich, line 22: So this is how you get new dependencies into sub classes... Very neat!

src/Entity/Hashtag.php
```diff
@@ -25,6 +27,51 @@ class Hashtag
     #[Column(type: 'boolean', options: ['default' => false])]
     public bool $banned = false;
 
+    #[OneToMany(mappedBy: 'hashtag', targetEntity: HashtagSubscription::class, fetch: 'EXTRA_LAZY', cascade: [
+        'persist',
+        'remove',
```
BentiGorlich, line 32: remove is not correct in this regard (I think). The doctrine removal is always really confusing for me, but it might be that it also removes the hashtag, when you're removing the subscription via `$entiyManager->remove`...

src/Entity/Hashtag.php
```diff
@@ -25,6 +27,51 @@ class Hashtag
...
+    public function __construct()
+    {
+        $this->subscriptions = new ArrayCollection();
+        $this->linkedPosts = new ArrayCollection();
+    }
+
+    public function subscribe(User $user): void
+    {
+        if (!$this->isSubscribed($user)) {
+            $subscription = new HashtagSubscription($user, $this);
```
BentiGorlich, line 48: do you need to persist it here?

src/Entity/Hashtag.php
```diff
@@ -25,6 +27,51 @@ class Hashtag
...
+
+        /** @var HashtagSubscription $subscription */
+        $subscription = $this->subscriptions->matching($criteria)->first();
+
+        if ($this->subscriptions->removeElement($subscription)) {
+            if ($subscription->hashtag === $this) {
+                $subscription->hashtag = null;
+            }
+            $user->subscribedHashtags->removeElement($subscription);
+        }
```
BentiGorlich, line 67: don't you have to remove the whole element as well via an entity manager?

src/Entity/HashtagBlock.php
```diff
@@ -0,0 +1,50 @@
...
+use Doctrine\ORM\Mapping\Entity;
+use Doctrine\ORM\Mapping\GeneratedValue;
+use Doctrine\ORM\Mapping\Id;
+use Doctrine\ORM\Mapping\JoinColumn;
+use Doctrine\ORM\Mapping\ManyToOne;
+use Doctrine\ORM\Mapping\Table;
+use Doctrine\ORM\Mapping\UniqueConstraint;
+
+#[Entity]
+#[Table]
```
BentiGorlich, line 18: What is this attribute for? We don't have it anywhere else and the [docs](https://www.doctrine-project.org/projects/doctrine-orm/en/3.6/reference/attributes-reference.html#attrref_table) basically say it is only to give the table another name 🤔 

> Attribute describes the table an entity is persisted in. It is placed on the entity-class level and is optional. If it is not specified the table name will default to the entity's unqualified classname.

src/Entity/HashtagBlock.php
```diff
@@ -0,0 +1,50 @@
...
+#[Cache(usage: 'NONSTRICT_READ_WRITE')]
+class HashtagBlock
+{
+    use CreatedAtTrait {
+        CreatedAtTrait::__construct as createdAtTraitConstruct;
+    }
+
+    #[ManyToOne(targetEntity: User::class, inversedBy: 'blockedHashtags')]
+    #[JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    public ?User $user;
```
BentiGorlich, line 29: why is the field nullable if the column definition says it is not?

src/Entity/HashtagBlock.php
```diff
@@ -0,0 +1,50 @@
...
+    #[ManyToOne(targetEntity: Hashtag::class)]
+    #[JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    public ?Hashtag $hashtag;
```
BentiGorlich, line 32: why is the field nullable if the column definition says it is not?

src/Entity/HashtagSubscription.php
```diff
@@ -0,0 +1,50 @@
...
+    use CreatedAtTrait {
+        CreatedAtTrait::__construct as createdAtTraitConstruct;
+    }
+
+    #[ManyToOne(targetEntity: User::class, inversedBy: 'subscribedHashtags')]
+    #[JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    public ?User $user;
+    #[ManyToOne(targetEntity: Hashtag::class, inversedBy: 'subscriptions')]
+    #[JoinColumn(nullable: false, onDelete: 'CASCADE')]
+    public ?Hashtag $hashtag;
```
BentiGorlich, line 32: same thing here

src/Entity/User.php
```diff
@@ -671,7 +680,42 @@ public function unblockDomain(Domain $domain): void
...
-                $this->blockedMagazines->removeElement($domainBlock);
+                $this->blockedDomains->removeElement($domainBlock);
+            }
+        }
+    }
+
+    public function blockHashtag(Hashtag $hashtag): self
+    {
+        if (!$this->isBlockedHashtag($hashtag)) {
+            $this->blockedHashtags->add(new HashtagBlock($this, $hashtag));
```
BentiGorlich, line 691: no persisting required?

src/Entity/User.php
```diff
@@ -671,7 +680,42 @@ public function unblockDomain(Domain $domain): void
...
+
+        /**
+         * @var HashtagBlock $hashtagBlock
+         */
+        $hashtagBlock = $this->blockedHashtags->matching($criteria)->first();
+
+        if ($this->blockedHashtags->removeElement($hashtagBlock)) {
+            if ($hashtagBlock->user === $this) {
+                $hashtagBlock->hashtag = null;
+                $this->blockedHashtags->removeElement($hashtagBlock);
```
BentiGorlich, line 718: is this element removed automatically from db this way?

src/Repository/ContentRepository.php
```diff
@@ -299,6 +334,26 @@ private function getQueryAndParameters(Criteria $criteria, bool $addCursor): arr
...
+                } else {
+                    $blockingClauseEntry = $blockingClauseEntry.' AND NOT EXISTS (SELECT 1 FROM hashtag_link hl WHERE hl.entry_id = c.id AND hl.hashtag_id IN (:cachedUserBlockedHashtags))';
+                    $blockingClausePost = $blockingClausePost.' AND NOT EXISTS (SELECT 1 FROM hashtag_link hl WHERE hl.post_id = c.id AND hl.hashtag_id IN (:cachedUserBlockedHashtags))';
+                    $blockingClauseEntryComment = $blockingClauseEntryComment.' AND NOT EXISTS (SELECT 1 FROM hashtag_link hl WHERE hl.entry_comment_id = c.id AND hl.hashtag_id IN (:cachedUserBlockedHashtags))';
+                    $blockingClausePostComment = $blockingClausePostComment.' AND NOT EXISTS (SELECT 1 FROM hashtag_link hl WHERE hl.post_comment_id = c.id AND hl.hashtag_id IN (:cachedUserBlockedHashtags))';
+
+                    $parameters['cachedUserBlockedHashtags'] = $criteria->cachedUserBlockedHashtags;
+                }
```
BentiGorlich, line 354: this should let everything the logged in user authored through. I believe that that is not the case right now

src/Repository/EntryCommentRepository.php
```diff
@@ -223,6 +225,13 @@ private function filter(QueryBuilder $qb, Criteria $criteria): QueryBuilder
                 'c.magazine NOT IN (SELECT IDENTITY(mb.magazine) FROM '.MagazineBlock::class.' mb WHERE mb.user = :blocker)'
             );
 
+            $qb->andWhere(
+                'NOT EXISTS ('
+                    .'SELECT 1 FROM '.HashtagBlock::class.' hb INNER JOIN '.HashtagLink::class.' hbl ON hb.hashtag = hbl.hashtag '
+                    .'WHERE hbl.entryComment = c AND hb.user = :blocker'
+                .')'
```
BentiGorlich, line 232: this also should let their own authored comments through

src/Repository/EntryRepository.php
```diff
@@ -238,6 +242,13 @@ private function filter(QueryBuilder $qb, EntryPageView $criteria): QueryBuilder
                 );
             }
 
+            $qb->andWhere(
+                'NOT EXISTS ('
+                .'SELECT 1 FROM '.HashtagBlock::class.' hb INNER JOIN '.HashtagLink::class.' hbl ON hb.hashtag = hbl.hashtag '
+                .'WHERE hbl.entry = e AND hb.user = :blocker'
+                .')'
+            );
```
BentiGorlich, line 250: this also should let their own authored comments through

src/Repository/PostCommentRepository.php
```diff
@@ -160,12 +163,31 @@ private function filter(QueryBuilder $qb, Criteria $criteria)
...
         if ($user && !$criteria->moderated) {
             $qb->andWhere(
                 'c.user NOT IN (SELECT IDENTITY(ub.blocked) FROM '.UserBlock::class.' ub WHERE ub.blocker = :blocker)'
             );
 
+            $qb->andWhere(
+                'NOT EXISTS ('
+                    .'SELECT 1 FROM '.HashtagBlock::class.' hb INNER JOIN '.HashtagLink::class.' hbl ON hb.hashtag = hbl.hashtag '
+                    .'WHERE hbl.postComment = c AND hb.user = :blocker'
+                .')'
```
BentiGorlich, line 188: this also should let their own authored comments through

src/Repository/PostRepository.php
```diff
@@ -205,12 +209,19 @@ private function filter(QueryBuilder $qb, Criteria $criteria): QueryBuilder
...
-                'NOT EXISTS (SELECT IDENTITY(mb.magazine) FROM '.MagazineBlock::class.' mb WHERE mb.user = :magazineBlocker AND mb.magazine = p.magazine)'
+                'NOT EXISTS (SELECT IDENTITY(mb.magazine) FROM '.MagazineBlock::class.' mb WHERE mb.user = :blocker AND mb.magazine = p.magazine)'
             );
-            $qb->setParameter('magazineBlocker', $user);
+
+            $qb->andWhere(
+                'NOT EXISTS ('
+                .'SELECT 1 FROM '.HashtagBlock::class.' hb INNER JOIN '.HashtagLink::class.' hbl ON hb.hashtag = hbl.hashtag '
+                .'WHERE hbl.post = p AND hb.user = :blocker'
+                .')'
```
BentiGorlich, line 221: this also should let their own authored comments through

templates/tag/_panel.html.twig
BentiGorlich: I also still get an exception here, maybe because the tag is not in the DB, so the tag needs to get created

templates/components/hashtag_sub.html.twig
```diff
@@ -0,0 +1,26 @@
+<aside{{ attributes.defaults({class: 'domain__subscribe', 'data-controller': 'subs'}) }}>
```
BentiGorlich, line 1:
```suggestion
<aside{{ attributes.defaults({class: 'hashtag__subscribe', 'data-controller': 'subs'}) }}>
```

BentiGorlich (commented):
I am probably wrong about the persisting issue, but it looks weird to me, being used to having to do that manually 😅

src/Entity/Hashtag.php
blued-gear: The Hashtag subscription and block entities + mechanisms I copied mostly from `Domain` and Co. So if it works there, it should work here too. \
In this instance my best guess is that `persist` is not needed as the collection has a cascade with `persist`.

If you still have doubts (which are legit as our Domain subscribe feature is not used much) I can go over it again, including Domain.

src/Entity/Hashtag.php
blued-gear: The collection has `orphanRemoval: true`, that should do the trick.

blued-gear: (Also here and for the following comments again: this was taken from Domain::unsubscribe())

src/Entity/HashtagBlock.php
blued-gear: I removed it (also from Domain)

blued-gear: ... but FYI we have this on many entities, just search for `#[Table]` (for example Embed, Image, MagazineBlock, Moderator; in total 18 classes).

src/Entity/HashtagBlock.php
blued-gear: See User line 717, there it is set to null, but the value is not persisted as the entity is removed at the same time. I don't know why it is set to null in the first place, but it is done so in our other block types.

src/Entity/HashtagSubscription.php
blued-gear: see comment above

src/Entity/User.php
blued-gear: `cascade: ['persist', ...`

src/Entity/User.php
blued-gear: should be

blued-gear:
> I also still get an exception here, maybe because the tag is not in the DB, so the tag needs to get created

Should we create hashtags on the fly on request or return a 404? Creation on user-request can be abused to flood the DB.

src/Entity/HashtagBlock.php
BentiGorlich: Oh okay, the ones I looked at didn't have it. Then we can remove it in another PR

src/Entity/HashtagBlock.php
BentiGorlich: maybe because of a remove cascade that is avoided that way?

BentiGorlich:
Yes I think you're right and it could be abused :thinking: 
We could just return a not persisted entity. As it has 0 posts it shouldn't change anything and if someone clicks subscribe or block we could create the entity then and rate limit it...

blued-gear:
Ad-hoc creation on subscription will not work, as the subscription controller expects an entity object in its args. So I will just throw a `NotFoundHttpException`.

src/Entity/HashtagBlock.php
blued-gear: Maybe, but I can't say I know *why* it works, just that it does 😅 .

src/Entity/Hashtag.php
blued-gear: According to a Stack Overflow answer, this should be correct

> cascade={"remove"} meaning that removing entity A, Doctrine will also remove all B entities in the Collection.

Also from [Doctrine Doc](https://www.doctrine-project.org/projects/doctrine-orm/en/3.6/reference/working-with-associations.html)

> Thanks to cascade: remove, you can easily delete a user and all linked comments without having to loop through them:

The inverse side `HashtagSubscription` has a `onDelete: 'CASCADE'` which will add a cascade to the foreign key constraint, so no hashtag-deletion by deleted subscription here too.

PR #2181, opened 2026-07-27
fix creation of temporary images in tests

blued-gear:
Until now, many images were created in the root project dir when test ran (and weren't deleted). This change introduces a central function for providing image copies for uploads (images are still not deleted after tests but are in one central place for easy cleanup).


tests/FactoryTrait.php
```diff
@@ -609,26 +609,43 @@ public function getKibbyFlippedImageDto(): ImageDto
...
+    }
 
-        if (!file_exists(\dirname($this->kibbyPath).'/copy')) {
-            if (!mkdir(\dirname($this->kibbyPath).'/copy')) {
+    public function getKibbyFlippedImageUpload(): UploadedFile
+    {
+        return $this->getKibbyImageVariantUpload('_flipped');
+    }
+
+    private function getKibbyImageVariantUpload(string $suffix): UploadedFile
```
garrettw, line 622: what's up with `$suffix`? It looks like it's not used.

blued-gear: It is used in line 632 `$srcPath = \dirname($this->kibbyPath).'/'.basename($this->kibbyPath, '.png').$suffix.'.png';`.
There are two variants of Kibby in the test assets and one has a suffix in its filename.

garrettw: Oh, I see it now. My mistake.

garrettw (approved):

PR #2184, opened 2026-07-30
fix and speed up migration of 'last_boosted_at' column

blued-gear:
On larger instances this migration step will still take much time. Therefore, I will add to the upcoming release-notes that the SQL script of the postUp() function should be manually executed in the background for such instances.


BentiGorlich (approved):

melroy89:
This migration can take up to **4-6 hours** depending on our hardware and database size. YOu have been warned!

> On larger instances this migration step will still take much time

Yes...

Mainly due ot the fact it goes row by row, single threaded.? And we have many rows on `entry_comment` table.

PR #2185, opened 2026-07-31
Consolidate search and add search for hashtags

blued-gear:
- Add a search form for Magazines to the search page, as the currently separated forms can be confusing for users (the form at the Magazines page is kept, as it offers more search options)
- add a search form for Hashtags to the search page
- it will list all Hashtags when no search-query exists
- implement the same for the REST API

tag_search

Build upon #2180


src/Service/TagManager.php
```diff
@@ -177,4 +184,51 @@ public function isAnyTagBanned(?array $tags): bool
...
+    }
+
+    public function searchWithCriteria(HashtagSearchView $criteria): PagerfantaInterface
+    {
+        return $this->search($criteria->query, $criteria->page, $criteria->perPage ?? TagRepository::PER_PAGE);
+    }
+
+    public function search(string $term, int $page, int $perPage): PagerfantaInterface
+    {
+        $name = $this->tagExtractor->transliterate($term);
```
blued-gear, line 231: TODO: `strtolower($term)`

PR #2189, opened 2026-08-04
fix: video embed of stream/playlist URL

garrettw:


melroy89:
Thanks for starting this, @garrettw. I carried the fix forward in #2224, preserving your original commit and authorship.

The replacement is rebuilt on current `main`, explains the original #487 failure and the gap left by #2172, preserves the video formats added since this draft was opened, expands manifest-path handling and tests, and has full scoped PHPStan/formatting validation. Closing this draft in favor of #2224.

Issue #2191, opened 2026-08-04
Microblog newest + boosted content includes comments, but they don't show the thread


BentiGorlich:
**Describe the bug**
When including a comment, I expect the parent thread and maybe also comment to be somewhat visible in the feed. Like including the short text of it in the rendering including a link to get there. Also they still color the element red if the commenter is also the op

**On which Mbin instance did you find the bug?**
[gehirneimer.de]

**Which Mbin version was running on the instance?**
[latest main]

**Screenshots**

<img width="1205" height="994" alt="Image" src="https://github.com/user-attachments/assets/0b54e9fa-ce9e-4517-a554-70fe64d119e7" />

PR #2193, opened 2026-08-04
fix: show parent post for microblog comments in feed (Fixes #2190, #2191)

### LINKED ISSUE #2190: Microblog Newst with boosted content sometimes mixes designs
**Describe the bug**
There are 2 different designs of microblogs, one for the combined view and one for the microblog view, but they get mixed in the microblog view

Image

[gehirneimer]

[latest main]

**To Reproduce**

**Expected behavior**
They should use one design

waterWang:
## Description

When boosted content is shown in the microblog feed (newest sort), post comments appear without their parent post. This causes two issues:

1. **Mixed designs** (#2190): Comments look different from regular posts because they lack the parent post context
2. **Missing thread context** (#2191): Users can't see what thread the comment belongs to

## Fix

In `templates/layout/_subject.html.twig`, pass `withPost: true` to the `post_comment` component when it's rendered in the non-combined (microblog) view. This causes the parent post to be rendered above the comment, providing thread context and making the visual design consistent.

The `PostCommentComponent` already supports the `withPost` boolean property (defaults to `false`). When set to `true`, it renders the parent post using the `post` component before the comment.

## Before

Comments in the microblog feed appear as standalone items without showing the parent post they belong to.

## After

Comments in the microblog feed show the parent post above the comment, providing full thread context and a consistent visual design.

### DIFF
--- a/templates/layout/_subject.html.twig
+++ b/templates/layout/_subject.html.twig
@@ -45,7 +45,7 @@
     {% if forCombined is same as true %}
         {{ component('post_comment_combined', {comment: subject}|merge(attributes)|merge(postCommentAttributes)) }}
     {% else %}
-        {{ component('post_comment', {comment: subject}|merge(attributes)|merge(postCommentAttributes)) }}
+        {{ component('post_comment', {comment: subject, withPost: true}|merge(attributes)|merge(postCommentAttributes)) }}
     {% endif %}
 {% elseif subject is magazine %}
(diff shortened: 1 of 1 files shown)


blued-gear:
`templates/layout/_subject.html.twig` is used for all instances where an entry / post / comment ... is shown, so you can't set properties in this file as it would affect parts of the UI where it is not intended to show the parent.

PR #2197, opened 2026-08-04
fix: emoji autocomplete inserts 'null' after colon + newline (Fixes #2151)

### LINKED ISSUE #2151: Colon + New Line on Microblog Textbox Appends "null" to Previous Line
**Describe the bug**
When attempting to write a new microblog post, after hitting Enter at the end of the line **where the last character on the line is a colon (:)**, the last word in the line is transformed into "null". This is intermittent, not every Enter event does this. I have a video but Github doesn't seem to allow it to upload.

After a little more testing, I think the issue is the emoji picker, which is triggered by the colon character. I think if you add a newline, then go back to the previous line, there is an attempt to bring up an emoji picker, but if the word has no corresponding emoji, then "null" is inserted instead.

kbin.earth

1.9.1

**To Reproduce**
1. Start a new microblog post.
2. Write a line of text **ending in a colon**.
3. Hit Enter to add a new line.
4. Hit Backspace to go back to the previous line.
5. Hit Enter again.
6. Observe the previous line of text now ending with "null".

**Expected behavior**
Words ending in a colon should not be replaced by "null" if no corresponding emoji exists for that word.

**Screenshots**
Image
Image

 - OS: Windows 10
 - Browser: Firefox
 - Browser Version: 151.0.2

 - Device: N/A
 - OS: N/A
 - Browser: N/A
 - Browser Version: N/A

**Additional context**

waterWang:
Fixes #2151

## Root Cause
When a user types a word ending with `:` and presses Enter, the emoji autocomplete (`rich_textarea_controller.js`) attempts to replace the search string with the selected suggestion. If no emoji matches the typed text, `getSelectedSuggestionReplacement()` returns `null`, and `replaceAutocompleteSearchString(null)` calls `document.execCommand('insertText', false, null)` which inserts the literal string "null".

## Fix
1. Added null check in the `handleInput` Enter handler: only call `replaceAutocompleteSearchString` when there's a valid suggestion
2. Added defensive null check at the start of `replaceAutocompleteSearchString`

Both changes ensure that pressing Enter with no matching emoji suggestion simply clears the autocomplete and inserts a normal newline, as expected.

### DIFF
--- a/assets/controllers/rich_textarea_controller.js
+++ b/assets/controllers/rich_textarea_controller.js
@@ -64,8 +64,13 @@ export default class extends Controller {
                 this.markSelectedSuggestion();
                 event.preventDefault();
             } else if ('Enter' === key) {
-                this.replaceAutocompleteSearchString(this.getSelectedSuggestionReplacement());
-                event.preventDefault();
+                const replacement = this.getSelectedSuggestionReplacement();
+                if (null !== replacement) {
+                    this.replaceAutocompleteSearchString(replacement);
+                    event.preventDefault();
+                } else {
+                    this.clearAutocomplete();
+                }
             } else {
                 this.fetchAutocompleteResults(this.getAutocompleteSearchString(key));
             }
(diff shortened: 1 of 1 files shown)


BentiGorlich (approved):

BentiGorlich:
Thanks for fixing this :)

melroy89 (approved):
Should we merge it?

PR #2198, opened 2026-08-05
docs: update PHP instructions to 8.5 + Ubuntu to 24.04

melroy89:
- Bump docs from PHP 8.4 to the soon new required PHP 8.5 version.
- I would also recommend at least Ubuntu 24.04 today or better Ubuntu 26.04 LTS

### DIFF
--- a/docs/02-admin/01-installation/01-bare_metal.md
+++ b/docs/02-admin/01-installation/01-bare_metal.md
@@ -19,7 +19,7 @@ You can start with a smaller server and add more resources later if you are usin
 
 ## Software Requirements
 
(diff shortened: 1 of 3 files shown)


blued-gear (approved):

BentiGorlich:
I thought we wait until the next release to put php 8.5 in the docs... With our current setup it works, but there are a huge amount of warning when running composer install right now

blued-gear:
Oh, I didn't consider this. But it is also a bit of a dilemma, the docs for the main branch would be outdated in one scenario and too recent for the latest release in the other one.

PR #2199, opened 2026-08-05
intoduce new async_slow queue for image deletion

blued-gear:
On some platforms the deletion of many images when a user gets deleted takes quite a while. To decouple the deletion from the transaction which removes the user, this PR adds a new queue and new message for this job.

### DIFF
--- a/src/Message/DeleteImageMessage.php
+++ b/src/Message/DeleteImageMessage.php
@@ -6,6 +6,9 @@ namespace App\Message;
 
 use App\Message\Contracts\AsyncMessageInterface;
 
+/**
+ * Attempts to delete the image entity from the database, but not the file from storage.
+ */
 class DeleteImageMessage implements AsyncMessageInterface
 {
     public function __construct(public int $id, public bool $force = false)
--- a/src/MessageHandler/DeleteImageV2Handler.php
+++ b/src/MessageHandler/DeleteImageV2Handler.php
@@ -1,0 +1,60 @@
+<?php
+
+declare(strict_types=1);
+
+namespace App\MessageHandler;
+
+use App\Message\Contracts\MessageInterface;
+use App\Message\DeleteImageV2Message;
+use App\Repository\ImageRepository;
+use App\Service\ImageManagerInterface;
+use Doctrine\ORM\EntityManagerInterface;
+use Symfony\Component\HttpKernel\KernelInterface;
+use Symfony\Component\Messenger\Attribute\AsMessageHandler;
+
+#[AsMessageHandler]
+class DeleteImageV2Handler extends MbinMessageHandler
+{
+    public function __construct(
+        KernelInterface $kernel,
+        private EntityManagerInterface $entityManager,
+        private ImageRepository $imageRepository,
+        private ImageManagerInterface $imageManager,
+    ) {
+        parent::__construct($this->entityManager, $kernel);
+    }
+
+    public function __invoke(DeleteImageV2Message $message): void
+    {
+        $this->workWrapper($message);
+    }
+
+    public function doWork(MessageInterface $message): void
+    {
+        if (!$message instanceof DeleteImageV2Message) {
+            throw new \LogicException();
+        }
+
+        // querying each image individually is inefficient but lowers the probability of a race-condition
+        foreach ($message->images as $hash => $filepath) {
+            $img = $this->imageRepository->findOneBySha256(hex2bin($hash));
+            if (null !== $img) {
+                if (!$this->imageRepository->isImageReferenced($img)) {
+                    // this should lock the row and prevent concurrent re-referencing
+                    $this->entityManager->remove($img);
+                    $this->entityManager->flush();
+
+                    $deleteFile = true;
+                } else {
+                    $deleteFile = false;
+                }
+            } else {
+                $deleteFile = true;
+            }
+
+            if ($deleteFile && null !== $filepath) {
+                $this->imageManager->remove($filepath);
+            }
+        }
+    }
+}
--- a/src/MessageHandler/DeleteImageV2Handler.php
+++ b/src/MessageHandler/DeleteImageV2Handler.php
@@ -4,56 +4,92 @@ declare(strict_types=1);
 
 namespace App\MessageHandler;
 
-use App\Message\Contracts\MessageInterface;
+use App\Entity\Image;
 use App\Message\DeleteImageV2Message;
 use App\Repository\ImageRepository;
 use App\Service\ImageManagerInterface;
 use Doctrine\ORM\EntityManagerInterface;
+use Psr\Log\LoggerInterface;
 use Symfony\Component\HttpKernel\KernelInterface;
 use Symfony\Component\Messenger\Attribute\AsMessageHandler;
 
 #[AsMessageHandler]
-class DeleteImageV2Handler extends MbinMessageHandler
+readonly class DeleteImageV2Handler
 {
+    private int $chunkSize;
+
     public function __construct(
         KernelInterface $kernel,
         private EntityManagerInterface $entityManager,
         private ImageRepository $imageRepository,
         private ImageManagerInterface $imageManager,
+        private LoggerInterface $logger,
     ) {
-        parent::__construct($this->entityManager, $kernel);
+        $this->chunkSize = ('test' !== $kernel->getEnvironment()) ? 256 : 2;
     }
 
     public function __invoke(DeleteImageV2Message $message): void
     {
-        $this->workWrapper($message);
+        try {
+            /*
+             * Split workload into chunks to balance overhead with probability of race-conditions. A race-condition can look like the following:
+             *   1. one image in a chunk is detected as an orphan
+             *   2. entity is deleted and transaction commited
+             *   3. new content is created with the same image (resulting in the same file path) while $filesToDelete is iterated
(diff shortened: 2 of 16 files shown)


blued-gear:
I reworked the new code to lower the chance of race conditions between entity and file deletions.

Also, the tests now cover the MessageHandler.

src/MessageHandler/DeleteImageV2Handler.php
```diff
@@ -0,0 +1,96 @@
...
+                    $filesToDelete[] = $img->filePath;
+                    $this->entityManager->remove($img);
+                }
+            }
+
+            $this->entityManager->flush();
+            return $filesToDelete;
+        });
+
+        foreach ($filesToDelete as $path) {
```
melroy89, line 85: The transaction commits before the loop below removes the files. In that gap, another request can upload the same content and `ImageRepository::create()` can create or reuse an image pointing to one of these content-addressed paths. This worker will then delete a file that is referenced again. Locking the existing `Image` rows does not cover the "row does not exist" case either, because there is no row to lock.

The reference check and storage removal need coordination based on the image hash/path that is also used by image creation. For example, the overall shape could be:

```php
$this->imageLock->withHash($hash, function () use ($hash, $path): void {
    $image = $this->imageRepository->findOneBySha256($hash);
    if ($image && $this->imageRepository->isReferenced($image)) {
        return;
    }

    $this->imageManager->remove($path);

    if ($image) {
        $this->entityManager->remove($image);
        $this->entityManager->flush();
    }
});
```

The important part is that both upload/create and deletion acquire the same per-hash lock and that the lock remains held through the file removal. The exact lock implementation can differ, but ending the database transaction before storage deletion leaves the race open.

src/MessageHandler/DeleteImageV2Handler.php
```diff
@@ -0,0 +1,96 @@
...
+            try {
+                $this->imageManager->remove($path);
```
melroy89, line 87: `Image::filePath` is nullable, and both message factories can therefore put a null value in this batch. When that value reaches `ImageManagerInterface::remove(string)`, PHP throws a `TypeError`. The current `catch (\Exception)` does not catch it because `TypeError` implements `Throwable`, not `Exception`. The handler then fails and retries instead of continuing with the remaining images.

A simple guard keeps null paths away from the string-only API:

```php
foreach ($filesToDelete as $path) {
    if (null === $path) {
        continue;
    }

    try {
        $this->imageManager->remove($path);
    } catch (\Exception $e) {
        // existing logging
    }
}
```

It would also help to stop adding null paths to `$filesToDelete` in the first place, but the removal boundary should still be safe because the incoming message may already contain null.

src/MessageHandler/DeleteUserHandler.php
```diff
@@ -122,6 +121,10 @@ public function doWork(MessageInterface $message): void
 
             throw $e;
         }
+
+        // dispatch at end or else reference-check would keep images
+        // because of the reference check this call can be safely placed outside the try{}
+        $this->bus->dispatch(new DeleteImageV2Message($deleteImagesPayload));
```
melroy89, line 127: The user deletion is committed before this message is dispatched. If the transport throws here, Messenger retries `DeleteUserMessage`, but the retry reaches the `$user->isDeleted && null === $user->markedForDeletionAt` branch and returns immediately. At that point the original user's image list is no longer available, so no later attempt schedules the cleanup and those files can remain orphaned permanently.

The cleanup intent needs to be saved atomically with the user deletion. One possible shape is to write it to a transactional outbox inside the same database transaction:

```php
$this->entityManager->persist(
    OutboxMessage::from(new DeleteImageV2Message($deleteImagesPayload))
);
$this->entityManager->flush();
$this->entityManager->commit();
```

A separate worker can then publish that outbox record with retries. If this project already uses a transaction-aware Messenger transport, dispatching through that mechanism inside the transaction would provide the same guarantee. Another valid fix is to make the already-deleted retry path retain enough cleanup data to dispatch the message, but the current replacement user does not contain that data.

melroy89 (dismissed):
Found three correctness issues in the image-deletion flow.

src/MessageHandler/DeleteImageV2Handler.php
blued-gear: It is possible to check the DB before each delete. But this can quickly get an expensive operation, especially when a user with its content gets removed.
We would need to gauge the probability of a wrong deletion by race against the additional DB load.

And checking all images at once and then deleting them within the transaction could recreate the problem of long-term blocked transaction this PR tries to solve.

src/MessageHandler/DeleteUserHandler.php
blued-gear: I would argue that the failing of `bus->dispatch()` is so unlikely that building a whole DB based dispatch buffer is overkill.

src/MessageHandler/DeleteImageV2Handler.php
melroy89: ok I agree, maybe checking all the files before deletion would create additional load again, and then possibility re-creating the locking problem again  🤕 .

So either this solution we have now, or we can still have some link between image creation and deletion with a so called **shared per hash** lock.

If you _want_ to know how to can create a [symphony shared lock](https://symfony.com/doc/current/components/lock.html), here is an example of such a thing. Again this is then **NOT** a db lock, but just an run-time application lock (eg. symphony can store this in Redis I believe as well as PostgreSQL or just file on disk):

```php
$lock = $lockFactory->createLock('image:'.$hash);

$lock->acquire(true);

try {
    $image = $imageRepository->findOneBySha256($hash);

    if ($image && $imageRepository->isReferenced($image)) {
        return;
    }

    if ($image) {
        $entityManager->remove($image);
        $entityManager->flush();
    }

    $imageManager->remove($path);
} finally {
    $lock->release();
}
```

I checked the configs we currently use (see `config/packages/lock.yaml`):

```yaml
framework:
    lock: "%env(LOCK_DSN)%"
```
And the current environment configuration is (eg. see `.env.example`):

```yaml
LOCK_DSN=flock
```
Which is just file using PHP flock (disk) lock: https://www.php.net/manual/en/function.flock.php

src/MessageHandler/DeleteUserHandler.php
melroy89: > I would argue that the failing of `bus->dispatch()` is so unlikely that building a whole DB based dispatch buffer is overkill.

yea ok you are right, the chances of this happening is very low. Although orphaned image consequence is real.. 

We can also add just some logging around dispatch failures (if not already), so if this happen in practice (production) we can later see it in the logs and always do a follow-up later if needed.

src/MessageHandler/DeleteImageV2Handler.php
blued-gear: If we use this, we might want to change it to Redis or the location of the lockfiles, as in container setups the files must be stored in shared volumes.

Then again, reference-checking and deletion is done in batches of 256 images, so the probability of a wrong deletion might be low.

melroy89: Yea it was just an idea and an example. We can always do it later as well if we wish.

Or use this locking mechanism in other places in our project..

src/MessageHandler/DeleteImageV2Handler.php
melroy89: is this done..?

BentiGorlich: It is because if the path is null, then the image is skipped now. If the image was not referenced the entity in the db got deleted earlier as well

src/Message/DeleteImageMessage.php
```diff
@@ -6,6 +6,9 @@
 
 use App\Message\Contracts\AsyncMessageInterface;
 
+/**
+ * Attempts to delete the image entity from the database, but not the file from storage.
```
BentiGorlich: That is not correct... The handler has this at the bottom:

        if ($image?->filePath) {
            $this->imageManager->remove($image->filePath);
        }

BentiGorlich (commented):
It looks good to me. What I don't understand is why you're keeping the old message and the old handler. Either it is a v2 or it is a special handler for big batches, which the name should reflect

blued-gear:
> What I don't understand is why you're keeping the old message and the old handler. Either it is a v2 or it is a special handler for big batches, which the name should reflect

It is a V2. I kept the old Message version because it is to be expected that queues are not empty when admins update their Mbin version. Then Symfony would not be able to deserialize the old messages, and we would lose data (data here means desired actions). \
There are many more usages of the old version, which I would like to keep until the V2 has shown to be effective in prod.

BentiGorlich:
CI is failing with this error: 
> Fatal error: Cannot apply #[\Deprecated] to class App\Message\DeleteImageMessage in /__w/mbin/mbin/src/Message/DeleteImageMessage.php on line 13

melroy89:
> CI is failing with this error:
> 
> > Fatal error: Cannot apply #[\Deprecated] to class App\Message\DeleteImageMessage in /__w/mbin/mbin/src/Message/DeleteImageMessage.php on line 13

yea indeed taht should be just become a doctstring like `@deprecated`.. should I try to fix this?

blued-gear:
> should I try to fix this?

I'm currently at it.

melroy89 (approved):

PR #2200, opened 2026-08-07
fix ImageManager::compressUntilSize()

blued-gear:
The loop checks if the compression had any effect, but it does so with the original file size. If the image got compressed in the first iteration, the condition then would always be false.

### DIFF
--- a/src/Service/ImageManager.php
+++ b/src/Service/ImageManager.php
@@ -107,7 +107,7 @@ class ImageManager implements ImageManagerInterface
         $tempPath = "{$filePath}_temp_compress.$extension";
         $compressed = false;
         $quality = 0.9;
-        if (0.1 <= $this->imageCompressionQuality && 1 > $this->imageCompressionQuality) {
+        if (0.3 <= $this->imageCompressionQuality && 1 > $this->imageCompressionQuality) {
             $quality = $this->imageCompressionQuality;
         }
         while ($bytes > $maxBytes && $quality > 0.3) {
--- a/src/Service/ImageManager.php
+++ b/src/Service/ImageManager.php
@@ -95,7 +95,7 @@ class ImageManager implements ImageManagerInterface
      */
     public function compressUntilSize(string $filePath, string $extension, int $maxBytes): bool
     {
-        if (-1 === (int)$this->imageCompressionQuality || filesize($filePath) <= $maxBytes) {
+        if (-1 === (int) $this->imageCompressionQuality || filesize($filePath) <= $maxBytes) {
             // don't compress images if disabled or smaller than max bytes
             return false;
         }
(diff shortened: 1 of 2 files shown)


src/Service/ImageManager.php
```diff
@@ -95,33 +95,35 @@ public function store(string $source, string $filePath): bool
      */
     public function compressUntilSize(string $filePath, string $extension, int $maxBytes): bool
     {
-        if (-1 === $this->imageCompressionQuality || filesize($filePath) <= $maxBytes) {
+        if (-1 === (int)$this->imageCompressionQuality || filesize($filePath) <= $maxBytes) {
```
blued-gear: IDE showed me that this comparison would always fail as -1 is an int and imageCompressionQuality is a float

src/Service/ImageManager.php
```diff
@@ -95,33 +95,35 @@ public function store(string $source, string $filePath): bool
...
         $initialBytes = $bytes;
+        $lastBytes = $bytes;
         $tempPath = "{$filePath}_temp_compress.$extension";
         $compressed = false;
         $quality = 0.9;
         if (0.1 <= $this->imageCompressionQuality && 1 > $this->imageCompressionQuality) {
             $quality = $this->imageCompressionQuality;
         }
-        while ($bytes > $maxBytes && $quality > 0.1) {
+        while ($bytes > $maxBytes && $quality > 0.3) {
```
blued-gear, line 113: I changed it to 0.3 to shorten the loop. IMO even at this quality level the images would be well fried already.

src/Service/ImageManager.php
```diff
@@ -95,33 +95,35 @@ public function store(string $source, string $filePath): bool
      */
     public function compressUntilSize(string $filePath, string $extension, int $maxBytes): bool
     {
-        if (-1 === $this->imageCompressionQuality || filesize($filePath) <= $maxBytes) {
+        if (-1 === (int) $this->imageCompressionQuality || filesize($filePath) <= $maxBytes) {
```
BentiGorlich, line 98: if you put -1 in the env file, I believe it might be converted to an int? At least that was my idea

BentiGorlich (approved):
I am approving this, but have you though about adding a test? If this code might be flawed from the beginning, maybe that would be a good idea

src/Service/ImageManager.php
blued-gear: The property is declared as `private readonly float $imageCompressionQuality`, so always a float.

blued-gear:
> but have you though about adding a test?

I agree that a test would be nice to have. But testing image manipulation is not as easy as testing an algorithm, so (and as I'm somewhat lazy currently) I'm skipping them for now.

PR #2201, opened 2026-08-07
fix some AP stuff

blued-gear:
- ActivityPubManager::getActorFromAttributedTo(): `array_filter` outputs either an array of a Person or a string with an AP URL, but the next line only handled the array version.
- for some incoming posts which have multiple image attachments, the wrong image is skipped when generating the extra markdown containing the embeddings resulting in the already embedded image replacing it
- support some more video types
- fix faulty parsing of `inReplyTo` attribute (#2208)
- fix bug where mention is not rendered if the mentioned user has a different apId and username

### DIFF
--- a/src/Service/ActivityPubManager.php
+++ b/src/Service/ActivityPubManager.php
@@ -531,9 +531,9 @@ class ActivityPubManager
             try {
                 $imageObject = $images[array_key_first($images)];
                 if (isset($imageObject['height'])) {
-                    // determine the highest resolution image
+                    // determine the highest resolution image for the same image (equality is ducktyped by comparing the alt text)
                     foreach ($images as $i) {
-                        if (isset($i['height']) && $i['height'] ?? 0 > $imageObject['height'] ?? 0) {
+                        if (isset($i['height']) && $i['height'] ?? 0 > $imageObject['height'] ?? 0 && $i['name'] ?? '' === $imageObject['name'] ?? '') {
                             $imageObject = $i;
                         }
                     }
@@ -965,14 +965,13 @@ class ActivityPubManager
         return null;
     }
 
-    public function handleExternalImages(array $attachment): ?array
+    public function handleExternalImages(array $attachment, ?\App\DTO\ImageDto $consumedImage): ?array
     {
+        $imageUrlToSkip = $consumedImage?->sourceUrl;
         $images = array_filter(
             $attachment,
-            fn ($val) => $this->isImageAttachment($val)
+            fn ($val) => $this->isImageAttachment($val) && $val['url'] !== $imageUrlToSkip
         );
-
-        array_shift($images);
 
         if (\count($images)) {
             return array_map(fn ($val) => (new ImageDto())->create(
@@ -1080,7 +1079,7 @@ class ActivityPubManager
     private function isImageAttachment(array $object): bool
     {
         // attachment object has acceptable object type
-        if (!\in_array($object['type'], ['Document', 'Image'])) {
+        if (!\in_array($object['type'], ['Document', 'Image']) || !isset($object['url']) || !is_string($object['url'])) {
             return false;
         }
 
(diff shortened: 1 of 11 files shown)


src/Service/ActivityPubManager.php
```diff
@@ -531,9 +531,9 @@ public function handleImages(array|string $attachment): ?Image
             try {
                 $imageObject = $images[array_key_first($images)];
                 if (isset($imageObject['height'])) {
-                    // determine the highest resolution image
+                    // determine the highest resolution image for the same image (equality is ducktyped by comparing the alt text)
                     foreach ($images as $i) {
-                        if (isset($i['height']) && $i['height'] ?? 0 > $imageObject['height'] ?? 0) {
+                        if (isset($i['height']) && $i['height'] ?? 0 > $imageObject['height'] ?? 0 && $i['name'] ?? '' === $imageObject['name'] ?? '') {
```
melroy89: The null-coalescing operators need parentheses here. In PHP this expression evaluates truthy for every attachment with a non-zero `height`, regardless of whether its height is greater or its `name` matches (for example, both a 50px same-name image and a 200px different-name image pass when the current image is 100px). The loop therefore selects the last height-bearing attachment, so `sourceUrl` can identify the wrong consumed image and the markdown builder can still omit/duplicate the wrong image. Please compare the normalized operands explicitly, e.g. `($i["height"] ?? 0) > ($imageObject["height"] ?? 0) && ($i["name"] ?? "") === ($imageObject["name"] ?? "")`. A focused regression test with multiple images of differing heights/names would catch this.

melroy89 (changes requested):
One blocking issue remains in the multi-image selection logic: the new height/name condition is defeated by PHP operator precedence, so it can choose the last height-bearing attachment rather than the highest-resolution variant of the same image. I left a verified inline comment with a reproducer and suggested correction. The existing checks are green at the reviewed head, but there is no focused regression test covering this case.

blued-gear:
The brackets were added and I implemented some test cases for the function.

melroy89 (dismissed):
Great!

src/Service/ActivityPubManager.php
```diff
@@ -965,15 +975,14 @@ public function handleVideos(array $attachment): ?VideoDto
...
     }
 
-    public function handleExternalImages(array $attachment): ?array
+    public function handleExternalImages(array $attachment, ?\App\DTO\ImageDto $consumedImage): ?array
     {
+        $imageUrlToSkip = $consumedImage?->sourceUrl;
         $images = array_filter(
             $attachment,
-            fn ($val) => $this->isImageAttachment($val)
+            fn ($val) => $this->isImageAttachment($val) && $val['url'] !== $imageUrlToSkip
```
melroy89: This only removes the exact URL selected as the primary image, so alternate-resolution attachments for that same image remain and are appended to the Markdown body. For example, with `A-100.jpg` (`name: A`) and `A-200.jpg` (`name: A`), `handleImages()` consumes `A-200.jpg`, but this filter keeps `A-100.jpg`, displaying the same image twice. Please exclude the whole consumed-image variant group (using the same identity rule as selection) and add a test covering `handleImages()` together with `handleExternalImages()`.

SOmething like this..?

```php
public function handleExternalImages(array $attachment, ?\App\DTO\ImageDto $consumedImage): ?array
  {
      $images = array_filter(
          $attachment,
          fn ($val) => $this->isImageAttachment($val)
              && $val['url'] !== $consumedImage?->sourceUrl
              && (
                  null === $consumedImage?->altText
                  || ($val['name'] ?? null) !== $consumedImage->altText
              )
      );

      if (\count($images)) {
          return array_map(fn ($val) => (new ImageDto())->create(
              $val['url'],
              $val['mediaType'],
              !empty($val['name']) ? $val['name'] : $val['mediaType']
          ), $images);
      }

      return null;
  }
```

blued-gear: done (with a slightly different impl)

melroy89 (approved):

PR #2206, opened 2026-08-10
show a summary of the entry / post a comment is from when displaying it in Combined or Microblog

### LINKED ISSUE #2191: Microblog newest + boosted content includes comments, but they don't show the thread

blued-gear:
microblog
combined
boosts

Mobile is a bit "squished"
mobile

Comments in the *User Overview* are now attached to their parents
overview

Closes #2191 and #2190 (somewhat)

### DIFF
--- a/src/Twig/Runtime/SubjectExtensionRuntime.php
+++ b/src/Twig/Runtime/SubjectExtensionRuntime.php
@@ -4,17 +4,72 @@ declare(strict_types=1);
 
 namespace App\Twig\Runtime;
 
+use App\Entity\Entry;
+use App\Entity\Post;
+use App\Entity\User;
+use Symfony\Bundle\SecurityBundle\Security;
 use Twig\Extension\RuntimeExtensionInterface;
 
-class SubjectExtensionRuntime implements RuntimeExtensionInterface
+readonly class SubjectExtensionRuntime implements RuntimeExtensionInterface
 {
-    public function __construct()
-    {
-        // Inject dependencies if needed
+    public function __construct(
+        private Security $security,
+    ) {
     }
 
-    public function doSomething($value)
+    public function userCanSeeEntry(Entry $entry): bool
     {
-        // ...
+        /* @var ?User $user */
+        $user = $this->security->getUser();
+
+        if (null !== $user) {
+            if ($user->isBlocked($entry->user)
+                || $user->isBlockedMagazine($entry->getMagazine())
+                || (null !== $entry->domain && $user->isBlockedDomain($entry->domain))) {
+                return false;
+            }
+        }
+
+        if ($entry->isVisible() && $entry->user->isVisible()) {
+            return true;
+        }
+
+        if (null !== $user) {
+            if ($user->isAdmin() || $user->isModerator()) {
+                return true;
+            }
+            if ($entry->getMagazine()->userIsModerator($user)) {
+                return true;
+            }
+        }
+
+        return false;
+    }
+
+    public function userCanSeePost(Post $post): bool
+    {
+        /* @var ?User $user */
+        $user = $this->security->getUser();
+
+        if (null !== $user) {
+            if ($user->isBlocked($post->user) || $user->isBlockedMagazine($post->getMagazine())) {
+                return false;
+            }
+        }
+
+        if ($post->isVisible() && $post->user->isVisible()) {
+            return true;
+        }
+
+        if (null !== $user) {
+            if ($user->isAdmin() || $user->isModerator()) {
+                return true;
+            }
+            if ($post->getMagazine()->userIsModerator($user)) {
+                return true;
+            }
+        }
+
+        return false;
     }
 }
--- a/src/Twig/Runtime/SubjectExtensionRuntime.php
+++ b/src/Twig/Runtime/SubjectExtensionRuntime.php
@@ -36,6 +36,12 @@ readonly class SubjectExtensionRuntime implements RuntimeExtensionInterface
     {
         $author = $content->user;
 
+        // don't show deleted content
+        if($content->isSoftDeleted() || $author->isSoftDeleted() || $author->isDeleted) {
+            return false;
+        }
+
+        // handle user blocks
         if (null !== $user) {
             if ($user->isBlocked($author)
                 || $user->isBlockedMagazine($content->getMagazine())
@@ -48,6 +54,7 @@ readonly class SubjectExtensionRuntime implements RuntimeExtensionInterface
             }
         }
 
+        // show visible or allowed private content
         if ($content->isVisible() && $author->isVisible()) {
             return true;
         }
@@ -55,6 +62,7 @@ readonly class SubjectExtensionRuntime implements RuntimeExtensionInterface
             return true;
         }
 
+        // admins and moderators may see everything (except deleted)
         if (null !== $user) {
             if ($user->isAdmin() || $user->isModerator()) {
                 return true;
--- a/src/Twig/Runtime/SubjectExtensionRuntime.php
+++ b/src/Twig/Runtime/SubjectExtensionRuntime.php
@@ -37,7 +37,7 @@ readonly class SubjectExtensionRuntime implements RuntimeExtensionInterface
         $author = $content->user;
 
         // don't show deleted content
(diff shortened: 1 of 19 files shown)


melroy89:
I fixed the merge conflcit in translations/messages.en.yaml for you. just helping you out a bit .

templates/post/comment/context.html.twig
```diff
@@ -0,0 +1,18 @@
+<span>
+    {{ 'in_reply_to'|trans }}
+    {% with {post: comment.post} %}
+    <span>
+        {{ component('user_inline', {user: post.user, fullName: SHOW_USER_FULLNAME is same as V_TRUE, showNewIcon: true}) }}:
+        <a href="{{ path('post_single', {magazine_name: post.magazine.name, post_id: post.id, slug: post.slug|length ? post.slug : '-'}) }}">
+            {% if post.image is not same as null %}
+                <i class="fa-solid fa-photo-film"></i>
+            {% endif %}
+            {{ post.getShortTitle() }}
```
melroy89: This renders the parent body even when the post is private, deleted, blocked, or otherwise hidden from the viewer. `ContentRepository` filters comment visibility independently, so a visible or boosted comment can still reach this template after its parent has been hidden; the normal post component deliberately suppresses the body in those cases. Please apply the parent visibility/access checks before calling `getShortTitle()`. The direct `entry.title` rendering in the entry context partial has the same problem.

blued-gear: I wrote functions (in `SubjectExtensionRuntime`) for visibility handling to get rid of this messy long checks inside the templates.

I'm not sure if the check for mod / admin should be after or before the block check. And maybe a check for blocked content text (and hashtag) might be added, or would this be overkill?

blued-gear:
An unrelated but small fix: see commit ` this is redundant as the same is done in countUpvotes()`

templates/post/comment/context.html.twig
melroy89: > I'm not sure if the check for mod / admin should be after or before the block check. 

I do not think the admin / moderator status should bypass blocks or content visibility at all. Private and soft deleted content should not become visible I think because somebody has a moderator role. 

> And maybe a check for blocked content text (and hashtag) might be added, or would this be overkill?

It's a valid concern, but lets do a separate filtering logic rather then trying to fix this now as part of the visibility helper. 

----

Finally back to the issue itself, that is not fully resolved actually. 

Followed private content is not correctly hidden because `isVisible()` exclude private. Like I said, admin and moderators can see non0visiable state (which include soft delete and un-followed private content). The parent author is rendered before the visibility check and finally the _adult-content_ preference is not yet part of the code check. 


What about:

```php
  private function canShowParentContext(
      Entry|Post $subject,
      bool $moderationContext = false,
  ): bool {
      /** @var User|null $user */
      $user = $this->security->getUser();
      $author = $subject->getUser();
      $magazine = $subject->getMagazine();

      if (!$author->isVisible()) {
          return false;
      }

      if (null !== $user) {
          if ($user->isBlocked($author)
              || $user->isBlockedMagazine($magazine)
              || ($subject instanceof Entry
                  && null !== $subject->domain
                  && $user->isBlockedDomain($subject->domain))) {
              return false;
          }

          if ($user->hideAdult && ($subject->isAdult || $magazine->isAdult)) {
              return false;
          }
      }

      if ($subject->isVisible()) {
          return true;
      }

      if ($subject->isPrivate()) {
          return null !== $user && $user->isFollowing($author);
      }

      if ($subject->isTrashed()) {
          return $moderationContext
              && null !== $user
              && $this->security->isGranted('moderate', $subject);
      }

      // Includes soft-deleted content.
      return false;
  }
```

And then using these two Twig functions that uses this helper method:

```php
  public function userCanSeeEntry(Entry $entry): bool
  {
      return $this->canShowParentContext($entry);
  }

  public function userCanSeePost(Post $post): bool
  {
      return $this->canShowParentContext($post);
  }
```

And finally used in the actually twig templating like so (just an example):

```twig
  {% if user_can_see_post(post) %}
      {{ component('user_inline', {
          user: post.user,
          fullName: SHOW_USER_FULLNAME is same as V_TRUE,
          showNewIcon: true
      }) }}:

      <a href="{{ path('post_single', {
          magazine_name: post.magazine.name,
          post_id: post.id,
          slug: post.slug|length ? post.slug : '-'
      }) }}">
          {{ post.getShortTitle() }}
      </a>

      {% if post.magazine.name is not same as 'random' %}
          {{ 'in'|trans }}
          {{ component('magazine_inline', {
              magazine: post.magazine,
              fullName: SHOW_MAGAZINE_FULLNAME is same as V_TRUE
          }) }}
      {% endif %}
  {% else %}
      <span class="text-muted">{{ 'invisible_or_deleted'|trans }}</span>
  {% endif %}
```
What do you think?

melroy89:
Also your test is failing, due to the change withPost true to withPost false with context True.. while the test til lexpects two full parent post elements:

```php
 $this->assertEquals(2, $crawler->filter('#main .post')->count());
```

Just saying...

templates/post/comment/context.html.twig
blued-gear: > I do not think the admin / moderator status should bypass blocks or content visibility at all. Private and soft deleted content should not become visible I think because somebody has a moderator role.

There is a usecase where it is staying visible is required: to moderate the Magazine / Instance. Then a mod needs to see all content in order to find offending content.

> It's a valid concern, but lets do a separate filtering logic rather then trying to fix this now as part of the visibility helper.

Alright. Hope we won't forget it later.

> Followed private content is not correctly hidden

Could you explain to me what the meaning of the `private` status is?

> The parent author is rendered before the visibility check

This is by design, so that a user has a hint to why the title shows as `invisible or deleted`. The check can be extended to include the user too, but I think without it might not be clear what was deleted there.

> adult-content preference is not yet part of the code check

Will be fixed in the next push.

melroy89: > Could you explain to me what the meaning of the private status is?

Sure, sorry if I was unclear about this. Honestly the user can't set the visibility status to "private". but it is possible via federation actually. SO this applies to the `visibility` column we have on the tables: entry, post, entry_comment and post_comment.  It doesn't happen a lot, but private is a valid string value for this column:

private_visibility

On posts it happens more often:

image

The same column can also, as you might now, have values like `visible`, `soft_deleted` and `trashed`. Hopefully this clears things up a bit? There is also no UX element afaik that the user can use to set it to private at the moment in Mbin itself...

blued-gear: I added comments to the VisibilityInterface constants to make the meaning (hopefully) clear. Also soft deleted content is now hidden from mods / admin.

melroy89:
pipeline is red again, should I fix it? its just a stupid space

src/Twig/Runtime/SubjectExtensionRuntime.php
```diff
@@ -4,17 +4,74 @@
...
+            if ($user->hideAdult && $content->isAdult()) {
+                return false;
+            }
+        }
+
+        // show visible or allowed private content
+        if ($content->isVisible() && $author->isVisible()) {
+            return true;
+        }
+        if (($content->isPrivate() || $author->isPrivate()) && $user->isFollowing($author)) {
```
melroy89, line 61: When an anonymous visitor encounters a comment whose parent entry or post has `visibility = private`, or whose parent author has `visibility = private`, `$user` is `null`, but this line still calls `$user->isFollowing($author)`. Here, private specifically means the `private` visibility state used for followers-only content or accounts. This causes the whole list page to return a 500 instead of showing the fallback. Please check that `$user` is not null before checking whether they follow the author.

melroy89: Mabye something like this?

```php
  if (
      null !== $user
      && ($content->isPrivate() || $author->isPrivate())
      && $user->isFollowing($author)
  ) {
      return true;
  }
```

melroy89 (approved):

PR #2210, opened 2026-08-15
show who boosted content in feed

### LINKED ISSUE #2192: Included boosted content should show that it was boosted and when and by whom
When including boosted content in the newest timeline, boosted content looks quite odd, as it doesn't say that it is boosted nor by whom or when. So you end up with a post in your timeline, by someone you don't know and which date doesn't match the surrounding posts

It should be obvious which posts where boosted, by including that information at the top: "BentiGorlich boosted ... 19 minutes ago" or something like that

none

**Additional context**
For this the query has to be adapted quite a bit, but I think this is essential

blued-gear:
If content gets included in the feed (Combined + Subscribed + Newest), then show who of the followees boosted it.
img

This is currently only for the UI. If you have an idea how this info can be included in API responses, please let me know (@jwr1).

As the content-query can become white heavy when including comments, this PR also adds a setting for the admins to disable comments in the feed.

Closes #2192

### DIFF
--- a/src/Pagination/Transformation/ExtendedContentPopulationTransformer.php
+++ b/src/Pagination/Transformation/ExtendedContentPopulationTransformer.php
@@ -11,7 +11,6 @@ use App\Entity\User;
 use App\Repository\Criteria;
 use App\Repository\UserRepository;
 use App\Utils\SqlHelpers;
-use Doctrine\DBAL\ParameterType;
 use Doctrine\ORM\EntityManagerInterface;
 
 readonly class ExtendedContentPopulationTransformer extends ContentPopulationTransformer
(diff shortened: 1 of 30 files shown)


jwr1:
For the API, we could modify the existing end points to accept an `includeBoosts` parameter. Then I think from there we'd need to include two more properties in the response objects, one for the boosting user, and one for the boosting date time. Does that sound like it could work?

src/Pagination/Transformation/ExtendedContentPopulationTransformer.php
```diff
@@ -0,0 +1,115 @@
...
+            $rewritten = SqlHelpers::rewriteArrayParameters($parameters, $sql);
+
+            $boostsQuery = $this->entityManager->getConnection()->prepare($rewritten['sql']);
+            foreach ($rewritten['parameters'] as $key => $value) {
+                $boostsQuery->bindValue($key, $value, SqlHelpers::getSqlType($value));
+            }
+        }
+
+        $boostInfo = $boostsQuery->executeQuery()->fetchAllAssociative();
+        $boostUsers = $this->userRepository->findBy(['id' => array_map(fn ($row) => $row['user_id'], $boostInfo)]);
```
melroy89, line 104: This performs two additional database queries for every boosted item: `executeQuery()` fetches the votes, then `UserRepository::findBy()` fetches the corresponding users. A full page of boosted content therefore adds up to 2N queries (and the cursored endpoint can do the same), which compounds the feed-performance concern called out in this PR. Please collect boost rows for all page items in batches, ideally one query per content/vote type, and resolve all user IDs in one query before decorating the items. A query-count regression test would help keep this bounded.

melroy89:
@jwr1’s API point is still outstanding. The combined collection endpoints already expose an `includeBoosts` query parameter and pass it into `Criteria`, so no new request parameter appears necessary. What this PR still needs is response metadata: the current serializer converts each decorated entity into the existing response DTO and drops `extendedContentProperties` entirely. Since multiple followed users can boost the same item, I suggest an optional `boostedBy` array of records such as `{ user: UserSmallResponseDto, boostedAt: datetime }` on each content response DTO (with OpenAPI and endpoint tests), rather than two scalar properties.

Let me try to add this directly here, while I'm at it.. So I will extend the PR with the requested API extention.

melroy89:
Implemented the API portion in 3dd16f95a4d7f4d59b0ebab2daedfebb08a33783. The existing `includeBoosts` parameter now causes each combined-feed item to expose optional `boostedBy` records containing the boosting `UserSmallResponseDto` and `boostedAt` timestamp. This covers entries, posts, and both comment types, with OpenAPI metadata and functional assertions. The separate 2N-query review finding remains open.

blued-gear:
I changed the `ExtendedContentPopulationTransformer` so now there are at most 4 + <number of unique boosting users> additional queries issued.

src/Pagination/Transformation/ExtendedContentPopulationTransformer.php
```diff
@@ -0,0 +1,204 @@
...
+            if (isset($userCache[$user])) {
+                $user = $userCache[$user];
+                $fetchUser = false;
+            } else {
+                $usersToFetch[] = $user;
+                $fetchUser = true;
+            }
+
+            $item = ['user' => $user, 'time' => new \DateTimeImmutable($row['created_at'])];
+            $boostExtensions[$row['item_id']][] = &$item;
```
melroy89: Each element added here references the same loop-local `$item` variable. On the next iteration, assigning a new array to `$item` changes every previously stored element as well, so two boost rows of the same content type become duplicate copies of the final row (and `itemsToFix` has the same alias). Please store distinct values/indices instead of references and add a test with at least two boost rows in one type, ideally from different users or for different content IDs.

src/Pagination/Transformation/ExtendedContentPopulationTransformer.php
```diff
@@ -0,0 +1,204 @@
...
+
+                foreach ($itemsToFix as $item) {
+                    if ($item['user'] === $user->getId()) {
+                        $item['user'] = $user;
+                    }
+                }
+            }
+        }
+
+        foreach ($boostExtensions as $boosts) {
```
melroy89: This sorts only the copied `$boosts` array, so the reordered value is discarded at the end of each iteration and the returned boost lists retain the database's unspecified row order. The previous implementation explicitly returned boosts ordered by time, which also determines which users survive the five-item template limit. Please iterate by reference or assign the sorted array back, and cover the ordering with more than one boost row.

blued-gear: As a sidenote: who in the right mind put in the PHP specification that arrays are always passed by value?!

melroy89 (approved):

PR #2212, opened 2026-08-25
fix overflow handling when a post or comment contains an overly wide table

blued-gear:
Before the text in the table would be extremely squeezed (wrapped) or partly invisible. Now the table scrolls while the surrounding content keeps it width.


melroy89 (approved):
👍🏽

PR #2216, opened 2026-09-09
Fix abandoned magazine ordering test flake

melroy89:
## Summary

- give one abandoned magazine a subscriber so the default `subscriptionsCount DESC` ordering is deterministic
- preserve the original positional assertions and default-sort coverage
- leave the explicit `ownerLastActive` ordering test unchanged

## Cause

The test created two abandoned magazines with equal subscription counts, then assumed a stable order. PostgreSQL may return tied rows in either order because the query has no secondary ordering, producing the intermittent failure seen in [this workflow run](https://github.com/MbinOrg/mbin/actions/runs/34353636142/job/102472724588).

The fixture now gives `magazine1` one subscriber, removing the tie while continuing to verify the endpoint’s documented default sort.


blued-gear:
You could also add a subscriber to one of the magazines. This way the test also covers the expected sort order.

melroy89:
> You could also add a subscriber to one of the magazines. This way the test also covers the expected sort order.

That is true indeed.. let me do that instead

blued-gear (approved):

PR #2218, opened 2026-09-09
Restore Redis DSN support for session storage

melroy89:
# Summary

Reuse Symfony's existing DSN-backed Redis provider for session storage instead of creating a second connection from mandatory `REDIS_HOST` and `REDIS_PORT` variables.

This restores support for all existing `REDIS_DNS` connection forms, including Unix sockets, while keeping TCP, authentication, and Docker configurations on the same canonical setting.

# Additional information

PR #2127 added Redis-backed session storage but created a separate `Redis` service that requires `REDIS_HOST` and `REDIS_PORT`. That made previously valid Unix-socket configurations fail during environment-variable resolution, even though `REDIS_DNS` remained configured as Symfony's default Redis provider.

`RedisSessionHandler` now receives `cache.default_redis_provider`, which Symfony builds through its Redis DSN connection factory. The redundant host and port variables are removed from the environment templates.

# Related issues

Follow-up to #2127.


BentiGorlich:
If it works, this is awesome. I didn't find a solution to stay on the dsn variable...

melroy89:
> If it works, this is awesome. I didn't find a solution to stay on the dsn variable...

Yes indeed. So we basically reuse the default redis provider variables documented here as well https://symfony.com/doc/current/cache.html#configuring-cache-with-frameworkbundle

melroy89:
And yes this should work again for  both TCP DNS and socket DNS. Since the `default_redis_provider` resolves from `%env(REDIS_DNS)%`. I tested it already.

see: https://github.com/MbinOrg/mbin/blob/037b9b296a87ab032be067f463b422fd9ed84543/config/packages/cache.yaml#L12

Symfony’s framework config then uses that provider here: `vendor/symfony/framework-bundle/Resources/config/cache.php:139` (see locally on disk if you have executed composer install):

```php
  'provider' => 'cache.default_redis_provider',
```

Please approve.

melroy89:
plz?

blued-gear (approved):

PR #2219, opened 2026-09-09
Tune autovacuum thresholds for high-churn tables

melroy89:
## Problem

On large Mbin instances, the global `autovacuum_vacuum_scale_factor = 0.2` lets high-churn tables accumulate millions of changed or dead rows before PostgreSQL schedules routine maintenance. This delays space reuse and leaves planner statistics stale longer than needed.

For example, a `favourite` table with roughly 190 million rows (on kbin.melroy.org) would wait for about 38 million changed/dead rows under the global default.

The production PostgreSQL guide also sets `maintenance_work_mem = 2GB` without setting `autovacuum_work_mem`. Since autovacuum inherits that value by default, several workers could each reserve the larger amount.

## Change

Add a PostgreSQL migration with conservative per-table autovacuum settings:

- `favourite`: vacuum and analyze at 50,000 rows plus 0.5% of table size (about 1 million rows at 190 million rows).
- `entry_comment`: vacuum and analyze at 20,000 rows plus 2% of table size.

The migration affects only table storage parameters. It does not run a vacuum, rewrite tables, change schema columns, or modify global PostgreSQL settings. Its down migration resets the overrides so an instance returns to its own global policy.

Update the PostgreSQL guide to cap each autovacuum worker at `1GB` while retaining `2GB` for manual maintenance.

## Why a migration

These are PostgreSQL-specific storage parameters, rather than Doctrine ORM mapping. Shipping them in a Doctrine migration gives every upgraded Mbin instance the same sensible defaults while preserving administrators' global worker, I/O, and resource policies.

### DIFF
--- a/docs/02-admin/02-configuration/04-postgresql.md
+++ b/docs/02-admin/02-configuration/04-postgresql.md
@@ -12,7 +12,7 @@ sudo nano /etc/postgresql/16/main/postgresql.conf
 
 These settings below are more **an indication and heavily depends on your server specifications**. As well as if you are running other services on the same server.
 
-However, the following settings are a good starting point when your serve is around 12 vCPUs and 32GB of RAM. Be sure to fune-tune these settings to your needs.
+However, the following settings are a good starting point when your server is around 12 vCPUs and 32GB of RAM. Be sure to fine-tune these settings to your needs.
 
 ```ini
 # Increase max connections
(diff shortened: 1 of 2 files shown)


blued-gear:
I think post_comment can also be added.

melroy89:
@blued-gear I checked the live PostgreSQL statistics on my instance, the current values are (on my instance kbin.melroy.org):

- `favourite`: 191.7M live rows, 102k dead rows, and 376k modifications since analyze. It has no recorded autovacuum, so the lower thresholds in this PR are justified.
- `entry_comment`: 11.05M live rows, 73k dead rows, and 227k modifications since analyze. This also has enough scale and churn to justify tuning.
- `post_comment`: 157.7k live rows, only 1k dead rows, and 1.5k modifications since analyze. It was autovacuumed and autoanalyzed recently on 2026-09-10.

So I would keep this PR limited to `favourite` and `entry_comment`. `post_comment` is much smaller and its default autovacuum/analyze policy is currently keeping it healthy; adding a table-specific override without evidence of stale statistics or accumulating dead tuples would add policy without a demonstrated benefit. We can revisit it if those live metrics change. What do yo uthink after seeing this values?

blued-gear (approved):

PR #2223, opened 2026-09-11
Fix CI failure when main advances during tests

melroy89:
## Summary

Fix the CI helper failure seen when `main` advances while a workflow is running.

The helper used `git diff --name-only main^ HEAD` after fetching the moving `origin/main` ref. With the checkout's shallow history, a concurrent commit could leave the tested commit as the shallow boundary, causing `main^` to fail before the thread-safe integration tests started.

This changes the main-branch path to compare the checked-out commit (`HEAD^..HEAD`) and configures the build and automated-test checkouts with `fetch-depth: 2`.

## Failing job addressed

This is the failure being resolved:
https://github.com/MbinOrg/mbin/actions/runs/34630637289/job/103366325277

The job passed unit tests and non-thread-safe integration tests, then failed with:

```text
fatal: ambiguous argument 'main^': unknown revision or path not in the working tree.
```

### DIFF
--- a/ci/skipOnExcluded.sh
+++ b/ci/skipOnExcluded.sh
@@ -7,9 +7,6 @@ git config --global --add safe.directory "$(realpath "$GITHUB_WORKSPACE")"
 
 ignoredPatterns="$(cat "$GITHUB_WORKSPACE"/ci/ignoredPaths.txt)"
 if [[ "${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" == 'main' ]]; then
-    # Compare the commit checked out for this run with its parent. Fetching the
-    # moving origin/main ref here can make the checked-out commit the shallow
-    # boundary when main advances during the workflow, leaving main^ unknown.
     changedFiles="$(git diff --name-only HEAD^ HEAD)"
 else
     git fetch origin main:main --depth 1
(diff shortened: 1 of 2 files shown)


ci/skipOnExcluded.sh
```diff
@@ -7,8 +7,10 @@ git config --global --add safe.directory "$(realpath "$GITHUB_WORKSPACE")"
 
 ignoredPatterns="$(cat "$GITHUB_WORKSPACE"/ci/ignoredPaths.txt)"
 if [[ "${GITHUB_HEAD_REF:-${GITHUB_REF#refs/heads/}}" == 'main' ]]; then
-    git fetch origin main --depth 2
-    changedFiles="$(git diff --name-only main^ HEAD)"
+    # Compare the commit checked out for this run with its parent. Fetching the
+    # moving origin/main ref here can make the checked-out commit the shallow
```
blued-gear: Please remove or boil down this comment. The reference to the fetching issue makes only sense in context of this issue and not for general functionality of the code.

melroy89: I agree, comment fully removed.

ci/skipOnExcluded.sh
```diff
@@ -7,8 +7,10 @@ git config --global --add safe.directory "$(realpath "$GITHUB_WORKSPACE")"
...
+    # boundary when main advances during the workflow, leaving main^ unknown.
+    changedFiles="$(git diff --name-only HEAD^ HEAD)"
```
blued-gear, line 10: ~~Does this work with multi-commit branches or would it just compare with the second-last commit of the PR branch?~~

I missed that this is only for runs on *main*. So this should work.

blued-gear (approved):

ci/skipOnExcluded.sh
melroy89: `HEAD^` Refers to the immediate parent commit (one commit before HEAD), so only the  single most recent commit.

PR #2226, opened 2026-09-18
Pin generated URLs to the configured domain

melroy89:
## Summary

- Pin the Symfony router context to the configured instance domain before absolute URLs are generated.
- Normalize the configured host, scheme, and ports once when the subscriber is constructed.
- Log structured warnings when a request-derived router context diverges from the configured canonical context.
- Reapply the canonical context for main requests, sub-requests, and after Symfony restores a parent request on `kernel.finish_request`.
- Add focused regression coverage for hostile hosts, custom ports, logging, and sub-request lifecycle resets.

Related to GHSA-v54w-hc3m-68pq.

## Testing

- `vendor/bin/phpunit --no-configuration --bootstrap vendor/autoload.php tests/Unit/EventSubscriber/RouterContextHostSubscriberTest.php`
- PHP CS Fixer dry run for the subscriber and its unit test


src/EventSubscriber/RouterContextHostSubscriber.php
```diff
@@ -0,0 +1,60 @@
...
+            ? $this->kbinDomain
+            : 'https://'.$this->kbinDomain;
+        $parts = parse_url($canonicalUrl);
+
+        if (false === $parts || !isset($parts['host'])) {
+            return;
+        }
+
+        $scheme = $parts['scheme'] ?? 'https';
+        $port = $parts['port'] ?? null;
```
blued-gear: This can be moved into the constructor, `kbinDomain` made a normal arg (not property) and scheme, etc. made to private class properties.

src/EventSubscriber/RouterContextHostSubscriber.php
```diff
@@ -0,0 +1,60 @@
...
+        $context = $this->router->getContext();
+
+        $context->setHost($parts['host']);
+        $context->setScheme($scheme);
+        $context->setHttpPort('http' === $scheme && null !== $port ? $port : 80);
+        $context->setHttpsPort('https' === $scheme && null !== $port ? $port : 443);
```
blued-gear: You could also check if the values in `context` diverge from the expected values before overwriting them. This way it can be logged if it happens so that we can observe if some URL generation in the Codebase is vulnerable.

src/EventSubscriber/RouterContextHostSubscriber.php
```diff
@@ -0,0 +1,60 @@
...
+    }
+
+    public static function getSubscribedEvents(): array
```
blued-gear: Just as a suggestion: most event subscribers have this function right next to the constructor. It is better readable what events are subscribed by the given class.

src/EventSubscriber/RouterContextHostSubscriber.php
melroy89: good idea, I moved this PR from the advisory git clone, into our own mbin pull request.. so I think the code can be more optimized. So yes I agree.

src/EventSubscriber/RouterContextHostSubscriber.php
melroy89: Implemented in e9006d05: divergent host, scheme, or port values are now logged with structured context before the canonical values are restored. Tests cover both divergence and the no-warning case.

src/EventSubscriber/RouterContextHostSubscriber.php
melroy89: Implemented in e9006d05. `getSubscribedEvents()` now sits directly after the constructor, matching the dominant subscriber layout in this repository.

src/EventSubscriber/RouterContextHostSubscriber.php
melroy89: Implemented in e9006d05: the configured domain is parsed once in the constructor and only the normalized host, scheme, and port are retained.

src/EventSubscriber/RouterContextHostSubscriber.php
blued-gear: One more thing: I'm not sure if a warning is enough or if we should make the request fail, if a divergence is detected. The current setup can be theoretically abused to redirect a request to an arbitrary path of the instance domain.

PR #2225, opened 2026-09-18
Block private network access in embed fetches

melroy89:
# Summary

Protect embed metadata and preview fetching against SSRF by routing `embed/embed` requests through Symfony's `NoPrivateNetworkHttpClient`. Private, loopback, link-local, reserved, and DNS-resolved internal destinations are blocked for both initial requests and redirects.

Related to:  GHSA-pmfw-qrm9-x75w, GHSA-58m5-w66j-f7cm and GHSA-f9cq-2j36-vqjg

PR #2228, opened 2026-09-20
prevent HTTP request to internal networks by user content

blued-gear:
Introduce new HttpClient factory which produces a client that blocks requests to internal and localhost networks using Symfonys `NoPrivateNetworkHttpClient` decorator.
All HttpClients in the project were replaced with instances produced by this factory.

fixes GHSA-779p-4f7q-pg32 \
supersedes #2225


src/Controller/AjaxController.php
```diff
@@ -207,7 +209,7 @@ public function fetchOnline(
         HttpClientInterface $httpClient,
         CacheInterface $cache,
     ): JsonResponse {
-        $resp = $httpClient->request('GET', $mercurePublicUrl.'/subscriptions/'.$topic, [
+        $resp = $this->httpClientFactory->getClient($httpClient)->request('GET', $mercurePublicUrl.'/subscriptions/'.$topic, [
```
melroy89: Since `MERCURE_PUBLIC_URL` is a trusted, configured endpoint, routing this request through `NoPrivateNetworkHttpClient` can reject valid split-horizon or local deployments when the public URL resolves to a private address. Could this status request use the unfiltered client (or the internal `MERCURE_URL`) instead?

src/Factory/WwwHttpClientFactory.php
```diff
@@ -0,0 +1,60 @@
...
+ * Creates HttpClients which block requests to non-WWW (e.g. localhost and LAN) destinations.
+ */
+readonly class WwwHttpClientFactory
+{
+    private HttpClientInterface $defaultClient;
+
+    public function __construct(
+        private HttpClientInterface $httpClientBase,
+    ) {
+        $this->defaultClient = $this->buildFilteredClient($this->buildConfiguredClient($this->httpClientBase));
```
melroy89, line 24: `defaultClient` is created once in the factory constructor, so its `NoPrivateNetworkHttpClient` DNS cache can live for the entire lifetime of a FrankenPHP or Messenger worker. If the remote host rotates DNS, image, embed, and visibility requests may remain pinned to the first address. Could the decorator be created per use, or could the factory implement/reset the cached client alongside the underlying resettable client?

melroy89 (dismissed):

melroy89:
Maybe just two small minor nits..:

1. Transitive dependency used directly,  `UrlUtils::checkUrlSubpathNotAscending()` uses `GuzzleHttp\Psr7\Uri/UriResolver`, but `guzzlehttp/psr7` isn't in composer.json. It's only present transitively via embed/embed. It works today, but a future dependency swap could break it. Consider adding "guzzlehttp/psr7": "^2.x" to require explicitly, agreed? Or do not depend on it ;P
2. Test base class, `UrlUtilsTest` extends `WebTestCase` (boots the full kernel) but only calls static methods with no container access needed. `EmbedTest` in the same directory extends plain `PHPUnit\Framework\TestCase` ->  matching that would make it a true unit test and faster to run.

melroy89 (approved):

